PR #10, opened 2015-10-12
Updating the error message on failIfDirectoryExists.

zormit:
The new error message describes another existing method to prevent the failure
that happens when the output directory exists on a newly started simulation.


zormit:
I tried to revise the error message. My goal was to be concise, as no one wants to read long texts in the console and I suppose this error message should be more a reminder than a tutorial on how to do it.

Another option would be to formulate it more tutorial-ish:

    A solution would be to delete the directory or its content. 
    With access to the code, another option is to change the configuration to another behavior via setOverwriteFileSetting() in the ControlerConfigGroup.

PR #11, opened 2015-10-19
Check sequence of activities and legs in DTD

krlmlr:
This actually works on my system when I pass this DTD to `xmllint`. YMMV. Has this been discussed before?


mrieser:
The strict act-leg-act sequence was required a while ago, but is no longer mandatory. Thus, this PR would result in correct, valid and supported files to no longer be readable. I think there are even tests that check that such files are correctly supported.

PR #25, opened 2016-01-06
Matrix bases pt input test

gabrielthunig:
fixed an issue of the MatrixBasedPTInputTest that caused an error on travis while the test didn't failed on a pc


gabrielthunig:
didn't change travis.yml

PR #27, opened 2016-01-07
inprovements in OsmNetworkReader

michalmac:
Two major changes are:
1.  the OSM lanes tag specifies the total no of lanes (i.e. in both directions). Therefore, for 2-way roads the numbers of lanes in each direction is 'lanes/2'. The osm maps of Poznan and Berlin follow this rule.
2. the access tag was added to prevent building links where no traffic is allowed
In particular, change 1 may result in reduced flow and storage capacities of some links in newly converted networks. 

These changes are also (being?) introduced by Michael Z. in josm-matsim-plugin

Besides that, a few smaller improvements/bugfixes were introduced.

### DIFF
--- a/matsim/src/main/java/org/matsim/core/utils/io/OsmNetworkReader.java
+++ b/matsim/src/main/java/org/matsim/core/utils/io/OsmNetworkReader.java
@@ -102,8 +102,6 @@ public class OsmNetworkReader implements MatsimSomeReader {
 
 	private boolean slowButLowMemory = false;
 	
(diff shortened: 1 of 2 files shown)


michaz:
Sounds good. For reference: http://wiki.openstreetmap.org/wiki/Key:lanes
I already put this into the JOSM plug-in.
Eventually, I'd like to factor out the "business rules" out of the OsmMatsimReader so that I can re-use them from the plug-in.

PR #40, opened 2016-07-11
File name attributes in core config groups are now relative URLs

michaz:
tl;dr
File names in the config file are now relative to the location of the config file (or else absolute, but not relative to the working directory). Just say `"network.xml"` instead of `"/test/scenario/berlin/network.xml"` if you are the `config.xml` in that directory. You can also put in absolute URLs like `"http://some.where/my-scenario/network.xml"`. You can also do things like `"../../otherScenario/network.xml"`, but that is discouraged because it isn't very portable, for example it doesn't work for Java resources across sub-modules.

- The `Config` class now has a `context` property, which is the URL it was loaded from. The name `context` seems to be official nomenclature, see the `URL` constructors.

- The core config attributes which are input file names are now taken to be relative to this `context`. This is done by the code which gets the attribute and passes it to the file reader, using the Java `URL` class. No rewriting of strings happens; the string properties always contain exactly what is written in the config file.

- A `Config` which is not loaded, but created, has the current directory as `context`. This means that I would now, for example
`config.network().setInputFile("network.xml");`
instead of
`config.network().setInputFile(` /* get current directory +*/ "network.xml");`

- "Additively loading" another config into a loaded config does *not* change the `context`, but only sets the strings to the new values, verbatim. This is because it is analogous to loading a config and then setting some attributes in code, where the same happens.

- I added extra getters, like `URL getInputFileAsURL(URL context)`, to the relevant config groups, which returns an URL resolved with `context`, **but I think I should remove them again**:

 1. They have to be passed the context, because config groups don't have back pointers to the `Config`. 
 2. `config.getInputFileAsURL(config.getContext())` is not better or clearer than `new URL(config.getContext(), config.getInputFile())`.

- All `config.xml` files covered by tests were changed, i.e. file name attributes were changed to be relative to the location of the `config.xml`.

- Nothing precludes adding some switch which results in the Config being loaded with the `context` set to the current directory, emulating the previous behaviour for everyone who uses the `context`, including the standard `Controler`. But maybe that is not even requested. A small survey among the practitioners indicates that the previous behavior was mostly considered to be causing trouble, with people using **only absolute** paths in reaction (which is still possible, of course).

- There is nothing automatic that can be done about non-core file readers: Interpreting the `Config` and reading things happens entirely in contrib-developer-land. We can only establish conventions.

- Nothing was done about the output directory. I'm not sure what (if anything) we should do there. Supporting URLs for writing is a whole different story and probably not what one wants.

- All this means that one can now load a scenario entirely from the network:
```
Config config = ConfigUtils.loadConfig(new URL("https://raw.githubusercontent.com/matsim-org/matsimExamples/master/tutorial/lesson-3/config.xml"));
Scenario scenario = ScenarioUtils.loadScenario(config);
assertThat("Network has expected size.", scenario.getNetwork().getLinks().size(), equalTo(12940));
assertThat("Population has expected size.", scenario.getPopulation().getPersons().size(), equalTo(8760));
assertThat("There are the expected number of activity facilities.", scenario.getActivityFacilities().getFacilities().size(), equalTo(10281));

```

- All this also means that the examples now work with the GUI, as they could all be changed to config-location-relative paths.

- I would like some help with the playgrounds, please.


michaz:
I fixed **some** playgrounds, but at some points, there are complicated things going on, like reading back a written config -- I'll need some help there. @tduberne @boescpa

PR #43, opened 2016-10-21
Freefloating car sharing service area

paschke:
this PR introduces freefloating service areas as discussed with F. Ciari and M. Balac on 2016-08-16. This could be achieved easier than I expected by extending the FreeFloatingVehiclesContainer much like the OneWayContainer:
- findClosestAvailableParkingLocation was implemented. The method now checks if the destinationLink is within the company's ff service area. If it is not, the closest link that is still within the service are is determined (The JTS Topology Suite has a convenience method for that), and the agent walks to the destination.
- a maximum walking distance was not implemented. Long egress walks will score badly and be eliminated.
- The FreefloatingAreas object is a map of MultiPolygon objects with carsharing provider names as keys. Each freefloating area can consist of several polygons, hence the MultiPolygon
- The FreefloatingAreas object is loaded by a separate loader extending MatsimXmlParser. It must be present in CarsharingXmlReaderNew when the CompanyContainers are created, so it must be loaded in CarsharingSupplyContainer.
- if FreefloatingAreas are not configured or loaded, the behaviour of the CarsharingSupplyContainer and of the VehiclesContainers should not be affected.
- the FreefloatingAreas service does not prevent the initial placement of, or agent access to, freefloating vehicles outside of the service area. This is a configuration issue the user must resolve by himself.


balacmi:
This "applay gitignore" commit includes a lot of changes in the cadyts contrib? This needs to be reversed

PR #60, opened 2017-03-30
Clruch14

claudioglasses:


mrieser:
Pull requests should generally be much smaller, and not contain over 300 commits, unless they are major new features. But even then, a description would be very helpful so one knows what new features are to be merged.

A lot of the changes (e.g. in QNode.java) seem to be just automatic formatting of code. As MATSim does not have strict formatting rules, I oppose such changes, as this could end up in commit-wars about the formatting. I don't care about the playground classes, but there are also files in org.matsim.* (as the aforementioned QNode) included in this PR.

Why is there a MANIFEST file added to examples/scenario that links to a playground? If one wants to build a runnable jar, the manifest should be configured in the maven build, e.g. by providing a custom profile or assembly descriptor, but not by committing a manually written MANIFEST file.

In its current form, I would propose to **NOT** merge this PR.

sebhoerl:
This PR just got here by accident

PR #71, opened 2017-05-10
remove dependency cycle

neuma:
Can we remove that cyclic dependency? Makes it somewhat difficult to build matsim from scratch.
`matsim-tutorial` depends on `matsim `and the maven-javadoc-plugin of `matsim `depends again on `matsim-tutorial`

### DIFF
--- a/matsim/pom.xml
+++ b/matsim/pom.xml
@@ -185,13 +185,6 @@
 							<packages>org.matsim.vis:org.matsim.vis.*</packages>
 						</group>
 					</groups>
-					<additionalDependencies>
-						<additionalDependency>
-							<groupId>org.matsim</groupId>
-							<artifactId>matsim-tutorial</artifactId>
-							<version>0.9.0-SNAPSHOT</version>
-						</additionalDependency>
-					</additionalDependencies>
 					<detectLinks>true</detectLinks>
 					<additionalparam>-Xdoclint:none</additionalparam>
 				</configuration>


michaz:
Thanks -- how does it make it difficult do build matsim from scratch? What are you doing that doesn't work?

The line that you are proposing to remove makes it possible to link to the tutorial documentation from the core documentation.

michaz:
Try

```
mvn package --projects matsim -DskipTests
```

in the root (!) directory.

michaz:
(Or even ``-Dmaven.javadoc.skip=true``, if you like.)

michaz:
(With multi-module projects, it's better to let Maven sort out the internal dependencies. Previously, they were *three* multi-module projects, but that was mainly because they were formerly three repositories.)

neuma:
> Thanks -- how does it make it difficult do build matsim from scratch? What are you doing that doesn't work?

My failing scripted workflow had been: Go through each required project and do
1. mvn clean -U
2. mvn source:jar
3. mvn install -U -DskipTests=true
4. mvn -Prelease -DskipTests=true

```
mvn install --projects matsim -DskipTests
```
works fine. Thanks.

michaz:
If you say

```
mvn package --projects xxsome-contribxx --am -DskipTests
```

it will figure out the dependencies xxsome-contribxx needs and build them before. It's probably better to always add the ``--am``. (It means "also make" :-})

neuma:
Well, I tried some of the more advanced reactor switches. However, I failed to simply build all of MATSim with the exception of the playgrounds, i.e. the naive approach of
```
mvn install --projects !playgrounds -DskipTests
```
still puts the playgrounds in the reactor build order. Any idea?

michaz:
> Any idea?

No. I think that's the limit of Maven, where it shines through that sub-modules, were put in _after_ the initial design.

``playground`` here just means the "playgrounds artifact" (which is just an "empty" POM), and can't refer to "playgrounds, the thing which has sub-modules".

We'll move the playgrounds to another repository one day.

michaz:
There _are_ workarounds. For example, one could put in another profile switch (´´-PwithoutPlaygrounds``) which puts in the playgrounds-submodule conditionally.

But I _really_ want to get our build more conventional, not more sophisticated.

neuma:
ok, thanks. So I build matsim and contribs separately

PR #75, opened 2017-07-07
Count sim comparison

cdobler:
I extended the outputs created by MATSim's counts package since there was some information I was missing when working with the current output:

- CountSimComparisonKMLWriter:
     - Count station id's are now included in the popups. This often helps to identify which direction is counted in case both directions are counted and the coordinate is the same.
     - Another set of markers is included that represents the count stations hourly GEH values (see: https://en.wikipedia.org/wiki/GEH_statistic). Not visible by default.
     - Yet another set of markers is included that represents something that we call "normalized relative error". The currently used "relative error" uses the counted values as 100% level. As a result, the calculated value is not symmetric. It can be e.g. -100% (no vehicles in the simulation) but also +200%. In our definition, the calculation is like Abs(sim - count) / Max (sim, count). As a result, the range is from 0%...100%. I also adapted the symbols. In case the fit is very good (e.g. error <10%), then a circle is used. So far, even in case the fit was (almost) perfect, always a +/- sign is used. This set of markers is also not visible by default.

I am not sure what other people think about currently used formular for the "relative error". At least I never liked it :? If other people have the same opinion, it might be removed?

- countscompare.txt:
     - "new" relative error is included
     - GEH value is included
     - count station id is included
     - Values are now between 0 and 1. Before, the were in % - but this was not inidicated in the header.

- countscompareAWTV.txt:
     - "new" relative error is included
     - count station id is included

Any ideas how to further improve the output?


kainagel:
Dear Christoph, thanks a lot.
Re "/max(sim,count)" ... I would rather have this as a config switch.
I am fine with the other changes.
Cheers!

kainagel:
Oops, pressed the wrong button above.  re-opening this

mrieser:
Looks like there are test failures:
```
Failed tests: 
  CountsControlerListenerTest.testFilterAnalyzedModes:344 expected:<150.0> but was:<5.0>
  CountsControlerListenerTest.testReset_CorrectlyExecuted:323 expected:<3.5> but was:<5.0>
```
They should be fixed before this can be merged.

cdobler:
Tests should be fine now. They did not like the additional columns in the output files...

PR #135, opened 2017-11-09
Fixing a bad link to the questions section

3vivekb:
If you aren't logged into atlassian, the current link leads you to an error page.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,8 +14,7 @@ For more information, see the project website at http://www.matsim.org/.
 
 ## Questions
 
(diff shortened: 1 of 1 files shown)


kainagel:
Dear 3vivekb, 

thank you for your effort.  There is indeed a problem.

For my personal taste, I don't really like links that seem to be related to underlying hardware or implementations.  So I would rather prefer finding some other solution.

Also, when I try your suggestion while not being logged in, I end up at the same place.  

Best wishes, Kai

3vivekb:
Darn you are right.

PR #158, opened 2017-12-15
fix .gefx file export in contrib/minibus/stats

vsp-gleich:
The previous implementation only wrote setStart (that means start time
respectively start iteration) for XMLAttvalue values, so these
dynamically changing (at PConfigGroup.WRITE_GEXF_STATS_INTERVAL) values
overlapped, because no end time was defined and more and more values
were added. This caused Gephi to complain about the overlaps and not to
load the exported gefx files.

solution:
Add corresponding setEnd(iteration) to each XMLAttvalue
setStart(iteration), so start and end time/iteration are equal and can
not overlap.

This might lead to gaps if PConfigGroup.WRITE_GEXF_STATS_INTERVAL > 1,
for example:

first value: start=1, end=1; 
next value: start=10, end=10;
...

### DIFF
--- a/contribs/minibus/src/main/java/org/matsim/contrib/minibus/stats/GexfPOperatorCount.java
+++ b/contribs/minibus/src/main/java/org/matsim/contrib/minibus/stats/GexfPOperatorCount.java
@@ -181,7 +181,9 @@ final class GexfPOperatorCount extends MatsimJaxbXmlWriter implements StartupLis
 			}
 			
 			operatorIdValue.setStart(Double.toString(iteration));
-			operatorCountValue.setStart(Double.toString(iteration));			
+			operatorIdValue.setEnd(Double.toString(iteration));
+			operatorCountValue.setStart(Double.toString(iteration));		
+			operatorCountValue.setEnd(Double.toString(iteration));	
 
 			entry.getValue().getAttvalue().add(operatorIdValue);
 			entry.getValue().getAttvalue().add(operatorCountValue);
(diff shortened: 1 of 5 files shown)


neuma:
@vsp-gleich Thanks for looking into this.

I might be wrong, but as far as I remember this has been Gephis way of updating an attribute. It kept a value from start until infinity and updated it once a new start event had been parsed. Back then the dynamic parts of Gephi had some issues. Probably they cleaned the code. Had you checked what Gephi expects now?

vsp-gleich:
@neuma 
Thanks for this hint. On github only gefx specifications 1.1draft and 1.2draft are available, although Gephi claims gefx version 1.3 support since Gephi 0.9.0. (so there should be somewhere a 1.3 specification). I see that in the 1.2 specification (https://gephi.org/gexf/1.2draft/gexf-12draft-primer.pdf) not setting an end time effectively makes the value valid until infinity (p. 15) and overlaps are solved by overriding the previous value after the next value had its start time (p. 20).

I started to look into this when I had run the minibus example and tried to open the output gefx files in Gephi 0.9.2 (the most recent version) and in the older version 0.8.2 on Linux and on Windows. Unfortunately none of these versions could open the gefx files. All of them gave the same error message that overlapping intervals are not allowed. So adding an end time seemed like an obvious solution and did indeed solve the overlap problem for standard gefx output of the minibus example. 

In the open file dialog Gephi warns that it could not detect the gefx version and therefore assumed version 1.3. So maybe it would still work without setting end times if Gephi had an option to set the gefx reader to version 1.2. However, neither the open file dialog nor the program options allow to set the gefx version. 

We could explicitly write the gefx version into the file header, but I'm not sure for how long Gephi will support older gefx versions. There is an example file in version 1.2 and Gephi already complains that the file format would be deprecated. The whole Gephi community does not seem to be very active.

vsp-gleich:
On the separate fixMinibusGefxFileExport2 branch I only added the version, but did not add endTimes. The resuiting gefx files could not be opened. There was the same error message "Overlapping intervals aren't allowed". So, even though the 1.2 specification (https://gephi.org/gexf/1.2draft/gexf-12draft-primer.pdf) allows for overlapping and the version was set and recognized by Gephi, Gephi still refuses to open the gefx file. I merged the fixMinibusGefxFileExport2 branch into this branch, so now both endTimes and the version are set.

PR #162, opened 2018-01-05
Matsim 768

vsp-gleich:
fix of raptor startup (initalize RaptorDisutility, ...)
fix of minibus schedule updating (prior to that the router used the schedule of the previous iteration)
bugfix in raptor (trip from transit stop to the same transit stop)

see Matsim-768 on Atlassian


neuma:
@vsp-gleich @amit2011 I really enjoyed the efficient bug hunting with you two. Thanks.

PR #168, opened 2018-01-10
travis increase memory and allow longer time out

vsp-gleich:
see MATSim-769 on Atlassian
travis_wait 15 increases time until time out (due to no output received from the test, i.e. a TestClass has completed all its test methods) from 10 to 15 min
increased memory could make tests run faster, avoiding time outs

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,7 @@ stages:
   - test
 
 # this is the script run in the "test" stage:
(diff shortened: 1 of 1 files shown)


vsp-gleich:
The changes tried here had little effect. The higher memory setting was obviously ignored and travis_wait x min seems to be started only once when the test package is started and seems to run only for these x min. It is not restarted each time a time out period starts (that means there was output on the console and the time out period re-starts from 0). The long build times have disappeared without any change on our side, so the problem was probably out of our control.

PR #192, opened 2018-02-07
Updates to emissions and road pricing contribs for SCCER work

joemolloy:
These two commits just make working with the emissions and road pricing packages easier. In particular, the generalisation of the HEBFA package allows for the direct correlation with OSM road types (What the swiss networks are built from, instead of just VISUM)

### DIFF
--- a/contribs/emissions/src/main/java/org/matsim/contrib/emissions/EmissionModule.java
+++ b/contribs/emissions/src/main/java/org/matsim/contrib/emissions/EmissionModule.java
@@ -38,6 +38,8 @@ import org.matsim.core.events.EventsUtils;
 import org.matsim.core.utils.io.IOUtils;
 import org.matsim.vehicles.VehicleType;
 import org.matsim.vehicles.Vehicles;
+import roadTypeMapping.HbefaRoadTypeMapping;
+import roadTypeMapping.VisumHbefaRoadTypeMapping;
 
 
 /**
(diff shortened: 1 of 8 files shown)


joemolloy:
The build failed on the minibus contrib, so just trying to run the checks again, and I didnt touch the minibus contrib and the ITests pass on my machine

amit2011:
On 7. Feb 2018, at 17:15, Joe Molloy <notifications@github.com<mailto:notifications@github.com>> wrote:


The build failed on the minibus contrib, so just trying to run the checks again, and I didnt touch the minibus contrib and the ITests pass on my machine

This happens because one of the test takes more than 10 mins on the server which triggers the failure. You can just restart ‘minibus’ contrib run.

https://travis-ci.org/matsim-org/matsim/jobs/338502055

Thanks.

Regards,
Amit

joemolloy:
@michaz, just a thought... could we maybe speed up CI testing by only running tests on contribs that depend on those modified in the commit? Ie ive changed emissions and roadpricing, but it's running the tests for every single contrib, even though they should be theoretically separate.

jfbischoff:
all safe and sound, so closing this one :)

michaz:
> just a thought... could we maybe speed up CI testing by only running tests on contribs that depend on those modified in the commit? Ie ive changed emissions and roadpricing, but it's running the tests for every single contrib, even though they should be theoretically separate.

@joemolloy Yeah, I know.. However, the more combinatorics the build has, the more likely things are going to go wrong. Even more so when it's connected to deploying artifacts. (Which our Travis build isn't, but in principle..)

Currently I think the benefits of having a CI build which does exactly the same thing as ``mvn package`` (or equivalent) does on the command line outweigh most other things, and that's what I tried to come close to. (Not sure what the exact current state is.)

PR #206, opened 2018-02-25
QSim Modularization (MATSIM-756)

sebhoerl:
DO NOT MERGE YET

**Playgrounds**
- This PR probably involves a few changes in playgrounds that build custom QSims. In most cases it should be no more than adding some @Inject annotations, because now, instead of QSim.get*() most objects must be obtained via injection / constructor passing.

**Current update:** (to keep track of the changes)
* MobsimTimer and AgentCounter are global objects for now (and they are reset by the QSim constructor, everytime a new QSim is constructed). Later these will become Mobism-scope objects.
* Mobsim is not passed through MobsimEvents anymore
* All service getters like Mobism.get*(), Netsim.get*(), QSim.get*() are gone, the relevant object can be obtained via injection
* There is no MobsimEngine.setInternalInterface() anymore, the InternalInterface can now be obtained via injection
* NetsimNetwork must now be obtained via injection
* NetsimInternalInterface has been removed, InternalInterface already covered all the functionality before and there was no real decoupling from it anyway
* ActiveQSimBridge has been introduced. It is useful for those cases where a global scope object got hold of the QSim via MobsimEvents. For the refactoring phase, these classes can access the current QSim via ActiveQSimBridge.getActiveQSim(). Whenever the injector creates a new QSim object it will be registered there


sebhoerl:
Closing for now. I'll integrate everything step by step.

PR #223, opened 2018-03-15
fixed the javadoc

neuma:
minor change

Is there any way to commit such atomic changes without going to a branch?


jfbischoff:
unfortunately, not.
We could possibly overthink our policy on that. I'll write it down for the Developer Meeting.

PR #270, opened 2018-05-16
DRT insertion filtering

michalmac:
Speed up DRT by applying insertion filtering. In short, the following steps are taken:
1. generate (pickup&dropoff) insertions that satisfy vehicle capacities
2. (new step) calculate insertion costs using `DetourTimeEstimator` and dismiss insertions that violate time constraints
3. execute the proper insertion algorithm for all filtered insertions

It is important that the `DetourTimeEstimator` is computationally-fast and gives good estimates. If the estimates are too high, valid insertions may get filtered out (-> may lead to more request rejections). If the estimates are too low, many initially accepted insertions will be unnecessarily processed in step 3 (-> longer computational times).


michalmac:
An example of a very simple detour times estimator that uses the existing DRT config parameters:
```
(from, to) -> DistanceUtils.calculateDistance(from, to) / (1.5 * drtCfg.getEstimatedDrtSpeed()
				/ drtCfg.getEstimatedBeelineDistanceFactor())
```
A more advanced approach could use landmarks, for instance.

====

Speed improvements for the TaxiBerlin scenario (27k to 108k requests per day, 1.6k available taxis over the day on average):
- Before: DRT 6-8x slower than the the _fastest_ taxi algorithm (rule-based dispatching)
- After: DRT is only 2-3x slower than the _fastest_ variant of taxi.

PR #274, opened 2018-05-18
further speed improvements in DRT

michalmac:


michalmac:
QSim runtime for autonomous taxi services with ride sharing in the 10% Berlin scenario (270k requests, 6k vehicles) down to 1h 50min.

PR #282, opened 2018-06-22
DrtRoute

michalmac:
- introduce DrtRoute
- calculation of unshared (direct) route is moved from `DrtRequestCreator` to `DrtRoutingModule`, which means it happens during replanning (not in simulation, as was the case), and thus can be easier to parallelise


michalmac:
10% Berlin AV scenario with ridesharing - QSim runtime down by 25 min from 1:50 to 1:25 at the cost of extra 2.5 min during PrepareForSim (12 threads).

PR #291, opened 2018-07-06
Minibus allow sbb raptor

vsp-gleich:
throw TransitScheduleChangedEvents, which the vsp-gleich SBB raptor fork (to be moved later) retrieves and updates schedule afterwards. Remove (duplicate) PTransitRouterFactory and corresponding PConfig setting for router. Other router implementations can use these TransitScheduleChangedEvents, too. 

@sebhoerl : This is a specific issue with the minibus contrib, which changes the transit schedule before each iteration. It used to have a PTransitRouterFactory class which did the schedule updating, but this seemed partly a duplicate of the normal Router Factory and was hard to combine with the external SBB raptor, which needed to be altered to retrieve the new TransitScheduleChangedEvents and update the schedule correspondingly.


Owen295:
@vsp-gleich I am running the minibus contrib for Nairobi paratransit's and I am not sure if the problem is the  PTransitRouterFactory because pStopsOnly are not returned after the first iteration , is it because the new transit schedule has not been updated for the iteration , I fail to see where it breaks in my simulation. When I run logs for the data the input have been well initialized.

vsp-gleich:
@Owen295 could you please open an issue here [https://github.com/matsim-org/matsim-code-examples/issues](https://github.com/matsim-org/matsim-code-examples/issues) and elaborate more on what your error message is, what matsim version you are using and provide a minimum reproducible example (a link to input files which give the error you describe, not including unrelated features)? It is hard to comment with the information you provided here and old code pull requests are not the place to discuss issues running the software.

PR #295, opened 2018-07-10
Added unit test to show rounding up of netsim link travel times

sebhoerl:


michalmac:
BTW. We use this TT when we run "uncongested" scenarios and want to have exact TTs:

https://github.com/matsim-org/matsim/blob/master/contribs/dvrp/src/main/java/org/matsim/contrib/dvrp/trafficmonitoring/QSimFreeSpeedTravelTime.java

sebhoerl:
Okay, good to know, we'll use it!

Any clue why those AppVeyor tests are failing now? Seems completely unrelated.

sebhoerl:
@michalmac Btw, I looked into this because we were wondering if the estimated travel times from the TravelTimeCalculator are consistent with the simulated ones when they are used for routing. I could imagine (though not sure) that there is a systematic bias that underestimates link travel times. I imagine the case where we have e.g. a travel time of 300s (everything freespeed) and suddenly in the Netsim the agent needs 301s. If this problem adds up for each link a route with let's say 30 links will already have a bias of 30s because of the rounding. On the other hand I assume that the TravelTimeCalculator would give 301s then in the first place. Did you investigate this before or are you aware that anybody did this?

michalmac:
Yes, there is always 1 second lost when "more/less" nobody travelled during a given time bin (more/less - because there are 2 strategies for collecting TTs). And yes - this is more noticeable for longer routes. That's why for "precise" DRT tests with time windows on empty network we started using this modified `QSImFreeSpeedTravelTime`.

The good thing is that the TT calculator uses time periods between 2 consecutive link-enter events, so 1 second used for moving over the node is then included. So "typical" matsim runs should not be that much affected.

I do not know how much it is a bug in TT calculation or in QSim. We could also argue that QSim keeps vehicles for 1 second too long on a link.

sebhoerl:
Ok, thanks for the info. Btw, is there a way to restart those AppVeyor builds? They seem to have failed at completely unrelated spots

michalmac:
> is there a way to restart those AppVeyor builds? They seem to have failed at completely unrelated spots
Looks like the windows IO thing? We had such problems last year and then @jfbischoff  and @mrieser fixed them. Don't really remember -- but could be related to parameterized tests (where the output directory returned from MatsimTestUtils is the same for a set of tests)

jfbischoff:
I will try to have a look into it within the next days. Simply ignore Appveyor until then :)

PR #296, opened 2018-07-11
Test consistency for routed travel times and Netsim travel times

sebhoerl:
See test comment


michalmac:
Yes, this is another problem that we tried to answer inside DVRP. So we add 1 sec for the first link and free-flow TT for the last one:
https://github.com/matsim-org/matsim/blob/master/contribs/dvrp/src/main/java/org/matsim/contrib/dvrp/path/VrpPaths.java

Using this approach together with the QSimFreeFlowTravelTime, we were able to get exactly the same  times both in routing and simulation (in an empty network).

PR #297, opened 2018-07-12
Hbefa road type

amit2011:
* MATSIM-785 (i.e. hbefaRoadType can be set directly to link).


joemolloy:
Hi @amit2011 , Would you care to shed some light on why these changes were necessary? and why they were merged without any comment from others involved? unforunately i was working on other things around when this push request was proposed

I made some modifications to the contrib earlier this year, to make it work with osm road types as well, and your changes have broken that functionality. I appreciate the marking of my functionality as 'depreciated', but 'depreciated' only means depreciated if it still works.... 

Additionally, this code is full of unnecessary imports and dubious naming of variables, which could be improved.

joemolloy:
Thanks for the context, @amit2011 . I've reopened the issue and we can continue the discussion there.

Regarding pull requests, with the amount of concurrent development happening concurrently on MATSim, it can be easy to miss things that are relevant (which happened to me here). May i suggest adding recent committers to affected files/contribs as PR reviewers in the future? This might help avoid future similar situations.

PR #298, opened 2018-07-13
Facilities related changes

amit2011:
Here is a summary of the important changes:

- MATSIM-801: coord for facility is optional (however, there will be an exception if neither link nor coord is available)
- similar to XY2Links, there is now also XY2LinkForFacilities which sets linkId for facilities if not available already
- slowly moving towards the convention in which facilities are used before activities and activities are used as fall-back option
- same convention is used to identify the subTours in TripStructureUtils

### DIFF
--- a/matsim/src/main/java/org/matsim/core/controler/PrepareForSimImpl.java
+++ b/matsim/src/main/java/org/matsim/core/controler/PrepareForSimImpl.java
@@ -1,6 +1,11 @@
 package org.matsim.core.controler;
 
 
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import javax.inject.Inject;
+import javax.inject.Provider;
 import org.apache.log4j.Logger;
 import org.matsim.api.core.v01.Id;
 import org.matsim.api.core.v01.Scenario;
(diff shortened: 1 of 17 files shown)


matsim/src/main/java/org/matsim/core/controler/PrepareForSimImpl.java
```diff
@@ -113,6 +113,10 @@ public void run() {
 				throw new RuntimeException("Facilities source '"+this.facilitiesConfigGroup.getFacilitiesSource()+"' is not implemented yet.");
 		}
 
+		// get links for facilities
+		// using car only network to get the links for facilities. Amit July'18
+		XY2LinksForFacilities.run(carOnlyNetwork, this.activityFacilities);
```
neuma: @amit2011 This adds another layer of automagic. I personally dislike the idea that PrepareForSim assigns on-the-fly link Ids to all facilities that lack one.

I understand that thanks to FacilitiesFromPopulation we now have the possibility to move the (location) info from activities to facilities. FacilitiesFromPopulation can optionally assign links to facilities that only have a coordinate. That is, if an activity features only coordinates the newly created facility also only features the coordinate - and misses a link. Adding link Ids via automagic to facilities should be an active decision by the modeler. If the modeler forgets to assign some link Ids to activities/facilities AND did not opt for the link assigning in FacilitiesFromPopulation the code should crash and not fix it with a single line warning.

kainagel: I can see this.  So we would have

(1) activity provided with coord but not link ➝ autogenerate facility with coord and link; remove coord and link in activity
(2) activity provided with link but not coordinate ➝ autogenerate facility with link and coordinate; remove coord and link in activity
(3) everything else, in particular when facilities provided which have only coordinates but no links ➝ do nothing; fail when inconsistent

PR #303, opened 2018-07-24
new conversion method in CoordUtils

dziemke:


neuma:
@dziemke reverted in #322

PR #313, opened 2018-07-31
Minibus fix routes find terminus

vsp-gleich:
new algorithm to determine the 2nd terminus stop. Instead of direct beeline distance from base stop (1st terminus, per definition at index 0 in stopsToBeServed), sum up all beeline distances between two following stops starting from the base stop. Then choose the stop half way on the circular route from the base stop over all other stops back to the base stop.

Appears to yield more plausible, i.e. less complex minibus routes.


Owen295:
@vsp-gleich I am still trying to understand what happened with the route modifications in the minibus module, I have probably brought this up on another comment, but my simulation does not run but seems all the stops in the network are initialized but not identified at the pStopOnly . Is it possible to get help understand how this specific implementation works for route design ?

vsp-gleich:
@Owen295 could you please open an issue here [https://github.com/matsim-org/matsim-code-examples/issues](https://github.com/matsim-org/matsim-code-examples/issues) and elaborate more on what your error message is, what matsim version you are using and provide a minimum reproducible example (a link to input files which give the error you describe, not including unrelated features)? It is hard to comment with the information you provided here and old code pull requests are not the place to discuss issues running the software.

PR #321, opened 2018-08-16
Dvrpvehiclereader

jfbischoff:


michalmac:
I suggest to keep the backward compatibility only for cases when the double-precision `capacity` retrieved from XML satisfies: `capacity == (int)capacity` (otherwise throw an exception). Actually, this should give us a 100% backward compatibility, as I do not remember anybody has used fractional capacities in the past.

PR #330, opened 2018-08-28
Remove "Identifiable" from "Facility"?

kainagel:
as discussed on dev mtg

### DIFF
--- a/matsim/src/main/java/org/matsim/facilities/Facility.java
+++ b/matsim/src/main/java/org/matsim/facilities/Facility.java
@@ -41,6 +41,6 @@ public interface Facility extends BasicLocation, Customizable {
 
 	public static final String FACILITY_NO_LONGER_IDENTIFIABLE = "After refactoring, facility " +
 													 "does not longer automatically " +
-										     "implement Identifiable.  Don't know what to do.";
-
+													 "implement Identifiable.  Don't know what to do.";
+	
 }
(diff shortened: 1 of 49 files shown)


matsim/src/main/java/org/matsim/facilities/Facility.java
```diff
@@ -31,11 +31,16 @@
...
 	//   interface Facility extends BasicLocation<Facility>, ...
 	// kai, dec'15
 
 	public Id<Link> getLinkId();
 
+
+	public static final String FACILITY_NO_LONGER_IDENTIFIABLE = "After refactoring, facility " +
+													 "does not longer automatically " +
+										     "implement Identifiable.  Don't know what to do.";
+
```
mrieser: hmm, terrible code formatting...

matsim/src/test/java/org/matsim/core/router/TripRouterFactoryImplTest.java
```diff
@@ -231,10 +231,10 @@ public Coord getCoord() {
...
 		}
 
-		@Override
-		public Id<Link> getId() {
-			return l.getId();
-		}
+//		@Override
+//		public Id<Link> getId() {
+//			return l.getId();
+//		}
```
mrieser: why not just delete as in other places?

mrieser:
After looking through the diff, it looks fairly straight forward.
Despite my comments, for me it would be okay to merge it (assuming it compiles and runs the tests).

PR #336, opened 2018-08-30
MATSIM-852 Additional utilities

tduberne:


tduberne:
@kainagel I added you as a reviewer, as I am unsure about the idea of a "ProjectionUtils" class and you are the one with the clearest concept about those "utils" classes. This PR introduces static methods to work with projections in container attributes (support for it was part of a previous PR)

PR #355, opened 2018-09-12
using the network mode from the dvrp config group instead of 'car'

ikaddoura:


ikaddoura:
After discussion with Michal and Joschka: Ok, let's keep it as it is. Closing this pull request without merging.

PR #363, opened 2018-09-18
improve request rejection handling

michalmac:
Main changes:
- standardise handling of both types of rejections (see #362 for discussion)
- add utils for calculating time constraint violations
- DrtRequestValidator return type changed from boolean to Set<String> that lists causes of constraint violations


michalmac:
DrtRequestRejectedEvent.getCause() returns the cause of rejection (comma-separated if more then one)

PR #365, opened 2018-09-20
add mobsimMode to support other modes than "car"

michalmac:
Now we can physically simulate truck or bike deliveries.

`RunOneTruckExample` shows how to set up a scenario where we use modes other than "car" and we do not transport passengers.


michalmac:
Note that there are 3 different modes in DVRP now:
1. `mode` is used to handle request of that mode (e.g. `taxi`, `drt`,...)
2. `networkMode` is used to route vehicles along mode-allowed links (this way we can, for instance, restrict operations of AVs to some selected links/areas)
3. `mobsimMode` is mostly used in traffic simulation, i.e. `QSim` (`mainModes` and `seepModes`) or  `TravelTimeCalculator` (`analyzedModes` and `filteredModes`).

PR #367, opened 2018-09-21
fare computation: use mode from fare config group instead of the taxi config group

ikaddoura:
This allows us to use the fare handler for taxi or drt.

Previously it used to be the mode from the dvrp config group, but since the mode was moved to the drt / taxi config group this handler was only working for taxi.

### DIFF
--- a/contribs/av/src/main/java/org/matsim/contrib/av/robotaxi/scoring/TaxiFareConfigGroup.java
+++ b/contribs/av/src/main/java/org/matsim/contrib/av/robotaxi/scoring/TaxiFareConfigGroup.java
@@ -29,10 +29,8 @@ import org.matsim.core.config.ReflectiveConfigGroup;
 
 /**
  * @author  jbischoff
- * Config Group to set taxi fares.
- */
-/**
- *
+ * Config Group to set taxi or drt fares.
+ * 
  */
 public class TaxiFareConfigGroup extends ReflectiveConfigGroup {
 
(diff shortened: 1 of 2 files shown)


michalmac:
Thanks, @ikaddoura . I thought it was used only for taxi (because of the class name).

BTW. Maybe we should rename it then?

ikaddoura:
re rename: yes, probably that makes sense... on the other hand DRT may be considered as some kind of (shared) taxis. So the term 'taxi' is not completely wrong (and maybe not worth the refactoring).

((BTW: Until today, I don't understand why there is a taxi contrib at all. As I understand, DRT with vehicle capacity = 1; and door2door service mode should do exactly the same as what the taxi contrib is doing. Right?))

michalmac:
Taxi is there mostly for historical reasons - it was first, and then DRT was created. Right now the optimisers in taxi are optimised for capacity=1 (faster than their DRT counterparts, and the results may be slightly better). That is why we still keep taxi alive. At some point in time we will move them into DRT (along with some analysis code) and then say good-bye to taxi. But there is no fixed date yet.

ikaddoura:
Ah, I see.

Maybe there is a way then to let the DRT module decide to either use the faster (and maybe slightly better) taxi optimizer in the case of vehicle capacity = 1 or the slower drt optimizer in the case of vehicle capacity > 1?

michalmac:
Unfortunately, not. Once the taxi optimisers get migrated into DRT, there will be some params in the DRT config group to switch between them. But then the choice is made once. Of course, I imagine one could switch optimisers even between iterations, but rather not in the middle of one.

PR #385, opened 2018-10-11
WIP: Configuring QSim via annotations

sebhoerl:
@michalmac See the `QSimComponentsTest` can you maybe try to put Drt / Taxi in in the way we were discussing? https://matsim.atlassian.net/browse/MATSIM-867

Right now it is a bit unfortunate that all the DepartureHandlers etc. are `QSimComponent`s now, while the configuration thing is a `QSimComonents` object... I guess we need to find a better name :)


sebhoerl:
Actually, I also would like to have `bindComponent(...).to(...)` instead of one method for all the different types, this will need a bit more refactoring. Especially it is not possible to know the actual type of the component until it is actually instantiated.

As a sidenote: I think generally it should also work now to do something like this:

```java
bind(MySpecificImplementationOfAMobsimEngine.class).annotatedWith(Drt.class);
```

That way it is also possible to bind multiple Components under the same annotation (which is especially useful for MobsimListeners, for instance in the WithinDayModule). In that sense the bind*(..) in the AbstractQSimModule are a bit misleading, because they bind a `MobsimListener` with a specific annotation to an implementation. However, only one `MobsimListener` with that specific annotation can exist in Guice...

sebhoerl:
@michalmac Eclipse was tricking me, there are still many errors in the contribs... I will fix them, then you can have a look.

michalmac:
Thanks. Looks good. I will adapt taxi/drt next week. Just have a few notes here:

For guice `Class<XxxAnnotation>` is equivalent to `XxxAnnotation` if it is a marker annotation. Probably the easiest way is to internally operate on `Key.get(FakeClass.class, annotation)` and `Key.get(FakeClass.class, annotationType)` for:
- uniqueness checks in `QSimComponents`
- for keeping the registry of components in `ComponentRegistry` (then we would need only one map: `Map<Key<FakeClass>, List<Key<? extends QSimComponent>>>`)

> Right now it is a bit unfortunate that all the DepartureHandlers etc. are QSimComponents now, while the configuration thing is a QSimComonents object... I guess we need to find a better name :)

Oh, yes. Initially I got lost in your second longer comment because I was thinking about the other component:) `QSimComponentsActivator`? (although it does not directly activates stuff) `ActiveQSimComponentsRegistry`??

> Actually, I also would like to have bindComponent(...).to(...) instead of one method for all the different types, this will need a bit more refactoring

Makes sense to me as now everything is `QSimComponent` and `QSimProvider` takes care of interpreting what is what. And yes, we would not need to invent strings if we want to have more then one listener. That would be a really cool feature.

One more thing: This should be probably a **not**-null check: https://github.com/matsim-org/matsim/pull/385/files#diff-fc4c78a0290fdaab5cc0af7d4bf430e2R36.

sebhoerl:
> For guice `Class<XxxAnnotation>` is equivalent to `XxxAnnotation` if it is a marker annotation. Probably the easiest way is to internally operate on `Key.get(FakeClass.class, annotation)` and `Key.get(FakeClass.class, annotationType)` for:
> 
>     * uniqueness checks in `QSimComponents`
> 
>     * for keeping the registry of components in `ComponentRegistry` (then we would need only one map: `Map<Key<FakeClass>, List<Key<? extends QSimComponent>>>`)

Not sure exactly, what you mean. Right now both options are already included? Maybe I don't understand the point correctly, but also for Guice it makes a different if one binds with `Class<XXXAnnotation>` or `XXXAnnotation`, no? The difference is relevant for `@Named` because it is actually a `NamedImpl` object created through `Names.named(...)`.

michalmac:
>  for Guice it makes a different if one binds with Class<XXXAnnotation> or XXXAnnotation, no? 
It is equivalent only  if this is a marker interface: https://github.com/google/guice/blob/master/core/src/com/google/inject/Key.java#L321.
So `@Drt` is equivalent to `Drt.class`. But for `Named` it is exactly as you said.

I need to look into this a bit deeper next week. For instance, if I (maybe accidentally) call `QSimComponents.addComponent(Named.class)` it will result in automatic registration of all components registered with Named annotations regardless of their actual names. (Please, correct me if I am wrong??). Whereas my expectation is that it should not register anything.

Thus my idea was to use "fake" Keys so that operations on annotations are handled in exactly the same way as guice does this.

sebhoerl:
Ok, I see now how it can get confusing, I agree. However, in Guice it works similar, if you do
```java
bind( ... ).to( ... ).annotatedWith( ... )
```
there are two different signatures for `annotatedWith`, one expects `Annotation annotation`, the other `Class<? extends Annotation> annotationType`.

Maybe it depends on how the annotations can be used? Is there maybe some Annotation-Magic mechanism that does not allow to annotate anything with `Named` itself, but only with `NamedImpl` ?

michalmac:
@sebhoerl , I just pushed two commits with changes to `QSimComponents` and `ComponentRegistry`, where `Key`s (generated for `Object.class`, but the actual annotations) are used internally to compare annotations. Please feel free to modify it.

To illustrate the change, I modified `ComponentRegistryTest`. There are 4 different keys and each considered distinct (also by guice):
```
Key[type=org.matsim.core.mobsim.qsim.interfaces.MobsimEngine, annotation=@com.google.inject.name.Named(value=A)]
Key[type=org.matsim.core.mobsim.qsim.interfaces.MobsimEngine, annotation=@com.google.inject.name.Named(value=B)]
Key[type=org.matsim.core.mobsim.qsim.interfaces.MobsimEngine, annotation=@com.google.inject.name.Named]
Key[type=org.matsim.core.mobsim.qsim.interfaces.MobsimEngine, annotation=[none]]
```

sebhoerl:
Looks good, I'm just running through a version where I added the "multibinding" feature, and I spotted another little bug. I think then we can merge it.

sebhoerl:
Okay, I did some further refactoring to really just have `addComponent` or `bindComponent`. The only thing that remains is to rename stuff. I would propose `QSimComponents` -> `QSimComponentConfig`. 

But something is strage... the changes are already in the branch, but somehow they don't show up here in the PR, no clue why?

sebhoerl:
Okay, I did some further refactoring to really just have `addComponent` or `bindComponent`. The only thing that remains is to rename stuff. I would propose `QSimComponents` -> `QSimComponentConfig`. 

But something is strage... the changes are already in the branch, but somehow they don't show up here in the PR, no clue why?

sebhoerl:
Okay, I did some further refactoring to really just have `addComponent` or `bindComponent`. The only thing that remains is to rename stuff. I would propose `QSimComponents` -> `QSimComponentConfig`. 

But something is strage... the changes are already in the branch, but somehow they don't show up here in the PR, no clue why?

sebhoerl:
Okay, I did some further refactoring to really just have `addComponent` or `bindComponent`. The only thing that remains is to rename stuff. I would propose `QSimComponents` -> `QSimComponentConfig`. 

But something is strage... the changes are already in the branch, but somehow they don't show up here in the PR, no clue why?

michalmac:
> Okay, I did some further refactoring to really just have addComponent or bindComponent. The only thing that remains is to rename stuff. I would propose QSimComponents -> QSimComponentConfig.

Cool. Thanks!

`QSimComponentConfig` looks okay. Would be great if you could also rename variables of this type and methods like `provideQSimComponents()` to `provideQSimComponentConfig()` so that there is no confusion.

> But something is strage... the changes are already in the branch, but somehow they don't show up here in the PR, no clue why?

Strange. Also the 'View it on GitHub' links in notification email were pointing to (half-)broken pages. There is though one hint form github: "We went looking everywhere, but couldn’t find those commits.
Sometimes commits can disappear after a force-push. Head back to the latest changes here." Maybe you have force-pushed them?

sebhoerl:
Apparently, it was Github's problem, they had some storage problems or whatever.

sebhoerl:
I'd say this is ready to merge now, but I'd like to wait until the other PRs are merged so we don't confuse anything with the new components configuration.

michalmac:
@sebhoerl  I just looked into the other 3 PRs. Not sure if they will get merged soon (at least not all of them seem to) and new PRs may pop up in the mean time. So maybe let's merge it now??

sebhoerl:
Done!

PR #393, opened 2018-10-15
Drt mode setup bugfix

vsp-gleich:
Let drt really use the mode set in DrtConfigGroup, set DrtStageActivityType accordingly

### DIFF
--- a/contribs/drt/src/main/java/org/matsim/contrib/drt/routing/DrtMainModeIdentifier.java
+++ b/contribs/drt/src/main/java/org/matsim/contrib/drt/routing/DrtMainModeIdentifier.java
@@ -26,7 +26,7 @@ public class DrtMainModeIdentifier implements MainModeIdentifier{
 	public String identifyMainMode(List<? extends PlanElement> tripElements) {
 		for (PlanElement pe : tripElements) {
 			if (pe instanceof Activity) {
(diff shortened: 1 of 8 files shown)


michalmac:
Thanks, @vsp-gleich !
The only thing I am not sure about is whether we really want to have `drtStageActivityType` inside `Config`. For me, `Config` is something that directly comes from the input file. It is a bit like adding `Scenario` to `Config`. Is there anything stopping us from injecting `DrtStageActivityType` via guice??

jfbischoff:
Good thought, @michalmac . 
> Is there anything stopping us from injecting DrtStageActivityType via guice??
The complicated thing is, that we need the Stage Activity names before the controler is created (namely, in DrtControlerGenerator, where we adapt the config). 

What's your intuition on this?

michalmac:
`DrtStageActivityType` is cheap to build so it is now created locally when needed. Since we have distinct local instances, I added `hashCode()` and `equals()` (see guidelines for `StageActivityTypes`).

Still one issue is unsolved -- see the commented out line: dc347c2

jfbischoff:
Walk Router is used in DRT Routing module, so there is no need to add routing parameters for DRT_walk to the config. Sorry about the confusion @vsp-gleich

vsp-gleich:
Thank you @michalmac and @jfbischoff !

PR #394, opened 2018-10-17
QuadTree performance and memory improvements

mrieser:
Reduce memory consumption of QuadTree by not wrapping every single Leaf in a Node, but store multiple Leafs per Node (known as fan-out in tree data structures). Due to less random memory access, and less object creation, put- and get-performance actually improves measurably, as does the memory overhead imposed by the QuadTree.


sebhoerl:
*Test* (somehow I cannot write comments in another PR right now)

PR #404, opened 2018-10-29
WIP: MATSIM-787

sebhoerl:
Discussion at https://matsim.atlassian.net/browse/MATSIM-787

### DIFF
--- a/matsim/src/main/java/org/matsim/core/controler/Controler.java
+++ b/matsim/src/main/java/org/matsim/core/controler/Controler.java
@@ -110,16 +110,12 @@ public final class Controler implements ControlerI, MatsimServices {
 
 	private com.google.inject.Injector injector;
 	private boolean injectorCreated = false;
+	private List<AbstractModule> modules = new LinkedList<>();
 
 	@Override
 	public IterationStopWatch getStopwatch() {
 		return injector.getInstance(IterationStopWatch.class);
 	}
-
-	// DefaultControlerModule includes submodules. If you want less than what the Controler does
-    // by default, you can leave ControlerDefaultsModule out, look at what it does,
-    // and only include what you want.
-    private List<AbstractModule> modules = Collections.singletonList(new ControlerDefaultsModule());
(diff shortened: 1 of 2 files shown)


matsim/src/main/java/org/matsim/core/controler/Controler.java
```diff
@@ -452,31 +465,51 @@ public void install() {
...
+	}
+	
+	public final void addOverridingModule(AbstractModule module) {
+		checkBeforeInjectorCreated();
+		overridingModules.add(module);
+	}
+
+    public final void setOverridingModule(AbstractModule module) {
+    	checkBeforeInjectorCreated();
+    	overridingModules.clear();
```
tduberne:
```suggestion
    	if (!overridingModules.empty()) throw new IllegalStateException("can only set OverridingModule once");
```

I find that a setter that clears a list of values is confusing. If I was to add 3 modules and then set 1, I am not sure what I would expect, as a user. So I would rather have the setter fail if the list is non-empty.

Of course, if we manage to get rid of `addOverridingModule`, one could then make the setter more "standard"

michalmac: What about this call sequence: `setOverridingModule()` and then `addOverridingModule()`. Should the second one fail? I would expect that probably yes.

Maybe `clearAndAddOverridingModule()` would be clearer (before Thibaut's change).

sebhoerl: I would agree with Thibaut, `clearAndAddOverrridingModule` I find a bit long :)

One could be a bit more radical and just have:

```
void setOverridingModules(AbstractModule... modules);
void setOverridingModule(AbstractModule module);
```

Which both raise an exception if the other has been called before and `setOverridingModules` soon to be removed. This would entail that everybody has to change their list of `addOverridingModule` calls to the one `setOverridingModules`... Maybe we should discuss this on JIRA?

PR #409, opened 2018-11-07
Kaibranch

kainagel:


kainagel:
This is a commit which unfortunately mixes too many things, sorry about that.  They were triggered/moved forward by Mads' work on the bicycle link, and I originally thought they were related; turns out they were not, but my commits where so intertwined that I didn't want to back out and do half the work again.  Sorry.

> Just saw this in the feed. I would say the initial version makes more sense. There is no need to generalize the AgentSource or make it "overridable". If one does not want the standard source, one should remove "PopulationAgentSource" from the active components, and create one's own AgentSource component and activate it though the QSimComponentsConfig.

I reverted the change.  Sebastian and I agreed to talk a bit about which things should be "explicitly bound" and which ones should rather be "named".  Not yet sorted out.

PR #416, opened 2018-11-13
WIP - Emissions contrib feature improvements

joemolloy:
There are two main feature enhancements, which result in an essentially new emissions contrib, but with much broader and consitent behaviour:

1. Handling the full range of pollutants:
- Pollutant types were previously hardcoded in two enums. This meant that the pollutants to be studied couldn't be selected at run time. 
- Now, whichever pollutants are in the HBEFA files will be recorded. They are detected from the input files automatically 
- This required ripping out the WarmPollutant and ColdPollutant enums (as they could only be static) - hence the large number of file changes

2. The contrib will now handle all 4 HBEFA traffic situations
- Previously, only stopandgo and freeflow were handled, with states inbetween being calculated from a weighted average of the two. The HBEFA database infact has 'heavy' and 'saturated' states as well. These are now considered, if they are available in the input files. 
- The old behaviour would no longer be supported. Old versions of the contrib can still be used.
- Numerous tests were added to checkt the new behaviour.


joemolloy:
@amit2011 would you like to have a look at this and see what you think? does it fit with your workflows?

amit2011:
Thanks @joemolloy. Comments explain it well. Unfortunately, I will not be able to look on it until end of next month. Please proceed as per your convenience.

PR #426, opened 2018-11-26
[do not merge] Qsim injection

kainagel:
Sebastian, 

thanks for all the work with the qsim injection.  

Michael Zilske and I today had a closer look, also towards (1) that I want to be able to do this myself, and (2) towards teaching.  Attached is a (in our view relatively small) proposal by the two of us to bring the qsim binding syntax
(a) closer to what we already have in the "controler level" injector, and
(b) closer to guice (which I find easier to teach).  
In the sense of
```
this.addQSimComponentBinding("name").to...
```
Could you please have a look?

Thanks and greetings to Zurich ...


sebhoerl:
Also see this discussion here:
https://github.com/matsim-org/matsim/pull/385

michalmac:
I do not think narrowing allowed annotations to `@Named` only is a good idea.

After introducing multi-mode DRT and taxi (also in a mixed setup), I heavily depend on per-mode bindings. However, instead of `@Named(mode)`, I decided to use `@DvrpMode(mode)`. By doing this we can prevent potential conflicting or, even worse, unexpected bindings.

Additionally it is much easier to navigate through code in IDE (helpful when refactoring dozens of DRT or taxi-dependent projects). This argument is more important for per-mode-type annotations: `@Drt` and `@Taxi`, which are scattered in many places. But so far there is no QSim component registered with such annotations.

kainagel:
(1) Our suggested `addQSimComponentBinding( String label )` just expands to `binder().bind(QSimComponent.class).annotatedWith(Names.named(label))`.  Can't you then just use Guice directly if you want other annotation types?  Is there a technical reason why that would not be possible with our proposal?
(2) The 2nd part of our proposal changes the "instanceof" check where the qsim modules are inserted: The version that MZ and I looked at checked the type of the key, we suggest to check the type of the implementation.  
Maybe I just go forward and refactor the changes on the present branch to see potential problems.

michalmac:
> (1a) Our suggested addQSimComponentBinding( String label ) just expands to binder().bind(QSimComponent.class).annotatedWith(Names.named(label)). Can't you then just use Guice directly if you want other annotation types? Is there a technical reason why that would not be possible with our proposal?

@kainagel , sure we could do this can. But IMO it would be cleaner if we leave the `bindComponent()` methods.

michalmac:
> (2) The 2nd part of our proposal changes the "instanceof" check where the qsim modules are inserted: The version that MZ and I looked at checked the type of the key, we suggest to check the type of the implementation.

This change makes things much cleaner.

michalmac:
One more important comment: At some point we decided to have more than one component with the same annotation. For instance, in DVRP and related, we have `PassengerEngine`, `AgentSource` and many other taxi/Drt-specific components (incl. mobsim listeners) annotated with the same `@DvrpMode(mode)` annotation (see e.g. `DvrpModeQSimModule`). I like this feature very much.

The `addQSimComponentBinding( String label )` method allows only for one component per label, which is fine in many cases. But if we leave only this method, we will remove the feature I mentioned from the "official API" (== hide from users).

kainagel:
I have the following two issues with the `bindComponent` method: (1) It only works for the Xy.class syntax, not for instances or providers.  (2) For the experienced users, I would rather teach them Guice syntax directly.  For me, the methods in `AbstractModule` (taking the older example) are not meant to "shield" the user from using plain Guice syntax directly, but rather to provide help towards which extension points exist. 
However, if you really prefer working with `bindComponent` rather than with `binder().bind(...).to(...)`, then be that so. :-)

michalmac:
@kainagel probably you meant `addComponent`, not `bindComponent`? This one is limited to `Xy.class` and has a call chain with `... .to(...)`

===

If we do not want to have too many methods (doing almost the same), maybe we could leave `bindComponent` only and remove `addComponent`?? But then lots of simple calls will look like this:
```
bindNamedComponent(DefaultTeleportationEngine.class, COMPONENT_NAME).to(DefaultTeleportationEngine.class);
```
instead of this:
```
addNamedComponent(DefaultTeleportationEngine.class, COMPONENT_NAME);
```
Hard to say...

michalmac:
> (2) The 2nd part of our proposal changes the "instanceof" check where the qsim modules are inserted: The version that MZ and I looked at checked the type of the key, we suggest to check the type of the implementation.

This change made me thinking about the following case:
```
DvrpMode dvrpMode = ...
bind(TaxiOptimizer.class).annotatedWith(dvrpMode).toProvider(...).asEagerSingleton();
bind(VrpOptimizer.class).annotatedWith(dvrpMode).to(TaxiOptimizer.class);
```
Before the mentioned change (2), it works fine, but will not work after it, because the taxi optimiser will be processed twice by `QSimProvider`, so will be added twice as a mobsim listener. The `VrpOptimiser` interface does not extend any interfaces, which is why there is no side effect in the original version.

However, if I modify `VrpOptimizer` to extend `QSimComponent`, both bindings will be processed also before change (2), but the second binding will never trigger any of the if-statements.

However, if I modify `VrpOptimizer` to extend `MobsimListener`, both bindings will be processed before and after the change, and in both cases the optimiser will be added twice to the collection of mobsim listeners.

====

Seems that we need first to convert bindings into an ordered set of components (thus getting rid of duplicates) and only then iterate through those components. This should work as long as we assume that bindings are effectively "singletons" (which is a reasonable assumption).

michalmac:
Alternatively, to solve the above issue and to have less automagic, we could use a multibinder and assume that we need to explicitly add a direct QSimComponent binding.
```
multibinder.bind(QSimComponent.class).annotatedWith(...).to(...);
```
In this case, there will be no unexpected side effects like in the example above, where a change in one place of code causes troubles in QSimProvider.

The above example could be coded like this (which is very close to what @kainagel proposes):
```
DvrpMode dvrpMode = ...
bind(TaxiOptimizer.class).annotatedWith(dvrpMode).toProvider(...).asEagerSingleton();
bind(VrpOptimizer.class).annotatedWith(dvrpMode).to(TaxiOptimizer.class);
addQSimComponentBinding(dvrpMode).to(Key.get(VrpOptimizer.class, dvrpMode))
```

Note: I suggest the use of a multibinder, so that more stuff could be bound under one annotation, as is the case with DVRP.

kainagel:
My current intuition is that we might be able to get all we want with a `MultiMap`.  Something like
```
			protected void configure(){
				MapBinder<Annotation, MyInterface> mapBinder = MapBinder.newMapBinder( this.binder(), Annotation.class, MyInterface.class );
				mapBinder.permitDuplicates() ;
				mapBinder.addBinding( Names.named("abc") ).to( MyImpl1.class ) ;
				mapBinder.addBinding(Names.named("abc") ).to( MyImpl2.class ) ;
			}
```
can later be retrieved by something like
```
		Set<Provider<MyInterface>> set = map.get( Names.named("abc") ) ;
		for( Provider<MyInterface> provider : set ){
			provider.get() ;
		}
```
To me this looks like the current `ComponentsRegistry` would then be taken over by that `MultiMap`.  

I just pushed the current status but am far from being done.

michaz:
> 	mapBinder.addBinding(Names.named("abc")).to(MyImpl1.class);

That is too meta, @kainagel. :-) Using annotations as Map keys would probably technically work, but the whole point of using annotations as keys (injection keys, not map keys) is that you can _actually use them as annotations_ when not using the ``bind(..).annotatedWith(Names.named("wurst")`` syntax, but the ``@Provides @Named("wurst")`` syntax for modules.

Not sure how that would translate to the ``Key`` as map key solution.

michaz:
Whaaaaaaaaaat, no, you're right, sorry, this works after all. Just found it in the Javadoc for ``@MapKey``. So many solutions.....

michaz:
I think we can even still change the solution on the other branch to that, easily. Would save the for-loop to find all the annotated multi-binders, but not more I think.

kainagel:
Yes I tried around using Annotation as key for the MapBinder but my intuition was the same as I think you are having: one should either use a MapBinder with String as key, or a MultiBinder with Annotation as "key" (albeit with a somewhat different syntax, at least internally).  So if we want to have Annotations as key, should go with MultiBinder in my intution as well.

PR #439, opened 2018-12-19
Slight change in QSimComponent implementation

michaz:
Okay, this would be my suggestion.

The driver of this proposal was that @kainagel and I wanted to keep the implementation class of a ``QSimComponent`` out of the injection key, mostly in order to be able to e.g.

```
addComponentBindingNamed("something").toInstance(instance);
```

Instead of additionally requiring the class to go in there.

Why was it in there? It was doing two jobs at once,
a) allowing multiple things to go under one name, and 
b) allowing @michalmac to pull individual things out of the injector again, under their concrete implementation class key.

a) is now done by Multibinders. (One per annotation -- I left that part unchanged.)
b) is now done by moving the binding to the implementation class key into @michalmac's own code, doing a chained binding: It's what you would have to do normally anyways: e.g. if I want to add a ``MyEventHandler``, and in addition to binding it to the Set of EventHandlers, I want to get back to it by injecting ``MyEventHandler`` (with annotation or without), I have to add an additional binding to ``MyEventHandler.class`` myself, privately, as well. *) I'm not sure it needs to be a special feature of ``QSimComponent`` that it gets bound to something other than ``QSimComponent`` by design.

*) Those chained bindings work as expected. In particular, for a ``@Singleton``, you still get _one_ instance, not two, as @sebhoerl confirmed by a test.


michaz:
(All tests working (but removed the one for the component registry, because that one's gone.))

michaz:
Same goes for overriding. If you _really want to_ pick and override one certain class out of one certain named component set, you can still allow and do that by first binding your class, then adding a binding of your class to QSimComponent Set. And then overriding the binding to your class.

But maybe that is one level on indirection too many, after all. :-)

michaz:
Btw @sebhoerl I really like how careful you were here. Not spreading this out too much, so that at one point we can still find a "hard" solution with Guice scopes and real, self-invented optional bindings or such. :-)

kainagel:
Looking good.  Thanks a lot.  Some suggestions concerning naming conventions:

(1) Could we please call it addQSimComponentBinding(...)?  Then it would be the same as addPlanStrategyBinding(...), addRoutingModuleBinding(...), etc.

I would then tend to overload that, i.e. addQSimComponentBinding(String), addQSimComponentBinding(Annotation), addQSimComponentBinding(Class<? extends Annotation>).

(2) I think that we should leave addNamedComponent(...) in place and mark it deprecated.  Clearly, you can (and should) inline it in all reachable code, but there is probably sufficient code already out there where you will not reach it (including code by myself), and I don't think that it will hurt to leave it around for another year or so.

PR #449, opened 2019-01-04
clean taxi and drt up from unused annotations

michalmac:


michalmac:
The "github" build on our jenkins stopped working (I have already notified @mrieser )

@jfbischoff, could you maybe set travis as the "required" check for the time being? The issue with jenkins stops me from propagating recent changes in matsim to other projects.

jfbischoff:
Done @michalmac 
I can also restart jenkins tomorrow morning, if required.

michalmac:
Thanks, @jfbischoff .

> I can also restart jenkins tomorrow morning, if required.

Yes, please. There is something wrong with Jenkins. Hope a restart will fix it.

michalmac:
Wasn't aware of that, but we have not been run IT on travis since mid September. So actually my suggestion to temporarily switch to travis was not so good.

BTW. This has happened after adding `-DskipTests` to the test phase script. Given @kainagel 's comment, I guess the intention was to turn off non-IT tests only (???).

But if that is not the case and we do not want to run ITs then the whole CI setup with matrix does not make sense and the build could take only about 15 min.

kainagel:
In principle, it should run the normal tests during the first phase, and the matrix-expanded IT tests during the second phase.

Unfortunately, I could not get travis to remember what was done in the first phase so that the mvn install has to be done in the second phase again, which is totally stupid. 

Yes, and then the intention was to run the IT tests only.  I thought that it was doing that at some point in time, but who knows.

If someone has time and patience to fix all this, please go ahead, I would be grateful.

michalmac:
> Unfortunately, I could not get travis to remember what was done in the first phase so that the mvn install has to be done in the second phase again, which is totally stupid.

Seems that we would need an external storage to reuse the built jars in the other phase: 
https://docs.travis-ci.com/user/build-stages/#data-persistence-between-stages-and-jobs

michalmac:
BTW. Yesterday evening jenkins again stopped executing builds...

PR #462, opened 2019-01-23
Provide the ability to choose between the fraction and average-speed approaches

joemolloy:
The default is the fraction method validated in Hülsmann et al. (2011).

This pull request would return the previous method which took a fraction of freeflow and stop&go to reflect queuing time on links. An approach using all 4 traffic states is still available, but must be enabled in the emissions config with `isUsingFractionalApproach`

The functionality that was restored is (removed in commit [5c80fd ](https://github.com/matsim-org/matsim/commit/5c80fdf9849e995a1adeee9a62d0720e991287fa):
```
double distanceStopGo_km = (linkLength_km * stopGoSpeedFromTable_kmh * (freeFlowSpeed_kmh - averageSpeed_kmh)) / (averageSpeed_kmh * (freeFlowSpeed_kmh - stopGoSpeedFromTable_kmh));
double distanceFreeFlow_km = linkLength_km - distanceStopGo_km;
generatedEmissions = (distanceFreeFlow_km * efFreeFlow_gpkm) + (distanceStopGo_km * efStopGo_gpkm);
```

### DIFF
--- a/contribs/emissions/src/main/java/org/matsim/contrib/emissions/WarmEmissionAnalysisModule.java
+++ b/contribs/emissions/src/main/java/org/matsim/contrib/emissions/WarmEmissionAnalysisModule.java
@@ -72,6 +72,7 @@ public final class WarmEmissionAnalysisModule {
 	private int saturatedCounter = 0;
 	private int heavyFlowCounter = 0;
 	private int stopGoCounter = 0;
+	private int fractionCounter = 0;
 	private int emissionEventCounter = 0;
 	
 	private double kmCounter = 0.0;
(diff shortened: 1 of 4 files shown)


kainagel:
Dear Joe, thanks a lot.  

Could we please use an enum rather than a boolean: 
```
enum EmissionsComputationMethod = {StopAndGoFraction,AverageSpeed} ;
private EmissionsComputationMethod emissionsComputationMethod = EmissionsComputationMethod.StopAndGoFraction ;
```
etc.?  This leaves the option of adding more options later without having to touch the config file again.

joemolloy:
Dear @kainagel , 
of course, good idea. I've implemented that now in a new commit.

PR #473, opened 2019-02-05
fix EV to work on paths with spaces in it.

mrieser:
the new EV contrib failed the weekly builds as it did not properly support running in paths that contain spaces. Fixed.


jfbischoff:
Thanks, Marcel!

PR #476, opened 2019-02-08
make DRT respect controler's createGraphs setting.

mrieser:
Controler has a setting to enable/disable the creation of charts/graphs/plots (for performance reasons and for reducing dependency on X11 on headless systems). Make DRT respect that setting and not output charts according to the setting.


michalmac:
Thanks, @mrieser!

I wasn't aware of this switch. Will apply it as well in taxi and EV...

jfbischoff:
That switch only came to the matsim core some time ago, after a special request.

mrieser:
Might be that the switch came only recently to the config group, but in the Controler itself it's there for [at least 9 years](https://github.com/matsim-org/matsim/blob/fa1bd9927d79b40ba51e0e30168f509d8687e022/matsim/src/main/java/org/matsim/core/controler/Controler.java#L935).
The option also comes handy to speed up tests that do iterations, as each chart-generation takes close to a second.

PR #495, opened 2019-03-09
set QSim insertingWaitingVehiclesBeforeDrivingVehicles = true (default)

michalmac:
Jira discussion: https://matsim.atlassian.net/browse/MATSIM-911


michalmac:
Automatically added "legs starting and ending at the same link now take 0 seconds (instead of 1)" to all commit comments. Actually, it was true only for the first fixes. For the remaining tests, where a bigger scenarios were run, it was not the only cause, but the only one I could easily detect comparing old vs new event files...

PR #497, opened 2019-03-15
move router type check from WithinDayTravelTime to WithinDayTravelTimeModule

michalmac:
within-day TT could be used by other routers (not defined inside the controller config), so checking against the config makes sense only in the module


michalmac:
seems that jenkins is down...

PR #505, opened 2019-03-29
allow to write the facilities without the description

mdziakowski:
the description may cause some problems with special characters and so on and is not really required...


mrieser:
I've applied a proper fix in #510 such that special characters can be used without posing problems when writing/reading facilities.

PR #507, opened 2019-04-03
Vehicles definition v2.0

kt86:
This PullRequest only includes changes in .xsd-Files (mainly the new vehiclesDefinitionv2.0.xsd) in preparation of merging the vehicles branch.
It also includes moving "attributeType" to the more centralized matsimCommon.xsd

### DIFF
--- a/matsim/src/main/resources/dtd/households_v1.0.xsd
+++ b/matsim/src/main/resources/dtd/households_v1.0.xsd
@@ -59,16 +59,6 @@
         </xs:complexContent>
     </xs:complexType>
 
-    <xs:complexType name="attributeType">
-        <xs:simpleContent>
-            <xs:extension base="xs:string">
-                <xs:attribute name="name" type="xs:string"/>
-                <xs:attribute name="class" type="xs:string"/>
-            </xs:extension>
-        </xs:simpleContent>
-    </xs:complexType>
-
-
     <xs:complexType name="incomeType">
         <xs:simpleContent>
             <xs:extension base="xs:decimal">
--- a/matsim/src/main/resources/dtd/laneDefinitions_v2.0.xsd
+++ b/matsim/src/main/resources/dtd/laneDefinitions_v2.0.xsd
@@ -76,10 +76,4 @@ the simulation -->
   </xs:complexContent>
 </xs:complexType>
 
-<xs:complexType name="attributeType">
-  <xs:attribute name="key" type="xs:string"/>
-  <xs:attribute name="clazz" type="xs:string"/>
-  <xs:attribute name="value" type="xs:string"/>
-</xs:complexType>
-
 </xs:schema>
--- a/matsim/src/main/resources/dtd/matsimCommon.xsd
+++ b/matsim/src/main/resources/dtd/matsimCommon.xsd
@@ -137,4 +137,16 @@ xml:lang="en">
   </xs:restriction>
 </xs:simpleType>
 
+
+<xs:complexType name="attributeType">
+	<xs:simpleContent>
+		<xs:extension base="xs:string">
+			<xs:attribute name="name" type="xs:string"/>
+			<xs:attribute name="class" type="xs:string"/>
+		</xs:extension>
+	</xs:simpleContent>
+</xs:complexType>
+
+
+
 </xs:schema>
--- a/matsim/src/main/resources/dtd/vehicleDefinitions_v1.0.xsd
+++ b/matsim/src/main/resources/dtd/vehicleDefinitions_v1.0.xsd
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 <xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema"
-targetNamespace="http://www.matsim.org/files/dtd"
-xmlns="http://www.matsim.org/files/dtd" elementFormDefault="qualified"
-xml:lang="en">
+ targetNamespace="http://www.matsim.org/files/dtd" xmlns="http://www.matsim.org/files/dtd" 
+ elementFormDefault="qualified" 
+ xml:lang="en">
 <!-- Editor: Dominik Grether, VSP, Berlin Institute of Technology -->
 <!-- This xml schema contains xml definitions for household information in the MATSim framework  -->
 
--- a/matsim/src/main/resources/dtd/vehicleDefinitions_v2.0.xsd
+++ b/matsim/src/main/resources/dtd/vehicleDefinitions_v2.0.xsd
@@ -1,0 +1,130 @@
+<?xml version="1.0" encoding="utf-8"?>
+<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema"
+ targetNamespace="http://www.matsim.org/files/dtd" xmlns="http://www.matsim.org/files/dtd"
+ elementFormDefault="qualified" 
+ xml:lang="en">
+  <!-- Editor: Kai Martins-Turner, VSP, Berlin Institute of Technology -->
+  <!-- This xml schema contains xml definitions for vehicles and vehicleTypes in the MATSim framework  -->
+
+  <xs:include schemaLocation="http://www.matsim.org/files/dtd/matsimCommon.xsd"/>
+  <xs:include schemaLocation="http://www.matsim.org/files/dtd/vehicleDefinitionsEnumTypes.xsd"/>
+
+  <xs:element name="vehicleDefinitions">
+    <xs:complexType>
+      <xs:sequence>
+        <xs:element name="vehicleType" type="vehicleTypeType" minOccurs="1" maxOccurs="unbounded" />
+        <xs:element name="vehicle" type="vehicleDefType" minOccurs="0" maxOccurs="unbounded" />
+      </xs:sequence>
+    </xs:complexType>
+  </xs:element>
+  <xs:complexType name="vehicleTypeType">
+    <xs:complexContent>
+      <xs:extension base="matsimObjectType">
+        <xs:sequence>
+          <xs:element name="attributes" minOccurs="0" maxOccurs="1">
+            <xs:complexType>
+              <xs:sequence>
+                <xs:element name="attribute" type="attributeType" minOccurs="0" maxOccurs="unbounded"></xs:element>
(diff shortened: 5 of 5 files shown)


michalmac:
@kt86, I quickly looked at the new vehicle definition.

It's good that "fuel consumption" has been removed (compared to 1.0) - I am thinking here about EVs.

As for `freightCapacity`: I am not sure about the purpose, but probably it's used together with jsprit. I suggest using integers for weight/volume and not to assume they are always provided in tonnes/m3, when the vehicle capacity is a hard constraint in optimisation. (adding N times 0.1 will sometimes give values lower than N*0.1, and sometimes higher --> as a result we may not be able to fit all goods because of the imprecise FP arithmetic).

JWJoubert:
@michalmac I'm in favour of the current use of `double` instead of integers. I can see your argument for using integers, but I think (from a freight perspective) there are stronger counter-use cases. 

@kt86 I also agree with @michalmac that we may want to think more about the units of measure. True, in our current freight work tons and m^3 are common, but we will likely see more and more use cases for much smaller (freight) modes, like cargo-bikes (interest from TUM also), or UberEats and Deliveroo. If ten pizzas can fit in a carrier-bag, are we comfortable with values of ± 0.03m^3 and 0.003 tonnes ? Probably fine?

michalmac:
Re doubles
Yes, `double` should not be a problem if carefully transformed into integers for the optimisation purposes. Then "± 0.03m^3 and 0.003 tonnes" can be transformed into "3 pizzas" or so.

Re units
Isn't it similar to the units of costs (not only in the vehicle specification)? We assume these are monetary costs, but there is no assumption about the actual currency.

kainagel:
I think that we talked about this some weeks/months ago (including Johan, I think).  jsprit, in fact, just has constraint1, constraint2, ..., which is completely free form.  I seem to remember that we said we would hardcode the most common units (weight, volume), and leave other ones to free form (=use attributes).  

I would be equivalently happy with having all of them free form.  Presumably, we would then "type" them as we go via FreightUtils, i.e.

set/getVehicleFreightCapacityInKg(...)
set/getVehicleVolumeCapacityInM3(...)
set/getVehicleCapacityInNumberOfItems(...) // this one potentially as integer

Would that be your preference?  Can someone forsee a problem with this approach?

---

As long as one feeds integer numbers into IEEE arithmetic, there should be no rounding errors, so using double should not be a problem here.  (As long as the numbers don't get too large, and as long as one does not divide them.)

---

Assuming we go with the typed arguments, then according to matsim standard, the units should be kg and m^3.  Not sure if we want to switch this.

mrieser:
v2 has done away with `accessTime`, `egressTime` and `doorOperation`. While they are very transit-specific and likely badly named, they are still quite essential for simulations of transit operators.
How would that information be stored now? Just as custom attributes?

One could argue that `standingRoom` and `seats` are also transit-specific, and they were kept. And `freightCapacity` is freight-specific. According to which criteria were attributes kept or removed in v2.0?

Should `fuelType` really be of a limited enum `FuelTypeType`? In many other places (activities, leg-modes, ...) we deliberately chose generic Strings instead of Enums, because who knows what future FuelTypes there will be. What about compressed-natural-gas (CNG) as fuel type as used in may small indian cars? that's currently not part of the enum. Is this fuelType-attribute actually used somewhere? I would vote to either remove it as well, or at least get rid of the enum and just support any kind of String.

michalmac:
> I think that we talked about this some weeks/months ago (including Johan, I think).  jsprit, in fact, just has constraint1, constraint2, ..., which is completely free form.  I seem to remember that we said we would hardcode the most common units (weight, volume), and leave other ones to free form (=use attributes).

I did not participate in those talks. Now I see the intentions. It makes sense.

> As long as one feeds integer numbers into IEEE arithmetic, there should be no rounding errors, so using double should not be a problem here.  (As long as the numbers don't get too large, and as long as one does not divide them.)

I have never used the freight contrib, so maybe this is not an issue (and actually, it should not when used with jsprit, because jsprit would require casting those values to integers). I just wanted to point out that if a vehicle capacity is 1.4 tonnes and all shipments are 100kg (0.1 tonnes), we will be able to fit only 13 goods into a vehicle (14 would weigh 1.4000000000000001 tonnes).

JWJoubert:
> One could argue that standingRoom and seats are also transit-specific, and they were kept. And freightCapacity is freight-specific. 

All vehicles have some capacity, and it can be persons (seated or standing) or cargo (freight), or _both_. I know some of the City Logistics work in Japan is using portions of passenger trains/trams to also move cargo (mainly smaller units).

> According to which criteria were attributes kept or removed in v2.0?

I agree, maybe this is an agenda point for the Developer Meeting? Because I can see a use case (using differently named variables) for `accessTime` and `egressTime` in freight: either manoeuvring the vehicle, or the `serviceDuration` (which is currently in the `freight` contrib's `CarrierService` and `CarrierShipment`)

kt86:
Dear all, thanks for your feedback and the discussion!

Some of the answers can be found at our starting point here: [https://matsim.atlassian.net/browse/MATSIM-896](MATSIM-896) We were focussing on:

- bringing in the usecasese freight vehicles, electric vehicles, self-driving vehicles to the main vehicleType definition. 
- move specific fields out to Attributes

to have an more general and common usable vehiclesDefinition.

By now, the specific usecases have in some ways own specification or even an own vehicleType definition (see e.g. `CarrierVehicleType` in the `freight `contrib). This somehow brings us some challanges when thinking on multi-use of vehicles: At the same time (e.g. the cargo tram example from @JWJoubert) or during the day: e.g. an AV drives a person, after this it transports some goods from B-C, ...  later it drives a person again.
At the same time we want to make the vehiclesDefintions more open and orientated for the future (AV, electric driving)

So, we tried to figure out, which parameters are useful for vehicles/vehicleType definition in general and which ones are more specific for a certain use. While the first ones remain / or were newly added, the other ones were sorted out to Attributes. 

We put them including getters/setters to the `VehicleUtils`-class (which is still on the `vehicles`-branch and was planned to merge into the master after the *.xsd files are online. (Until that the tests will fail and we cannot merge it in)) :
	DOOR_OPERATION_MODE = "doorOperationMode" ;
	EGRESSTIME = "egressTimeInSecondsPerPerson";
	ACCESSTIME = "accessTimeInSecondsPerPerson";
	FUELCONSUMPTION = "fuelConsumptionLitersPerMeter";
	FREIGHT_CAPACITY_UNITS = "freightCapacityInUnits";  --> Maybe "freightCapacityInNumberOfItems" ??

`fuelConsumption` we sorted out, because it is here used independed of the traffic conditions, independend of the engine state (warm/cold start). It is just an faktor that is mutliplied with the distance driven. We (can) compute the fuelConsuption more accurate with the `emissions` contrib. 

Other ones like the meantioned `freightCapacity` we let in, because even normal cars have a trunk (or in future) some other part of luggage space. But as mentioned from @kainagel it should be also possible to move them to Attributes as well.

Also normal cars have something like number of `seats` and it is an important value when thinking of ride-sharing,.. Therefore we let it in as as own field.

@JWJoubert : I think that the `serviceDuration` is something other than `acessTime` or `egressTime`, because it also depends and therefore includes on the number/volume of the handled goods. So for me this is more related to the 'service' or  'shipment'. But of cause, one can argue, that the person (un)loading the truck also has an `acessTime` or `egressTime`, which can be sorted out from the `serviceDuration`

I hope I could bring you a bit closer to our thoughts. 
Best, Kai (MT)

kainagel:
We have started https://docs.google.com/document/d/1hxVqY4fy1VkNgtG63FGj8eGX63biUfYhcRDLDAXUccs/edit?usp=sharing where each suggestion can be discussed individually.

kainagel:
@michalmac : floating point: The "weight" problem could be removed by denoting it in kilograms instead of in tons.  Unfortunately, this is not as easy for volume.  Maybe in liters?  On the other hand, I would argue that the real world problem is also in real numbers, and the carrier needs to make a decision if an epsilon of possible overweight (including measurement uncertainty) is acceptable or not.

kainagel:
PS: If possible, we would not want to wait until the user meeting.  We have diverging vehicle types between passenger vehicles, freight vehicles, dvrp/taxi.  On the other hand, we have projects where we need consolidated emissions computations.  Having the vehicle types unified is a pre-requisite for unifying the emissions calculations.

neuma:
Since I authored the ComplexTransitStopHandler a decade ago, some background info concerning the "transit vehicle attributes"

> DOOR_OPERATION_MODE = "doorOperationMode" ;
EGRESSTIME = "egressTimeInSecondsPerPerson";
ACCESSTIME = "accessTimeInSecondsPerPerson";

ACCESSTIME is the time needed to load one item into the vehicle, i.e. a person into a bus. EGRESSTIME - same but for unloading an item. Loading and unloading can block each other (serial door operation mode) or can be performed independently (parallel door operation mode).

This whole thing probably needs an overhaul not only for the naming. Nonetheless, the same concept can be used for loading and unloading a delivery truck at the supermarket - basically substitute "person" by "pallet" or "skid" and set the door operation to serial.

Bottom line: Not sure what the common denominator is here.

kainagel:
I have to say that I did not notice until just now that the pull request here only contains the V2 xsd.  The point, presumably, is to get it such that it will be served via www, which is the prerequisite for some other tests to pass.  That is, the whole proposed design is actually in the `vehicles` branch.

mrieser:
There is a system option `matsim.preferLocalDtds` that could be set, e.g. either by specifying `-Dmatsim.preferLocalDtds=true` or by using `System.setProperty("matsim.preferLocalDtds", "true");` directly in the code. I think at least Jenkins should have that property enabled by default in some of the jobs, as does travis. So it should be possible to test everything in one branch.

kainagel:
One thing that I am now (based on some other work) recognizing as missing is `mode`.  Which would mean the network mode, defined by the attributes on the links.  

We have always wondered a bit if we should pull "traveller mode" and "network mode" apart.  This here would be a "network mode".  Should we thus call it `mode`, or `networkMode`?  Any intuitions?

JWJoubert:
> Should we thus call it `mode`, or `networkMode`? Any intuitions?

For me the latter, `networkMode`, would be more descriptive.

PR #518, opened 2019-04-18
network filter manager did not recreate link attributes

dts-ait:
add this and also write a unit test
clearer logging of filtering result


dts-ait:
I accidentally also committed "fix crash in energy calculation", this should not be merged here as its solution is discussed in https://matsim.atlassian.net/browse/MATSIM-924

dts-ait:
it would be great if somebody could merge this (or give feedback) - thanks!

kainagel:
The "automerge" label should have merged it automatically.  Looks more like there is some problem with jenkins, or the jenkins <-> github integration.

jfbischoff:
Maybe the automerge tag does not work if the PR author does not have merge rights?

michalmac:
I do not think this PR should be merged before the cleaning it up (it contains 1 fix commit, 1 commit that should not be merged and then a series of merge commits).

michalmac:
I created a PR from this one: #528 . Please (auto)merge it if it looks okay to you. Thanks!

PR #520, opened 2019-04-24
WIP Set typed event as "oneof" in protocol buffers

tduberne:
This is what was implemented by hand. The docs
(https://developers.google.com/protocol-buffers/docs/proto#updating)
say:

> Changing a single optional value into a member of a new oneof is safe
> and binary compatible. Moving multiple optional fields into a new oneof
> may be safe if you are sure that no code sets more than one at a time.
> Moving any fields into an existing oneof is not safe.

We are in the second case, so that should be safe.
What surprises me is that the code required adaptations, because
checking which field is set is done differently.

As it was now redundant, I also removed the "type" field. Again, the
docs say

> Non-required fields can be removed, as long as the field number is not
> used again in your updated message type. You may want to rename the
> field instead, perhaps adding the prefix "OBSOLETE_", or make the field
> number reserved, so that future users of your .proto can't accidentally
> reuse the number.

So that should be fine.

I am a bit unsure about what type of compatibility it relates to,
however. They speak of "not breaking existing code", but here the
concern would rather be "being able to ingest old format". I understand
compatibility goes in both directions, but I am not sure.

### DIFF
--- a/contribs/protobuf/src/main/java/org/matsim/contrib/protobuf/Event2ProtoEvent.java
+++ b/contribs/protobuf/src/main/java/org/matsim/contrib/protobuf/Event2ProtoEvent.java
@@ -42,7 +42,7 @@ public abstract class Event2ProtoEvent {
 					.setTime(event.getTime())
 					.setLinkId(ProtobufEvents.LinkId.newBuilder().setId(((LinkLeaveEvent) event).getLinkId().toString()))
 					.setVehId(ProtobufEvents.VehicleId.newBuilder().setId(((LinkLeaveEvent) event).getVehicleId().toString()));
-			eb.setType(ProtobufEvents.Event.Type.LinkLeave).setLinkLeave(ll);
+			eb.setLinkLeave(ll);
 		}
 		else {
 			if (event instanceof LinkEnterEvent) {
@@ -50,7 +50,7 @@ public abstract class Event2ProtoEvent {
 						.setTime(event.getTime())
 						.setLinkId(ProtobufEvents.LinkId.newBuilder().setId(((LinkEnterEvent) event).getLinkId().toString()))
 						.setVehId(ProtobufEvents.VehicleId.newBuilder().setId(((LinkEnterEvent) event).getVehicleId().toString()));
-				eb.setType(ProtobufEvents.Event.Type.LinkEnter).setLinkEnter(ll);
+				eb.setLinkEnter(ll);
 			}
 			else {
 				if (event instanceof ActivityEndEvent) {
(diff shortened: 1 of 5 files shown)


kainagel:
Dear Thibaut, 

what triggered these changes?  The commit messages don't say.

In general, this is for interchanging information between matsim (in java) to some external simulation (in cpp).  The code that is actually using it is presumably in https://github.com/matsim-org/hybridsim-it .  I tried to set it up as an example that would be regularly regression tested, and complain when failing, but I think that I did not succeed.

What we _could_ do is treat the protobuf contrib as an example of how to convert certain matsim data structures into protobuf.  It would then need tests similar to our reader and writer tests, i.e. convert something from matsim into proto and then back into matsim and see if the material is still the same.  I am skeptic if that was done ... but also am not sure.

So overall, this is more of a construction site than it should be.  Not sure what to do with it.

Best wishes, Kai

tduberne:
Hi,

what triggered those changes is some work I am doing on getting a usable Python API working for my next MATSim class, including event handling. I am using protocol buffers to keep communication cost between the two VMs to a minimum. I did experiments on a toy simulation, and it seems to work quite nicely. Once the MATSim version is ready, I expect performance of event handling in Python to be similar to performance in Java.

The reason for this change is that using the "oneof" protocol buffer feature:
- improves memory footprint, as the format knows only one of the various options will ever be here and can improve its layout
- encodes this property in the format itself, rather than in the "business" code (see how the tests were simplified)

I could have gone forward and just implemented my own protocol buffer converters, but I figured I would instead try and see if I can use the "official" version (it is distributed on maven, so implicitly supported). As I was not so sure what the status is since Gregor's departure, I put this here for discussion exactly to get the kind of comments you provided.

> What we could do is treat the protobuf contrib as an example of how to convert certain matsim data structures into protobuf. 

I do not have a problem with this, but then this is not a contrib, and clearly not a maven artifact. I personally find it useful to have this deployed as a maven artifact: reimplementing this is trivial, but painful. Having this done somewhere once and for all is nice.

> It would then need tests similar to our reader and writer tests, i.e. convert something from matsim into proto and then back into matsim and see if the material is still the same. I am skeptic if that was done ... but also am not sure.

This repository contains such tests.

> In general, this is for interchanging information between matsim (in java) to some external simulation (in cpp). 

Is it? I looked at Hybridsim, and it does not seems to depend on this module. It does use protocol buffers, but for other things than events. As far as I can see, this module contains experiments to write events to files using protocol buffers, but was most likely never used (the useful methods are package protected...).

tduberne:
Created https://matsim.atlassian.net/browse/MATSIM-925 to discuss the purpose of this repository (hybridsim discussion above)

kainagel:
Looked slightly more at it. The repo hybridsim-it is just there to regularly run the regression test (in principle, probably either not switched on, or not monitored).  So we can ignore it.  Some runnable code is `org.matsim.contrib.hybridsim.run.RunExample`.  You are right, the pom does not depend on `protobuf`.  It still seems to me that there was some code autogeneration in protobuf that needed to run before `hybridsim` became functional.  But I never tested this myself; as said, I wanted to have it runnable from Gregor/MZ before looking into it, but somehow we ran out of time and so I don't know the status.

kainagel:
@tduberne arranged with @rakow to take over the "protobuf" work, and to hopefully also resolve this PR one way or other

rakow:
Can be closed, because it is superseded by #876

PR #522, opened 2019-04-24
agent wakeups etc.

kainagel:
Towards having prebooking for DRT.  We implemented this by having a `TripInfoRequest`, which is modelled after how one requests info from BerlKoenig (Berlin MaaS) or from German Rail.  Then there is now `TripInfo`, which is what comes back.  There is also an `ActivityEngineWithWakeups`, which is a first implementation of, well, an activity engine with wakeups, i.e. agents can schedule a wakeup to call a taxi.  (Note: this should, in the longer run, become more general; there is no reason to restrict this to agents; `MessageEngine` might be the way to go; also put the regular wakup from activities there.) . There is also a `PreplanningEngine`, which handles the resulting behavior.

There is also corresponding code on the dvrp side.  There is also one integration test there (more to come).

There is also code maintenance left and right that came across our way.


michalmac:
Just noticed this was not mentioned in the PR description:

Request rejection in DVRP/DRT handling have changed
from:
"the optimizer sets request.setRejected(true) and emits `PassengerRequestRejectedEvent` & `PersonStuckEvent`"
to:
"the optimizer emits only `PassengerRequestRejectedEvent` and then `PassengerEngine` either emits `PersonStuckEvent` (immediate request) or notifies the preplanning engine about rejections (advance request -> the agent can still react and try other options)

Since the rejection notification channel is not different, as a result, `Request.isRejected()` got removed.

PR #523, opened 2019-05-01
make Route extend Attributable

michalmac:
Instead of using `DrtRoute` (and adding soon `TaxiRoute`, etc.), we (@kainagel and I) decided to use the route types from the matsim core after making them more flexible. Besides fewer classes, this also means no need to supply customised route factories before scenario loading.

### DIFF
--- a/contribs/socnetsim/src/main/java/org/matsim/contrib/socnetsim/jointtrips/population/DriverRoute.java
+++ b/contribs/socnetsim/src/main/java/org/matsim/contrib/socnetsim/jointtrips/population/DriverRoute.java
@@ -32,8 +32,8 @@ import org.matsim.api.core.v01.network.Link;
 import org.matsim.api.core.v01.population.Person;
 import org.matsim.api.core.v01.population.Route;
 import org.matsim.core.population.routes.NetworkRoute;
-import org.matsim.core.population.routes.NetworkRoute;
 import org.matsim.core.population.routes.RouteUtils;
+import org.matsim.utils.objectattributes.attributable.Attributes;
 
 /**
  * @author thibautd
@@ -41,6 +41,7 @@ import org.matsim.core.population.routes.RouteUtils;
 public class DriverRoute implements Route , NetworkRoute {
 	private final NetworkRoute netRoute;
 	private final Set<Id<Person>> passengers = new TreeSet<Id<Person>>();
+	private final Attributes attributes = new Attributes();
 
 	public DriverRoute(final Id<Link> startLink , final Id<Link> endLink) {
 		netRoute = RouteUtils.createLinkNetworkRouteImpl(startLink, endLink);
@@ -205,7 +206,12 @@ public class DriverRoute implements Route , NetworkRoute {
 	public String getRouteType() {
 		return "driver";
 	}
-	
+
+	@Override
+	public Attributes getAttributes() {
+		return this.attributes;
+	}
+
 	@Override
 	public String toString() {
 		return "[DriverRoute: delegate="+netRoute+"; passengers="+passengers+"]";
--- a/contribs/socnetsim/src/main/java/org/matsim/contrib/socnetsim/jointtrips/population/PassengerRoute.java
+++ b/contribs/socnetsim/src/main/java/org/matsim/contrib/socnetsim/jointtrips/population/PassengerRoute.java
@@ -24,6 +24,7 @@ import org.matsim.api.core.v01.network.Link;
 import org.matsim.api.core.v01.population.Person;
 import org.matsim.api.core.v01.population.Route;
 import org.matsim.core.utils.misc.Time;
+import org.matsim.utils.objectattributes.attributable.Attributes;
 
 /**
  * A route for passenger trips.
@@ -35,6 +36,7 @@ public class PassengerRoute implements Route {
 	private Id<Link> startLink = null;
 	private Id<Link> endLink = null;
 	private Id<Person> driver = null;
+	private final Attributes attributes = new Attributes();
 
 	private PassengerRoute() {}
 
@@ -108,6 +110,11 @@ public class PassengerRoute implements Route {
 	@Override
 	public String getRouteType() {
 		return "passenger";
+	}
+
+	@Override
+	public Attributes getAttributes() {
+		return this.attributes;
 	}
 
 	@Override
--- a/contribs/socnetsim/src/main/java/org/matsim/contrib/socnetsim/sharedvehicles/replanning/VehicleOnlyNetworkRoute.java
+++ b/contribs/socnetsim/src/main/java/org/matsim/contrib/socnetsim/sharedvehicles/replanning/VehicleOnlyNetworkRoute.java
@@ -24,6 +24,7 @@ import java.util.List;
 import org.matsim.api.core.v01.Id;
 import org.matsim.api.core.v01.network.Link;
 import org.matsim.core.population.routes.NetworkRoute;
+import org.matsim.utils.objectattributes.attributable.Attributes;
 
 /**
  * to put vehicle Id in legs without route.
@@ -132,6 +133,11 @@ class VehicleOnlyNetworkRoute implements NetworkRoute {
 
 	@Override
 	public VehicleOnlyNetworkRoute clone() {
+		throw new UnsupportedOperationException();
+	}
+
+	@Override
+	public Attributes getAttributes() {
 		throw new UnsupportedOperationException();
 	}
 }
--- a/matsim/src/main/java/org/matsim/api/core/v01/population/Route.java
+++ b/matsim/src/main/java/org/matsim/api/core/v01/population/Route.java
@@ -23,13 +23,13 @@ package org.matsim.api.core.v01.population;
 import org.matsim.api.core.v01.Id;
 import org.matsim.api.core.v01.network.Link;
 import org.matsim.core.api.internal.MatsimPopulationObject;
-import org.matsim.core.scenario.Lockable;
+import org.matsim.utils.objectattributes.attributable.Attributable;
 
 /**
  * @author nagel
  *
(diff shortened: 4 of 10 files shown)


mrieser:
Thanks for putting this up for review.
Just to clarify that I understand the change correctly:

Instead of having different route classes, each with a different route type (e.g. (Link)NetworkRoute, ExperimentalTransitRoute, ...) where the codes usually works with `if (route instanceof MyRoute) {...`, basically all routes could now be of GenericRouteImpl with different types and just put whatever data they need to store in attributes? And the code would more work like `if (route.getRouteType().equals("myRouteType") {...`?

Pro:
- Easier storing of custom data, no need to develop a custom serialization format
- less custom classes
- easier loading of custom route types from population files

Con:
- less type safety, as code would compile but attributes at runtime could be missing if it's of a wrong route type. This might result in crashes at unexpected places, instead of at the location of the current instanceof of cast expressions. (On the other hand: one could still use custom classes along with the attributes to store data, to have the flexibility of storing data and still have type safety).
- accessing attributes might be slower as they are stored in Attributes that requires a more complex lookup, and potentially also use more memory.

neuma:
@michalmac How do you plan to keep track of all the magic attributes - say a DRT route might require compared to a regular transit route. Personally, I find it difficult to track changes in the code if there is only one never changing route attribute container.

kainagel:
This change was "nudged" by me, so let me comment.
@mrieser : We have problems with route types that are not in the matsim main distribution, since one needs to register them manually when scenario is read outside Controler, which I find rather awkward.  So I was suggesting to use GeneralRoute plus attributes until the design is sorted out.  We can still make it type-safe afterwards.  Alternative would be to put `DrtRoute` into matsim main.  I would also be ok with this, but it may not yet be fully stable (e.g. we are not sure if the direct distance/travel time should be kept here, or rather somewhere else).
@neuma : The route is per leg, so that should address the comment by @neuma . Or am I misunderstanding something?

sebhoerl:
What would be the downside of having multiple Route types? My only issue with the current system is that anything falls back to `GenericRouteImpl` when loading a population. If somebody forgets to add the right factories, this usually leads to confusion. (Or let's say, I load a population, do something unrelated with it, write it out again, and magically all routes have become "generic"). The `Attributable` approach mitigates the problem (because we don't need a special serializer), but if that is the intention of the PR, maybe an alternative would just be to require route types to be registered explicitly (i.e. throw an error if an unknown route type is encountered)?

michalmac:
> maybe an alternative would just be to require route types to be registered explicitly (i.e. throw an error if an unknown route type is encountered)?

👍  I fully agree - explicit type->factory mappings should be required.

===

As for adding attributes, the idea was not to replace the existing route types. It's more about addressing the following issues:
- the need to pre-register route factories (well, arguable)
- DVRP-based dynamic modes may require different attributes depending on a specific setup (incl. the applied VRP algorithm). Currently, for DRT these are: `maxWaitTime` (maybe could go to `Leg`, which is already attributable) and `directRideTime`. But the set of attributes is not super stable.
- I could imagine adding calculated energy consumption to routes for EVs. Creating subclasses of (one or more) existing route classes to include only consumption may not be optimal

Yes, there is a risk that attributes could be misused leading to less clean/stable design or higher computational burden, as pointed by @mrieser and @neuma . Since I do not work on day-to-day replanning I have no experience to judge it, but it seems that almost everything in the population file is "attributable". Have there been any problems with this design so far?

mrieser:
Thanks for the feedback.
In principal, I can live with the proposed changes (and thus will accept the PR). Just some additional notes as my opinion was requested:

> since one needs to register them manually when scenario is read outside Controler, which I find rather awkward

For me, that's a feature. If I want to work with a Scenario, I want to know what kind of routes I'm going to expect, so I better be surprised when trying to load the data, than later on when working with the data. But I can see that this is a personal preference. And I see that there are rare use cases where this information is not available and one still want's to parse a population file (although an explicit mapping to GenericRouteImpl could solve this).

> We can still make it type-safe afterwards. 

Yeah, "we can make .... afterwards" is something we did a lot in MATSim, and then often didn't do it, which left us with a sometimes confusing code base because there are often different ways to achieve the same. I understand the reasoning, but am not convinced it's best for the users.

> But the set of attributes is not super stable.

For me, that would be even more a reason to use fixed types instead of Attributes, because then I clearly see in the code at compile time where an attribute is used, and I see when trying to load the data if the attributes are there. With the proposed solution, you basically can write out a population file with a "drt route" which has some attributes, and in a month you read it in again but the code will crash during the simulation because some attributes were renamed, new attributes are required or whatever. Basically you're missing a version information when you modify your route types, which I would not consider very stable (that's the reason why the serialization of a TransitRoute starts with `PT1===`, that's a version number encoded right there).

I really like the idea of having attributes in routes, as this will make storing custom route types much more easy than when everybody needs to try to come up with its own serialization format as I did 10 years ago with the transit route. (It would actually be cool to implement a custom attributes container that, instead of storing the data in the generic lookup map, would just add the attributes to actual class members, similar as it is done in config groups, and throw exceptions when unsupported attributes are requested. Then at run-time it would still be type-safe and fast, but make use of the better serialization technology to save it.)

But I fear the misuse potential it provides. But I guess that's something everybody estimates differently.

neuma:
@kainagel I meant exactly what @mrieser answered quite nicely to @michalmac super stable comment. Thanks.

> > But the set of attributes is not super stable.
> 
> For me, that would be even more a reason to use fixed types instead of Attributes, because then I clearly see in the code at compile time where an attribute is used, and I see when trying to load the data if the attributes are there. With the proposed solution, you basically can write out a population file with a "drt route" which has some attributes, and in a month you read it in again but the code will crash during the simulation because some attributes were renamed, new attributes are required or whatever. Basically you're missing a version information when you modify your route types, which I would not consider very stable (that's the reason why the serialization of a TransitRoute starts with `PT1===`, that's a version number encoded right there).

Some additional remarks:

From the user's perspective, I would want a proper documentation of the required attributes and a migration path once new attributes are introduced, or old ones are dropped. I would also like to safeguard against the case that the new code runs without any errors but produces different results - say - because an old set of attributes is provided and some of them are ignored or newly introduced ones silently fall back to some defaults. Don't know if that applies here right now. Though, it will eventually happen. And it is a nightmare to debug.

kainagel:
We could, if that is preferred, add `DrtRoute` into matsim-main.  If so, could you please clearly vote for that option?  Also, if so, would you rather have type-safe arguments that change from time to time, or have the more experimental changes via attributes.

I (strongly) suggest to use attributes that are standard in a contrib via `DrtUtils.get/putXxx(...)`.  Then this is at least semi-typesafe.

If we have situations that we did not make typesafe "later", could we list them somewhere in jira?  I can't think of an example right now.  

I agree with failing when encountering route types that are not registered.

michalmac:
Thanks for lots of valuable input.

> For me, that would be even more a reason to use fixed types instead of Attributes, because then I clearly see in the code at compile time where an attribute is used, and I see when trying to load the data if the attributes are there. 

Good point. Interestingly, the initial motivation of simplifying the process of setting up DRT etc. for end users, may not actually be worth it (and the more experienced end users may actually prefer type-safety and code-navigability over a slightly simplified setup).

After thinking it over again, my suggestion is to:
 - keep things the way they are now (close the issue, but keep the branch in case we change the decision)
 - create a Jira issue re failing for unregistered route types
 - wait and see how the DRT route structure evolves after adding pre-booking or new constraints/objectives to the optimisation algorithm etc. and then maybe approach this issue again

tduberne:
Sorry for being late at the party. Here are my thoughts about the general idea:
- I am generally strongly in favor of typed XYZ. Java does not provide great support for this, but what we do is actually very similar to Scala's "case classes": it is actually so useful some languages build this kind of "there are several types of XYZ with their own set of fields" directly into the language. And when in doubt,I generally prefer trying to be close to Scala rather than Javascript ;-)
- I agree with Marcel, that in general the "let's start dirty and clean it later" does not work. Once you get something in the open, fixing it becomes hard. Just look at pt interactions. It was already like this with the monorepo, and becomes it much more now that the community becomes less and less centralized. Static utils classes might help, but who knows how users might abuse them.
- That being said, the idea of having attributable routes is not necessarily bad. There might be usecases where one wants to attach additional, scenario-specific things to a route. This is just again the old trade-off flexibility vs memory consumption. `Attributes` is designed to be light on memory, but even when empty, it still takes a few bytes per instance, which can add up to a few Gb for a big scenario with millions of agents with 5 plans in memory. But I have ideas on how to solve this if this becomes a problem, so I would favor flexibility over performance in this particular case.

Now to contribute to the issue of automatic conversion: I had similar problems in the past, not only with routes, but also with events and object attributes. I agree with Kay that having to register the right converters in my analysis script felt awkward, making it much more scenario specific as could be.

The thing is, our "everything is explicit" approach is not the only one. Spring Boot, for instance, is much more in a "everything is implicit" approach. Basically, every time it needs something (let's say a route serializer), it will search on the classpath to see if it can find an object with the required specification, and (I hope) fail if it finds more than one. This makes working with it very often feel like black magic, which is bad, but maybe in a few every limited cases, it might make sense. For instance: if I have the DVRT extension on the classpath, dvrt routes are understood, otherwise reading fails. So I can use the same pre-/post-processing script in various applications with various extensions, rather than re-writing a new one to change only the registered converters.

I am not necessarily suggesting to go this way, and I honestly do not know how to implement this right for our case, but I thought it might be good to have this option on the table. For the moment, given our architecture, I still favor explicit registering of converters. You can still make your analysis/pre-processing/whatever re-usable enough by separating the heavy lifting and the "call script". 

Actual code review comes later today.

mrieser:
Closing this PR based on @michalmac 's comment above:
> keep things the way they are now (close the issue, but keep the branch in case we change the decision)

PR #525, opened 2019-05-03
Config validation

michalmac:
Use bean validation for simple config group validation instead of longish (often out-of-date and partially missing) sequences of 'if' statements in config checkers.

I moved (and enhanced a bit) this functionality from DVRP, the idea is described here: #201

Additionally, I added annotations to fields in `GlobalConfigGroup` and `QSimConfigGroup`, e.g. `numberOfThreads` is required to be positive by annotating it with `@Positive`:
```
	@Positive
	private int numberOfThreads = 2;
```

So if there are some incorrect values in config (e.g. `numberOfThreads` set to 0), we should get a message in the following format during config validation:
```
javax.validation.ConstraintViolationException: 3 error(s) found in the config:
1) org.matsim.core.config.groups.GlobalConfigGroup(name=global).numberOfThreads: must be greater than 0
2) org.matsim.core.config.groups.QSimConfigGroup(name=qsim).flowCapFactor: must be greater than 0
3) org.matsim.core.config.groups.QSimConfigGroup(name=qsim).snapshotPeriod: must be greater than or equal to 0
```

Everything is backward compatible as long as there are no annotations added to a config group - the behaviour will be exactly as it was before. Once there are annotations added, some errors will be caught earlier then before.


michalmac:
A small update: `QSimConfigGroup.numberOfThreads` can actually be set to 0 (changed annotation from `@Positive` to `@PositiveOrZero`)

tduberne:
That looks really nice, thank you! I always found that consistency checking error-prone, and that makes it much, much better.

Just a small question from looking at the diff: do we really need all these new dependencies? I only see import of `javax.validate`. I now of some `javax` extensions just providing the interface and expecting something else to provide the implementation, but 5 new dependencies for such a small feature sounds rather extreme. I would argue that if one can get rid of one or two of them at the expense of a few additional lines of code, we should do it.

michalmac:
Thanks for looking into this. Yes, it is 5 new dependencies, 4 related to the new feature and the last one is for unit testing.

Here is a short description:
1. `validation-api` - contains the definitions of annotations (so this is what actually is used in code)
2. `hibernate-validator` - contains the actual implementation of validators (so it's a must)
3. `hibernate-validator-annotation-processor` - it is used during compilation to check the annotation errors (before runtime); can be skipped or moved to `plugins` in this (and other?!) `pom.xml`s (https://docs.jboss.org/hibernate/validator/5.0/reference/en-US/html/validator-annotation-processor.html)
4. `javax.el` - used for message interpolation and cannot be skipped, but could be replaced by some other implementation of message interpolation (https://docs.jboss.org/hibernate/validator/5.1/reference/en-US/html/chapter-message-interpolation.html#chapter-message-interpolation)
5. assertj-core - this is only a test dependency (so not really harmful) unrelated to this feature, but super helpful for writing unit tests (I would call it a better/modern version of hamcrest, which we also have as a test dependency)

So yes, I agree, it looks like quite a lot. However, all dependencies (except for the 2nd) have no transitive dependencies, so are relatively thin (the 2nd one brings 3 additional ones). So these are not like GeoTools dependencies, which are very heavy.

jfbischoff:
Mmhm,
presumably part of this PR: Please let's not check that messages contain exactly a certain String
In BeanValidationConfigConsistencyCheckerTest there's something like
	```
Assertions.assertThatThrownBy(() -> new BeanValidationConfigConsistencyChecker().checkConsistency(config))
					.isExactlyInstanceOf(ConstraintViolationException.class)
					.hasMessageStartingWith("3 error(s) found in the config:")
					.hasMessageContaining(
							") org.matsim.core.config.groups.GlobalConfigGroup(name=global).numberOfThreads: must be greater than or equal to 0")
					.hasMessageContaining(
							") org.matsim.core.config.groups.QSimConfigGroup(name=qsim).flowCapFactor: must be greater than 0")
					.hasMessageContaining(
							") org.matsim.core.config.groups.QSimConfigGroup(name=qsim).snapshotPeriod: must be greater than or equal to 0");
		}
	}
```

This will fail on any other Locale than English, as @kt86  just found out.

michalmac:
Thanks. Did not know the message interpolation is multi-lingual. Will adapt the test.

PR #528, opened 2019-05-08
network filter manager did not recreate link attributes

michalmac:
This is a cleaned up PR #518 (without the commit relating to EV)


michalmac:
Merged. Thanks @dts-ait for your bugfix!

PR #530, opened 2019-05-11
adapt BeanValidationConfigConsistencyCheckerTest for message internationalisation

michalmac:
to address https://github.com/matsim-org/matsim/pull/525#issuecomment-490868812


kt86:
thanks @michalmac for adapting this. Works fine with my Locale.

PR #531, opened 2019-05-15
Ev instance in energy consumption

jfbischoff:
In many circumstances, the consumption of a vehicle may depend on more things than Link and travelTime or other, static (i.e., known at time of creation) vehicle information, but on the actual State of Charge of the Vehicle. Thus, I'm adding a setter for it. Not exactly the nicest solution, but perhaps comparable to Plan, which also has a getPerson. 
Do you have any objections, @michalmac ?


michalmac:
I am aware of this problem. The energyConsumption-electricVehicle coupling is has not been fully sorted out yet. But well, let's try to do it now :-)

An alternative to the bi-di coupling could be passing `ElectricVehicle` as an argument to `calcEnergyConsumption()` instead.

But if we want to stick to the bi-di coupling (which makes sense for me), my suggestion is as follows:

Instead of exposing the setter method in the interface (actually in 2 interfaces, because there is an `AuxEnergyConsumption` too), I would set it all up inside the `ElectricVehicleImpl` class using a static factory method that:
1. creates an EV
2. creates both consumption models for this EV (the EV is passed to the factory as an argument, and the factory decides if the consumption model needs to remember the reference to the EV)
3. sets both consumption models to the EV
4. returns the fully initialised EV

This should be also more consistent with the existing interfaces, where we have `ElectricVehicle.getXxxEnergyConsumption()`, but not `XxxEnergyConsumption.getElectricVehicle()`

jfbischoff:
This makes sense  - adding it to the Factory was my first thought, but I didn't think of a static factory method. I will go forward with implementing.

michalmac:
There is still one problem though with `XxxEnergyConsumption.Factory`ies - they accept `ElectricVehicleSpecification` as an argument. And actually this makes sense for energy consumption calculations done during replanning, where we do not have an EV (which only lives in simulation).

So for replanning we would need a different consumption model that takes all possible inputs as a (possibly long?) list of arguments... or we create fake EVs during replanning (and update them accordingly as if they were inside simulation...). I first need to dive into the code and think a bit about it...

jfbischoff:
Yes, I'm aware of this. For the time being, I would opt for the second (creating a fake vehicle).

michalmac:
> Yes, I'm aware of this. For the time being, I would opt for the second (creating a fake vehicle).

👍

michalmac:
Thanks @jfbischoff

PR #534, opened 2019-05-17
Misc ev

jfbischoff:


jfbischoff:
after discussing with @michalmac , updated DriveEnergyConsumption  interface.

PR #543, opened 2019-06-01
remove the separate person attributes 

kainagel:
This relates to https://matsim.atlassian.net/browse/MATSIM-726 .  

Setting the person attribute file to some value will make the scenario loader integrate the material in the existing population, and then abort.  The recommended behavior thus would be to use that newly generated population file, without the separate person attribute file, and start over.  

There is a switch that you can insist on using the separate person attribute file.  The material is then integrated into the person via Attributable anyways.  Direct calls to `scenario.getPersonAttributes().get/put/...` lead to an exception; replacing those direct call by a static method is possible.

The tests pass, but I don't know how many other cases there are in derived code.

PS: If someone is wondering why I am spending time on this: I encounter this every year in teaching, and explaining all these different methods to achieve similar things is a waste of time. And one cannot get away with not teaching it since, for example, the subpopulations stuff up to now depends on the separate person attribute file.


matsim/src/main/java/org/matsim/core/population/PersonAttributes.java
```diff
@@ -0,0 +1,41 @@
+package org.matsim.core.population;
```
tduberne: I am not sure about that. If any call on it results in an exception anyways, I would rather directly remove the ObjectAttributes getter from the Population, or at least have it throw an exception, and add a package-private getter for PopulationUtils. I will try a proposal in that direction.

matsim/src/main/java/org/matsim/core/population/PopulationUtils.java
```diff
@@ -1093,4 +1080,49 @@ public static boolean comparePopulations( Population population1, Population pop
 		boolean result = PopulationUtils.equalPopulation( population1, population2 );
 		return result ;
 	}
+
+	// ---
+
+	public static Object getPersonAttribute( HasPlansAndId person, String key, Population population ) {
```
tduberne: Why HasPlanAndId, when all other methods accept a `Person`? Shouldn't then all other methods accept a `HasPlansAndId` as well?

matsim/src/main/java/org/matsim/core/population/PopulationUtils.java
```diff
@@ -1093,4 +1080,49 @@ public static boolean comparePopulations( Population population1, Population pop
...
+			return result1 ;
+		}
+		if ( Objects.equals( result1, result2 ) ) {
+			return result1 ;
+		}
+		throw new RuntimeException("inconsistent object attributes state; don't know what to do.  Try using PopulationUtils" +
+							     ".get/put/...PersonAttribute instead of population.getPersonAttributes... .. kai, " +
+							     "mar'19") ;
+	}
+	@Deprecated  // this command is a bit dangerous, since it might clear someone else's attributes.  Maybe best don't use.  kai, may'19
```
tduberne: What is meant here by "someone else's attributes"? Of another `Person`? Why?

matsim/src/main/java/org/matsim/core/population/PopulationUtils.java
```diff
@@ -1093,4 +1080,49 @@ public static boolean comparePopulations( Population population1, Population pop
...
+		if( result1==null ) {
+			return result2 ;
+		}
+		if ( result2==null ) {
+			return result1 ;
+		}
+		if ( Objects.equals( result1, result2 ) ) {
+			return result1 ;
+		}
+		throw new RuntimeException("inconsistent object attributes state; don't know what to do.  Try using PopulationUtils" +
```
tduberne: I do not understand that message. Didn't the user do exactly what the message is asking her to do?

matsim/src/main/java/org/matsim/core/scenario/ScenarioLoaderImpl.java
```diff
@@ -196,9 +200,50 @@ private void loadPopulation() {
...
+								 "are not in the population.  This is not necessarily a bug so we will continue, but note that such material " +
+								 "will no longer be contained in the output_* files.  The material that is still there will follow.  kai, " +
+								 "jun'19" ;
+				log.warn( message ) ;
+				log.warn( personAttributes.toString() ) ;
+			}
+
+			final String outputDirectory = this.config.controler().getOutputDirectory();
+			final File outDir = new File( outputDirectory );
+			if ( outDir.exists() && outDir.canWrite() ){
```
tduberne: I do not like this. I would never expect a reader to have this kind of side-effect. In particular, this will ignore the setting that says "crash immediately if output directory exists".

For most users, taking ITERS/it.0/0.plans.xml.gz should be good enough.

tduberne:
I just pushed two commits:
- moving the constants for attribute xml tag names outside of the Attributable
- actually removing completely the person attributes container

all test passed on my machine. Was there any specific reason why you still did access  attributes in the ObjectAttributes container from PopulationUtils? If there is a good reason, we can just revert my commit

matsim/src/main/java/org/matsim/core/network/io/NetworkReaderMatsimV2.java
```diff
@@ -96,8 +95,8 @@ public void startTag(final String name, final Attributes atts, final Stack<Strin
 			case LINKS:
 				startLinks(atts);
 				break;
-			case ATTRIBUTES:
-			case ATTRIBUTE:
+			case Attributable.ATTRIBUTES:
```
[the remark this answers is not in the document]
kainagel: Not a decision I took; I just centralized the `ATTRIBUTES` constant.  I can't say why I also removed the `ATTRIBUTE` tag.

matsim/src/main/java/org/matsim/core/population/PopulationUtils.java
kainagel: Well, if you are done with your personAttributes, and clear them by using this method, then you will (evidently) also clear everybody else's person attributes.  Which means that we may have two codes that work separately, but not when combined.  Thus I think we could decide to not offer such a method.

matsim/src/main/java/org/matsim/core/population/PopulationUtils.java
kainagel: What I wanted to say is that users should in particular use `PopulationUtils.put...` to define attributes.  That static method should then take care of keeping things consistent.  So maybe the message should be "make sure you have used PopulationUtils.put(...) to define person attributes".

matsim/src/main/java/org/matsim/core/population/PersonAttributes.java
kainagel: I think the main reason for not doing that is that it throws the exception at the wrong place.  There is a lot of code that first calls `population.getPersonAttributes()`, then passes the result around, and only much later calls `personAttributes.put/get/...`.  But it is the latter place where one needs to replace it with the static method.  But I have no strong opinion here.

matsim/src/main/java/org/matsim/core/population/PopulationUtils.java
kainagel: There is actually a comment in `HasPlansAndId`:
https://github.com/matsim-org/matsim/blob/7ca3b868329c137060e8a0d245b4c2a7f451354b/matsim/src/main/java/org/matsim/api/core/v01/population/HasPlansAndId.java#L8-L14
The issue is that the main reason for the existence of `HasPlansAndId` is the strategy manager, and the strategy manager is also used by the carriers (and should also be used by some other things, such as paratransit).  And so it is _not_ only a person that is passed into it.

Now evidently we could decide that the subpop functionality is only there for the persons, but I don't see why.

matsim/src/main/java/org/matsim/core/scenario/ScenarioLoaderImpl.java
kainagel: Only this: `ITERS/it.0/0.plans.xml.gz` is already routed.  So it is a totally different animal.  But I don't really care; I myself won't have any difficulties with this.

kainagel:
Can't give a specific reply to your last comment, so it comes here.  80% of this was probably historical: I started by hedging against both approaches in the static methods.  And I ended with resolving already in the scenario loader.  If you take your argument one step further, one could now as well inline the static methods and then remove them.  (And give a good error message when someone calls `population.getPersonAttributes()`.)

PR #547, opened 2019-06-05
Cleanup aux discharging

michalmac:
Major EV-related changes:
- rewrite `AuxDischargingHandler`, use `VehicleProvider` to filter EVs that consume energy during certain activities
- add linkId to `AuxEnergyConsumption` -> no need for no-arg `TemperatureService.getCurrentTemperature()`
- adapt `OhdeSlaskiAuxEnergyConsumption` to use `TemperatureSevice`
- remove decision on including `AuxEnergyConsumption` to `DriveEnergyConsumption` from EV config (this is vehicle-specific and should be handled programmatically using `XxxEnergyConsumption.Factory`s)

Major non-EV changes:
- add `DvrpMode` "auto-registration"
- add `DvrpVehicleLookup` for a global lookup (many Dvrp fleets vs. one EV fleet)


michalmac:
Despite many changes in interfaces, bindings etc., the only **backward incompatible change** from the user perspective is the removal of `auxDischargingSimulation` from the EV config group.

PR #549, opened 2019-06-07
added constructor to MutableScenario

carloscaldas:


kainagel:
Looks like this was not merged.  I would certainly be against a merge that refers, in pom.xml, to a non-matsim version.  

In terms of design, the implementation classes (e.g. `MutableScenario`) are deliberately not accessible, so please don't change that.  There is `createMutableScenario` in `ScenarioUtils`, which should fulfil the necessary purpose.  If that is not sufficient, please let us know, and we can discuss adding another creational method in `ScenarioUtils`.  Thanks a lot.

kainagel:
PS: Am just seeing what you wanted to do.  Not sure why you need that.  The implementations of the containers inside scenario are not very heavyweight, so there is no problem leaving them empty.  You can also just replace them.  And if that is not enough, you could just program your own implementation of `Scenario` behind the interface.

PR #552, opened 2019-06-09
Move battery charging from charger to electric vehicle

michalmac:
The goal: Move battery charging model from `Charger` to `ElectricVehicle` - as it mostly depends on the vehicle (battery SOC, age and temperature, charging profile, on-board charger, ...)

Changes:
- extract battery charging model from `ChargingStrategy`
- battery charging can be modelled at two time horizons: `ChargingPower` calculates instantaneous charging power (enough for simulating charging process) and `BatteryCharging` (a sub interface of the former) calculates also the charging time necessary to reach a given SOC level (necessary for electric VRP etc.)


michalmac:
Also added validation to the static factory methods in `Charger`, `DvrpVehicle` and `ElectricVehicle` for fail-fast upon invalid input data -- related to matsim-org/matsim-code-examples#141 (and similar issues reported in the past)

PR #558, opened 2019-06-11
minor change in the command line functionality

ikaddoura:
allow for more than one hierarchy level in the parameter sets, allow for
the (quite common) case subpopulation=null in the scoring parameters


ikaddoura:
@sebhoerl what do you think?

kainagel:
@tduberne : Not sure if this pull request interacts but I remember that we wanted to get rid of the alternative one vs two level deep hierarchies and just allow the more comprehensive one.

sebhoerl:
I think for now it is fine. For the future, I could have a look to build a proper Lexer + Parser for the command line arguments, so this stays consistent. (For instance, it would also good to properly understand strings, e.g.

    --config:abc.asdf[attribute="my value with spaces"].xyz = 5

Of course, if nested `parameterset`s would not be allowed by then anymore, this would make things a bit easier :)

PR #560, opened 2019-06-13
Vehicle handler

sebhoerl:
This PR adds a new (optional in the sense of `LinkSpeedCalculator`) component to the Netsim. Basically, it represents a condition that decides whether a vehicle is actually allowed to arrive at a link if it `isWantingToArrive`. In the default case, the handler will always return `true`, which replicates the current behaviour that there is no restriction on where and when vehicles can arrive on a link:

```java
public interface VehicleHandler {
   void handleVehicleDeparture(QVehicle vehicle, Link link);
   boolean handleVehicleArrival(QVehicle vehicle, Link link);
   void handleInitialVehicleArrival(QVehicle vehicle, Link link);
}
```

With this handler it is now possible to define additional dynamics: One can count the number of vehicles that have already entered the link and block once a certain capacity limit is reached. The interface is quite versatile, in a current project we, for instance, have defined two different capacities for "pickup and dropoff" and for "parking" for DVRP vehicles. 

The important point is that there is backspill when a dropoff point is already crowded, for example. While it was possible before to count vehicles and maybe artificially add delays in the dispatching of vehicles, this is new now.

Later on, I can add examples for DVRP.

### DIFF
--- a/contribs/hybridsim/src/main/java/org/matsim/core/mobsim/qsim/qnetsimengine/HybridNetworkFactory.java
+++ b/contribs/hybridsim/src/main/java/org/matsim/core/mobsim/qsim/qnetsimengine/HybridNetworkFactory.java
@@ -33,6 +33,8 @@ import org.matsim.core.mobsim.qsim.interfaces.AgentCounter;
 import org.matsim.core.mobsim.qsim.qnetsimengine.QNetsimEngine.NetsimInternalInterface;
 import org.matsim.core.mobsim.qsim.qnetsimengine.linkspeedcalculator.DefaultLinkSpeedCalculator;
 import org.matsim.core.mobsim.qsim.qnetsimengine.linkspeedcalculator.LinkSpeedCalculator;
+import org.matsim.core.mobsim.qsim.qnetsimengine.vehicle_handler.DefaultVehicleHandler;
+import org.matsim.core.mobsim.qsim.qnetsimengine.vehicle_handler.VehicleHandler;
 import org.matsim.vis.snapshotwriters.SnapshotLinkWidthCalculator;
 
 public final class HybridNetworkFactory implements QNetworkFactory {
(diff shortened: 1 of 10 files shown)


sebhoerl:
One thing I did not think about is lanes (because I also don't know a lot about the implementation). Nevertheless, the way I understand it is that the Lane is asking the Link to `letVehicleArrive`. So if in this case the Link is blocking now (because of capacity constraints e.g.) all lanes will be blocked? But I guess this would open a whole new discussion, because then it would be important on which Lane vehicles can arrive at all. Or is this considered somehow in the current state?

mrieser:
Just out of curiosity: What happens if the handler returns `false` and the vehicle is not allowed to arrive? The route is likely at the end, what does the QSim do then? Or how can an alternative action for what happens afterwards be defined? Would that allow for implementing e.g. parking-search?

sebhoerl:
Currently, it will just wait until the handler eventually returns `true`. So the idea here is really just to have the simple "Netsim side" covered. On the other hand, with our taxis we would then have a counter at a separate place, which knows the capacity and would then dispatch vehicles in a way that they don't have to queue (while for pickup/dropoff they *will* queue). 

In `QueueWithBuffer` the call to `letVehicleArrive` is done after the vehicle says `isWantingToArrive`. If it cannot arrive, processing of the current *queue* is stopped. In the next time step, the *queue* will processed again, so the agent could then return `false` for `isWantingToArrive`, which would move the agent further to the buffer. Only then it would need to tell Netsim where it wants to go next. So in principle, I think it could work. But probably better to test this in a unit tests :)

kainagel:
Re lanes: vehicle departure/arrival is upstream of where the link may split into several lanes, so it should work there as well.  Thing to keep in mind (in general with lanes) is that the arrival/departure location is moved upstream with lanes.

PR #565, opened 2019-06-20
Bicycle contrib injection

Janekdererste:


Janekdererste:
Hey, @dziemke should I also make the BicycleLinkSpeedCalculator use BicycleTravelTime.getLinkTravelTime?

Currently the calculator is only using the speed-factor-attribute. The TravelTime module also reflects this factor but also takes other parameters such as surface or gradient into account.

What do you think?

contribs/bicycle/src/main/java/org/matsim/contrib/bicycle/BicycleTravelTime.java
```diff
@@ -69,66 +63,78 @@ public double getLinkTravelTime(Link link, double time, Person person, Vehicle v
...
+		return factor;
 	}
-	
+
 	// TODO combine this with comfort
-	private double computeSurfaceSpeed(double vehicleLinkSpeed, String surface, String type) {
-		if (type == null || type.equals("cycleway")) {
-			return vehicleLinkSpeed; // Assuming that dedicated cycleways are on good surface like ashalt
+	private double computeSurfaceFactor(Link link) {
+		if (hasNotAttribute(link, BicycleLabels.WAY_TYPE)
```
Janekdererste: so, I've included this because the old code did too. Do we want the suface factor to be 1.0 if we don't know the waytype? One could imagine that a user only supplies surface attributes but no waytypes

PR #567, opened 2019-06-25
simulate pseudo EV consumption when calculating charging breaks

jfbischoff:
This leads to more accurate results when simulating ERS


contribs/ev/src/main/java/org/matsim/contrib/ev/routing/EVNetworkRoutingModule.java
```diff
@@ -207,11 +207,16 @@ public EVNetworkRoutingModule(final String mode, final Network network, RoutingM
 
 			double consumption = driveEnergyConsumption.calcEnergyConsumption(l, travelT, Time.getUndefinedTime())
 					+ auxEnergyConsumption.calcEnergyConsumption(basicLeg.getDepartureTime(), travelT, l.getId());
+			if (consumption > 0) {
+				pseudoVehicle.getBattery().discharge(consumption);
+			} else {
+				pseudoVehicle.getBattery().discharge(-consumption);
```
michalmac: @jfbischoff,  did you mean this?
`pseudoVehicle.getBattery().charge(-consumption);`

jfbischoff: whoop, indeed.

PR #569, opened 2019-06-25
Drt access egress routing

kainagel:
The main change is that the config setting `includingAccessEgressRouting` is now extended to drt.  This is done directly in the `DrtRoutingModule`, so the design is different from the one with network routing.  Reason is that `DrtRoutingModule` is injected and then used elsewhere, and thus cannot be replaced by a differently named class.

Changes that happened along the way:
* I combined `TransportMode.access_walk` and `TransportMode.egress_walk` into `TransportMode.non_network_walk`, since what is egress from drt may be access to pt and overall that distinction does not make sense.  The code currently leaves the old constants in place and just re-maps them to the same string, but this should be changed rather soon if no other problems emerge.
* coloring change of otfvis routes.  The different modes are now pulled apart in color-space.
* some consistency checks that are IMO no longer meaningful are commented out.
* I introduced proxy vehicles by mode into the router.  This should probably be replaced by passing on the vehicles from the scenario, but that other change is still a bit in the future, but I needed different speeds for walk/bike/drt on the same network.
* Some changes in `NetworkInclAccessEgressRoutingModule` that were triggered from the cleaning up.
* Some debugging output (which in one or the other case may still have to be switched off).


contribs/drt/src/main/java/org/matsim/contrib/drt/routing/DrtRoutingModule.java
```diff
@@ -61,15 +66,22 @@
...
+		// we have at least some flexibility without having to change the argument list.  kai, jul'19
 
-	public DrtRoutingModule(DrtConfigGroup drtCfg, @Named(DvrpRoutingNetworkProvider.DVRP_ROUTING) Network network,
-			@Named(DvrpTravelTimeModule.DVRP_ESTIMATED) TravelTime travelTime,
-			TravelDisutilityFactory travelDisutilityFactory, PopulationFactory populationFactory,
-			@Named(TransportMode.walk) RoutingModule walkRouter) {
-		this.drtCfg = drtCfg;
+		LOGGER.setLevel( Level.DEBUG );
+
+		this.drtCfg = drtCfg ; // can't take out of normal config; don't know why.  kai, jul'19
```
michalmac: There may be multiple drt config groups. So it needs to be explicitly passed via constructor.

PR #579, opened 2019-07-01
Make (Dvrp) FleetWriter concurrent, so that it can be used by paralle…

Janekdererste:
…l streams

I want a parallel stream, since my filtering logic of where to put vehicles is slow, due to complitcated geometries


michalmac:
Thanks. Makes sense.
Actually this should be also applied to other dvrp/ev writers (that operate on streams). Will take care of it later...

BTW. Shouldn't  automerging be blocked until the requested review is done?? @kainagel

PR #594, opened 2019-07-12
Replace access egress walks in person prepare for sim

vsp-gleich:


kainagel:
Context: I had recently combined "access_walk" and "egress_walk" into "non_network_walk".  The reason was that we are now planning drt--pt--drt trips, and what is an "egress_walk" from the point of view of the first drt subtrip is an "access_walk" from the point of view of the pt trip.  I had always wondered if the differentiation between access_walk and egress_walk made sense, and now it is clear that it does not (and I should have introduced non_network_walk in the first place).

Now evidently, files that have "access_walk" or "egress_walk" legs will evidently fail with this new setup.  Thus that addition.

From my perspective, the change is ok.  Sorting out this data model will be a bit of a battle, but I don't think that we have a choice.  And now we are well in between releases.

PR #596, opened 2019-07-16
Have vehicle in router

Janekdererste:


matsim/src/main/java/org/matsim/core/router/NetworkRoutingInclAccessEgressModule.java
```diff
@@ -299,7 +288,9 @@ public String toString() {
 
 		if (toLink != fromLink) { // (a "true" route)
```
Janekdererste: So, if one constucts a scenario as I did in the test, this will be executed before PrepareForSim. Therefore in the 0th iteration the vehicle of a user is not known since it hasn't been added to the scenario in PrepareForSim. What do you think @kainagel ?

kainagel: Could you please let me know which test so I could try myself/see the problem?  Thanks ...

Janekdererste: https://github.com/matsim-org/matsim/pull/596/files#diff-0210a2a5d9f242ce5ed2105852b9c5fe
Sorry about the lengthy file. I meant to clean it up before I merge it into master

matsim/src/main/java/org/matsim/vehicles/VehicleUtils.java
```diff
@@ -51,17 +55,52 @@ public static VehicleType getDefaultVehicleType() {
 		return DEFAULT_VEHICLE_TYPE;
 	}
 
-    public static Id<Vehicle> getVehicleId(Person person, String mode, Config config) {
+	/**
```
Janekdererste: I have moved most of the logic of attaching the vehicle ids into this utility class. Do you think this is the way to go @kainagel ?

matsim/src/test/java/org/matsim/core/router/NetworkRoutingInclAccessEgressModuleTest.java
```diff
@@ -49,86 +55,101 @@ private static Scenario createScenario() {
...
+                .filter(element -> element instanceof Leg)
+                .map(element -> (Leg) element)
+                .filter(leg -> !leg.getMode().equals(TransportMode.non_network_walk))
+                .flatMap(leg -> ((NetworkRoute) leg.getRoute()).getLinkIds().stream())
+                .collect(Collectors.toList());
+    }
+
     @Test
-    public void calcRoute() {
+    public void calcRoute_speedOfIndividualVehicle() {
```
Janekdererste: I've put an integration test into this class to test, whether the router respects the vehicles and their speeds. Is this the right place @kainagel ?

matsim/src/main/java/org/matsim/vehicles/VehicleUtils.java
kainagel: Looking good to me.

matsim/src/test/java/org/matsim/core/router/NetworkRoutingInclAccessEgressModuleTest.java
kainagel: I guess that for tests we don't have a lot of convention.  I move tests if by moving it I can make some methods package-private, but that is about it.

PR #602, opened 2019-07-24
Update JFreeChart to version 1.5.0

sebhoerl:
There are a couple of features missing in the old JFreeChart from 2014. This PR updates the version to current 1.5.0 from 2017.

@kainagel, @mrieser, @tduberne, @michalmac  This may probably have impact on third-party analysis tools. Do you think it is good to merge?

Some guidelines for fixing code after the update:
- Function calls containing the keyword `Base` or which have a general purpose have been transformed to `Default`, e.g. `setLinesVisible` -> `setDefaultLinesVisible`
- `ChartUtilities` was renamed to `ChartUtils`

### DIFF
--- a/contribs/analysis/src/main/java/org/matsim/contrib/analysis/vsp/traveltimedistance/TravelTimeValidationRunner.java
+++ b/contribs/analysis/src/main/java/org/matsim/contrib/analysis/vsp/traveltimedistance/TravelTimeValidationRunner.java
@@ -23,7 +23,7 @@
 package org.matsim.contrib.analysis.vsp.traveltimedistance;
 
 import org.jfree.chart.ChartFactory;
-import org.jfree.chart.ChartUtilities;
+import org.jfree.chart.ChartUtils;
 import org.jfree.chart.JFreeChart;
 import org.jfree.chart.annotations.XYAnnotation;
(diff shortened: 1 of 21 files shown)


sebhoerl:
The `freight` contrib fails, because it complains that it doesn't find `ChartUtils`. So it seems that a dependency of `freight` overrides the newer JFreeChart with an older version. @JWJoubert, do you have an idea which one?

kt86:
@sebhoerl attached you can find the dependency hierachy of the freight pom.xml.
The old version of jfreechart comes with jsprit-analysis.
grafik

michalmac:
I like the idea of upgrading to 1.5.0

As for the mentioned issue:  What about declaring jfreechart 1.5.0 in freight pom.xml? From what I can see jsprit-analysis compiles without errors with 1.5.0 (not sure about graphstream that is used by jsprit-analysis and depends on 1.0.14)

BTW. I think the dependency to jcommon (in matsim pom.xml) can be removed after switching to 1.5.0

sebhoerl:
@kt86 Nice, I have never discovered this tab :)
@michalmac I'll check jcommons and try to add 1.5.0 to freight. I can also try to use an exclusion

PR #604, opened 2019-07-24
IOUtils cleanup + bz2 compression

sebhoerl:
I wanted to add bz2 compression for I/O, especially because then pt2matsim could use bz2 packaged osm files, which can be downloaded directly from Geofabrik.

While doing this I noticed the many different methods in IOUtils, at the end I ended up cleaning up the whole class + making it easy to add new compression algorithms. Actually, compression now is based on Apache Commons Compress, so it should be fairly stable for the future.

Please have a look at `IOUtils`, where I documented the rationale behind the new class and which I/O methods I left there and which I removed. The idea was that everything is based solely on `URL`s and only at the very end of the "chain of utility" methods one can use `String` paths.


sebhoerl:
@kainagel I could also rename it back. The `newUrl` was in most cases called as `newUrl(null, "some path")`, that's why I split this functionality up into `getFilePath(String)`, which simply gets a path and `extendUrl(URL, String)`, which extends a given URL. This avoids the many ugly calls with the `null` argument. The I decided to change it to "extend" because it better describes the functionality. Plus, almost all methods in MATSim follow the `[verb][Something]` scheme. What do you think?

sebhoerl:
@kainagel Any comment on the previous?

mrieser:
Thanks for cleaning IOUtils up a bit.
It will also help regarding my proposed change here: [MATSim-946](https://matsim.atlassian.net/projects/MATSIM/issues/MATSIM-946)

Have you tested the performance of Apache's LZ4 compression compared to the native one it replaces? At least for the Snappy compression, the apache implementation is a lot slower than the one provided by another library, so it sometimes actually makes sense to have a dedicated dependency instead of using Apache compress. I don't know how it is with LZ4, though.

mrieser:
> Actually, compression now is based on Apache Commons Compress, so it should be fairly stable for the future.

But it only works if Apache Commons supports the compression format, which is not the case for e.g. ZStandard. I guess it's okay for the moment, but I would prefer a solution where we are not limited to one compression-provider (aka Apache Compress).

sebhoerl:
Not sure about the speed for LZ4. I just noticed that there are two versions in Apache compress, I think lz4 and l4-block. The latter one is what has been used so far in MATSim it seems (with another external library), but lz4 created smaller files, actually. To keep it compatible I sticked to lz4-block.

mrieser:
I just ran my compression benchmark, comparing the old lz4 implementation with the one from apache. In multiple runs, testing different files, apache lz4 is around **500x slower** than the lz4 implementation used before.
(Others find the same thing, e.g. [here](http://mail-archives.apache.org/mod_mbox/commons-user/201710.mbox/%3C877ew9vc5o.fsf@v45346.1blu.de%3E) is noted that the apache implementation takes more than 1 minute, while other implementations do the same job in less than 10ms).

I thus strongly advise **against** merging this change, as it replaces a well-working lz4 compression with a basically unusable one.

sebhoerl:
Hm, interesting. So I guess we could extend lines 218/219 and 278/279 to return an InputStream / OutputStream backed by any other library in a more modular way. At this point we already know from the extension which compression we want. I could put back in the old LZ4 library. Maybe then you could directly put in the other compression algorithm that you suggested, before we merge this?

mrieser:
> Maybe then you could directly put in the other compression algorithm that you suggested
Sounds good! Will likely do so at the DevMtg.

PR #605, opened 2019-07-26
Public emission attributename

JWJoubert:
The attribute name used for HBEFA link types needs to be public to ensure consistent use of the attribute name when creating networks from outside the contrib


JWJoubert:
My apologies, this should not be needed. The HBEFA road type can (and probably SHOULD) rather be set using `EmissionUtils.setHbefaRoadType(Link link)`

So, I'm going to close this pull-request without merging.

PR #612, opened 2019-08-13
increase max speed to 80 m/s in Ohde-Slaski EV consumption

michalmac:
fix https://matsim.atlassian.net/browse/MATSIM-924


jfbischoff:
Isn`t 80 still rather low for a Nissan Leaf?

michalmac:
It's 80 m/s :-)

jfbischoff:
haha, okay, that is one bullet train of a Nissan then :-D

PR #616, opened 2019-08-16
identify main mode based on stage activity if no main mode leg in trip

vsp-gleich:
In the current DrtRoutingModule design and in the VSP fork of SwissRailRaptor we try to reduce (and eventually eliminate) direct walks such as "drt_walk". Instead we have trips of the form  non_network_walk --> pt interaction --> non_network_walk. There is no leg of the main mode, so we try to find out the main mode using the interaction activity.
MATSIM-943, MATSIM-945


vsp-gleich:
We arehaving 2nd thoughts about this (KN+GL)

PR #618, opened 2019-08-19
move fallback routing into framework?

kainagel:
Routers that cannot find routes in matsim typically return something else, such as transit_walk or drt_walk.  Our proposal is to allow `RoutingModule`s to return `null`, and address the situation in `TripRouter`.  A reason for this is that it may indeed be the case in reality that a route cannot be found (drive from Berlin to New York, or walk from Amsterdam to London). 
Also, forcing implementers of `RoutingModule` to find fixes for this themselves leads to quite inconsistent situations (as we have found over the last weeks).  

To achieve this, additionally a `FallbackRoutingModule` is registered, and then used.  Some things we are unsure about:
* Would one want separate such fallback routing modules for each mode?  We think not since it would (once more) make things a lot more brittle.

To note:
* `<mode>_fallback` would be the convention for fallback modes.  Since we are moving towards more such conventions with "_interaction" and "subpopulation", we would consider this acceptable.


matsim/src/main/java/org/matsim/core/router/MainModeIdentifierImpl.java
```diff
@@ -42,6 +42,9 @@ public String identifyMainMode( final List<? extends PlanElement> tripElements)
 			// Marcel's SBB raptor returns access/egress_walk to and from pt, and transit_walk only
 			// for direct walk (and presumably in between pt legs, if necessary).  kai/gregor, sep'18
 		}
+		if ( mode.endsWith( "_fallback" ) )  {
+			return mode.split( "_fallback" )[0] ;
```
mrieser: I would probably use `mode.substring(0, mode.length() - 9)` instead of `split`, as `split` is (a) a more expensive method, and (b) while it may works it looks counter-intuitive to me (we're not splitting, we want to cut off a part).

matsim/src/main/java/org/matsim/core/router/FallbackRoutingModule.java
```diff
@@ -0,0 +1,4 @@
+package org.matsim.core.router;
+
+public interface FallbackRoutingModule extends RoutingModule {
+}
```
mrieser: Do we really need this interface? It does not provide additional functionality. The fallback routing module is just a routing module that is used as a fallback. The fact that it is used as a fallback comes from its designation of a specific routing module, not from implementing a specific interface.

One could argue that the fallback routing module is not allowed to return `null`, compared to a regular routing module. But I don't see this constraint here either.

matsim/src/main/java/org/matsim/core/router/NetworkRoutingInclAccessEgressModule.java
```diff
@@ -263,7 +263,7 @@ private static Activity createInteractionActivity( final Coord interactionCoord,
 	}
 
 
-	private static double routeBushwhackingLeg(Person person, Leg leg, Coord fromCoord, Coord toCoord, double depTime,
+	static double routeBushwhackingLeg(Person person, Leg leg, Coord fromCoord, Coord toCoord, double depTime,
```
mrieser: Not sure, would it probably make sense to move this to a separate `TeleportationUtils` class or similar? Because that's what it calculates in the end.

At least I find it confusing that the `FallbackRoutingModuleDefaultImpl` uses functionality from the `NetworkRoutingInclAccessEgressModule`, when all it actually does is to calculate a hard-coded teleportation.

Also, this method has hard-coded assumptions about teleportation speed, with a `yyyyyy`-remark to take the values from the config. For the fallback it might not make a difference, although one could argue that the fallback is "teleported walking" and that it should try to use the walk-parameters, if `walk` is defined for teleportation.

matsim/src/main/java/org/matsim/core/router/FallbackRoutingModule.java
mrieser: Ah, now I see, dependency injection... It's easier to bind a specific implementation to this specific interface, so it will be used as fallback routing module. I understand it now, but it kind of confirms my opinion that dependency injection does not necessarily help to write clean code, because clearly this interface is only useful due to dependency injection being used, and is kind of useless otherwise.

matsim/src/test/java/org/matsim/core/router/FallbackRoutingModuleTest.java
```diff
@@ -0,0 +1,93 @@
...
+		StrategyConfigGroup.StrategySettings sets = new StrategyConfigGroup.StrategySettings(  ) ;
+		sets.setStrategyName( DefaultPlanStrategiesModule.DefaultStrategy.ReRoute ) ;
+		sets.setWeight( 1. ) ;
+		config.strategy().addStrategySettings( sets );
+
+		Scenario scenario = ScenarioUtils.createScenario( config ) ;
+
+		Network network = scenario.getNetwork() ;
+		NetworkFactory nf = network.getFactory();
+
```
mrieser: I would have expected consistent code style in a test class which I assume would have been written by a single person. But about 50% of the lines end with a semicolon with a space in fron of it, and the other 50% of the lines do not have a space before. Some of the method arguments have spaces around the parentheses like above (`( config )`), some not (like below: `new Coord(0.,10.)`). Sometimes there are spaces after (and even before) commas separating parameters, sometimes not at all (both cases can be seen on the next line.

I would really prefer if at least within one file the formatting would be consistent, as it makes reading code much more easier.

matsim/src/main/java/org/matsim/core/router/MainModeIdentifierImpl.java
kainagel: perfectly fine with me.  done

matsim/src/main/java/org/matsim/core/router/FallbackRoutingModule.java
kainagel: The main reason for this actually was not that I wanted this to be replaceable, but that even the simple `FallbackRouterDefaultImpl` needs access to material (`Population`, `Network`) that `TripRouter` does not have.  And I didn't really want to touch the arguments of the public constructor of `TripRouter`.  So in that sense I would say that DI is in fact not so bad here since it allows each `RoutingModule` (including this one here) to use arbitrary information without having to pass it through `TripRouter` explicitly.

matsim/src/main/java/org/matsim/core/router/NetworkRoutingInclAccessEgressModule.java
kainagel: Agreed.  I just didn't want to move it before we agree in general.  Do we all like `TeleportationUtils`?

matsim/src/test/java/org/matsim/core/router/FallbackRoutingModuleTest.java
kainagel: I guess that my manually typed coding style and the intellij autoformatting are not consistent.  I have now intellij-autoformatted the test class; hoping that that will make it better.

neuma:
I like the proposal. It is the kind of approach I used in my own implementions, e.g. https://github.com/matsim-org/matsim/commit/ec6f8e64f6de0c70225a47cb93b052be9414876c. 

For me, the router anwers a request for a certain type of service. The answer may be negative if
-  there is no service in the area
-  the service has no capacity left
-  the service is too slow to answer
-  whatever the reason

The service should not invent a solution by providing a route that requires a different kind of service. The solution is up to the user, i.e. the TripRouter. A best practice example would be the search engine provided by [Trafi](https://www.trafi.com/site/en) or its implementation for Berlin called Jelbi. Their `TripRouter` (App) integrates a bunch of `RoutingModule`s (PT, Taxi, car|bike|scooter sharing providers) and always provides a regular walk trip as fallback - because more often than not a couple of those providers will not be available for your particular request and just return null.

PR #619, opened 2019-08-20
add service area shapefile name for each individual drt config group

kainagel:
It would, for the time being, allow to set the file name, but not do anything else.  But derived code (e.g. the matsim berlin scenario) would use it.

We have thought to and fro, but this is really the best place.  Otherwise, it ends up first in some `AvoevConfigGroup`, next in some `BerlinConfigGroup`, and eventually it will end up in `DrtConfigGroup` anyways.  Best put it there from the beginning.

We have other cases where adding a file name will not do anything (e.g. the households file, which, to be true, will be read in the scenario loader, but this is more a shortcoming of the scenario loader which is not able to load files that are in non-core config groups).


contribs/drt/src/main/java/org/matsim/contrib/drt/run/DrtConfigGroup.java
```diff
@@ -50,9 +51,40 @@
...
 	@SuppressWarnings("deprecation")
 	public static DrtConfigGroup get(Config config) {
-		return (DrtConfigGroup)config.getModule(GROUP_NAME);
+//		return (DrtConfigGroup)config.getModule(GROUP_NAME);
+		return ConfigUtils.addOrGetModule( config, DrtConfigGroup.class ) ;
+		// yyyy I think that this method should be inlined and then removed to become consistent with how it is done elsewhere.  kai, aug'19
+		// yy I have to say that I am not sure if the method should exist at all.  It seems to me that DrtConfigGroup can exist both at the base level
+		// _and_ as one of many such config groups inside MultiModeDrtConfigGroups.  Our experience with such dual use approaches is not so good.  kai,
+		// aug'19
 	}
```
michalmac: This method cannot have a side effect (i.e. creating a module if missing), otherwise checks in `MultiModeDrtConfigGroup` will get broken.

Re disallowing a standalone `DrtConfigGroup`: probably the only risk is that most people use a single DRT setup. So by doing this, we will break lots of scenarios (which we do from time to time anyway) and we will make the typical setup a bit more complicated (no injection of `DrtConfigGroup`, and obtaining it by providing additional argument: `String drtServiceName`). But maybe worth doing. Will add this to my task list for the upcoming (remote) DevMeet:-)

contribs/drt/src/main/java/org/matsim/contrib/drt/run/MultiModeDrtConfigGroup.java
```diff
@@ -38,7 +39,9 @@
 
 	@SuppressWarnings("deprecation")
 	public static MultiModeDrtConfigGroup get(Config config) {
-		return (MultiModeDrtConfigGroup)config.getModule(GROUP_NAME);
+//		return (MultiModeDrtConfigGroup)config.getModule(GROUP_NAME);
+		return ConfigUtils.addOrGetModule( config, MultiModeDrtConfigGroup.class ) ;
+		// yyyy I think that this method should be inlined and then removed to become consistent with how it is done elsewhere.  kai, aug'19
 	}
```
michalmac: same as above -- no side effects when getting a group

contribs/drt/src/main/java/org/matsim/contrib/drt/run/DrtConfigGroup.java
jfbischoff: Mhm, I'm not against it, but maybe we should then get the naming right:
MultiModeDrtConfigGroup --> let's name it drt or DrtConfigGroup or whatever.
DrtConfigGroup --> rename it to DrtOperatorConfigGroup, cause that's what it is.

jfbischoff:
> Not sure if DrtConfigGroup should be used as a placeholder for something that is not officially supported. Makes sense only if we eventually add this functionality to the DRT contrib.

I strongly support Michal's oppinion. Adding placeholders for something to the DrtConfigGroup will rather sooner than later lead to people asking about the functionality. That said, I'm not against adding that functionality and would find it useful and easy to implement. 

>We have thought to and fro, but this is really the best place. Otherwise, it ends up first in some AvoevConfigGroup, next in some BerlinConfigGroup, and eventually it will end up in DrtConfigGroup anyways. Best put it there from the beginning.

At SBB they use a separate ZonesConfigGroup where a global shape file is set that is then accessed by many different functionality implementations. Perhaps a way to go for you in Berlin as well?

kainagel:
> Not sure if DrtConfigGroup should be used as a placeholder for something that is not officially supported. Makes sense only if we eventually add this functionality to the DRT contrib.
> 
> I strongly support Michal's oppinion. Adding placeholders for something to the DrtConfigGroup will rather sooner than later lead to people asking about the functionality. That said, I'm not against adding that functionality and would find it useful and easy to implement.
> 
> We have thought to and fro, but this is really the best place. Otherwise, it ends up first in some AvoevConfigGroup, next in some BerlinConfigGroup, and eventually it will end up in DrtConfigGroup anyways. Best put it there from the beginning.
> 
> At SBB they use a separate ZonesConfigGroup where a global shape file is set that is then accessed by many different functionality implementations. Perhaps a way to go for you in Berlin as well?

The way it works for us is actually dead simple: we go through the network and add the corresponding drt mode to each link in the corresponding service area.  The rest comes out via access/egress walk and iterations.  (I.e. if you want to use drt in a far away area, you first have to access-walk to it.) .  For a more general implementation, however, there are some issues:

1. Since this is a scenario modification, I prefer doing this after loading the scenario and before instantiating `Controler`.  Maybe this is a minor issue, but:
2. different people may want different implementations of this.  For example, I personally don't like the "reject requests outside service area" approach, since it means that someone who starts 5m outside the service area cannot use it any more.  So I rather have a drt router that has non-network-walk to closest drt link and then drt from there.

I guess that we (at vsp) could put one default implementation (which would be the network modification) into the drt contrib.  And add an enum config switch with name "handlingOfServiceArea" with settings such as "ignore", "rejectRequestsOutsideServiceArea" and "beAbleToWalkToServiceArea".  I somehow find it too early to decide on this.

In contrast, I am very certain that the service area shape file per drt mode will be useful.

Also note that I have commented out the StringGetter/Setter things so it will not show up in xml for the time being.

---

But clearly, we can have such config options elsewhere if that is your preference.  Let me know.

jfbischoff:
>I guess that we (at vsp) could put one default implementation (which would be the network modification) into the drt contrib. And add an enum config switch with name "handlingOfServiceArea" with settings such as "ignore", "rejectRequestsOutsideServiceArea" and "beAbleToWalkToServiceArea". I somehow find it too early to decide on this.

That sounds pretty wise to me. Perhaps @vsp-gleich and I could implement this functionality at the DevMeeting.

vsp-gleich:
@jfbischoff Great! I would be happy to implement that together at the Developer Meeting.

michalmac:
I suggest that the DrtShpUtils and ServiceAreaNetworkModification should be pulled up somewhere higher the module hierarchy (e.g. the matsim module?). They are very generic, so we could use them for taxi too, but also for any other purposes (not related to DVRP). And right now, we would not be able to use them for taxis since the taxi and drt contribs do not depend on each other.

BTW. I am working on forbidding standalone single-taxi and single-drt configs (discussed earlier in this thread)

jfbischoff:
After thorough discussion, we concluded to design this a bit differently:
A service area shape will interpreted as an additional OperationalMode. We basically have all the infrastructure and there is no need to tag the network with modes. 
Instead, one stop per link (at the toNode of each link) will be created and the stop based module will be used. This will lead to *mostly* the same behavior.

jfbischoff:
okay, this should run now.

jfbischoff:
okay, this should run now.

jfbischoff:
(hen  / egg problem - I've added test files to the examples project, but the snapshot is not up-to-date). Setting test to ignore, then I will enable it afterwards.

PR #629, opened 2019-08-27
ID<?> with integer representation

rodrigo-bruno:
This patch provides an integer representation of the ID<?> which already has a String representation. In addition, an event handler using this new integer representation is provided.


mrieser:
Travis got stuck again, testing a PR from a fork... Tracking a new PR at #635.

PR #631, opened 2019-08-28
listener for noise records.  also a more general infrastructure for xyt records

kainagel:
@nkuehnel :

See `NoiseOnlineExampleIT#testNoiseListener()` for an example.

This is not yet finished, but maybe you could (1) already make sense out of it, (2) find and fill in the missing values (in particular the times) where the record is generated, (3) add some plausible "asserts" to the test.


JWJoubert:
I see _many_ use cases (in our code) for the `XYTRecord` class. I am just not sure about the `startTime` and `endTime` attributes. Is the plan that this class is only used as an _after-the-fact_ format before writing it out... or also _during_ a simulation run: the container that is updated as `EventListeners` are processed?

nkuehnel:
@ikaddoura @kainagel 
I was looking for a way to send receiver points from SILO to MATSim by code, without needing to write out all current dwelling locations to file and reading them back in before the MATSim simulation.
In 9e786d5 I tried providing a third way of generating the grid by adding a list of receiver points as a scenario element which is then read in the grid class. 

However, this way I will still have to recalculate every receiver point preparation (correction terms etc) for every MATSim simulation, would be even nicer if one could just re-use the prepared receiver points throughout multiple MATSim runs (although it does not seem obvious to me how to do it).

Any opinion?
Thanks in advance

kainagel:
Seems like a good approach.  Comments:

(1) The object you use as ScenarioElement should preferably be typed.  Then you can get the typed object via a cast:

   TypedObject to = (TypedObject) scenario.getScenarioElement( NOISE_RECEIVER_POINTS ) ;

and don't need any more casts afterward.

(2) A plausible object IMO would be the receiver points already in the final format.  This would mean something like

public final class NoiseReceiverPoints implements Map<Id<ReceiverPoint>,ReceiverPoint> {
   private Map<Id<ReceiverPoint>,ReceiverPoint> delegate = new LinkedHashMap<>() ;
}

and delegate all the Map methods to the delegate.  Then you would have

  NoiseReceiverPoints nrp = (NoiseReceiverPoint) scenario.getScenarioElement( NOISE_RECEIVER_POINTS ) ;

Note that this is already in the format that you need, so all the "put" methods in "loadGridFromScenario" will no longer be necessary.  Instead, you can probably say

  this.receiverPoints = nrp ;

and be done with it.

(3) You could then construct and keep the NoiseReceiverPoints object on the SILO side, and add it into the scenario every time you call matsim.

Would that work?

Best wishes

Kai

On 27. Sep 2019, at 13:03, Nico Kühnel <notifications@github.com<mailto:notifications@github.com>> wrote:


@ikaddoura<https://github.com/ikaddoura> @kainagel<https://github.com/kainagel>
I was looking for a way to send receiver points from SILO to MATSim by code, without needing to write out all current dwelling locations to file and reading them back in before the MATSim simulation.
In 9e786d5<https://github.com/matsim-org/matsim/commit/9e786d53616b68bd9b111a5db077770106aedf9d> I tried providing a third way of generating the grid by adding a list of receiver points as a scenario element which is then read in the grid class.

However, this way I will still have to recalculate every receiver point preparation (correction terms etc) for every MATSim simulation, would be even nicer if one could just re-use the prepared receiver points throughout multiple MATSim runs (although it does not seem obvious to me how to do it).

Any opinion?
Thanks in advance

nkuehnel:
Thanks @kainagel for the comments, will work on that!

nkuehnel:
@kainagel @ikaddoura 

dd7bdda implemented the noise receiver points as discussed and got it to work such that one can see the calculated daily emissions in SILO. 

Therefore I had to make the NoiseReceiverPoint class public.
If we keep it like that I would not even need the XYT-Records-Listener infrastructure.

Another proposal I was trying is to add a boolean to the NoiseReceiverPoints that indicates whether the correction terms have already been calculated. This allows to re-use receiver points in multiple MATSim runs without needing to recalculate correction terms.

Happy to get comments!

Also, one last question is if it is possible to run the online calculation only during the last iteration of a matsim run?

contribs/noise/src/main/java/org/matsim/contrib/noise/NoiseReceiverPoint.java
```diff
@@ -224,11 +224,17 @@ void resetTimeInterval() {
 	}
 
 	void processImmission() {
+		// yy why is it meaningful to not compute them in the getters?  (If the getters are called very often, then precomputing the relatively expensive
```
nkuehnel: thanks for pointing that out. we can actually save three fields for each receiver point in that case. I will update accordingly

PR #632, opened 2019-08-28
ID<?> with integer representation

mrieser:
Same as #629, but from within the matsim-repository so travis works...


mrieser:
failed approach to fix issues with EventHandlers, thus closing and NOT merging this PR. Newer version is tracked at #635.

PR #637, opened 2019-08-29
Only multi mode taxi setup

michalmac:
Two most important changes:
- no standalone TaxiConfigGroup (must be inside MultiModeTaxiConfigGroup)
- removed (E)TaxiModule (use MultiMode(E)TaxiModule instead)

Config files (in this and dependant projects) were updated accordingly.


michalmac:
Have a look at this diff to see the little rearrangement at the level of config files:
https://github.com/matsim-org/matsim/pull/637/files#diff-16f9d278281a8b778386d90aa289389f

For single-mode taxi setups, instead of:
```
<module name="taxi">
    ...
</module>
```
we have now:
```
<module name="multiModeTaxi">
    <parameterset type="taxi">
        ...
    </parameterset>
</module>
```
which is a multi-modal setup with only one taxi mode provided.

michalmac:
Same changes will be applied to DRT

PR #639, opened 2019-08-31
Only multi mode drt config

michalmac:
Similar to #637 

For single-mode drt setups, instead of:

<module name="drt">
    ...
</module>
we have now:

<module name="multiModeDrt">
    <parameterset type="drt">
        ...
    </parameterset>
</module>
which is a multi-modal setup with only one drt mode provided.


michalmac:
Additionally:
- re-enable area-based DRT tests
- fix ShpGeometryUtils
- move SHP read/write test class (+test input) so they are in the same package as the production classes
- upgrade geotools 20.2->21.1 (compliant with JDK11 modules, e.g. https://osgeo-org.atlassian.net/browse/GEOT-6335)

PR #640, opened 2019-09-01
Guice Grapher

michaz:
This adds the Guice configuration grapher from my dissertation as a `ControlerListener` that puts the live configuration graph into the output directory.

Can be viewed with Graphviz, with something like

```
dot -Tpdf modules.dot > modules.pdf
```

Requires JGraphT dependency because I needed to filter down a graph, and to breadth-first-search  for a connected component -- could be done from scratch.

Contains some copied code from Guice-Graph that I couldn't extend/use/modify otherwise.

Contains two places where lots of MATSim classes are listed:
- One is just a list of stuff I arbitrarily decided I didn't want to see -- adapt at your pleasure. 
- The other lists all the multi-extension-points that we support, either with the regular Guice Map- or SetBinder, *or* with my custom version that allows map items to be overridden. This should be adapted when we add (or remove) such things. It's how I filter out the "technical" nodes created by multibinders and such. *Alas*, this requires tweaking for every new Guice version, since the (otherwise hidden) implementation details of those change every version, so with a new Guice version, you likely get lots of new clutter nodes.


mrieser:
I understand the usefulness of such a feature, but:

> Alas, this requires tweaking for every new Guice version, since [...]

Do we have the resources to maintain this?

kainagel:
I guess we could still switch it off if we can't maintain it. :-|

michaz:
Yeah, worst thing that happens is that the result becomes more cluttered. It doesn't fail or turn into a useless mess immediately.

michaz:
Ah. Now we're getting an exception for a Contrib that uses an optional binding, and the Guice Grapher doesn't seem to know about that. :-( Not good. 

https://github.com/google/guice/issues/998

kainagel:
I commented out the affected piece of code.  I am not sure if it makes so much sense to have optional bindings.  In the case here, it was a quick fix for a design that should in theory re-use the central `TripRouter`, but in practice does not and thus needs workarounds.

PR #651, opened 2019-09-06
add support for calculating emissions of public transport buses.

mrieser:
Mostly by copy/paste from the HGV vehicle type.
MATSIM-953.


mrieser:
I don't know what happened in the file ColdEmissionAnalysisModule that GitHub thinks the whole file has changed. IntelliJ only shows me the around 20 lines in the diff that I have actually changed...

kainagel:
Need to first merge it with the recent changes in Emissions (which now relies on vehicle type to get its input).

kainagel:
we are currently in the process of moving the emissions contrib forward once more.  see `knEmissionsReintroduceEnums` branch.  We encountered the same problem with missing vehicles types, and will try to introduce all of them consistent with hbefa 4.  This should then make this pull request here obsolete (and explain why there are conflicts since we will be touching the same material).

kainagel:
@mrieser : I think that this is now obsolete, since the more general overhaul of the emissions contrib contains the same thing.  Could you please check, or maybe just close this here and hope for the best.  Thanks

PR #653, opened 2019-09-09
Delete stage activity types (MATSIM-945)

vsp-gleich:
The idea is to remove StageActivityTypes and all related classes after we decided that every activity type which ends on "interaction" is a stage activity and remove the possibility to declare custom stage activities. Instead there is only a class StageActivityTypeIdentifier which (only) contains the static method isStageActivity(act).

In TripStructureUtils.getActivities() there were usages with empty StageActivityTypes (get all activities no matter whether they are a stage activity or not) and tried to solve this with an enum (StageActivityHandling) which defines whether stage activities shall be included (StagesAsNormalActivities) or excluded (ExcludeStageActivities). This is something we can discuss about. I tried to split these different usages up into different methods, but unfortunately lots of other classes use these methods and would have to be splitted up, too, so I turned back to the enum.

In the contrib socnetsim there is a class JointTripRemoverAlgorithm which makes use of two different sets of stageActivityTypes. This is the only place where our approach that all stage activities must end on "interaction" won't work (because we cannot differentiate different sets of stage activity types). So as a temporary measure (= until socnetsim resolves that differently) I added deprecated methods in TripStructureUtils where a Set of Strings can be passed with activity types where to split trips. I did this to remove the class StageActivityTypes completely, but maybe it would be nicer to keep StageActivityTypes as deprecated class for that purpose.

Any opinions on the code design, especially the StageActivityHandling enum and the deprecated methods for socnetsim?


kainagel:
There is a comment by Thibaut under https://matsim.atlassian.net/browse/MATSIM-945.

PR #670, opened 2019-09-17
Add FlowEfficiencyCalculator

sebhoerl:


sebhoerl:
@kainagel Only in the `QLinkInternalInterface` I wasn't sure if any design decision speaks against adding `getLink` there? (There are getters for nodes though).

PR #676, opened 2019-09-23
fix freight example and (re)enable ignored IT

kt86:


kt86:
Don't know why it fails on travis, while passing locally. Closing it for the time beeing.

PR #677, opened 2019-09-24
fix EventsToLegs

michalmac:
Fix bugs:
1. vehicle2passengers were not cleaned between iterations (should be fine as long as no trip gets "interrupted" by QSim endTime  (the critical one)
2. relPosOnDepartureLinkPerPerson and relPosOnArrivalLinkPerPerson is not handled properly (when we have intermediate stops when sharing rides)

Improvements:
1. record routes for vehicles (instead for each agent) and later assign a sublist of links to passengers/drivers upon their arrival
2. add more unit tests (shared rides and special boundary cases)

### DIFF
--- a/contribs/drt/src/main/java/org/matsim/contrib/drt/routing/StopBasedDrtRoutingModule.java
+++ b/contribs/drt/src/main/java/org/matsim/contrib/drt/routing/StopBasedDrtRoutingModule.java
@@ -43,7 +43,6 @@ import org.matsim.core.router.StageActivityTypes;
 import org.matsim.facilities.Facility;
 import org.matsim.pt.transitSchedule.api.TransitStopFacility;
 
-import com.google.inject.Inject;
 import com.google.inject.name.Named;
 
 /**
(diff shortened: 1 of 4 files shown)


michalmac:
I think it is impossible to move through the network without being in a vehicle. Then this (and related) part of the code could be removed:
https://github.com/matsim-org/matsim/blob/3160da47e0db216ebb0d58d27a1f0d57af7cddbe/matsim/src/main/java/org/matsim/core/scoring/EventsToLegs.java#L277-L291

Just asking, because maybe there is some 'pseudo' way of simulating walks along streets etc... I just cannot see anything like this in QSim.

kainagel:
Not sure if this still is an issue.  In "standard matsim" it should not be possible to move through the network without a vehicle; it is like that even for "walk" (think of "shoes" as "vehicle").

Right now, however, I fail to see how that relates to the incriminated code.  Probably my stupidity.  Maybe in that class routes should be registered to vehicles not persons, but I don't think that that is the case.  Or is it?

PR #680, opened 2019-09-29
Performance Improvements wrt Id

mrieser:
Make use of the optimized IdMap where is seems useful.


michalmac:
Appveyor started to fail in minibus. A windows-specific issue?

michalmac:
https://ci.appveyor.com/project/mrieser/matsim/history

michalmac:
The test passes on my Windows machine though.

mrieser:
I'll need to have a closer look at this. It seems as if the test is non-deterministic. It passed in Appveyor and Travis in 7e97ea6 but not on Jenkins, but failed in 1d9b74a in AppVeyor but passed in Travis, although I only merged in the master in that commit.
The weekly build (running on Jenkins) now also failed due to this test...

mrieser:
Looks like `SidewaysRouteExtensionTest` fails.

I did some tests on my machine locally.
- Running the test alone in IntelliJ works. 
- Running all tests in minibus-contrib in IntelliJ fails.
- Running all tests in minibus-contrib in Maven works with sure-fire (`mvn test`)
- Running all tests incl. integration tests in minibus-contrib in Maven works ('mvn verify')

Not sure why we get inconsistent behaviour between Travis and AppVeyor.
But the effects (failing when run in part of a larger test series, vs running stand-alone) looks similar to the problem I fixed in cf21594, so I'll try the same change again.

mrieser:
Hopefully fixed in #682.

PR #681, opened 2019-09-30
Freight collections to maps

kt86:
Use Maps<Id<..>, ... > instead of Collections for Carriers:

- CarrierShipment
- CarrierService
- CarrierVehicle

This is be done, to make it possible to get one specific element without iterating through the collection.

One the way: Extend CarrierUtils with methods to add/get this objects to/from the carrier.


kt86:
Note: The problems with the Appveyor and jenkins tests (minibus contrib) seems to me being related to the observation made in #680 . 
@mrieser In this PR here, again travis is happy, while appveyor and Jenkins were not.

PR #682, opened 2019-09-30
trying to fix non-deterministic Minibus-tests

mrieser:


neuma:
Good catch! Thanks a lot.

PR #686, opened 2019-10-07
Freight config group

tschlenther:
I introduced the FreightConfigGroup which basically contains the file names for carriers and carrierVehicleTypes input files and one more parameter. I deprecated CarriersConfigGroup.
In the CarriersModule, the input files now are read in.
@kainagel: Could you check if this is what is meant by the comments in Freight.java ?

@kt86: is there anything else that belongs into this config group?


kt86:
Even when we create the algorithm in many cases within the code, I think it would be useful to specify the location of the algorithms file in the freight config as well.

@kainagel Should we also prepare that one can switch the VRPSolver (by now its only jsprit, but maybe there will be something else in thew future)

tschlenther:
In order to let everything run, one would either 
a) have to read carriers manually and load the vehicle types and then hand the carriers object to the module or 
b) provide input file paths in the FreightConfigGroup

JWJoubert:
I'm happy with that arrangement too.

PR #688, opened 2019-10-09
CarrierPlanXmlReaderV2 and CarrierPlanXmlWriterV2 now handle attributes

tschlenther:
attributes are read in and written out for the following objects
a) carrier
b) shipment
c) service


kainagel:
I would not set the automerge label to things that are not yet ready ...

tschlenther:
after i merged in the master, the build of dvrp contrib timed out because no output was generated for 10 min..
can't see the reason why.
I'll trigger a new build with an empty line commit

PR #689, opened 2019-10-09
Fix carrier module

tschlenther:
the reading of freight input files in the carrier module was in the wrong place.
trying to fix the access on FreightConfigGroup in CarrierModule.
introduce a test for CarrierModule that instantiates the CarrierModule and runs a simulation for all existing constructors


kainagel:
I would not set the automerge label to things that are not yet ready; it then feels like fixing regression tests is the _only_ thing one programs against.

tschlenther:
@kainagel in addition to the previous discussion, the CarrierVehicleTypes also are loaded into the scenario. On first sight, this was basically only needed in RunChessboard.java (for creating some TravelDisutility later) but it might also be needed somewhere else. Please let me know how you feel about this.
Moreover, i changed RunChessboard so that the new infrastructure of loading everything into the scenario is used. That means, the CarriersModule is plugged in.

PR #699, opened 2019-10-18
Drt stop search radius

vsp-gleich:
Make stop finder in DRT more similar to those in public transit routers, re-apply hard constraint for access/egress walk distance to drt to reduce number of passengers walking into drt service area for very long distances

* interpret maxWalkDistance as a real max, use quad tree to find drt stops

* do not factor in beeline distance factor in maximum distance

* set default value of maxWalkDistance to Double.MAX_VALUE, so default
behaviour is very similar to situation before this commit

* change StopBasedDrtRoutingModuleTest to new (respectively initial)
interpretation of maxWalkDistance

* maxWalkDistance was already a hard constraint until commit
d99ef91


kainagel:
The conflict with the master concerns a test, where I cannot judge which result is the correct one.

vsp-gleich:
@kainagel I already saw that coming (after the fallback router merge) and resolved the conflict :-)

PR #708, opened 2019-11-01
override equals() and hashCode() in TimeWindow

tschlenther:


contribs/freight/src/main/java/org/matsim/contrib/freight/carrier/TimeWindow.java
```diff
@@ -62,4 +62,26 @@ public String toString() {
...
+	}
+
+	@Override
+	public int hashCode(){
+		int result = 59;
+		Long startLong  = Double.doubleToLongBits(start);
+		int startHash = (int) (startLong^(startLong>>>32));
+		result = 31 * result + startHash;
+		Long endLong  = Double.doubleToLongBits(end);
+		int endHash = (int) (startLong^(startLong>>>32));
```
michalmac: shouldn't `endHash` be: `endLong^(endLong>>>32)`?

tschlenther: yes it should! thanks

PR #712, opened 2019-11-08
Raptor config consistency check for intermodal access/egress

vsp-gleich:


vsp-gleich:
> Very nice. We should also add checks if maxRadius is smaller than the initial radius.

Hm, I see your point, but I set the default of initialSearchRadius to Double.POSITIVE_INFINITY (so users normally don't have to care about this). So the consistency checker would always complain, unless someone manually set a different initialSearchRadius.

jfbischoff:
> Very nice. We should also add checks if maxRadius is smaller than the initial radius.
> 
> Hm, I see your point, but I set the default of initialSearchRadius to Double.POSITIVE_INFINITY (so users normally don't have to care about this). So the consistency checker would always complain, unless someone manually set a different initialSearchRadius.

That leaves the question if infinity is a good default value :-)

vsp-gleich:
> That leaves the question if infinity is a good default value :-)

What would be your intiuition on a good default? I added the parameters _initialSearchRadius_ and _searchExtensionRadius_ and gave them default values, in order not to break old code. _maxRadius_ already existed before as _radius_. All the original parameters (including _maxRadius_) did not have any default values. So is that the way to go? No defaults in the ConfigGroup and a consistency checker which checks whether _initialSearchRadius_, _searchExtensionRadius_ and _maxRadius_ are set and larger than 0?

vsp-gleich:
Access / egress modes can be very heterogeneous (think of walk and say flying taxis ;-) ), so I don't see the default value which fits it all (besides _initialSearchRadius_=infinity which at least does not restrict any possible route, but slows down computation).

jfbischoff:
Not providing any default values comes with the advantage that people need to actively think of reasonable values, which perhaps is a good way to go, what do you think?

vsp-gleich:
Yes, I'll remove the defaults.

vsp-gleich:
maxRadius == 0 is definitely wrong, initialSearchRadius==0 and searchExtensionRadius==0 is most probably also wrong, I don't see any legitimate use case.

PR #721, opened 2019-11-14
Only StopBasedDrtRoutingModule

michalmac:
Since `StopBasedDrtRoutingModule` now can use different routing modules for access/egress and use different access/egress facility finders, we can simply emulate the original `DrtRoutingModule` with it, and hence the latter has been removed.

Thinking about renaming of `StopBasedDrtRoutingModule` to simple `DrtRoutingModule` (as it is now the only one)


jfbischoff:
>Thinking about renaming of StopBasedDrtRoutingModule to simple DrtRoutingModule (as it is now the only one)


Sounds wise to me, also to erase confusion.

kainagel:
Thanks a lot.  I like "reduction of lines of code" :-).  Yes, please rename.

neuma:
> Thanks a lot. I like "reduction of lines of code" :-). Yes, please rename.

+1

PR #724, opened 2019-11-15
Add DrtStopNetwork

michalmac:
To be used instead of TransitSchedule for keeping bus stops. Depending on the case, it is empty (door2door) or contains all drt-allowed links in a given area (serviceAreaBased) or contains all stops provided in a separate file (stopBased).

The input file for the 3rd case uses "transit schedule" format (as before), but this may be changed at some point in the future.

===

7555ad4 adds a simple `QuadTrees` util for creating quad trees - I have seen almost exactly the same code in dozens of places (incl. contribs, playgrounds etc.). Have not yet refactored that duplication though.

### DIFF
--- a/contribs/drt/src/main/java/org/matsim/contrib/drt/routing/ClosestAccessEgressFacilityFinder.java
+++ b/contribs/drt/src/main/java/org/matsim/contrib/drt/routing/ClosestAccessEgressFacilityFinder.java
@@ -26,26 +26,26 @@ import org.matsim.api.core.v01.Coord;
 import org.matsim.api.core.v01.network.Network;
 import org.matsim.contrib.drt.routing.DrtRoutingModule.AccessEgressFacilityFinder;
 import org.matsim.core.utils.collections.QuadTree;
+import org.matsim.core.utils.collections.QuadTrees;
 import org.matsim.core.utils.geometry.CoordUtils;
 import org.matsim.facilities.Facility;
+
+import com.google.common.base.Preconditions;
 
 /**
  * @author michalm
  */
 public class ClosestAccessEgressFacilityFinder implements AccessEgressFacilityFinder {
 	private final Network network;
-	private final QuadTree<? extends Facility> facilityQT;
+	private final QuadTree<DrtStopFacility> drtStopQuadTree;
 	private final double maxDistance;
 
-	public ClosestAccessEgressFacilityFinder(double maxDistance, Network network,
-			QuadTree<? extends Facility> facilityQT) {
-		if (facilityQT.size() == 0) {
-			throw new IllegalArgumentException("Empty facility QuadTree");
-		}
-
+	public ClosestAccessEgressFacilityFinder(double maxDistance, Network network, DrtStopNetwork drtStopNetwork) {
+		Preconditions.checkArgument(!drtStopNetwork.getDrtStops().isEmpty(), "Empty DrtStopNetwork");
 		this.network = network;
-		this.facilityQT = facilityQT;
 		this.maxDistance = maxDistance;
+
+		drtStopQuadTree = QuadTrees.createQuadTree(drtStopNetwork.getDrtStops().values());
 	}
 
 	@Override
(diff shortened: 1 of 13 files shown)


jfbischoff:
>The input file for the 3rd case uses "transit schedule" format (as before), but this may be changed at some point in the future.


I think I would prefer if we keep them in the transit schedule format. It is reasonably well-established within MATSim and several utils exist to read / write /display them.

kainagel:
I would (evidently) also prefer to stay with established formats.  But you know that before :-).

michalmac:
What I meant was: Currently, the transit schedule format works fine since stops from TransitSchedule are easy to convert to DrtStopFacilities (no extra info), so we will continue using it. However, at some point in the future this may not be the case...

kainagel:
Couldn't we make StopFacilities implement Attributable?

michalmac:
Motivation for changes:
- TransitSchedule for DRT was way too much for storing a collection of very simple facilities, hence DrtStopNetwork was introduced.
- TransitSchedule was also a bit misleading due to its use in DI bindings. I imagine the end user who wants to provide its own list of DRT stops via a TransitSchedule binding, being confused with all the remaining data in the transit schedule (and perhaps wondering whether/how to initialise the remaining fields)

Why only implementing Facility and not reusing TransitStopFacilities:
- DRT stops do not need to be physically transit stops or even be converted from transit stops, they can be all links in a given area ("serviceAreaBased" scheme).
- DRT buses can be routed in a very flexible way, so theoretically DRT stops could be attached to both links (=street) or a node (=intersection): so passengers may be suggested a waiting location
- sharing the same data object classes for two different purposes would unnecessarily couple two different functionalities: DRT and pt (changes/refactorings in pt may unnecessarily impact DRT and vice versa)

kainagel:
Without understanding all the details and ramifications: Couldn't this then be a file in the normal "facilities" format?  Again, the advantage would be that we have readers etc.  If it is not flexible enough, we could add Attributable (if not yet there).

csv always is easy initially.  Yet it runs into problems eventually, e.g. with versioning.

If you hate xml, we could consider having json as an alternative format.  Way to go would be:

1. allow reading/writing of facilities also in json
2. then use this to read/write DrtStopFacilities

michalmac:
Yes, the normal "facilities" format will work here too, and definitely will be less confusing.

I did not want to change the file format in this PR without talking to actual users, therefore only left a note. If there are no objections I will happily switch to the other format. Probably the sooner the better...

jfbischoff:
> Couldn't we make StopFacilities implement Attributable?

Transit Stops are attributable, adding a Node would not be a problem.

>Yes, the normal "facilities" format will work here too, and definitely will be less confusing.

I'm not sure it would be less confusing. From a user's perspective, having stop based drt in mind and not thinking of the underlying stop network, I would expect a stop to be a stop and a facility to be a facility. 
In either case, it's a recycling of a container that is not 100% fitting (Facilities allow for example capacities to be set). So one could also just leave it the way it is :-)
That said, I could probably live with the change, but it will require a proper documentation. There are more people using these things than we probably suspect.

kainagel:
My main comment in the end is that csv files lead to versioning problems in the long run.  Seems that you are already having them.  I am not sure if in core matsim we ever had a case where we could not read old files.  (Except config.)  But probably I am overlooking problems that we actually had.

vsp-gleich:
> > Couldn't we make StopFacilities implement Attributable?
> 
> Transit Stops are attributable, adding a Node would not be a problem.
> 
> > Yes, the normal "facilities" format will work here too, and definitely will be less confusing.
> 
> I'm not sure it would be less confusing. From a user's perspective, having stop based drt in mind and not thinking of the underlying stop network, I would expect a stop to be a stop and a facility to be a facility.
> In either case, it's a recycling of a container that is not 100% fitting (Facilities allow for example capacities to be set). So one could also just leave it the way it is :-)
> That said, I could probably live with the change, but it will require a proper documentation. There are more people using these things than we probably suspect.

I go with @jfbischoff here. Changing from TransitSchedule to Facilities causes work to be done, but does not seem to yield a real advantage.

vsp-gleich:
As far as I see, the linkId field of TransitStopFacilities is optional and TransitStopFacilities are already attributable. So it's easy to have drt stops as a TransitStopFacility without a linkId, but a nodeId or whatever attribute turns out to be useful.

PR #740, opened 2019-11-26
Make Bicycle Contrib use vehicle speeds during qsim

Janekdererste:


Janekdererste:
The regression tests have to be fixed, but this is difficult and I don't have time for it at the moment.

PR #743, opened 2019-12-03
Return to one layer of configuration

michaz:
The issue:
We have nested layers of configuration (for dependency injection): There is QSim, and there is Controler. Both are configured with Guice. QSim can run without Controler, but when it runs within Controler, it is instantiated from a Guice container that is a child of Controler's Guice container. The historical reason is that there are things, most typically `QSimNetwork`, that are _shared_ within one iteration, but must be re-created for a _new_ iteration. This problem is solved by using a child container.

This would be fine, except:
- Because the configuration of the child Injector for QSim is assembled from modules at the startup time of QSim (not of Controler), it is not visible to the layer of the Controler. In particular, it is not possible to inspect or graph the full runtime configuration at startup of the Controler, because it is not known yet.
- Conversely, the configuration for QSim cannot be controlled from the outer layer in the same way the configuration of other things can. In particular, we cannot override a binding for `QSimNetwork`, because it has not been bound at that point.
-- For this reason, `overridingQSimModules` are now kept separately by the controler, to which they are added through a separate method.
--- For *that* reason, there are now something like meta-modules, which encapsulate the adding of the two kinds of modules, while the actual modules are made `private` to facilitate this encapsulation.

All of this is too complicated. It re-introduces a custom language for precisely those things Guice is meant to provide a generic language for. So it is not only too complicated, but also too complicated to teach, because it never becomes clear what the problem was that DI was supposed to solve, and why DI solves it.

The projected solution:
- Only have one sort of modules: Modules.
- Use custom methods and custom classes like `AbstractMatsimModule` only to document what can be bound and how (providing shorthand-notations), not _require_ them to be used because we must keep bits of configuration in parallel to Guice. Always return Guice's own Builders from such methods, so all means of binding that Guice enables can be used by our users.
- Use `@Scope` to solve the initial issue instead of nested configuration and nested injectors.
- Remove as much extra infrastructure as possible.


michaz:
This should have little effect on what QSimModule does with annotations, modes and such. I think that's completely orthogonal to this issue.

michaz:
Related: Back then, I urgently wanted to magically have the `Config` available to Matsim `Modules`, in order to show that it is _allowed_ to rely on the config when deciding what to bind. This means that the normal Guice ways of combining modules (overrides, private modules) always require extra steps or they give a `NullPointerException`.

This must go away, too, because it creates a need for workarounds where none would be necessary.

sebhoerl:
Especially I'd be in favor of the last point, I already wanted to discuss this a while ago. I think it would actually *remove* a lot of confusion if we would have

```
controller.addOverridingModule(new SwissRailRaptorModule(config));
```

compared to "magically" injected config groups with the two layers of injection. And it only adds a minimal additional amount of code writing for a huge gain in clarity I think.

michalmac:
Nice! Yes - for simplifying as much as possible and being compliant with the standard guice.

A few thoughts:
- I once tried to use private modules (probably in QSimModules??) to hide some bindings, but gave up (too much customisation required)
- Wondering how QSim-scoped bindings will be accessible to EventHandlers (the controller scope). Currently this requires some workarounds. Maybe for this purpose we should have QSim-scoped event handlers that will be shut down at the end of QSim? 
(BTW. We also had some bugs in the past when even handlers did not clean up their state on `reset()` -- maybe another reason for having them short-lived)

kainagel:
A problem with
```
install( new XyzModule( config ) ) ;
```
is that people will use that as an opportunity to add more and more constructor arguments ...

... thus again leading to unstable API.

So if this is the path we take, then

(1) we need to decide on a dialect.  Some people pass only partial configs.  I have often ended up with situations where I needed other parts of the config.  Passing the whole config would alleviate this ... but would go against MZs point (personal communication) that the config hierarchy should reflect the object hierarchy.

(2) we need to decide if we regression-test the matsim API from outside the matsim repo ... so that, when people change constructor arguments, we get test failures.  (Which would also make it harder to change them, which would, I guess, be a desirable side effect.)  To some extent, matsim-code-examples is already assuming that role, but we would need to agree to make it more central; I can't do this alone.

michaz:
> people will use that as an opportunity to add more and more constructor arguments

Very true, thanks.

michaz:
Regarding `Config`, I think it _should_ be structured so its structure resembles the configured services, but we shouldn't build something that _relies_ on that by information-hiding bits of the Config from bits of the system. That's too restrictive.

Even though I could have used that yesterday, because I could have built two router instances seeing different versions of the router's (normally Singleton) config block. That's too messy. Will have to find another way.

sebhoerl:
> people will use that as an opportunity to add more and more constructor arguments

(1) But who is "people" in that case? Since the official contribs are all in this repository, it is easy to keep track of this, no? One could even add a comment to the constructors, telling people to please don't add new arguments to the official contrib modules. And if it still happens, it is easy to revert the change.

(2) For the part of changing API I'm also not sure, because *ideally* users would base their project on a fixed (weekly) version of MATSim (which I always advise to them). As soon as they update, they will see that something has changed because they will get compilation errors. In the case of constructors (although this should not happen) it would be particularly easy to find a solution, because it would be rather obvious which new things to pass? 

So given that (1) works, (2) would not be an issue, I'm not sure if understand exactly what you mean by "making the API unstable"?

[ Plus, on top *ideally* there would be a changelog in which one could look up the API changes that have happened since the last version update (and we could enforce extending a changelog with an additional check for PRs). But this discussion goes beyond giving the modules a simpler structure :) ]

kainagel:
Ok with me.

kainagel:
A couple of days later.  Having `Config` as a standard argument should not be a problem.  Re "adding further constructor arguments": I can follow Sebastian's argument; thank you for clarifying this.  I guess "the problem" starts when people give the same object into two of these constructors, and the first one modifies it in some way on which the second one relies:
```
install( new AbcModule( object1 ) ) ;
install( new AbcModule( object1 ) ) ;
```

PR #744, opened 2019-12-04
FixTripRouter

tschlenther:
deal with plans where the last activity is the same instance as the first activity (MATSim logic).

This fixes a bug which seems to occur when one creates plan with the default population factory.
Before, the trip router would not find the indexOfOrigin for the last trip of the day, since after setting the indexOfDestination to 0 (which is wrong) it would break out of the loop.
Now, we only break if both indices are altered and still have the consistency check if indices have the right order. Moreover, we check for the above mentioned special case.

This problem did not seem to occur for population which was parsed in from an input file.


tschlenther:
please forget about this. i did a mistake when creating the plans

PR #746, opened 2019-12-04
delete outdated (transit) routers

vsp-gleich:
Currently, we have multiple transit router implementations. For production runs (monomodal and intermodal pt) we almost always use SwissRailRaptor at Vsp, because it is faster, offers more features and/or is better maintained than other transit routers. I know that EventsBasedPtRouter is in use especially by the Singapore ETH group and many core matsim tests still use the old default transit router.

However, we have some more transit routers which - as far as I know - are no longer very actively maintained or used.

1. The [raptor implementation in minibus](https://github.com/matsim-org/matsim/tree/master/contribs/minibus/src/main/java/org/matsim/contrib/minibus/performance/raptor) is no longer referenced by core matsim code. We now have the SwissRailRaptor, another raptor implementation, in the matsim repository (thanks to SBB and Marcel Rieser!).
2. The [VariableAccessTransitRouterModule](https://github.com/matsim-org/matsim/tree/master/contribs/av/src/main/java/org/matsim/contrib/av/intermodal/router) is mostly a copy of the old default transit router which can be used for intermodal routing. However, SwissRailRaptor offers more features and is much faster and better maintained.

Furthermore, given the superior performance of the SwissRailRaptor, we might consider to 

3. remove the [old default transit router](https://github.com/matsim-org/matsim/tree/master/matsim/src/main/java/org/matsim/pt/router), too.

I started to work on removing 1) and 2) and might work on removing 3) later.

Are there any opinions on this? Is somebody using 1), 2) or 3) and would like to keep any of these transit routers?


sebhoerl:
Plus, I once wanted to pull back to MATSim something like this...:
https://github.com/eqasim-org/eqasim-java/blob/develop/core/src/main/java/org/eqasim/core/components/transit/routing/EnrichedTransitRoute.java

... and remove the `ExperimentalTransitRoute`. At the end I got stuck with the minibus Raptor to recover all the additional information for the interface above. So this may be a good chance, I can have a look again if you like.

vsp-gleich:
> https://github.com/eqasim-org/eqasim-java/blob/develop/core/src/main/java/org/eqasim/core/components/transit/routing/EnrichedTransitRoute.java
> ... and remove the `ExperimentalTransitRoute`

Yes, sure, I really like the idea :+1:  It would be great to have a more expressive (Passenger)TransitRoute, hopefully already in the next matsim release.

Maybe we should address this directly after deleting the old outdated transit routers. I want to wait  a bit for others to give there opinion before removing them.

mrieser:
I don't oppose to delete my old, slow router. So from my side, go ahead.
Also looking forward to Sebastian's `EnrichedTransitRoute`.

kainagel:
Please do 1 & 2.

I am not so sure about 3.  There is a "research group" about public transit routing, run by mathematicians plus Peter Vortisch and Markus Friedrich.  They find that once one wants more flexible routing, the Dijkstra search structure is better than the raptor search structure.

vsp-gleich:
Ok, so let's restrict to remove 1) and 2) and keep the old default transit router.

1) is already done (on this branch here)
2) before removing this one, I need @michalmac 's help to get the SwissRailRaptor running with taxi (=leg.setTravelTime to something different than NaN in taxi).

jfbischoff:
>before removing this one, I need @michalmac 's help to get the SwissRailRaptor running with taxi (=leg.setTravelTime to something different than NaN in taxi).

That would be wise anyway 👍

jfbischoff:
(though, one could also replace the example by DRT rather than taxi, then it works out of the box)

PR #748, opened 2019-12-05
fix a bug in NetworkSimplifier resulting in corrupt networks.

mrieser:
In special cases, where link attributes were different in opposite directions of links, the simplification algorithm merged links it shouldn't had, resulting in networks no longer being routable.


vsp-gleich:
manual "automerge"

mrieser:
thanks!

PR #753, opened 2019-12-06
minor possible simplification in FreeSpeedFactorRoutingModule

nkuehnel:
Am I missing something or can this calculation in FreespeedFactorRoutingModule be simplified without calculating the travel time of final link twice?


kainagel:
Sure.  Sorry that this took so long.  Setting automerge label right now.

PR #754, opened 2019-12-09
Changes for Drt Reservations

tschlenther:
1) protected access modifier for VehicleData.entries  (alternatively, one could think about providing encapsulated protected access via getter)

2) The initialisation of the schedules should be triggered from the RequestInsertionScheduler constructor, not by the RequestInserter. That is because now, if i plug in my own RequestInserter, the DefaultUnplannedRequestInserter still is instantiated (because it is still added as ModalComponent - however the guice binding is overwritten)
and consequently also triggers the schedule initialisation. So in my own RequestInserter, i would need to rely on the fact that the schedule initialisation is triggered by the DefaultUnplannedRequestInserter, which does not feel good, or trigger it myself. The latter leads to an exception in the current state,
because schedules are then initialized twice, leading to 2 stayTasks to be inserted in the schedule.
If the initialisation is moved to the RequestInsertionSchedulerconstructor however, everything seems to work fine. So is there any strong argument, not to do this?


michalmac:
> That is because now, if i plug in my own RequestInserter, the DefaultUnplannedRequestInserter still is instantiated (because it is still added as ModalComponent - however the guice binding is overwritten) and consequently also triggers the schedule initialisation.

Good point! Yes, you should not override bindings which are added to QSimComponents bindings due to possible side effects... 

DefaultUnplannedRequestInserter is a QSim component because it needs to shutdown a thread pool at the end of qsim. I will remove that functionality outside.

However, if you want to substantially adapt the DRT functionality to your needs, it's safest to create your own version of `DrtModeQSimModule`. Overriding bindings is tricky to add and even worse to maintain. I will extract some components (e.g. "inlined" providers) outside `DrtModeQSimModule` in order to encourage writing own guice modules instead of relying solely on overriding bindings.

> The initialisation of the schedules should be triggered from the RequestInsertionScheduler constructor, not by the RequestInserter

The cleanest approach would probably be triggering it on MobsimInitializedEvent, so that it happens "outside guice". Will look into this too :-)

tschlenther:
Okay sounds good. The fact that DrtModeQSimModule does bind many substantial things and i only need to change 2 of these led me to overriding instead of copy pasting..
From my understanding, this also how most of MATSim is designed.

michalmac:
(1) should be solved with a custom EntryFactory
(2) is solved via #757

michalmac:
> (1) should be solved with a custom EntryFactory

@tschlenther here is an example of a customised factory (for eDRT): https://github.com/matsim-org/matsim/blob/1c5ea076bfc3f20be4832a165bbbfb4f2f6e3fad/contribs/drt/src/main/java/org/matsim/contrib/edrt/optimizer/EDrtVehicleDataEntryFactory.java

PR #757, opened 2019-12-10
add QSimScopeForkJoinPoolHolder

michalmac:
and trigger RequestInsertionScheduler.initSchedule() from constructor

Addresses issues mentioned in #754


jfbischoff:
What a name :-D

PR #758, opened 2019-12-12
make interaction labels shorter in otfvis

kainagel:
cannot test this right now since matsim-berlin is behind matsim head.


kainagel:
Danke für den Hinweis.

PR #762, opened 2019-12-13
synchronise Id creation

michalmac:
In response to https://github.com/matsim-org/matsim-code-examples/issues/241

Creation of new `Id`s is synchronized. However still `IdMap` and `IdSet` may not work correctly. The 
read-access to the array lists inside the `cacheIndex` map is not synchronised in the method `Id<T> get(int index, final Class<T> type)`, which may return nulls for already inserted ids.

Regarding a potential fix: I am not sure what was the idea behind these two collections?
 - memory efficiency -- then we can add synchronisation on the array list inside the mentioned method
 - computational efficiency -- then probably an array of `Id`s should be stored internally in `IdMap` and `IdSet` (and we can remove this method)

### DIFF
--- a/matsim/src/main/java/org/matsim/api/core/v01/Id.java
+++ b/matsim/src/main/java/org/matsim/api/core/v01/Id.java
@@ -20,31 +20,31 @@
 
 package org.matsim.api.core.v01;
 
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+
 import org.matsim.api.core.v01.network.Link;
 import org.matsim.api.core.v01.network.Node;
 import org.matsim.api.core.v01.population.Person;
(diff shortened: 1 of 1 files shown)


michalmac:
> In Java, whenever a thread ends, its memory is consolidated. So to falsely return null in get(index, type), there would have to be at least two threads, started about the same time, one creating the id, the other accessing it by index. As soon as the thread creating the id ends before the second thread starts, it should not be a problem. Can this happen in your case?

Theoretically, a parallel QSim is a place that IDs could be created and read by different threads at the same time. Hard to find a very convincing example, though:-)

Anyway, I slightly modified the create method so that we do not need to synchronise getters.

michalmac:
@mrieser  thanks for looking into this PR! I think it's ready for merging now.

michalmac:
>  Can this happen in your case?

I think I have found a case where Id creation is mixed with querying them: https://github.com/matsim-org/matsim/pull/769 😄

PR #764, opened 2019-12-16
Affects VSPadjustments in OSMNetworkReader

dziemke:
Reduce link capacity increase threshold
Handle long roundabout links based on junction tag
Some comments and warnings


Janekdererste:
Hey @dziemke do you still want these changes to be incorporated into network parsing? If so, I feel responsible for this module so you could tell me how to go about this.

PR #779, opened 2019-12-28
Remove DynRoutingModule

michalmac:
DvrpRoutingModule offers a way broader functionality and can also work as the old DynRoutingModule.
All taxi plans will be pre-routed

@vsp-gleich no more NaN distances/times in taxi legs, which should unblock this PR: https://github.com/matsim-org/matsim/pull/746


vsp-gleich:
Thanks @michalmac :+1:

PR #783, opened 2020-01-06
bug fix EditTrips

smuellervsp:


michalmac:
@smuellervsp , was this bug related to getting `-Inf` only. Or were there cases with `+Inf`?

Just asking as I am trying to find out which calls to `Double.isInfinite()` could be reduced to `Time.isUndefinedTime()` and hence easily replaced with `OptionalTime`.

PR #790, opened 2020-01-13
Fix integration test

JWJoubert:
See JIRA #998. This is an intermediate fix while I’m still sorting out the roadpricingCleanup branch. With this fix the integration test at least is aware of the TollFactor that was loaded as part of the road pricing scheme.


JWJoubert:
Link to the issue: https://matsim.atlassian.net/browse/MATSIM-998

PR #806, opened 2020-02-04
Created new test class for ModeStatsControlerListener

vsp-gleich:


vsp-gleich:
Does not work directly, because continuous integration is not set up for aravind's fork. Instead 1) opening a new branch in original matsim repository, 2) pull request aravinds fork -> branch, 3) new pull request

PR #811, opened 2020-02-06
Person money event extension

vsp-gleich:
add information for what purpose a person received / paid money and from / to whom -> differentiate different kinds of payments, e.g. analyse toll revenue and drt fare revenue per operator etc.


vsp-gleich:
What is the best default value for the new fields?

- `"null"` (a String, not the null pointer!): This is what comes out if I read in an EventsFile with old personMoneyEvents without the new fields (`null` in the code which is written out as String `"null"`). Disadavantage: hard to tell whether it is null due to a (null pointer) bug in the code or because it should be null
- `""` the empty String: shows more that someone wanted to return this (differentiate from a bug).
- Let's not use the null pointer, because after writing out and reading in again it would change to a String `"null"` :-( 

Any opinions on this?

kainagel:
I think there is some ticket in jira that discusses this, and gives a clear answer for matsim conventions.  Can't find it right now because I am in "the East" where there is no Internet in trains.

However, the above would be for "required" attributes according to the dtd.  I think that for events the convention is to just not write out at all (look into the Event class).

What is the best default value for the new fields?

  *   "null" (a String, not the null pointer!): This is what comes out if I read in an EventsFile with old personMoneyEvents without the new fields (null in the code which is written out as String "null"). Disadavantage: hard to tell whether it is null due to a (null pointer) bug in the code or because it should be null
  *   "" the empty String: shows more that someone wanted to return this (differentiate from a bug).
  *   Let's not use the null pointer, because after writing out and reading in again it would change to a String "null" :-(

Any opinions on this?

vsp-gleich:
> I think that for events the convention is to just not write out at all
I've just changed that and replaced the empty Strings with `null` as default value for empty Strings.

vsp-gleich:
There is an issue with contrib protobuf. It apparently auto-generates `ProtobufEvents` based on the _old_ `PersonMoneyEvent` without the new fields. Does anyone know more about how protobuf works or has recently changed an event?

michalmac:
> There is an issue with contrib protobuf. It apparently auto-generates `ProtobufEvents` based on the _old_ `PersonMoneyEvent` without the new fields. Does anyone know more about how protobuf works or has recently changed an event?

Updated the proto definition. Should be fixed now.

michalmac:
Re default values:
In proto, the default value for strings is "", and you cannot set it to null (NullPointerException).

If we want to have null as the default value in the matsim events, some extra handling is necessary that bi-di maps:  ""<=> null while doing conversions.

vsp-gleich:
> Updated the proto definition. Should be fixed now.

Thanks a lot @michalmac :+1: !

vsp-gleich:
> Re default values:
> In proto, the default value for strings is "", and you cannot set it to null (NullPointerException).
> 
> If we want to have null as the default value in the matsim events, some extra handling is necessary that bi-di maps: _""<=> null_ while doing conversions.

Thanks for the hint, I was not aware of this. I have just looked through some other event types. `PersonStuckEvent` also has an optional String field (`legMode`). It is written analogous to the new `PersonMoneyEvent` fields both in matsim core and in contrib protobuf. So they should face the same difficulties.

[PersonMoneyEvent.java#L90-L102](https://github.com/matsim-org/matsim/blob/2f27f259d2a54137fa7b7eb7c2e216d2248a208e/matsim/src/main/java/org/matsim/api/core/v01/events/PersonMoneyEvent.java#L90-L102)

[PersonStuckEvent.java#L66-L77](https://github.com/matsim-org/matsim/blob/2f27f259d2a54137fa7b7eb7c2e216d2248a208e/matsim/src/main/java/org/matsim/api/core/v01/events/PersonStuckEvent.java#L66-L77)

[contribs/protobuf/src/main/proto/events.proto#L119-L132](https://github.com/matsim-org/matsim/blob/2f27f259d2a54137fa7b7eb7c2e216d2248a208e/contribs/protobuf/src/main/proto/events.proto#L119-L132)

Maybe we stick for this pull request to the current MATSim default (`null`, `null` is not written out) and move the more general discussion about default values to [JIRA 1007](https://github.com/matsim-org/matsim/pull/811#issuecomment-583747605).

PR #820, opened 2020-02-12
introducing commercialTrafficApplications contrib

tschlenther:
introduced a new contrib which is supposed to be the place for code that combines the functionality of the freight contrib with other contribs or extends it.
moved in the code which links the demand for passenger and freight transport, see package jointDemand.
I plan on combining some functionality of the freight module with the drt contrib by making some adjustmens on it; that could also take place here.

The plan is to keep the different applications for freight (or commercial traffic respectively) in their own package.

@kainagel Please let me know if you are fine with the structure.

Thank you :)


JWJoubert:
@tschlenther @kainagel  I just want to check - is this the preferred approach? To create a new contribution when two (or more) other contribs are "integrated" or combined? Is one consequence not that we end up with a contrib for every possible permutation of contrib-integrations? I just think that maintaining "freight", for example, will then become quite hard.

But if there are specific benefits to it, then great, maybe we can change our approach to better aligned.

tschlenther:
> @tschlenther @kainagel I just want to check - is this the preferred approach? To create a new contribution when two (or more) other contribs are "integrated" or combined? 

This is indeed the right question. I am not sure about the answer. The problem is, that some functionality (not yet) contained in here will be used at multiple other, external places.

> I just think that maintaining "freight", for example, will then become quite hard.

I am not sure if i understand why. The goal of this contrib would be to apply 'freight' and combine in the way it exists. We have contribs depending on various other contribs already, especially on drt. Indeed, it does not seem like the 'drt' contrib is the easiest to maintain..

Another option would be to integrate the "jointDemand" code, which is currently in this new contrib, into the freight contrib. But,
a) i am not sure if this what the 'freight' contrib was meant for (wasn't it rather to bridge the gap to vrp-solving algorithms?) 
b) it would leave me with the question where to put the code for combining dvrp with freight

Moreover, in any case, we still have to look where to put the UAM code @kainagel.
You had proposed to put anything which is combining several contribs into the 'integration' contrib, but it turned out that it is not really suitable as i had to manually specify which tests should be included.

JWJoubert:
> ...I am not sure if I understand why. The goal of this contrib would be to apply 'freight' and combine in the way it exists...

This makes sense. Sorry, maybe I misinterpreted the pull request then. It seemed like the idea was to change some of the freight code so as to accommodate the combination. If it is simply adding on, then it makes much (more) sense.

tschlenther:
Maybe we should give this contrib another name?
In a way that also the UAM (airDRT) stuff fits in..

A suggestion:  "intersegmentalTraffic". This would cover the meaning of freight+passenger and ground+air respectively, while the latter would normally rather be 'intersectoral'.

Or we should rethink the design of the integration contrib to make it useful for any code that combines contribs.

tschlenther:
Actually i have to close this pull request as the history was not properly cleaned before the movement of code took place. I opened another pull request #821 so the discussion can go on.
Sorry

kainagel:
Despite the fact that this is closed, a quick answer.  Tilmann already stated it; just to re-iterate to be clear: the new package both needs freight and dvrp.  However, we did not want dvrp to depend on freight, and we did not want freight to depend on dvrp.  It is also more than just an "example".  So it had to be a third contrib.  We also tried "integration", which would have been a good place, but it has some special setup for the way it runs tests, and we did not want that.

If we run into the "combinatorial combination" problem, then we should consider reconfiguring the "integration" contrib.

JWJoubert:
Thanks @kainagel, and yes, sure, I misunderstood the original intent
> the new package both needs freight and dvrp.  However, we did not want dvrp to depend on freight, and we did not want freight to depend on dvrp.

This indeed make sense.

PR #821, opened 2020-02-13
introducing commercialTrafficApplications contrib

tschlenther:
Please see discussion at #820


kainagel:
Internet in train is stuck.  IMO, it is ok to progress.  As we already discussed.  Thank you so much for putting this into a module.

PR #827, opened 2020-02-17
make automatic facility generation more standard

kainagel:
use methods `PopulationUtils.decide...`


kainagel:
more careful approach on different branch

PR #830, opened 2020-02-19
Fix location contrib WIP

kainagel:
can now handle activities with a duration instead of  an endTime
also the route to and from the changed activity is now calculated correctly


mdziakowski:
> todo
can now handle activities with a duration instead of  an endTime
also the route to and from the changed activity is now calculated correctly

kainagel:
I think that MD did not continue on this PR, but created #831 instead.

ikaddoura:
@mdziakowski can you please resolve the conflicts, thanks!

mdziakowski:
see pull request
Replace trip to legs in location choice contrip #895

PR #833, opened 2020-02-21
reduce public inheritance of dijkstra etc

kainagel:
The internals of the Dijkstra class are quite exposed.  I have now provided final public facades, and made the non-final implementation classes package-private.  

I can't say if this will lead to problems.  I would say that if external code inherits, we should either move it into the core, or discard it.  But maybe I am too neurotic here.


kainagel:
This does not seem to go into a good direction.

PR #842, opened 2020-02-24
make subpopulation attribute name no longer defineable

kainagel:
now absorbed in `PopulationUtils.getSubpopulation(person)`.  Is this what you had in mind re https://matsim.atlassian.net/browse/MATSIM-206 ?


kainagel:
@tduberne : Would this be ok with you?  It is done in the same way we do it elsewhere, which is a method `PopulationUtils.getSubpopulation( person )`, and the key that is needed is then a static constant inside `PopulationUtils`.

PR #872, opened 2020-02-26
Parkingproxy

Royal2Flush:
Introduces a third option in the parking contrib that does not actually simulate parking but rather estimates the time cost of searching for a parking spot (and the prolonged egress walk) based on the number of cars already there.


jfbischoff:
Very nice!
In long-term, we could possibly see where the old parking choice simulation is actually in use and consider removing it.

PR #880, opened 2020-02-26
Apply OptionalTime to times stored in Activities

michalmac:
- Introduce `OptionalTime` (similar to `OptionalDouble`, but not exactly the same)
   - impossible to create `OptionalTime.of(Double.NaN)`
   - `OptionalTime.isUndefined()`/ `OptionalTime.isDefined()` depends on the stored `seconds` field (whether it is equal/non-equal Time.UNDEFINED_TIME)
   - `OptionalTime.seconds()` throws an exception if the time is undefined
   - There are a couple of helper methods (similar to the ones in `OptionalDouble`) like: `orElse()`, `orElseGet()` etc.

- Apply `OptionalTime` for the activity start/end times and maximum duration fields (other potential places for replacement in the future: Routes and Legs)

- remove lots of `Time.getUndefined()`
- make `Time.UNDEFINED_TIME` package-protected


kainagel:
Given that with this we now have an idea how to store "undefined", a question is how to write it to file.  There is the following jira ticket: https://matsim.atlassian.net/browse/MATSIM-818 .  Is this also our convention for reading/writing "undefined"?

michalmac:
Add setters to `Activity`:
- `setStartTimeUndefined()`
- `setEndTimeUndefined()`
- `setMaximumDurationUndefined()`

Add some additional helper methods to `OptionalTime`, including `ifDefinedOrElse()` that is super useful in such cases like this: https://github.com/matsim-org/matsim-libs/blob/6f14809efd1dca65cdd38e48c00c317f221232be/matsim/src/main/java/org/matsim/core/population/PopulationUtils.java#L872-L874

PR #883, opened 2020-02-27
Trip csv writer + person hours travelled/waited statistics

vsp-gleich:


kainagel:
Note that I modified .travis.yml so that (hopefully) only the failing test will run, but with more output.

PR #895, opened 2020-02-28
Replace trip to legs in location choice contrip

mdziakowski:
pls merge this instead of Fix location contrib WIP #830
the changes in the Fix location contrib branch did only moved the problems to another place

 in this branch the problems are fixed

please keep in mind that i needed to change a test:

changed in the FrozenEpsilonLocaChoiceIT class the testFacilitiesAlongALine() test, because of a change in the calculation of the route. The parameters were in the wrong order, so no path was found and the travel time for the calculation of the score was 0, now it is done correctly


kainagel:
looks ok to me.  setting "automerge" label

PR #898, opened 2020-03-01
WIP: Parkingproxy

kainagel:
`RunWithParkingProxyIT` generates different events on travis than locally.  The branch here only runs that specific test on travis, and produced more output than normal.  The difference is 
```
2020-03-01T08:29:10,976  WARN EventsFileComparator$EventsComparator:166 The event:
2020-03-01T08:29:10,979  WARN EventsFileComparator$EventsComparator:167  | link=69 | time=32504.0 | type=entered link | vehicle=119 | 
2020-03-01T08:29:10,981  WARN EventsFileComparator$EventsComparator:168 is missing in events file:test/output/org/matsim/contrib/parking/parkingproxy/run/RunWithParkingProxyIT/testMain//output_events.xml.gz
2020-03-01T08:29:10,985  WARN EventsFileComparator$EventsComparator:186 The event:
2020-03-01T08:29:10,985  WARN EventsFileComparator$EventsComparator:187  | link=122 | time=32504.0 | type=entered link | vehicle=119 | 
2020-03-01T08:29:10,988  WARN EventsFileComparator$EventsComparator:188 is missing in events file:test/input/org/matsim/contrib/parking/parkingproxy/run/RunWithParkingProxyIT/testMain//output_events.xml.gz
```
implying that it is a "real" difference (vehicle entering a different link) rather than, say, just two exchanged events.

Errors that occur on the server only are unfortunately difficult to debug.  One thing one could do here is output the full plan of the affected vehicle on the console and compare the routes between local and travis run.

Possible causes are in our experience:
* race conditions from multi-threading, i.e. two threads writing to the same data structure, but doing this in a different sequence than locally.  (I have tried to switch this off for the present test, see the test class.)
* non-deterministic data structures.  A typical one is `HashSet`, which may result in different iteration orders on different machines.  (Only when one iterates over it; it is ok for lookups.  It is, however, also not thread safe.)
* Two routes with exactly the same cost, and different machines pick different variants.  It is not implausible that this is the case here since the underlying grid test case has that problem.  If it is verified that this is the problem (i.e. both routes have same cost), then I would suggest to indeed change the test case.

And at some point we should make the router deal deterministically with such cases.  But we first need to find out if that is indeed the problem.


Royal2Flush:
I have replaced one instance of MatsimRandom.getLocalInstance() which uses a seed depending on how often the method has been called before with an own, truly local instance of Random. The output before and after the rebase to the current master are identical (except for the fact that after there is an enterVehicleTie attribute to some elements in the experienced plans file).

I'm curious what the build server says...

PR #911, opened 2020-03-05
Add TransitPassengerRoute interface

sebhoerl:
This PR responds to [MATSIM-790](https://matsim.atlassian.net/browse/MATSIM-790) and replaces the deprecated `ExperimentalTransitRoute` with a general `TransitPassengerRoute` interface with a default implementation `DefaultTransitPassengerRoute`.

The `DefaultTransitPassengerRoute` uses the Jackson serializer to save access stop id, line id, etc. in the route description rather than a custom-made serializer as in the old `ExperimentalTransitRoute`. An example now looks like this:

```
{"transitRouteId":"1980","boardingTime":"08:00:00","transitLineId":"11","accessFacilityId":"5","egressFacilityId":"1055"}
```

Additionally, the new interface has a `boardingTime` attribute, which was valuable information that was missing in the routes before. Currently, **only** SwissRailRaptor writes this information into the route, **not** the old Dijkstra-based router, and also **not** the events-based public transport router. For me it is not completely clear how and where to recover this information properly. Maybe somebody can look into this before this PR is committed?

`EventsToLegs` recovers the `boardingTime` properly so it could potentially also be used in scoring in the future, rather than tracking the vehicle enter events additionally.


sebhoerl:
@mrieser Could you have a look? :)

PR #916, opened 2020-03-06
Fix attributes converter crashes

### LINKED ISSUE #841: ObjectAttributesConverter:75 No AttributeConverter found for class java.util.HashMap. Not all attribute values can be converted.
I think that we now have Maps as attributes, and so the above should be fixed.

Janekdererste:
fixes #841


vsp-gleich:
drt speed up works again :+1:

PR #919, opened 2020-03-06
Fix interaction activities

### LINKED ISSUE #917: Drt interaction activities have arbitrary durations
Drt interaction activities seem to be of arbitrary duration. This only concerns drt interaction activities after getting off the vehicle (i.e. after the dropoff). The BusStop activity duration for the driver is fine, just as the activity duration of customer's the interaction activity before the leg.

I checked the following things:
1) During routing before the sim, everything looks fine. The generated plans are consistent and do not show strange durations.
2) The new ScheduleTimingUpdater also does not have anything to do with this

The problem occurs in a newly created equil test scenario as well as when running the examples in matsim-maas.

vsp-gleich:
fixes #917

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -72,8 +72,6 @@ cache:
   - $HOME/.m2
 
 env:
-  global:
-    - MAVEN_OPTS="-Xmx2g"
   matrix:
     - MODULE=matsim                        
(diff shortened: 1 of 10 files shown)


kainagel:
@vsp-gleich: Could you please also remove @tschlenther's warning again in `DrtConfigGroup`?  Thanks ...

PR #928, opened 2020-03-13
avoid assert statements in non-test code. See MATSIM 920

kt86:
"assert" may be ignored outside of tests when not enabled explicitly or by the IDE.

Solutions:
1. Use `Gbl.assertIf(boolean)` -> not possible to hand over a message for the user OR
1. Replace assert by if statement (Suggestion from IntelliJ)

In this step I decided to go for the if statement.

@kainagel I think it would be useful to also have an `Gbl.assertIf(boolean, message)`. So one can use the MATSim infrastructure and also hand over a Message including some information about the assertion error, e.g. print out some information of the value that causes the error.


michalmac:
BTW. There is a nice support in guava for such things: https://github.com/google/guava/wiki/ConditionalFailuresExplained

Classes like `Preconditions` or `Verify` offer a lot of methods, some with template strings.

kt86:
Thanks, @michalmac.

PR #940, opened 2020-03-28
Speed up emission grid calculation

Janekdererste:
The processing of emissions took up to an hour depending on the size of the scenario. Now, things are processed in parallel and at least on my machine things run in 10 minutes.


michalmac:
Using `PreparedGeometry` gives usually a nice boost in processing speed (I wouldn't be surprised if it offers 1+ order of magnitude speedup)

Not sure if applies to your case, but have a look at this: 
https://github.com/matsim-org/matsim-libs/blob/6d6a998940838d3e286edf00a19236cc298f77fd/matsim/src/main/java/org/matsim/utils/gis/shp2matsim/ShpGeometryUtils.java#L42

Janekdererste:
Thanks for the advice @michalmac. I didn't know this data structure. Unfotunately this doesn't really speed up things in my use case. Though I haven't used a complex geometry but rather only a rectangle as bounds in my current analysis of the nemo project.

The bottle neck must be something else and would require an investigation by profiler. For now 10 minutes is fast enough, I think.

michalmac:
Ah, right. Makes sense. In my case is sped up things greatly, but that was probably due to the use of "real" zones. Thanks for trying out and reporting :-)

Issue #929, opened 2020-03-14
Calculation of departure times in PlanRouter


sebhoerl:
I observed a potential problem in the `PlanRouter` when the departure times for the routing are calculated.

Assume the plan:
```
     Activity (end time 08:00)
     Leg (duration 00:05)
     Activity (end time 09:00)
     Leg (duration 00:05)
     Activity (end time 10:00, duration 00:10)
     Leg
     Activity
```

With a value for `ActivityDurationInterpretation` of `tryEndTimeThenDuration` the calculation is easy: The first leg starts at 08:00 and the second leg has departure time 09:00 and the third 10:00.

Now with `minOfDurationAndEndTime` wouldn't we expect the first departure time to be 08:00, the second 09:00, but the third one 09:15? As far I understand, this is how things are interpreted in the QSim and in the TimeMutator. However, `PlanRouter` uses a departure time of `10:00` here.

The reason is that it calls `PlanRouter.calcEndOfActivity` which has a "shortcut" in the beginning, which *always* returns the activity end time *if* it exists (which *always* needs to be the case if `minOfDurationAndEndTime` is used).

Or did I interpret wrongly what should happen here? I added a branch with a unit test that shows the current behaviour (`PlanRouterTest`): 
https://github.com/matsim-org/matsim-libs/tree/plan-router-departure-times

sebhoerl:
There is a related issue (added in the same branch and class as `testSummingUpTravelTimesAndActivityEndTimes`).

Assume the plan
```
Activity (end time 1000)
Leg (routed travel time 1000)
Activity (end time 1200)
Leg (routed travel time 1000)
Activity (end time 1300)
Leg
Activity
```

Since the `PlanRouter` *routes* the trips, we know that the agent arrives at the second activity at 2000. Which would mean the activity has a duration of zero in the QSim, so we already know that the departure for the second trip would be at 2000. Again, for the third trip we know that the departure would be at 3000.

What the `PlanRouter` does, however, is to reset the departure times for the trips to 1200 and 1300, respectively. While routing we therefore assume the wrong departure time. This is unfortunate as the route is based on the time-of-day-dependent network travel times.

There is a trade-off: If we use the actual departure time, the written plan will be inconsistent, because the leg departure time will not fit to the activity end time. However, if we use the activity end time as the departure time, the route is not consistent.

Is this something that has already been discussed?

sebhoerl:
[...]

PR #943, opened 2020-03-30
Refine `PopulationUtils.decideOnTravelTimeForLeg()`

michalmac:
- The method `PopulationUtils.decideOnTravelTimeForLeg()` returns `OptionalTime`, specifically `OptionalTime.undefined()` instead of `-Inf`. The caller should decide what to do when the optional time is undefined (e.g. fail, use 0, etc.).

- The method was slightly adapted to return `leg.travelTime` if `leg.route` exists but `leg.route.travelTime` is undefined. This approach is already commonly used in many other places (so now in all of these places we have a call to `decideOnTravelTimeForLeg()`)

- Turns out that location choice had a bug. `Leg.travelTime` was not set and the undefined time was used instead of `route.travelTime`. The bug was revealed after `decideOnTravelTimeForLeg()` was refined.


sebhoerl:
There is actually a whole bunch of things that are related: #929 
Any comments on those?

michalmac:
> There is actually a whole bunch of things that are related: #929

I am not a re-planning expert, so hard to have a strong opinion on the inconsistencies mentioned there.

Currently, I am just gradually introducing `OptionalTime` to address the hazards of using `-Inf`. I try to do it in a way that reduces a potential overuse of `OptionalTime.orElse(0)` or `OptionalTime.orElse(-Inf)` in the future.

contribs/locationchoice/src/test/java/org/matsim/contrib/locationchoice/frozenepsilons/FrozenEpsilonLocaChoiceIT.java
```diff
@@ -413,16 +413,16 @@ int binFromVal( double val ) {
...
-						check( 376, cnt[1] ) ;
-						check( 414, cnt[2] ) ;
-						check( 386, cnt[3] ) ;
-						check( 242, cnt[4] ) ;
-						check( 108, cnt[5] ) ;
-						check( 36, cnt[6] ) ;
-						check( 20, cnt[7] ) ;
-						check( 6, cnt[8] ) ;
-
+						check( 1104, cnt[0] );
```
mrieser, line 416: these numbers seem to change quite a bit. Why? Shouldn't the change in time-representation be idempotent?

kainagel:
Re this one here: I think that Ihab and Marc found the same bug independently (already some time ago).  Possibly, they have not (yet) merged their changes.

Maybe you can check if you find (roughly) the same new results?

In general, this should approximately be an exponential function.  So an approximate division by two from one bin to the next is not obviously implausible.

Thanks and best wishes

Kai

On 4. Apr 2020, at 15:01, Marcel <notifications@github.com<mailto:notifications@github.com>> wrote:



@mrieser commented on this pull request.

________________________________

In contribs/locationchoice/src/test/java/org/matsim/contrib/locationchoice/frozenepsilons/FrozenEpsilonLocaChoiceIT.java<http://quarantine.tu-berlin.de:32224/?dmVyPTEuMDAxJiZlZDU2ODA1OWE1NmU0YWYyZj01RTg4ODUxRV8yMjA2MV8xMTI0Ml8xJiZmM2FkMGNmMjYxMTg5ZDI9MTIzMiYmdXJsPWh0dHBzJTNBJTJGJTJGZ2l0aHViJTJFY29tJTJGbWF0c2ltLW9yZyUyRm1hdHNpbS1saWJzJTJGcHVsbCUyRjk0MyUyM2Rpc2N1c3Npb24lNUZyNDAzNDY3NzI5>:

> @@ -413,16 +413,16 @@ int binFromVal( double val ) {
                                                }
                                                // Note that the following "check" method is deliberately a bit imprecise (see implementation), since we are only interested in the
                                                // (approximate) distribution.  kai, mar'19
-                                               check( 410, cnt[0] );
-                                               check( 376, cnt[1] ) ;
-                                               check( 414, cnt[2] ) ;
-                                               check( 386, cnt[3] ) ;
-                                               check( 242, cnt[4] ) ;
-                                               check( 108, cnt[5] ) ;
-                                               check( 36, cnt[6] ) ;
-                                               check( 20, cnt[7] ) ;
-                                               check( 6, cnt[8] ) ;
-
+                                               check( 1104, cnt[0] );


these numbers seem to change quite a bit. Why? Shouldn't the change in time-representation be idempotent?

—
You are receiving this because your review was requested.
Reply to this email directly, view it on GitHub<http://quarantine.tu-berlin.de:32224/?dmVyPTEuMDAxJiZlZDU2ODA1OWE1NmU0YWYyZj01RTg4ODUxRV8yMjA2MV8xMTI0Ml8xJiZmM2FkMGNmMjYxMTg5ZDI9MTIzMiYmdXJsPWh0dHBzJTNBJTJGJTJGZ2l0aHViJTJFY29tJTJGbWF0c2ltLW9yZyUyRm1hdHNpbS1saWJzJTJGcHVsbCUyRjk0MyUyM3B1bGxyZXF1ZXN0cmV2aWV3LTM4NzY5NzU3MQ==>, or unsubscribe<http://quarantine.tu-berlin.de:32224/?dmVyPTEuMDAxJiZlZDU2ODA1OWE1NmU0YWYxZj01RTg4ODUxRV8yMjA2MV8xMTI0Ml8xJiYxM2FjYWMwMjI1Zjg3ZDQ9MTIzMiYmdXJsPWh0dHBzJTNBJTJGJTJGZ2l0aHViJTJFY29tJTJGbm90aWZpY2F0aW9ucyUyRnVuc3Vic2NyaWJlLWF1dGglMkZBQ0M2N01PVktFVlNPQVZBU0hEN1o3RFJLNFZSWkFOQ05GU000TFdSVzZKUQ==>.

======================================================

Mobility traces and spreading of COVID-19 https://doi.org/10.1101/2020.03.27.20045302
https://matsim-vsp.github.io/covid-sim/

github.com/matsim-org/matsim-example-project<http://github.com/matsim-org/matsim-example-project> | github.com/matsim-org/matsim-code-examples<http://github.com/matsim-org/matsim-code-examples> | matsim.org/downloads<http://matsim.org/downloads> | matsim.org/docs/tutorials<http://matsim.org/docs/tutorials> | matsim.org/open-scenario-data<http://matsim.org/open-scenario-data> | matsim.org/faq<http://matsim.org/faq> .

---
www.vsp.tu-berlin.de<http://www.vsp.tu-berlin.de>

post-corona:
+49 30 314 23308 (phone) +49 30 314 26269 (fax)
postal: TU Berlin Sekr. SG 12, Salzufer 17-19, 10587 Berlin
physical: building SG 4.1, room 305

mrieser:
Thanks for the explanation regarding the changed values. So I assume it's fine and will accept the PR.

PR #945, opened 2020-03-30
change return type of PopulationUtils.decideOnActivityEndTime() to OptionalTime

michalmac:


michalmac:
I changed the base branch so that commits that were already included in the previous PR (#944) are not in the diff.

PR #954, opened 2020-04-03
Replace events file with a current version without any emission events

### LINKED ISSUE #939: emissions contrib contains outdated emissions-events file
https://github.com/matsim-org/matsim-libs/blob/master/contribs/emissions/scenarios/sampleScenario/5.events.xml.gz

Contains emissions-events which can't be parsed into the Pollutant-Enums, since at least NOx is spelled 'NOX'. As far as I can see this file is only used by the offline writing example.

Janekdererste:
fixes #939


kt86:
Hi @Janekdererste ,

looks like you only deleted the file instead of replacing it.

Janekdererste:
Thanks @kt86 I'd forgotten to git-add the file :-| 

Please merge, when you think this works now (or set automerge label)

PR #959, opened 2020-04-09
Optional times in pt

michalmac:
~~just a preview... not review-ready yet...~~


matsim/src/main/java/org/matsim/core/mobsim/qsim/pt/AbstractTransitDriverAgent.java
```diff
@@ -259,23 +257,22 @@ protected double longerStopTimeIfWeAreAheadOfSchedule(final double now,
...
 
 	private boolean isBadDouble(double d){
-		return Double.isNaN(d) || Double.isInfinite(d);
+		Preconditions.checkArgument(Double.isFinite(d));
+		return Double.isNaN(d) || Double.isInfinite(d);//FIXME: REMOVE
 	}
 
 	private void depart(final double now) {
 		double delay = now - this.getDeparture().getDepartureTime();
 		if (this.isBadDouble(this.getDeparture().getDepartureTime())){ //this is the case if the next stop is null
```
michalmac, line 266: @mrieser Can `departureTime` really be NaN/Inf ?? I could not find any such case in our code base. `Preconditions.checkArgument(Double.isFinite(d));` in `isBadDouble()` does not thrown any exception when running tests.

So I am not sure if `departureTime`should be `OptionalTime` or should we have the precondition check or should we optimistically assume the value is always finite?

michalmac:
Now ready for review. There are 2 commits: 1 big + 1 very small. The latter will be squashed into the former (once this is answered:https://github.com/matsim-org/matsim-libs/pull/959#discussion_r406654789)

matsim/src/main/java/org/matsim/core/mobsim/qsim/pt/AbstractTransitDriverAgent.java
vsp-gleich: IMHO opinion a Departure object must have a finite departure time, because otherwise all departure and arrival times at all stops would be infinite. However, a TransitRouteStop can have an (NaN or) infinite departureOffset at the last stop of the line (see Marcel's pt-tutorial scenario) https://github.com/matsim-org/matsim-libs/blob/7a57c242fe59a2b59b8f4f4b76ad30ce205933d4/examples/scenarios/pt-tutorial/transitschedule.xml#L14-L18) 
Real life public transit schedules also often have no departure time at the last stop of the line. Real life schedules also feature some trains without departure time at intermediate stops when boarding is prohibited at that stop (especially cross-border trains).
@mrieser Feel free to correct if I'm wrong.

michalmac: Thanks @vsp-gleich for your answer. I updated the PR. Will stick to `departureTime` as `double`.
As for the offsets, both are already converted into `OptionalTime`.

PR #981, opened 2020-05-07
DiscreteModeChoice contrib

sebhoerl:
Adds the DiscreteModeChoice extension as a contrib to `matsim-libs`.

Open question: Where does the documentation (currently in `/docs`) go?


rakow:
@sebhoerl could you please update your README to point to the updated maven artifacts and references. Thank you!

PR #988, opened 2020-05-24
Vehicle types in DVRP

sebhoerl:
Attempt to add vehicle types to DVRP, which may be useful for fleets with varying vehicles sizes. Probably also useful for later emissions analysis, etc. 

Tests in `dvrp` are running, but not yet in downstream contribs. I first want to check, if you think this is a good way to go (defining the type per vehicle), or do you see a better alternative, @michalmac ?

Currently, I removed the `capacity` attribute from the fleet files / FleetSpecification and added a `vehicleTypeId` attribute.

Probably, then we would also need to define `FleetReader_v1`, and `FleetReader_v2`.

And I'm not sure how this relates to the `ev` contrib, because I saw something like `battery_capacity`, which is defined in a separate file, and not in the central vehicles file. Is there a general strategy?

### DIFF
--- a/contribs/drt/src/main/java/org/matsim/contrib/drt/analysis/DrtAnalysisControlerListener.java
+++ b/contribs/drt/src/main/java/org/matsim/contrib/drt/analysis/DrtAnalysisControlerListener.java
@@ -35,6 +35,7 @@ import org.matsim.core.controler.MatsimServices;
 import org.matsim.core.controler.events.IterationEndsEvent;
 import org.matsim.core.controler.listener.IterationEndsListener;
 import org.matsim.core.utils.io.IOUtils;
+import org.matsim.vehicles.Vehicles;
 
 /**
  * @author jbischoff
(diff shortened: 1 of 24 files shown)


sebhoerl:
@michalmac Thinking about it, maybe this is too much for now. An option leading to fewer mess would be to have something like a `VehicleTypeProvider` similar, to where the `DvrpVehicleImpl` get constructed and where it is decided which link to give to the vehicle. Here we could fall back to the `DVRP_VEHICLE_TYPE`, but one could override this provider.

kt86:
Dear Sebastian, 
I jump in regarding the `battery_capacity`. 
We have done something similar within freight. Therefore we added in the first step the standard MATSim vehicle types to freight :)
In the second step we have now have attribute within `EngineAttributes`: 
grafik

I know that that approach is currently not totally in line with the EV-contrib. But it could be a could starting point to go forward.

Best, Kai

michalmac:
I think it would be nice to have an option of reading in vehicles from the standard matsim vehicle file and translating them into DVRP vehicles.

I think it should be done in a possibly similar way to how matsim vehicles are translated into jsprit vehicles. I would rather not add a direct dependence on the matsim `VehicleType` to `DvrpVehicle` (at least not in the first step). For instance, in some cases, the capacity is the number of seats while in other cases is the volume/weight/... of the load (these kind of projections would be made only once - during the translation).

kainagel:
That would be good.  It would also be good if the dvrp vehicles could be constructed in code ... what we are currently doing is to generate them in code, write that to file, and read back in.

The VehicleType has become considerably more flexible with the V2.  Could you please have a look?

michalmac:
> It would also be good if the dvrp vehicles could be constructed in code ... what we are currently doing is to generate them in code, write that to file, and read back in.

Good idea!

@kainagel  is it enough to have a fixed fleet specification for all iterations, or we want to modify it between iterations?

michalmac:
A different approach was taken (PR #1607), so closing this PR.

PR #1019, opened 2020-06-10
Event-based occupancy analysis for DRT

sebhoerl:
Transformed the schedule-based analysis into event-based analysis.

I tested with `RunMultiModeDrtExample` for 10 iterations. In each iteration, this implementation produces the exact same graphs. Not sure about the runtime implications for large scenarios as this now needs to cache a lot of information, like the vehicle occupancy, while waiting for the events.

@michalmac Let me know what you think!


sebhoerl:
Interesting, I thought the auto-merge would wait until the reviews are through.

PR #1026, opened 2020-06-16
Add "Coord" and "Coord[]" to the possible xml attributes.

Nitnelav:
The code for "Coord" was already there but was never added to the  ObjectAttributesConverter.
I also added Tests, though I'm not sure it's really useful in this case.

An exemple :
```
<link id="240502340001r" from="260834665" to="260834664" length="98.2801579998562" freespeed="4.166666666666667" capacity="1200.0" permlanes="1.0" oneway="1" modes="car" >
	<attributes>
		<attribute name="geometry" class="[Lorg.matsim.api.core.v01.Coord;">[(223972.307724;6757244.262287),(224025.990389;6757317.967732),(224030.167495;6757323.705577)]</attribute>
		<attribute name="origid" class="java.lang.Long">24050234</attribute>
		<attribute name="type" class="java.lang.String">residential</attribute>
	</attributes>
</link>
```


matsim/src/main/java/org/matsim/utils/objectattributes/attributeconverters/CoordConverter.java
```diff
@@ -44,7 +46,7 @@ public String convertToString(Object o) {
 		}
 		Coord c = (Coord)o;
 		
-		return String.format("(%.2f;%.2f)", c.getX(), c.getY());
+		return String.format(Locale.ROOT, "(%f;%f)", c.getX(), c.getY());
```
sebhoerl, line 49: I guess the `Locale` should be removed, because now you're writing "," instead "." as the decimal separator, which is dependent on you system settings.

matsim/src/test/java/org/matsim/utils/objectattributes/attributeconverters/CoordArrayConverterTest.java
```diff
@@ -0,0 +1,56 @@
...
+import org.junit.Test;
+
+import org.matsim.api.core.v01.Coord;
+
+public class CoordArrayConverterTest {
+
+    @Test
+    public void testFromToString() {
+        final CoordArrayConverter converter = new CoordArrayConverter();
+        String a = "[(223380.4988791829;6758072.4280857295),(223404.67545027257;6758049.17275259),(223417.0127605943;6758038.021038004),(223450.67625251273;6757924.791645723),(223456.13332351885;6757906.359813054)]";
```
sebhoerl, line 33: Here numbers, have point decimal separators, as it should be.

matsim/src/test/java/org/matsim/utils/objectattributes/attributeconverters/CoordArrayConverterTest.java
```diff
@@ -0,0 +1,56 @@
...
+        Assert.assertEquals(coords[3].hasZ(), false);
+        Assert.assertEquals(coords[3].getX(), 223450.67625251273, 0.00005);
+        Assert.assertEquals(coords[3].getY(), 6757924.791645723, 0.00005);
+        Assert.assertEquals(coords[4].hasZ(), false);
+        Assert.assertEquals(coords[4].getX(), 223456.13332351885, 0.00005);
+        Assert.assertEquals(coords[4].getY(), 6757906.359813054, 0.00005);
+
+        String b = converter.convertToString(coords);
+        Assert.assertEquals("[(223380,498879;6758072,428086),(223404,675450;6758049,172753),(223417,012761;6758038,021038),(223450,676253;6757924,791646),(223456,133324;6757906,359813)]", b);
+    }
```
sebhoerl, line 54: But here it is a comma, because of your Locale settings.

matsim/src/main/java/org/matsim/utils/objectattributes/attributeconverters/CoordConverter.java
Nitnelav: I'm not sure if I understand correctly but I want the formatted output to be independent of the user's system settings. So I should keep the `Locale.ROOT` right ?

matsim/src/test/java/org/matsim/utils/objectattributes/attributeconverters/CoordArrayConverterTest.java
Nitnelav: Agreed.

matsim/src/test/java/org/matsim/utils/objectattributes/attributeconverters/CoordArrayConverterTest.java
Nitnelav: Yes so there is the error. I forgot to update the Test after realizing there was this comma issue and fixing it with Locale.ROOT).
I'll update with the fix

matsim/src/main/java/org/matsim/utils/objectattributes/attributeconverters/CoordConverter.java
sebhoerl: No, if you don't provide any Locale it will always use the standard English-style formatting.

Nitnelav: Mmm, that doesn't seem to be true... At least on my machine :
```
public class Test {
    public static void main(String []args){
        double toto = 13531.65165165;
        System.out.println(String.format("%f", toto));
    }
}
```
Returns : `13531,651652`

sebhoerl: Interesting, never noticed that. But than it would probably make sense to use US Locale so it is formatted with a dot, like in the code?

Nitnelav: (this is a small detail, but) I would argue It makes more sense to keep the root locale. According to the docs, it "is regarded as the base locale of all locales, and is used as the language/country neutral locale for the locale sensitive operations."
It does what we want (ie. put dots instead of commas in formatted double strings) and is country agnostic, which is relevant here.
Anyway thanks for the approval ! :)

sebhoerl: But when you used the root locale, you were testing for "1231,412" so it was actually not using a dot, didn't it? But I think now I see why I was confused. I think usually we always use `Double.parseDouble(String)` to read stuff in and then `Double.toString(Double)`, which then always does the same, independent of the system (for instance of all the `PopuatlionWriterV*` do it this way). So maybe for consistency's sake it would be best to say `String.format("%s;%s", Double.toString(x), Double.toString(y))` :)

Now we just need to figure out, why Travis doesn't want to run the PR ... Maybe you need to merge in the latest commits?

Nitnelav: Oooo `String.format("%s;%s", Double.toString(x), Double.toString(y))` is actually better as it keeps the double precision ! I will update and push with that method.

Nitnelav:
It seems this PR is going to be stuck at "Travis CI - Branch Expected — Waiting for status to be reported" because it is from a fork instead of being from a branch... Am I right ?

kainagel:
Yes I think that is correct.  I think that it is a security feature so that people without permissions are not able to flood the build queue, but it totally screws up the workflow for pull requests.  Could you please try to find some help?  (Myself, I am totally under the water time-wise, sorry.)

sebhoerl:
I checked for some other repository where I have admin access on Travis. There is this button "active pull requests" in the settings. Could it be that it is disabled for the MATSim repository? Who has access?

mrieser:
Not sure if I found the right setting, but I found a "build pushed pull requests" setting in travis that was off, and I switched it to on. Let's observe if this helps.

michalmac:
> Not sure if I found the right setting, but I found a "build pushed pull requests" setting in travis that was off, and I switched it to on. Let's observe if this helps.

I think we have 5 concurrent jobs altogether for all repos under the matsim-org organisation. Building PRs will slow things down (even twice, pessimistically). IMO we should disable PR builds (esp. for repos which are heavy and/or very active).

sebhoerl:
I'm not sure what options there are exactly, but wouldn't the best thing be to disable automatic builds for all branches, *except* for updates in the PRs? Because this is our barrier for committing. And testing anyboy can do locally, and they will be notified if they want to send the PR and it fails because of system-dependent code, etc.

sebhoerl:
@mrieser I think you found the right setting, but it will only work for new PRs. I guess this one here we'll have to merge manually :)

michalmac:
> In that case we'd need to require "Travis Pull Request" instead of "Travis Branch" to pass in the Github branch settings.

👍 

Yes, that would also be a way to go, assuming that draft PRs are come under "Travis Pull Request". We could use draft PRs for some preliminary hacking. Maybe a bit less convenient to what we have now. On the other hand, by making draft PRs for early-stage development branches improves knowledge sharing.

PR #1027, opened 2020-06-17
Flexible vehicle occupancy time profiles

michalmac:
Now we can plot different activities not related to serving passengers (non-operating activities), such as charging:

0 drt_occupancy_time_profiles_StackedArea_drt


tschlenther:
very nice, i will have a look soon and try to use it for my custom drtFreightTasks

michalmac:
I am thinking of throwing TaskStarted and TaskEnded events (for DVRP agents). This would be super flexible and would allow easy tracking of both activities and legs. Now we can track charging, but we cannot track fleet rebalancing (as this is not an activity).

PR #1033, opened 2020-06-18
Define Access and Egress Walks by link attribute

jfbischoff:
This PR consists of two components:
- make the boolean to insert access and egress walks at main modes into an enum (and write it to the config)
- allow to set access time per mode as a link attribute. This can be extremely handy in many situations, e.g. if you want to approximate time spent on parking search in city centers.


matsim/src/main/java/org/matsim/core/router/NetworkRoutingInclAccessEgressModule.java
```diff
@@ -53,279 +54,301 @@
...
-				config ) ;
-			// yyyy might be possible to set the link ids to null. kai & dominik, may'16
-
-			result.add( accessLeg ) ;
-		} else {
-			Facility toFacility = FacilitiesUtils.wrapLink(accessActLink);
-			List<? extends PlanElement> accessTrip = accessEgressToNetworkRouter.calcRoute(fromFacility, toFacility, now, person);
-			for (PlanElement planElement: accessTrip) {
-				now = TripRouter.calcEndOfPlanElement( now, planElement, config ) ;
+    private static final Logger log = Logger.getLogger(NetworkRoutingInclAccessEgressModule.class);
```
michalmac, line 65: @jfbischoff this diff is hard to follow. Could you reformat spaces to tabs (for indentation)?

jfbischoff: whoops, that setting apparently did not survive my IDE update... thanks @michalmac

matsim/src/main/java/org/matsim/core/config/consistency/VspConfigConsistencyCheckerImpl.java
```diff
@@ -242,7 +243,7 @@ public void checkConsistency(Config config) {
 		// === plansCalcRoute:
 		
 		// added feb'16
-		if ( !config.plansCalcRoute().isInsertingAccessEgressWalk() ) {
+		if ( config.plansCalcRoute().getAccessEgressWalkType().equals(AccessEgressWalkType.None) ) {
```
Janekdererste, line 246: maybe you could correct the warning as well, since 'true' and 'false' can't be set anymore

matsim/src/main/java/org/matsim/core/network/NetworkUtils.java
```diff
@@ -792,8 +794,22 @@ public static void readNetwork(Network network, String string) {
...
+	}
+	public static void setLinkAccessTime(Link link, String mode, double accessTime){
+		String attribute = NetworkRoutingInclAccessEgressModule.ACCESSTIMELINKATTRIBUTEPREFIX+mode;
+		link.getAttributes().putAttribute(attribute,accessTime);
+	}
+
+
 	public static Network readNetwork(String string) {
-		Network network = ScenarioUtils.createScenario(ConfigUtils.createConfig()).getNetwork();
+		Network network = createNetwork();
```
Janekdererste, line 812: This has always bothered me. Thanks for chaning it

contribs/noise/src/test/java/org/matsim/contrib/noise/NoiseIT.java
```diff
@@ -473,7 +470,7 @@ else if(event.getActType().equals("work")){
 					log.warn( "receiverPointId:" + receiverPointId );
 					log.warn( "affected:" + list ) ;
 				}
-				if ( !runConfig.plansCalcRoute().isInsertingAccessEgressWalk() ) {
+				if ( !!runConfig.plansCalcRoute().getAccessEgressWalkType().equals(AccessEgressWalkType.None) ) {
```
vsp-gleich, line 473: double negation is not not here

tschlenther:
>     4\. After transitioning to routing modes we had some demands for setting different access/egress modes to the network dependent on the routing (main) mode. This pull request seems in line with implementing this some day.

This should be something which is technically solved and can be addressed by using a custom `NetworkRoutingProvider` (so it is not really solved on the configFile-user-level)

By the way, i am starting experiments with having a different access and egress modes, e.g. walk as access and DRT as egress...

vsp-gleich:
> > ```
> > 4\. After transitioning to routing modes we had some demands for setting different access/egress modes to the network dependent on the routing (main) mode. This pull request seems in line with implementing this some day.
> > ```
> 
> This should be something which is technically solved and can be addressed by using a custom `NetworkRoutingProvider` (so it is not really solved on the configFile-user-level)

I think this is a sllghtly different level of access/egress. Right now all network modes eventually delegate to the NetworkRoutingInclAccessEgressModule and use mode "walk" for access/egress to the network. Even if you have car delegating to drt for access/egress, drt will eventually use a walk router to access the drt network. My point 4. was about being able to replace "walk" with say "access_egress_car" for that last bit from the car network to the activity facility. Right now the standard implementation of NetworkRoutingProvider allows only mode "walk" for this last bit to the network.

> 
> By the way, i am starting experiments with having a different access and egress modes, e.g. walk as access and DRT as egress...

Interesting, maybe we should take about this (not here).

matsim/src/main/java/org/matsim/core/router/NetworkRoutingInclAccessEgressModule.java
```diff
@@ -98,144 +101,166 @@
...
+			Leg egressLeg = populationFactory.createLeg(TransportMode.non_network_walk);
+			egressLeg.setDepartureTime(now);
+			routeBushwhackingLeg(person, egressLeg, startCoord, toFacility.getCoord(), now, startLinkId, endLinkId, populationFactory, config);
+			result.add(egressLeg);
+		} else if (config.plansCalcRoute().getAccessEgressWalkType().equals(AccessEgressWalkType.readAccessTimeFromLinkAttribute)) {
+			Leg egressLeg = populationFactory.createLeg(TransportMode.walk);
+			egressLeg.setDepartureTime(now);
+			routeBushwhackingLeg(person, egressLeg, startCoord, toFacility.getCoord(), now, startLinkId, endLinkId, populationFactory, config);
+			double accessTime = NetworkUtils.getLinkAccessTime(filteredNetwork.getLinks().get(endLinkId), mode).orElse(0.0);
+			egressLeg.setTravelTime(accessTime);
```
vsp-gleich, line 183: this is a dangerous fallback to 0s access time. Dangerous because you won't notice if no access time link attribute is found for the mode (it neither fails nor gives a warning).

vsp-gleich:
I just had a meeting with @kainagel on this topic.

He prefers solving this issue in a slightly different way:
Keep the insertingAccessEgressWalk boolean in config as is (not in config file).
Instead make accessRoutingMode and egressRoutingMode configurable in NetworkRoutingProvider (by adding them to the constructor) and then add your own self-implemented "JoschCarAccess" respectively "JoschCarEgress" routing modes in code (=add routing modules in your run class) which you can use as access/egress routing modes for car. I'll help with making NetworkRoutingProvider more configurable. Those custom routing modules can then read link attributes.

This is preferred, because

- It is more flexible: anyone can implement his own access / egress routing module easily in the run class, no adding to enums in matsim-libs.
- It does not replace the current mode hierachy, e.g. if access/egress routing modes were already configurable, the existing infrastructure would already allow to have car delegate to bike for access/egress, bike delegate to boat for access/egress and boat to walk and so on. The current approach in this pull requests instead short cuts to replace all after car with a single hard-coded "walk" leg (with travel time set by link attribute). That can be what you want, but if it's not the preffered approach allows also to keep the hierachy.
- We need it anyway for Tilmann's projects.

We can make access/egress routing modes configurable in config later, when we have more experience and feel more comfortable with the design.

Some additional thoughts:

- We think it's probably better to just add the parking time to the normal walk travel time instead of replacing the walk travel time entirely. This avoids the inconsistency with non-network modes like pt.
- If adding parking time is the purpose, access and egress time are likely to be different. Parking a car takes more time than picking up an already parked car and drive away. Here we could solve this by setting different access and egress routing modes (e.g. plain "walk" for access and "parkCarAndWalk" for egress).
- The parking time is probably best located in the "car interaction" (or call it "parking search interaction") stage activity followed by a normal (egress) walk leg with normal travel time. This helps to keep the walk times consistent. Putting the parking time into the car leg is not possible because mobsim will ignore and update the leg travel time from the router with what happens during mobsim. To really use this with SwissRailRaptor, SwissRailRaptor should keep the stage activities instead of replacing them ([old JIRA issue 989](https://matsim.atlassian.net/browse/MATSIM-989)).

Some idea how the run class could look like:
`

        controler.addOverridingModule(new AbstractModule() {
            @Override
            public void install() {
                this.addRoutingModuleBinding(TransportMode.car).toProvider(new NetworkRoutingProvider(TransportMode.car, TransportMode.car, "JoschCarAccess", "JoschCarEgress"));
                this.addRoutingModuleBinding("JoschkaAccess").toInstance(new RoutingModule() { ...

`

vsp-gleich:
Since most of the code changes on this branch here handle the transition from isInsertingAccessEgressWalk from boolean to enum we might consider starting a clean branch and take over only the remaining changes.

matsim/src/main/java/org/matsim/core/router/NetworkRoutingInclAccessEgressModule.java
jfbischoff: Allright, let's throw an exception instead :)

jfbischoff:
> Keep the insertingAccessEgressWalk boolean in config as is (not in config file).

I honestly think that we can add a switch to the config file. After all, we have been experimenting with it for years and there is no need in keeping secrets here.

>Instead make accessRoutingMode and egressRoutingMode configurable in NetworkRoutingProvider

That is a good idea. If you need it for your projects anyway, we should implement it :)
That said, I don't see why we could not still use an enum and provide one or two meaningful defaults here. 
I would argue that fixed access-egress-times as introduced in this PR are likely to be one of them...and  if someone feels like overriding them, then that is still possible. 
We also discussed this at the Developer's Meeting 2018 (https://matsim.atlassian.net/browse/MATSIM-828), and the PR goes towards that direction.

>We think it's probably better to just add the parking time to the normal walk travel time instead of replacing the walk travel time entirely. This avoids the inconsistency with non-network modes like pt.

Mhm, I'm not sure about this. This will lead to a new inconsistency between neighbors. It is hard to reason why anyone in Salzufer 17 should have a longer access time than someone living next door who happens to sit closer to the node. 

> If adding parking time is the purpose, access and egress time are likely to be different. Parking a car takes more time than picking up an already parked car and drive away. Here we could solve this by setting different access and egress routing modes (e.g. plain "walk" for access and "parkCarAndWalk" for egress).

There are reasons for and against it. If you think of a garage in the city center, both are likely to take the same amount of time. But I wouldn't mind adding setting access and egress time differently.

> The parking time is probably best located in the "car interaction" (or call it "parking search interaction") stage activity followed by a normal (egress) walk leg with normal travel time. This helps to keep the walk times consistent. Putting the parking time into the car leg is not possible because mobsim will ignore and update the leg travel time from the router with what happens during mobsim. To really use this with SwissRailRaptor, SwissRailRaptor should keep the stage activities instead of replacing them (old JIRA issue 989).

So far, we have kept interaction activities at 0 seconds. I don't feel horribly comfortable changing this. 


To sum up, my preference would be:

- add configurable Access and Egress Routing Modules in NetworkRoutingProvider
- Switch to the enum as proposed here
- add it to the config file (at least in time for the next release)
- Allow different access and egress times

matsim/src/main/java/org/matsim/core/network/NetworkUtils.java
```diff
@@ -792,8 +794,22 @@ public static void readNetwork(Network network, String string) {
 		new MatsimNetworkReader(network).readFile(string);
 	}
 
+	public static OptionalTime getLinkAccessTime(Link link, String mode){
```
vsp-gleich, line 797: routingMode instead of mode

matsim/src/main/java/org/matsim/core/network/NetworkUtils.java
```diff
@@ -792,8 +794,22 @@ public static void readNetwork(Network network, String string) {
...
+		String attribute = NetworkRoutingInclAccessEgressModule.ACCESSTIMELINKATTRIBUTEPREFIX+mode;
+		Object o = link.getAttributes().getAttribute(attribute);
+		if (o!=null){
+			return OptionalTime.defined((double) o);
+		}
+		else return OptionalTime.undefined();
+	}
+	public static void setLinkAccessTime(Link link, String mode, double accessTime){
```
vsp-gleich, line 805: routingMode instead of mode

kainagel:
A comment to some of these things:

(1) enum instead of boolean is nearly always better.  So it is here.

(2) Making access-egress-insertion xml-configurable is certainly appropriate.  A setting of "walk" would be plausible.  I am not so sure if "directWalk" is the correct terminology, since, if walk is "on the network", there will be additional non_network_walks.  

(3) I am not so sure if it makes a lot of sense at this point to immediately make a third method public that we haven't really tested and used a lot.  One question here is if SBB is running their code "from java" or "from config-xml": if "from config-xml", then having it in config.xml would simplify life.  Otherwise, it would make not much difference.  The other question is if SBB would consider their setting already "well tested".  If the answer is yes, we should maybe discuss the naming convention and move forward. --  If we move forward, I am still a bit sceptical if we should really implement it by ifthenelse in the same class, or pull it apart into different classes, i.e. a separate implementation of RoutingModule.  (Evidently, the static methods in NetworkRoutingInclAccessEgressModule could be make package-private to access them from other classes in the same package.)

Joschka, does this make some sense, or is this horribly in your way?

jfbischoff:
@vsp-gleich and I came up with a plan on implementing this:

>(2) Making access-egress-insertion xml-configurable is certainly appropriate.  A setting of "walk" would be plausible.  I am not so sure if "directWalk" is the correct terminology, since, if walk is "on the network", there will be additional non_network_walks.

We will change the terminology of the enum-values a bit:
-@deprecated AccessEgressWalkType.None (this could then, at one point in time, be removed)
-AccessEgressWalkType.WalkToLink
-AccessEgressWalkType.ConstantTimeFromLink (with different access and egress times)
-AccessEgressWalkType.WalkToLinkPlusConstantTimeFromLink (sum of both)

... and also implement all the other comments in code.

In a next stage, overwriting the access/egress modes should also be possible, i.e.

`   controler.addOverridingModule(new AbstractModule() {
        @Override
        public void install() {
            this.addRoutingModuleBinding(TransportMode.car).toProvider(new NetworkRoutingProvider(TransportMode.car, TransportMode.car, "TilmannCarAccess", "FlyingSaucepan"));
            this.addRoutingModuleBinding("FlyingSaucepan").toInstance(new RoutingModule() { ...

`
This probably overwrites what would be installed via the enum (i.e. AccessEgressWalkType.WalkToLink would install for all network modes a NetworkRoutingProvider("networkMode", "networkMode", "walk", "walk"), then this can be (hopefully) overwritten using the syntax above).

In a later PullRequest we make AccessEgressWalkType.WalkToLink use orthogonal travel time to link (Matsim-966).

Regarding the config-Switch, I would say, we leave it out for now, but should add it in time for next release.
Internally, we have used a similar solution for roughly two years and have found it extremely handy.

kainagel:
Reads good to me.  Thank you so much for moving this forward.

tschlenther:
> In a next stage, overwriting the access/egress modes should also be possible, i.e.
> 
> ` controler.addOverridingModule(new AbstractModule() {
> @override
> public void install() {
> this.addRoutingModuleBinding(TransportMode.car).toProvider(new NetworkRoutingProvider(TransportMode.car, TransportMode.car, "TilmannCarAccess", "FlyingSaucepan"));
> this.addRoutingModuleBinding("FlyingSaucepan").toInstance(new RoutingModule() { ...
> 

As mentioned before, I have the corresponding changes on a (yet) local branch. I will push it later today and make it a PR, where we can discuss some things. One problem with a flexible access/egress routing module is, that it might return null, so we need the fallback routing module as another field in NetworkRoutingInclAccessEgressModule.
But more on that later..

kainagel:
The fallback routing should in principle work in general, i.e. returning null should be ok.   Could you please ask Gregor?

kainagel:
Right.  Thanks ...

matsim/src/main/java/org/matsim/core/router/NetworkRoutingInclAccessEgressModule.java
```diff
@@ -175,19 +177,27 @@ private void addBushwhackingLegFromLinkToFacilityIfNecessary(final Facility toFa
...
 			egressLeg.setTravelTime(egressTime);
 			egressLeg.getRoute().setTravelTime(egressTime);
 			result.add(egressLeg);
 		} else {
 			Facility fromFacility = FacilitiesUtils.wrapLink(egressActLink);
-			result.addAll(accessEgressToNetworkRouter.calcRoute(fromFacility, toFacility, now, person));
+			List<? extends PlanElement> egressTrip = accessEgressToNetworkRouter.calcRoute(fromFacility, toFacility, now, person);
+			if (this.accessEgressWalkType.equals(AccessEgressWalkType.walkToLinkPlusConstant)){
+				double accessTime = NetworkUtils.getLinkEgressTime(egressActLink,mode).orElseThrow(()->new RuntimeException("Access Time not set for link "+ egressActLink.getId().toString()));
+				Leg leg0 = TripStructureUtils.getLegs(egressTrip).get(0);
```
vsp-gleich, line 195: Hm, this is harder than I thought. It can go wrong if there is no leg in the egress trip (->null pointer exception, ok) or that first leg is not a teleported mode, but rather has the travel time updated (e.g. setting adifferent travel time on a network mode leg has no effect, because qsim will determine the real travel time; no exception, not good). One might argue that putting the constant time in some stage activity avoids that, but that causes other problems (should be scored whereas usually stage activities are not scored). 

In the end the enum shall provide some commonly used functionality in a convenient way. The common use case is having some teleported walk for access and egress, so we should be fine in most cases. Even with walk on the network there should be a (possibly 0 m length) non_network_walk leg first (with network_walk and another non_network_walk following), so there should be no problem. 

So IMHO let's leave it like this for the time being until we have a better idea and more time. I'll try to add a comment (here in code if we have no better idea).

Thanks @jfbischoff for pushing this forward!

PR #1041, opened 2020-06-24
WIP: flexibility for separate access and egress modes

tschlenther:
This PR proposes to have (possibly) different RoutingModules inside `NetworkRoutingInclAccessEgressModule `for access and egress routing. This is facilitates to model intermodal trips, i.e. it would be possible to define bike as access mode for car and walk as egress mode. Another use case could be to introduce an area where cars are banned and possibly replaced by SAV. This PR is also mentioned by and needed for #1033 which deals with access and egress (walk) times.

The main changes take place in `NetworkRoutingInclAccessEgressModule` and `DefaultRoutingModules` which is public 'interface' to access `NetworkRoutingInclAccessEgressModule`.
Besides the changes in these classes, I created an example scenario for a car-banned-area test case.
After all, this PR is rather meant for discussion - i can come up with a clean version when the discussion has finished.

Things that i want to leave to discussion now:

1. Before, access and egress trips were always calculated by the walk RoutingModule, which never returned null. As this is now flexible, the `NetworkRoutingInclAccessEgressModule` needs to handle cases in which null is returned. My proposal is to use the fallbackRoutingModule. Unfortunately, i was not able to inject it, so it needs to get passed via (a new) constructor.

2. In DefaultRoutingModules, we need another static method to provide the functionality of having different access and egress routing modulesfor outside users. As i decided here to use the fallbackRoutingModule as well, it has to be passed through this method, too. If you see another way, please propose it.

3. If we decide to use the fallbackRoutingModule inside `NetworkRoutingInclAccessEgressModule`, one could also consider to handle cases where for the main leg no route could be created (e.g. two non-connected car network parts). But maybe that is overpushing and dealing with problems that should be dealt with elsewhere...

Later, i will remove some logging statements and some comments.


tschlenther:
i have tests running in another (public) project repo. hit me up if you want to have a look

vsp-gleich:
>     1. Before, access and egress trips were always calculated by the walk RoutingModule, which never returned null. As this is now flexible, the `NetworkRoutingInclAccessEgressModule` needs to handle cases in which null is returned. My proposal is to use the fallbackRoutingModule. Unfortunately, i was not able to inject it, so it needs to get passed via (a new) constructor.

For an example let's consider car with bike access and walk egress. Let's assume the bike router returned null, because it has not found a route. Then we can either 
- try to deal with the access trip only (i.e. find a replacement for that missing bike access trip) (A) or 
- deal with it one level up, i.e. we say the main mode car with bike access and walk egress is not available (and find a fallback for the whole thing).(B)

I think the difficulties with using fallbackRoutingModule to only replace the access trip are a sign that this is not the right level to deal with this.

If we compare with how SwissRailRaptor intermodal pt routing works, we see why that makes sense:
SwissRailRaptor is one level up, i.e. it considers multiple access/egress modes at the same time to find the least cost path using the best combination of those. So you would expect that it should always return a route if any combination of the available access and egress modes gives a working route.

Now here with the NetworkRoutingInclAccessEgressModule we have only one available access mode and only one available egress mode. So, it's not about finding the best combination of all available access/egress modes, but about finding the best route using those specific access and egress modes - if such a route exists. That's one level down in respect to SwissRailraptor. If there is no route found using that defined access mode then there is no route to be found by this router. If you want to test other combinations of more available access and egress modes, you would have to define those combinations as separate   NetworkRoutingInclAccessEgressModules, e.g. walk access - car main - walk egress. Or you need to write another router which tries out multiple access/egress modes one level up similar to SwissRailRaptor.

So to me it seems we should simply return null for the whole route in NetworkRoutingInclAccessEgressModule. TripRouter will then receive that null pointer and automatically call the fallbackRoutingModule. Then you don't need to bother with getting fallbackRoutingModule into NetworkRoutingInclAccessEgressModule, solving some headaches.

tschlenther:
After discussing with @kainagel, we came up with the following thoughts:

- instead of passing through routingModules we should use the trip router directly. That would facilitate handling the access and egress routing. For example, the tripRouter would handle the fallbackRouting for access, main leg and egress itself..

michalmac:
@tschlenther Maybe worth looking into? This is how we implemented flexible access/egress modes for UAM:
https://gitlab.tubit.tu-berlin.de/vsp/vw-projects/blob/master/src/main/java/uam/run/RunUamDrtScenario.java#L110-144

tschlenther:
> 
> 
> @tschlenther Maybe worth looking into? This is how we implemented flexible access/egress modes for UAM:
> https://gitlab.tubit.tu-berlin.de/vsp/vw-projects/blob/master/src/main/java/uam/run/RunUamDrtScenario.java#L110-144

thanks, looks quite elegant. i will check and compare again next week

tschlenther:
I have updated the PR to the following approach (which i had discussed with @vsp-gleich and to most parts with @kainagel before):

1. `NetworkRoutingInclAccessEgressModule` uses separate RoutingModules for access and egress. @kainagel: we chose to not delegate to `TripRouter` for access and egress trips, as this would result in the `TripRouter` calling the `FallbackRoutingModule` only for one segment (i.e. access or egress) of the entire trip. In consequence, one agent could end up with a plan containing a leg of routing mode walkCarDrt where it would not be technically possible to use drt for egress. This would basically be equivalent to a car leg. So there could be two plans that have different routing modes but mean the same in terms of execution. 

2. The module holds information on _accessMode_ and _egressMode_ as String fields. That has the consequence of a method signature change `DefaultRoutingModules` (which was kind of the access interface for outside users so far) with no real backwards-compatibility..
Actually, these mode String fields are only needed in cases, where the access/egress trip should result in a leg with a constant time set by link attribute. So one could argue that for these cases, one could set a standard mode (non-network-walk e.g.) and these parameters are not really needed (so we could keep backwards-compatibility and just delegate calls with one _accessEgressRoutingModule_ in a way where the latter is used for both, access and egress)...
 
3. I restructured the code in a way that for each segment of the trip, null can be returned which then leads to null as the value to be returned for the entire trip. This is in line with the above mentioned logic that, if one segment of the entire trip could not get routed, the whole mode chain is stamped invalid and the `TripRouter` will call the `FallbackRoutingModule` for the entire trip.

4. I renamed the (new) enum in `PlansCalcRouteConfigGroup` from _AccessEgressWalkType_ to _AccessEgressType_, as now the mode(s) may differ from walk. I also renamed two values of the enum accordingly. I included warnings in the `NetworkRoutingInclAccessEgressModule` constructor for cases where the user has set the usage of a constant time but also has provided a network routing module for access/egress trips. @jfbischoff please let me know if this would cause you any troubles..

tschlenther:
@kainagel approved this, @jfbischoff and @vsp-gleich please have a look so i can hopefully merge.
Whenever i find some time, i will elaborate on modularizing everything even a little bit more - with the goal that @jfbischoff 's stuff es nothing but a AccessRoutingModule (better different AccessRoutingModule for different values of the AccessEgressType enum).
But that requires some time as one has to handle all the data structure mashup with links and facilities..
So far everything seems to be working - but the design has room for improvement.

vsp-gleich:
> Whenever i find some time, i will elaborate on modularizing everything even a little bit more - with the goal that @jfbischoff 's stuff es nothing but a AccessRoutingModule (better different AccessRoutingModule for different values of the AccessEgressType enum).

This sounds good, because in the long term AccessEgressType should be probably configurable for every network mode, with the current approach we cannot use AccessEgressType.walkConstantTimeToLink for say bike and set up a access routing mode different from walk for car. But let's keep that for when we have time to do this...

Issue #1044, opened 2020-06-25
IdMap very slow for larger scenarios


d-roeder:
Hi ,

we realized that IdMap becomes very slow for larger scenarios, i.e millions of agents, when filling it. I fixed that issue temporally (and unfortunatelly forgot to report) with commit fafc3ff6d0a3d9b84e4ccba441684a4f2430f087. I think my solution is not a reasonable fix for everyone as it would be a waste of memory for smaller scenarios. 

My suggestion would be something like 

```java
public class IdMap<T, V> implements Map<Id<T>, V>, Iterable<V> {

	private static final String KEY = "idMapSize";
	private static final int DEFAULT_SIZE = 100;
	
	private Class<T> idClass;
	private final int sizeIncrement;
	private int size = 0;
	private Object[] data;


	public IdMap(Class<T> idClass) {
		this(idClass, Math.max(Id.getNumberOfIds(idClass), System.getProperties().getOrDefault(KEY, DEFAULT_SIZE)));
	}
	...
}
```
Now the advanced user would be able to configure the initial size without any drawbacks for the standard-user via commandline with 

```shell 
java -DidMapSize=999999 ....
```

Does anyone see a problem here?

Best,
Daniel

michalmac:
Maybe the underlying array expansion should be relative, e.g. x1.5 or x2.0? So that we have some kind of self-adaptation mechanism?

d-roeder:
hm, this would solve the performance-issue. But wouldn't it lead to a memory issue for large scenarios?

michalmac:
- With the 1.5 factor we do not risk so much.
- It's an array of `Objects`, so in the worst case 8 bytes per empty cell (e.g. 8 MB for 1 million empty cells).

d-roeder:
ah, right. I forgot that it is only the memory for the pointer and not for the object ... thanks

PR #1046, opened 2020-06-29
adaptive IdMap-size

d-roeder:
* made ensure capacity adaptive to the number of Ids
* fix https://github.com/matsim-org/matsim-libs/issues/1044

### DIFF
--- a/matsim/src/main/java/org/matsim/api/core/v01/IdMap.java
+++ b/matsim/src/main/java/org/matsim/api/core/v01/IdMap.java
@@ -1,6 +1,13 @@
 package org.matsim.api.core.v01;
 
-import java.util.*;
+import java.util.Arrays;
+import java.util.Collection;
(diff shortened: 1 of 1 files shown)


d-roeder:
Hi @mrieser,

as mentioned in #1044 IdMap is very slow for larger datasets. Do you see any drawbacks with that fix?

Best,
Daniel

mrieser:
> Do you see any drawbacks with that fix?

Except the already mentioned memory overhead of all the empty cells, I don't see any, no.

As the memory overhead hopefully should be small enough as it's only object pointers, I think it's okay. The normal ArrayList uses a very similar strategy, IIRC. If we observe that it becomes a problem, one could later try to have an upper limit of the added size, e.g. add at most 1 million entries, or something. But for the time being it's okay.

Thanks a lot!

PR #1054, opened 2020-07-01
Removed assumption that newTripToNewActivity returns list of size > 1 always

dhixsingh:
It's unclear to me why the assertion is on index 1. At lease for us, there are cases where the returned list is of size 1, and therefore the assertion on index 1 (which assumes size 2) causes a runtime exception.

The proposed change passes all tests so I am hoping this can be merged into master. 

Note that a better idea might be to change the assertion altogether so the intent is clear. But because I don't understand the initial reasoning (re index on 1), I'm unclear on what a revised assertion should be. 

/cc @kainagel


vsp-gleich:
Unfortunately EditTrips is not well tested yet, so seeing no errors in tests does not mean that it works as expected. Running all matsim core tests with coverage showed the lines changed in this pull request are not covered  :-(

The assumption of having >1 PlanElements and the assert of a leg with NetworkRoute at position 1 came in with commit 6c884bf5453 from summer last year, when we were working on a prototype version of EditTrips for public transit trips. 

At VSP we always use access/egress walks to the network. This causes all trips with network modes to be routed with NetworkRoutingInclAccessEgressModule which usually gives a list of 5 PlanElements: access leg, stage activity, the network route, another stage activity and egress leg. Except if we route from a facility on the network, there is no access walk necessary. Which is the case here, because we are on a NetworkRoute and therefore route from a link on the right network. The same may apply for the egress from the network to the destination facility (although very unlikely, so it probably never occurred to us), so even with access/egress routing we can end up with only 1 PlanElement.
So the silent assumption of > 1 PlanElements here is indeed wrong, but never caused trouble to us with access/egress legs inserted because the destination facility of an entire trip is usually not on the network. The assumption is certainly wrong if not inserting access/egress walk legs to the network (although we discourage not inserting access/egress walks).
Still I don't see yet why the list element at position 1 (should be network mode leg [0] -> stage activity [1] -> ...) is checked to be a leg with a NetworkRoute.
Probably it should be `Gbl.assertIf( (((Leg)newTripElements.get(0)).getRoute() instanceof NetworkRoute) );` instead, more in line with the comment above `// (2) there should be no access leg even with access/egress routing`. I don't have a real test or at least a test scenario for this at hand. Could you maybe try whether that solves the problem?

dhixsingh:
Thanks for clarifying @vsp-gleich. 

For context, we use EditTrips for dynamic re-routing of vehicles for evacuation modelling (bushfires in Australia), where the change of plan can happen while the vehicle is on a link (the remaining plan is discarded immediately and a new one attached based on some agent reasoning).

I have changed the assertion as per your suggestion and that works fine also for us. 

Hopefully this is ok to merge and close now.

PR #1057, opened 2020-07-03
Drt-Speed-Up compatibility with drt rebalancing

vsp-gleich:
Motivation: Make PreviousIterationZonalDRTDemandAggregator compatible with drt-speed-up module.

To achieve that DepartureEvents of the drt-speed-up mode should be considered drt demand. Changes:
- add a param drtSpeedUpMode to rebalancing config
- consider DepartureEvents of the drt-speed-up in PreviousIterationZonalDRTDemandAggregator

I adapted the existing test, but I'm not sure whether zone 4 is really the right zone for the new "drt_teleportation" mode demand starting from link 147 (no way to graphically check this).

### DIFF
--- a/contribs/drt/src/main/java/org/matsim/contrib/drt/analysis/zonal/PreviousIterationZonalDRTDemandAggregator.java
+++ b/contribs/drt/src/main/java/org/matsim/contrib/drt/analysis/zonal/PreviousIterationZonalDRTDemandAggregator.java
@@ -54,7 +54,7 @@ public class PreviousIterationZonalDRTDemandAggregator implements ZonalDemandAgg
 	public PreviousIterationZonalDRTDemandAggregator(EventsManager events, DrtZonalSystem zonalSystem, DrtConfigGroup drtCfg) {
 		this.zonalSystem = zonalSystem;
 		mode = drtCfg.getMode();
-		drtSpeedUpMode = drtCfg.getMinCostFlowRebalancing().get().getDrtSpeedUpMode();
+		drtSpeedUpMode = drtCfg.getDrtSpeedUpMode();
 		timeBinSize = drtCfg.getMinCostFlowRebalancing().get().getInterval();
 		events.addHandler(this);
 	}
--- a/contribs/drt/src/main/java/org/matsim/contrib/drt/analysis/zonal/PreviousIterationZonalDRTDemandAggregator.java
+++ b/contribs/drt/src/main/java/org/matsim/contrib/drt/analysis/zonal/PreviousIterationZonalDRTDemandAggregator.java
@@ -45,6 +45,7 @@ public class PreviousIterationZonalDRTDemandAggregator implements ZonalDemandAgg
 
 	private final DrtZonalSystem zonalSystem;
 	private final String mode;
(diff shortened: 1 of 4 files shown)


vsp-gleich:
If we plan to move drt-speed-up into the drt contrib in the long term, it would be better to have the parameter drtSpeedUpMode in the DrtConfigGroup and not in the RebalancingParameters. Is that acceptable now (although for now we use it only in rebalancing)?

jfbischoff:
> If we plan to move drt-speed-up into the drt contrib in the long term, it would be better to have the parameter drtSpeedUpMode in the DrtConfigGroup and not in the RebalancingParameters. Is that acceptable now (although for now we use it only in rebalancing)?

fine with me (but better not in XML ;-D)

vsp-gleich:
> fine with me (but better not in XML ;-D)

ok ;-D

tschlenther:
Zone 4 goes right to the edge if link 147, so should be right (tests also run through).
(Each zone has a width of 500 m and nodes are placed every 200 m)

i included a comment.

Here is a way to take the idleVehiclesPerZoneXY.csv in the iteration output folder for graphical control

grafik

vsp-gleich:
Thank you!

michalmac:
@vsp-gleich @ikaddoura I am wondering what is the problem with using the same name for the DRT mode and only replacing a proper optimiser with the speedup "optimiser"?

michalmac:
I just looked into: https://github.com/matsim-vsp/drt-speed-up. I see there are lots of steps that could be avoided if we simply only replace the optimiser between iterations (+we would need to modify `PassengerEngine` too I guess).

Also having these checks `(event.getLegMode().equals(mode) || event.getLegMode().equals(drtSpeedUpMode))` across many places is not clean and rather hard to maintain (in a consistent way).

vsp-gleich:
> I just looked into: https://github.com/matsim-vsp/drt-speed-up. I see there are lots of steps that could be avoided if we simply only replace the optimiser between iterations (+we would need to modify `PassengerEngine` too I guess).
> 
> Also having these checks `(event.getLegMode().equals(mode) || event.getLegMode().equals(drtSpeedUpMode))` across many places is not clean and rather hard to maintain (in a consistent way).

I agree, it would be certainly nicer not to change the mode and instead to stay within normal drt and to send all the normal drt events (request, departure, ...). It would also help with drt fares where we had to copy code into drt-speed-up to emulate them. It's certainly cleaner not to replace all this and just kind of switch off the time-consuming insertion calculation part. But then I need your help and time to modify the optimiser and the PassengerEngine, because I don't want to interfere with current code restructuring and I don't know what is tested and therefore safe to modify. We can certainly keep this as a plan for when you have time. In the long run moving some drt-speed-up like feature into the drt/dvrp contrib would be a good idea. When we came up with the current design we did not see that possibility and being already behind schedule with a project we tried to get something running as soon as possible using existing matsim infrastructure, e.g. the simple teleportation engine.

michalmac:
@vsp-gleich Thanks for your feedback. Since we both agree, I am going to make it more flexible (somewhere this week).

ikaddoura:
Thanks @michalmac , sounds good!

PR #1058, opened 2020-07-03
Input file decryption

rakow:
This PR introduces a new CipherUtils class for decrypting input files and integrates it into IOUtils, so files ending with ".enc" are decrypted automatically.


mrieser:
Just out of curiosity: Where does MATSim use encrypted files?

rakow:
We have a project with input files that are very sensitive for privacy reasons. (Mostly population & plans generated from mobile data). We also do a lot of runs on a computing cluster, which can be accessed by many people. To keep our data secure we planned to encrypt it first, so only we will be able to make use of it.

neuma:
@rakow Since you mentioned running sensitive stuff on a cluster: How do you shield a) the pw and b) the MATSim output from other users or the admin?

rakow:
a) This depends on the platform. The password is only read from an environment variable. Platforms like kubernetes allow to populate these with [secrets](https://kubernetes.io/docs/concepts/configuration/secret/) that are not accessible by other users.
For a batch system one could query the password every time a job needs to be started.
The least we can do is not to store the password on disk at any time. However, it will always be in the memory at some point and an attacker could find a way to access.

b) In our use-case the output is not critical. This PR does not do anything is this regard. However, it would not be too complicated to enable encryption of output files in a similar way.

PR #1059, opened 2020-07-03
Modestats takes all main modes given by MainModeIdentifier

vsp-gleich:
Up to now Modestats takes all modes registered in PlanCalcScoreConfig. These are not necessarily the same modes an MainModeIdentifier might return, e.g. you can have specialized MainModeIdentifiers which differentiate intermodal trips to different "main modes", which not necessarily are individual modes in PlanCalcScoreConfig. The workaround of adding such modes to PlanCalcScoreConfig is messy.
Instead in this pull request Modestats writes out any mode returned by the MainModeIdentifier. This means that the output must be written from scratch in each iteration, because new modes might have to be added, which were encountered only in this iteration and not in previous iterations. Modes which are never used are not written out at all. To me this seems a feature, not a bug, but I'm open for other opinions on this.
The implementation is as close as possible to the original, although git does not match well some only slightly modified code fragments.


matsim/src/main/java/org/matsim/analysis/ModeStatsControlerListener.java
```diff
@@ -136,31 +127,50 @@ private void collectModeShareInfo(final IterationEndsEvent event) {
...
-				if ( modeHistory == null ) {
-					modeHistory = new TreeMap<>() ;
-					this.modeHistories.put(mode, modeHistory) ;
+				modeOut.write("\t" + mode);
+			}
+			modeOut.write("\n"); ;
+			for (int iter = firstIteration; iter <= event.getIteration(); iter++) {
+				modeOut.write( String.valueOf(iter) ) ;
+				for ( String mode : modes ) {
+					modeOut.write( "\t" + modeHistories.get(mode).get(iter)) ;
```
jfbischoff, line 165: since we are touching it :-)
Perhaps we could round this double to a more readable form, e.g. 4 digits?

vsp-gleich:
I had a meeting with @kainagel and he was ok with the changes.

matsim/src/main/java/org/matsim/analysis/ModeStatsControlerListener.java
vsp-gleich: I'm not so sure about this. We have cases which need more digits, e.g. you have your entire Switzerland model and enable drt in a small area only. With 4 digits you end up with a mode share of 00.00 for drt and don't know whether it was used at all whereas with the non-rounded value you can multiply with the share of trips in that small area of all trips in Switzerland and have a useful mode share for drt.

jfbischoff: Mhm :-) I would argue that anything after two-digits behind the percentage-comma is either noise or requires a dedicated analysis anyway

vsp-gleich: Yes, sure, in many cases it is rather noise, however in that rarely used mode problem it is not. We have this a lot with drt (mode share somewhere in the 10^⁻4 or 10^-5 area). Some information on intermediate iterations is often helpful. Unless you write out plan files in each iteration it is impossible to get mode shares for intermediate iterations and get some idea whether agents even tried out drt or never tried because something was broken or mode shares have not converged yet.

vsp-gleich: So maybe this should be made configurable? Maybe in a separate pull request (making it configurable for all output files)?

PR #1062, opened 2020-07-06
added getter for modeHistories

jakobrehmann:
This pull request adds a getter method for modeHistories. I need access to this collection for the dynamicShutdown module I am working on. I did not change the access modifier of modeHistories to private, since I am not sure what effects that would have down the line.


kainagel:
Could you please make the new method final, and since you are at it, try to make the class final?  Apart from that, it feels ok to me since it is just an observer.

jakobrehmann:
sounds good, all done!

PR #1063, opened 2020-07-07
Extendible taxi task types

michalmac:
The motivation is to have a sealed original set of taxi sets, but at the same time being able to add new task types.

**Background:** The standard taxi optimisers operate on the fixed pre-defined set of tasks: `EMPTY_DRIVE`, `PICKUP`, `OCCUPIED_DRIVE`, `DROPOFF` and `STAY`. Using enums to represent task types has lots of advantages (type safety, efficient `switch` statements and `switch` expressions from Java 12), and one big drawback: this set cannot be extended (by adding new task types). So whenever we add a new task class (e.g. charging task) we need assign it one of the existing types (e.g. `STAY` for the charging task) and be very careful to override the handling for this specific task class. 

**Solution:**
1. Make `TaxiTaskType` an extendable class (so we can add charging tasks, for instance)
2. Create an enum type `TaxiTaskBaseType` that contains the limited set of base types that are understood by the standard taxi optimisers
3. Create non-standard types, such as the aforementioned charging task
4. Assign each `TaxiTaskType` a `TaxiTaskBaseType` or null (if they need a special handling).
5. Adapt the standard taxi optimisation algorithms to use the (sealed) base types for handling tasks. Tasks without the assigned base type will not be handled by the standard algorithms (an exception will be raised in order to avoid incorrect/fragile/ill-defined behaviour).

### DIFF
--- a/contribs/dvrp/src/main/java/org/matsim/contrib/dvrp/schedule/Task.java
+++ b/contribs/dvrp/src/main/java/org/matsim/contrib/dvrp/schedule/Task.java
@@ -37,6 +37,7 @@ import org.matsim.contrib.dvrp.tracker.TaskTracker;
  */
 public interface Task {
 	interface TaskType {
+		String name();
 	}
 
 	enum TaskStatus {
--- a/contribs/dvrp/src/main/java/org/matsim/contrib/dvrp/vrpagent/AbstractTaskEvent.java
+++ b/contribs/dvrp/src/main/java/org/matsim/contrib/dvrp/vrpagent/AbstractTaskEvent.java
@@ -70,7 +70,7 @@ public abstract class AbstractTaskEvent extends Event implements HasPersonId {
 	public Map<String, String> getAttributes() {
 		Map<String, String> attr = super.getAttributes();
 		attr.put(ATTRIBUTE_DVRP_VEHICLE, dvrpVehicleId + "");
-		attr.put(ATTRIBUTE_TASK_TYPE, taskType + "");
+		attr.put(ATTRIBUTE_TASK_TYPE, taskType.name());
 		attr.put(ATTRIBUTE_TASK_INDEX, taskIndex + "");
 		return attr;
 	}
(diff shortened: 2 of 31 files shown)


michalmac:
Additional benefits:
- output events are more descriptive:
```
	<event time="22026.0" type="dvrpTaskStarted" person="taxi_5_5" dvrpVehicle="taxi_5_5" taskType="CHARGING" taskIndex="2"  />
```
instead of
```
	<event time="22026.0" type="dvrpTaskStarted" person="taxi_5_5" dvrpVehicle="taxi_5_5" taskType="STAY" taskIndex="2"  />
```
- genereated plots/stats/... can easily use the extended task type set:
0 taxi_status_time_profiles_taxi_StackedArea
instead of:
0 taxi_status_time_profiles_taxi_StackedArea

michalmac:
A vehicle relocation task is an opposite example to the charging task that has no base type (== special handling). For a relocation task:
- we want to model it as a separate task type (for instance, to obtain more meaningful stats, visualisation, event handling...)
- at the same time we **may** want the standard algorithms to handle them as regular empty drives

In this case, we need to assign `EMPTY_DRIVE` as the base type to the relocation task type.

tschlenther:
Pretty cool! This would have made my life much easier when i was coding the dual-use-taxis.

So, in summary: the standard algorithms and some analysis such as the TaxiStats rely on getBaseType(task) whereas the time profiles and any other, potentially customized, analysis can use the new custom task types!?

michalmac:
> Pretty cool! This would have made my life much easier when i was coding the dual-use-taxis.
> 
> So, in summary: the standard algorithms and some analysis such as the TaxiStats rely on getBaseType(task) whereas the time profiles and any other, potentially customized, analysis can use the new custom task types!?

Yes, except that `TaxiStats` work with with both types:
 - (extendable) task types are used to calculate base stats (e.g. task duration by types)
 - base types are used to compute derived-states based on the collected stats (such as empty-to-total drive, wait-to-total-time ratios)

michalmac:
@tschlenther I started adapting `pave` to work with this PR (before merging it). Will let you know when either it's done or I am stuck :-)

tschlenther:
> 
> 
> @tschlenther I started adapting `pave` to work with this PR (before merging it). Will let you know when either it's done or I am stuck :-)

Perfect, thank you! Hope that this (old) piece of spaghetti code does not cost you too much time.

PR #1068, opened 2020-07-08
customise TaskType of TaxiEmptyDriveTask (only base type must == EMPTY_DRIVE)

michalmac:
Additionally:
- introduce the `DRIVE_TO_CHARGER` task type, so we can differentiate such tasks from the standard "empty" drives to pickups  (both having `EMPTY_DRIVE` as the base type).


michalmac:
Plot with `DRIVE_TO_CHARGER` tasks (dark blue at the top, above charging tasks)
0 taxi_status_time_profiles_taxi_StackedArea

PR #1074, opened 2020-07-13
enable using DefaultPassengerEngine in multi-threaded QSim

michalmac:
Adds synchronisation on the optimiser to:
- calling `VrpOptimiser.submitRequest()` from `DefaultPassengerEngine.validateAndSubmitRequest()`
- `VrpAgentLogic.computeNextAction()`

In many cases the synchronisation could be even more fine-grained and probably done entirely on the optimiser side. However, there are already some "unusual" corner cases (a few examples: (1) switching to "an idle task" is immediately converted into a relocation to the "globally" best depot, (2) submitting a request immediately calls global re-optimisation). So I think it is better to move the responsibility of synchronisation to the "framework" rather than requiring (or rather hoping) that each optimiser implementation would provide thread-safety on its own.

Final notes:
- request ID creation uses `AtomicInteger` (already added in the previous PR)
- calling `PassengerRequestValidator` or `PassengerRequestCreator` is not synchronised. In most cases, these operations are stateless. Otherwise a special measures should be provided to ensure their thread safety. A javadoc comment added to both interfaces.

### DIFF
--- a/contribs/drt/src/main/java/org/matsim/contrib/drt/run/DrtConfigGroup.java
+++ b/contribs/drt/src/main/java/org/matsim/contrib/drt/run/DrtConfigGroup.java
@@ -229,7 +229,9 @@ public final class DrtConfigGroup extends ReflectiveConfigGroup implements Modal
 					+ "attempting to travel without vehicles being available.");
 		}
 
-		Verify.verify(config.qsim().getNumberOfThreads() == 1, "Only a single-threaded QSim allowed");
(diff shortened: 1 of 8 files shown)


michalmac:
I just merged this PR and marked it as an experimental feature (for taxi and drt -- you will get a warning when running them with more than 1 QSim thread).

Looking for volunteers to try it out and report potential speed improvements (or issues).

PR #1075, opened 2020-07-13
Drt analysis

tschlenther:
1) Fix idleVehiclesPerZoneXY.csv:
  do not decrease number of idle vehicles when AfterVrpSchedule activity started. This led to negative numbers at the end of the simulation
2) introduce zonal wait time analysis:
 some zonal wait time statistics are dumped out into the iteration directory. They contain aggregated stats on passenger wait times per rebalancing zone

Unfortunately, it seems like no EventsReader exists that can handle dvrp passenger events, which means that 2) can not be used for post-processung (as none of the DrtAnalysis classes ?)


michalmac:
I just noticed that the states do not discriminate modes. Is this intended?

Here is how we detect mode in the DRT occupancy (and additionally respond to task starts/ends instead of activity start/ends): https://github.com/matsim-org/matsim-libs/blob/60600d3bd0ce6751753ef9d3a0a65ba0584ecf5d/contribs/drt/src/main/java/org/matsim/contrib/drt/util/stats/DrtVehicleOccupancyProfileCalculator.java#L175-L200

tschlenther:
> I just noticed that the states do not discriminate modes. Is this intended?

The `DrtRequestAnalyzer` filters the mode. As createZonalStats() iterates over requestAnalyzer.getWaitTimeCompare().keySet(),
only requests of the corresponding mode should be considered..?

tschlenther:
> 
> 
> > I just noticed that the states do not discriminate modes. Is this intended?
> 
> The `DrtRequestAnalyzer` filters the mode. As createZonalStats() iterates over requestAnalyzer.getWaitTimeCompare().keySet(),
> only requests of the corresponding mode should be considered..?

Ah sorry you probably meant `ZonalIdleVehicleXYVisualiser` and not `DrtZonalWaitTimesAnalyzer`...
I will check on that!

tschlenther:
> 
> 
> I just noticed that the states do not discriminate modes. Is this intended?
> 
> Here is how we detect mode in the DRT occupancy (and additionally respond to task starts/ends instead of activity start/ends):
> 
> https://github.com/matsim-org/matsim-libs/blob/60600d3bd0ce6751753ef9d3a0a65ba0584ecf5d/contribs/drt/src/main/java/org/matsim/contrib/drt/util/stats/DrtVehicleOccupancyProfileCalculator.java#L175-L200

One problem is that the linkId is needed in order to retrieve the corresponding zone from the DrtZonalSystem and the TaskStartedEvent does not provide information on the link id. Using Fleet.class instead of FleetSpecification.class to retrieve the linkId won't work too, as it only is bound in qsim scope.

contribs/drt/src/main/java/org/matsim/contrib/drt/analysis/zonal/ZonalIdleVehicleXYVisualiser.java
```diff
@@ -66,37 +75,34 @@ private void initEntryMap() {
...
 	}
 
 	@Override
 	public void handleEvent(ActivityEndEvent event) {
 		if (event.getActType().equals(DrtActionCreator.DRT_STAY_NAME)) {
-			String zone = zonalSystem.getZoneForLinkId(event.getLinkId());
-			if (zone != null) {
-				Integer oldNrOfVeh = zoneEntries.get(zone).getLast().getSecond();
-				zoneEntries.get(zone).add(new Tuple<>(event.getTime(), oldNrOfVeh - 1));
+			if(this.fleet.getVehicleSpecifications().containsKey(Id.create(event.getPersonId().toString(), DvrpVehicle.class))){
```
tschlenther, line 92: I decided to still use ActivityEnd and ActivityStart and do the well-known-but-gross id-lookup via conversion of person and vehicle id...

michalmac:
> > I just noticed that the states do not discriminate modes. Is this intended?
> > Here is how we detect mode in the DRT occupancy (and additionally respond to task starts/ends instead of activity start/ends):
> > https://github.com/matsim-org/matsim-libs/blob/60600d3bd0ce6751753ef9d3a0a65ba0584ecf5d/contribs/drt/src/main/java/org/matsim/contrib/drt/util/stats/DrtVehicleOccupancyProfileCalculator.java#L175-L200
> 
> One problem is that the linkId is needed in order to retrieve the corresponding zone from the DrtZonalSystem and the TaskStartedEvent does not provide information on the link id. Using Fleet.class instead of FleetSpecification.class to retrieve the linkId won't work too, as it only is bound in qsim scope.

True. `TaskStartedEvent` does not have a link. Maybe it would make sense to add it? Another field we could add is `dvrpMode`, as this should greatly simplify filtering task events by (dvrp) mode? (There is already `mode` in all passenger request events)

tschlenther:
> 
> 
> > > I just noticed that the states do not discriminate modes. Is this intended?
> > > Here is how we detect mode in the DRT occupancy (and additionally respond to task starts/ends instead of activity start/ends):
> > > https://github.com/matsim-org/matsim-libs/blob/60600d3bd0ce6751753ef9d3a0a65ba0584ecf5d/contribs/drt/src/main/java/org/matsim/contrib/drt/util/stats/DrtVehicleOccupancyProfileCalculator.java#L175-L200
> > 
> > 
> > One problem is that the linkId is needed in order to retrieve the corresponding zone from the DrtZonalSystem and the TaskStartedEvent does not provide information on the link id. Using Fleet.class instead of FleetSpecification.class to retrieve the linkId won't work too, as it only is bound in qsim scope.
> 
> True. `TaskStartedEvent` does not have a link. Maybe it would make sense to add it? Another field we could add is `dvrpMode`, as this should greatly simplify filtering task events by (dvrp) mode? (There is already `mode` in all passenger request events)

I am not sure about whether to add the link id to the event, as it does not seem too obvious which link id should be returned in case of a DriveTask for example. While one could think about conventions like departure or arrival link, it would be hard to make the API very clear about it. On the other side it might make life easier at many places for analysis.
I do think that adding the mode field really helps and has no downside!

I merged this branch so we can move on with our rebalancing experiments (needed these changes in the master snapshot..)

michalmac:
> I am not sure about whether to add the link id to the event, as it does not seem too obvious which link id should be returned in case of a DriveTask for example. 

But the link is the place at which the event happens (not the whole task), so for the drive task start/end event it would be the start/end link, respectively

tschlenther:
> But the link is the place at which the event happens (not the whole task), so for the drive task start/end event it would be the start/end link, respectively

Makes perfect sense, you are right

PR #1077, opened 2020-07-14
OSM Reader: Public link properties

rakow:
Made LinkProperties and some utility methods public, so they can be re-used for network creation.

### DIFF
--- a/contribs/osm/src/main/java/org/matsim/contrib/osm/networkReader/SupersonicOsmNetworkReader.java
+++ b/contribs/osm/src/main/java/org/matsim/contrib/osm/networkReader/SupersonicOsmNetworkReader.java
@@ -196,7 +196,7 @@ public class SupersonicOsmNetworkReader {
         link.setLength(segment.getLength());
         link.setFreespeed(getFreespeed(segment.getTags(), link.getLength(), properties));
         link.setNumberOfLanes(getNumberOfLanes(segment.getTags(), direction, properties));
-        link.setCapacity(getLaneCapacity(link.getLength(), properties) * link.getNumberOfLanes());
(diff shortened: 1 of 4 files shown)


contribs/osm/src/main/java/org/matsim/contrib/osm/networkReader/SupersonicOsmNetworkReader.java
```diff
@@ -196,7 +196,7 @@ Link createLink(Node fromNode, Node toNode, WaySegment segment, Direction direct
         link.setLength(segment.getLength());
         link.setFreespeed(getFreespeed(segment.getTags(), link.getLength(), properties));
```
Janekdererste, line 197: Maybe, for consistency, it would be better to move the other methods, such as getFreespeed, getNumberOfLanes that extract link properties into the LinkProperties class?

The way it is now, we have some guessing of properties in the reader and some in the LinkProperties class

Janekdererste:
You have to convert the expected ConcurrentMap to Map in the Signals reader as well. Then your build should pass.

contribs/osm/src/main/java/org/matsim/contrib/osm/networkReader/SupersonicOsmNetworkReader.java
rakow: I was thinking these other methods depends heavily on OSM-Tags, so they would not be so useful outside this context. I don't think they would be likely to be used, so I left them there.

Janekdererste: Okay, that makes sence.

PR #1079, opened 2020-07-17
Fix calculation of speed if no speed tag is provided

Janekdererste:
* It calculates the speed in m/s now, instead of km/h
  * fix threshold to include links which are <= 300m long
  * fix threshold in calculateIfSpeedTag to only apply urbanSpeedFactor to links <= 50km/h
  * add unit tests for all this


Janekdererste:
grafik

changes look about right (the more red the more reduced speed compared to former network output)

PR #1088, opened 2020-07-24
Update Noise contrib to RLS 19

nkuehnel:
noise contrib can be used with RLS90 or RLS19
noise emissions, immissions and damages are separated in their own interfaces implementations
some memory optimizations...
receiver points are processed concurrently (large speedup)
...events file processing might be a bit slower at the moment


nkuehnel:
rls 19 calculation not yet under test

PR #1092, opened 2020-07-31
Pull up jsprit version & use cost values for waiting times and activity times

kt86:
This PR includes two parts:
- Pull up the jsprit version (is necessary for oncoming work of @tschlenther ) to 1.8.
- Give jsprit the time-dependent costs for scoring waiting times (for early arrivals and the performing activity times -> duration of (un)loading shipments or performing services. 
In earlier versions this was all part of the time dependent cost  parameter. Since 1.7.(?) this cost parameter can be set differently for transport/waiting/service times. We only set transport times, so the times for waiting and services were scored with 0 within jsprit.
These new fields can be defined in the vehicle type ->  Attributes of the costInformation. If these Attributes are not set, the value for costsPerSecond will be used.

Please note: 
(1) This will lead to differing scoring within of jsprit, which will also occur in the score of the carrierPlan after solving the VRP. -> I corrected the failing ITs after manually checking, that the score difference corresponds to the previously not scored times. 
(2) It is strongly recommend to not have too early depot opening times, because the vehicles will leave directly after opening the gate and may (depending on the settings) wait in front of the first activity when arriving before the activity time-window (which now means, that they get a bad score - which is ok, because one has to pay the driver for this time as well).


tschlenther:
@kt86 shall we proceed?

kt86:
Yes, let's merge it.

PR #1095, opened 2020-08-04
PassengerEventsReader for DVRP and DRT

tschlenther:
EventsReaders that use delegation for dvrp an drt passenger events.

@michalmac do you think they are at the right place? Should other Events be included (such as TaskStartedEvent etc.) and the readers be renamed and moved?

I came up with this, as i want to do some post processing analysis, which could actually also be done by reading in trip tables, but i prefer event-based analysis because that allows for custom interpretation of times etc.


tschlenther:
PassengerRequestSubmittedEvent had the same event type as PassengerRequestScheduledEvent!

I have not implemented any tests yet..
Is there a preferred way of testing EventsReaders?

contribs/drt/src/main/java/org/matsim/contrib/drt/passenger/events/DrtPassengerEventsReader.java
```diff
@@ -0,0 +1,79 @@
...
+
+    private DvrpPassengerEventsReader delegate;
+
+    public DrtPassengerEventsReader(EventsManager events) {
+        delegate = new DvrpPassengerEventsReader(events);
+        this.setValidating(false);
+        delegate.addCustomEventMapper(org.matsim.contrib.drt.passenger.events.DrtRequestSubmittedEvent.EVENT_TYPE, getDrtRequestSubmittedEventMapper());
+    }
+
+    private MatsimEventsReader.CustomEventMapper<org.matsim.contrib.drt.passenger.events.DrtRequestSubmittedEvent> getDrtRequestSubmittedEventMapper() {
```
michalmac, line 47: There are many looooong class names (prefixed with packages) in this class. They should go to imports.

contribs/dvrp/src/main/java/org/matsim/contrib/dvrp/passenger/DvrpPassengerEventsReader.java
```diff
@@ -0,0 +1,115 @@
...
+
+public class DvrpPassengerEventsReader extends MatsimXmlParser {
+
+    private EventsReaderXMLv1 delegate;
+
+    public DvrpPassengerEventsReader(EventsManager events) {
+        delegate = new EventsReaderXMLv1(events);
+        this.setValidating(false);
+        delegate.addCustomEventMapper(PassengerRequestRejectedEvent.EVENT_TYPE, getPassengerRequestRejectedEventMapper());
+        delegate.addCustomEventMapper(PassengerRequestScheduledEvent.EVENT_TYPE, getPassengerRequestScheduledEventMapper());
```
michalmac, line 46: probably using method references would be cleaner (one level of abstraction less). Here is an example:
```
        delegate.addCustomEventMapper(PassengerRequestScheduledEvent.EVENT_TYPE, this::passengerRequestScheduledEvent);
```
and the modified method:
```
 private PassengerRequestScheduledEvent passengerRequestScheduledEvent(GenericEvent event) {
            Map<String, String> attributes = event.getAttributes();

            double time = Double.parseDouble(attributes.get(PassengerRequestScheduledEvent.ATTRIBUTE_TIME));
			String mode = attributes.get(PassengerRequestScheduledEvent.ATTRIBUTE_MODE);
			Id<Request> requestId = Id.create(attributes.get(PassengerRequestScheduledEvent.ATTRIBUTE_REQUEST), Request.class);
			Id<Person> personId = Id.createPersonId(attributes.get(PassengerRequestScheduledEvent.ATTRIBUTE_PERSON));
			Id<DvrpVehicle> vehicleId = Id.create(attributes.get(PassengerRequestScheduledEvent.ATTRIBUTE_VEHICLE), DvrpVehicle.class);
			double pickupTime = Double.parseDouble(attributes.get(PassengerRequestScheduledEvent.ATTRIBUTE_PICKUP_TIME));
            double dropoffTime = Double.parseDouble(attributes.get(PassengerRequestScheduledEvent.ATTRIBUTE_DROPOFF_TIME));

            return new PassengerRequestScheduledEvent(time, mode, requestId, personId, vehicleId, pickupTime, dropoffTime);
    }
```

michalmac:
> @michalmac do you think they are at the right place? Should other Events be included (such as TaskStartedEvent etc.) and the readers be renamed and moved?

Right place! Good idea to add other event types! ;-)

BTW. I am going to add a few more attributes to task started/ended events (dvrp mode and linkId)

tschlenther:
> BTW. I am going to add a few more attributes to task started/ended events (dvrp mode and linkId)

Okay, let me know when you are done. I will resume this work after that..

michalmac:
> Okay, let me know when you are done. I will resume this work after that..

You can also finalise and merge this one without adding new event types and then add new types in another PR (so to keep dev branches small and short-lived). Up to you:-)

PR #1096, opened 2020-08-04
add support for qsim-scoped event handlers

michalmac:
Some event handlers are created before each mobsim (i.e. bound inside `AbstractQSimModule`s), mainly because they operate on data which is also created just before the mobsim. This is typical in DVRP, EV and related modules. Such handlers need to be removed after the mobsim ends.

This PR simplifies/standardises adding and removing such handlers. To add such a handler, use `AbstractQSimModule.addMobsimScopeEventHandlerBinding()` in exactly the same way "typical" handlers are added via `AbstractModule.addEventHandlerBinding()` and declare it as `MobsimScopeEventHandler` (which is a sub-interface of `EventHandler`). The rest happens automatically.

### DIFF
--- a/contribs/ev/src/main/java/org/matsim/contrib/ev/EvModule.java
+++ b/contribs/ev/src/main/java/org/matsim/contrib/ev/EvModule.java
@@ -31,9 +31,6 @@ public class EvModule extends AbstractModule {
 
 	@Override
 	public void install() {
-		bind(MobsimScopeEventHandling.class).asEagerSingleton();
-		addControlerListenerBinding().to(MobsimScopeEventHandling.class);
-
 		install(new ElectricFleetModule());
 		install(new ChargingInfrastructureModule());
 		install(new ChargingModule());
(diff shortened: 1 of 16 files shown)


michalmac:
This PR does not impact the existing functionality, so I will merge it without waiting for reviews (summer holiday season:-) ) as we need it for working on DRT rebalancing strategies. We can later get back to it and change stuff.

jfbischoff:
Looks good to me, if you feel like having some holiday-season-replacement blessing :-)

michalmac:
Thanks! Holiday-season-replacement blessing acknowledged :-)

jfbischoff:
Mhm, it seems travis does not approve. Strangely only on master:

https://travis-ci.com/github/matsim-org/matsim-libs/builds/178714834

tschlenther:
> 
> 
> Mhm, it seems travis does not approve. Strangely only on master:
> 
> https://travis-ci.com/github/matsim-org/matsim-libs/builds/178714834

in #1095 i faced the same failing test on travis (jenkins approved though). rerunning it twice fixed the problem ¯\_(ツ)_/¯

probably, that is an issue with non-determinism in dvrp..?

jfbischoff:
strange, but well.

michalmac:
What was the problem? Now after restarting all jobs are green so I cannot tell anything more :-/

jfbischoff:
An assertion error in RunOneTaxiExample in DVRP Config, IIRC

tschlenther:
I think that the assertion error occured in line 247 of RunOneTaxiWithPrebookingExampleIT because time was expected to be 2 but was 3

michalmac:
Thanks. Yes, `RunOneTaxiWithPrebookingExampleIT` fails from time to time :-( There are still a couple of things around prebooking that are not fully sorted out...

jfbischoff:
>Thanks. Yes, RunOneTaxiWithPrebookingExampleIT fails from time to time :-( There are still a couple of things around prebooking that are not fully sorted out...


Prebookings - troublesome since 2013 (c) ;-) ;-)

PR #1101, opened 2020-08-06
increase events queue size in "event-intense" QSim tests

michalmac:
To prevent failures on AppVeyor. I could not reproduce the issues locally, and the output from AppVeyor is too small to point to the right place in code, but this is quite likely the reason.


michalmac:
AppVeyor builds are back and green!

PR #1104, opened 2020-08-07
Refactor drt event handlers

michalmac:
Switch event handling from activity start/end events to task started/ended events. This is now possible after #1098. These task-oriented events are much handier for tracking dvrp vehicle states.


tschlenther:
thanks! 👍

PR #1106, opened 2020-08-11
Demand Aggregators

tschlenther:
The population based demand estimator now scales the estimate value down by total number of inhabitants in drt service area and multiplies it with fleet size. The resulting demand estimate value represents the number of vehicle desired in that zone.
That is in line with the other Demand Aggregators and leads to a consistenst meaning/interpretation for the alpha parameter.
The alpha parameter can thus be interpreted as "how many vehicles do i want to relocate per estimated demand unit" where estimated demand unit is either vehicles or drt rides (depending on which DemandAggregator you chose).

Moreover, as fleet size is now possibly subject to change over iterations (when enabling optDrt - currently located in an external repository), demand aggregators relying on fleet size are not bound as eager singleton any more such that they are reconstructed every iteration. This is not a problem concerning run time (in the Open Berlin Scenario, construction of the FleetSizeWeightedByPopulationShareDemandAggregator takes roughly 4 seconds).

### DIFF
--- a/contribs/drt/src/main/java/org/matsim/contrib/drt/analysis/zonal/FleetSizeWeightedByPopulationShareDemandAggregator.java
+++ b/contribs/drt/src/main/java/org/matsim/contrib/drt/analysis/zonal/FleetSizeWeightedByPopulationShareDemandAggregator.java
@@ -22,52 +22,43 @@
  */
 package org.matsim.contrib.drt.analysis.zonal;
 
-import org.apache.commons.lang3.mutable.MutableInt;
 import org.apache.log4j.Logger;
-import org.matsim.api.core.v01.events.ActivityEndEvent;
-import org.matsim.api.core.v01.events.handler.ActivityEndEventHandler;
 import org.matsim.api.core.v01.population.Activity;
 import org.matsim.api.core.v01.population.Population;
-import org.matsim.contrib.drt.run.DrtConfigGroup;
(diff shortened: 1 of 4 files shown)


michalmac:
@tschlenther removing the 'singleton' requirement does not ensure that there will be one instance per iteration. It means that now a new instance will be injected every time.

I suggest adapting the aggregator to the current fleet size before each mobsim start.

tschlenther:
Do you mean by making the aggregator an implementation IterationStartsListener (or MobsimInitializedListener) ?
I thought of that possibility, but was not sure how to inject the current FleetSpecification object into the aggregator - or does the FleetSpecification object remain the same over iterations? (i thought it had mobsim scope but maybe i was wrong - did not try around)

michalmac:
`FleetSpecification` evolves over time. One of examples: change the start location to the previous day's end location. Adding/removing vehicles "between iterations" should also be reflected there.

michalmac:
> Do you mean by making the aggregator an implementation of IterationStartsListener (or MobsimInitializedListener) ?

I was thinking of mobsim-related events, preferably https://github.com/matsim-org/matsim-libs/blob/master/matsim/src/main/java/org/matsim/core/controler/listener/BeforeMobsimListener.java (at the same phase event handlers are reset before each mobsim).

BTW. Your #1111 effectively solves the issue without even using listeners! 👍

PR #1110, opened 2020-08-14
Different Vehicle Types & More accurate network capacities in Hermes

jfbischoff:
**This PR brings a couple of changes into Hermes**

* support for different vehicle types with different PCUE and Flow Efficiency Factor values
* an improved representation of how downscaling of scenarios work (from now on, vehicles are scaled up and network capacities are kept constant. This can greatly reduce the pains of shared pt vehicles and cars on links, as one simply needs to set the PCU of the pt vehicle low enough)
* the accuracy of flowCapacity values is improved and is now just the way it is defined in the network (see FlowCapacityTest). For that, some values are now floats that were ints previously, but the overall performance stays similar
* some minor changes

Concluding remarks: For us, this adds all the functionality to Hermes to use it in production. The overall current speedup for a Switzerland 10% run is around 45% (300 iterations)


jfbischoff:
Thanks. Jenkins isn't entirely happy yet, let's see how I can fix that.

neuma:
> The overall current speedup for a Switzerland 10% run is around 45% (300 iterations)

A whopping speedup of 45% including non-teleported pt and replanning? That's impressive! Thanks for making this accessible!

jfbischoff:
Yes, that's the region of speedup. 

@rodrigo-bruno and  I are preparing a blog-post for the MATSim website which shows the effects a bit more in detail. Overall, we'd be very happy if people start using it (and help fixing bugs ;-) ).

neuma:
a subtle hint - I see :)

PR #1118, opened 2020-08-25
Feature/add walk marginto raptor

jfbischoff:


mrieser (approved):

vsp-gleich:
Hi @jfbischoff , I'm not fully sure I understand the motivation for this. Is it the idea that people walk faster to catch the connecting train if they arrive later than scheduled at the transfer stop? Or is it about seconds lost for interaction activities in between (which even at duration 0 take 1 sec in mobsim, because the following leg cannot start immediately)?
The deafult is 5s, so it will change simulation results for everyone not aware of this new parameter.

jfbischoff:
This used to be a niche bug which has been existing for ages, if an agent changed the PT facility when changing (i.e., from a train station to the bus station), but became more obvious with a recent change in [SwissRailRaptoreCore](https://github.com/matsim-org/matsim-libs/commit/ebc67f7830e4dd48fb785d9ada3ff97ccd33132e):
A walk leg that takes exactly the minimum transfer time between two stops will make agents miss their connection in the Mobsim, as the walk takes too long to board the connecting service even if everything runs bang-on-time.

Example: Imagine a service arriving at 7:59 and the connection departing at 8:00, with the minimal transfer time being 60 seconds. An arrival at 7:59:00, means that the agent starts walking one second after that. Adding 60 seconds walk will thuse make the agent available for the next leg at 8:00:01 and miss the connection.

In most cases, 1 second would be sufficient to solve the issue. Adding the default to five seconds also allows a buffer considering that in QSim (opposed to Hermes) not all people leave the vehicle at the same second (which is, however, accounted for, when creating those minimal transfer times and putting them into timetables in the first place). One could even argue that the Minimum Transfer Time should only be accounted for in the router, whereas the duration of the walk leg in between should always depend on the actual stop-to-stop-distance. But that would mix two data sources and probably create more headaches.

vsp-gleich:
>Adding the default to five seconds also allows a buffer considering that in QSim (opposed to Hermes) not all people leave the vehicle at the same second (which is, however, accounted for, when creating those minimal transfer times and putting them into timetables in the first place). One could even argue that the Minimum Transfer Time should only be accounted for in the router, whereas the duration of the walk leg in between should always depend on the actual stop-to-stop-distance.

Thank you, that makes sense. Have you looked into some simulation runs how long alighting from a vehicle typically lasts in QSim (I haven't yet)? I think it can easily take more than 5 sec. Typically we have something like 1s per alighting passenger set in the TransitVehicle type, so starting with the 6th passenger alighting the problem can still occur. For a bus terminating at the last stop of a line with all passengers alighting I would also expect rather more than 5 sec in reality. So for QSim 5 sec might even be not enough. Interesting. 
The mixing of real physically necessary walk time between stops and buffer time for slow alighting and delays makes things complicated. I added an issue to think about that later: #1128

PR #1120, opened 2020-08-26
round pkm and ph mode stats to improve readability

jfbischoff:


vsp-gleich:
Maybe it would be better to have a config param in global config group (something like "outputStatisticsDigits")?

jfbischoff:
Mhm, not sure. 
For mode stats, anything behind the comma is probably useful in many scenarios.
--
For accumulated PKM and PH, anything behind the comma is probably only meaningful in debugging and makes it harder to decipher the actual text file.

PR #1125, opened 2020-08-27
Introduce zonal system params and DrtModeZonalSystemModule

michalmac:
This functionality was (too) tightly coupled with rebalancing. Now it is extracted from there.

Rebalancing strategies may use this zonal system for collecting demand data and making decisions.


michalmac:
> This functionality was (too) tightly coupled with rebalancing

I meant here: zonal analysis and rebalaning

PR #1126, opened 2020-08-27
3 New Rebalancing Strategies implemented

luchengqi7:
3 new rebalancing strategies in the AMoDeus platform have been implemented in MATSim. The algorithms have been tested in Mielec and Open Berlin Scenario. There are many commits during the implementation process. It is not necessary to go through those commits one by one.


michalmac (commented):
Thanks. Really nice work! 🚀 

I quickly looked through the PR. I see that 2 out of 3 strategies share lots of similarities with the existing min-cost-flow rebalancing (esp. `AdaptiveRealTimeRebalncingStrategy` looks almost identical). But maybe I overlooked some details.

@luchengqi7 Could we schedule a pair programming session next week and try to go through these strategies and do some refactoring (also in the existing DRT contrib) to increase code reuse? That would be also a good opportunity for me to learn more about these new strategies.

luchengqi7:
example_config_for_rebalance_strategy
Hi Michal, here's the config file I am using for the local testing. I have made some changes. Now it seems that there is no error. So I guess it is correct. Can you help me double check? Thank you very much!

michalmac:
merged via #1169

PR #1129, opened 2020-08-27
Refactor drt zones and demand aggregators

michalmac:


tschlenther:
Should we now go ahead and rename `ZonalDemandAggregator` to `ZonalDemandEstimator` (and all implementations accordingly)?
I was waiting with this for you to finish restructuring first in order to avoid conflicts or prolonging PRs, but i am happy to do that.

BTW, i would probably delete `TimeDependentActivityBasedZonalDemandAggregator `or replace it with something which does the same as FleetSizeWeightedByPopulationShareDemandAggregator but with time-dependent activity numbers (which then assumes perfect knowledge of all people's trajectories by the drt operator)...

michalmac:
> Should we now go ahead and rename ZonalDemandAggregator to ZonalDemandEstimator (and all implementations accordingly)?
> I was waiting with this for you to finish restructuring first in order to avoid conflicts or prolonging PRs, but i am happy to do that.

Sure. I did not want to rename the interface without discussing it first with you :-) (as you already had some thoughts on that). `ZonalDemandEstimator ` looks good!

michalmac:
> with time-dependent activity numbers (which then assumes perfect knowledge of all people's trajectories by the drt operator)...

Nice! Sounds like a (temp) substitute of the prebooking functionality.

tschlenther:
> 
> 
> > with time-dependent activity numbers (which then assumes perfect knowledge of all people's trajectories by the drt operator)...
> 
> Nice! Sounds like a (temp) substitute of the prebooking functionality.

Interesting thought. Not quite the same though. Probably it is even further away from prebooking than `PreviousIterationDemandEstimator`, as it would take a variety of activity types into account. But for high mode choice values, it might be closer to prebooking.

Issue #1127, opened 2020-08-27
Emission calculation crashes (when using drt)


markusstraub:
We try to combine drt with offline emission calculation (with MATSim 12.0) but noticed that the emission calculation crashes quite often with the following error:

    Average speed has been calculated to 0.0 or a negative value. Aborting...

I traced this to the following line in `WarmEmissionHandler.handleEvent(LinkLeaveEvent)`:
https://github.com/matsim-org/matsim-libs/blob/94dd3bca2e13e14acc5960efee7ec952aa31ecff/contribs/emissions/src/main/java/org/matsim/contrib/emissions/WarmEmissionHandler.java#L185

The relevant snippet from the `events.xml`, which I am processing after the simulation, is here:

```xml
<event time="24673.0" type="entered link" link="63894" vehicle="_saev95"  />
<event time="24681.0" type="left link" link="63894" vehicle="_saev95"  />
<event time="24681.0" type="entered link" link="98615" vehicle="_saev95"  />
<event time="24689.0" type="vehicle leaves traffic" person="_saev95" link="98615" vehicle="_saev95" networkMode="car" relativePosition="1.0"  />
<event time="24689.0" type="PersonLeavesVehicle" person="_saev95" vehicle="_saev95"  />
<event time="24689.0" type="arrival" person="_saev95" link="98615" legMode="car"  />
<event time="24689.0" type="actstart" person="_saev95" link="98615" actType="DrtBusStop"  />
<event time="24689.0" type="PersonLeavesVehicle" person="107_7576_1_1" vehicle="_saev95"  />
<event time="24749.0" type="actend" person="_saev95" link="98615" actType="DrtBusStop"  />
<event time="24749.0" type="actstart" person="_saev95" link="98615" actType="DrtStay"  />
<event time="24749.0" type="actend" person="_saev95" link="98615" actType="DrtStay"  />
<event time="24749.0" type="departure" person="_saev95" link="98615" legMode="car"  />
<event time="24749.0" type="PersonEntersVehicle" person="_saev95" vehicle="_saev95"  />
<event time="24750.0" type="vehicle enters traffic" person="_saev95" link="98615" vehicle="_saev95" networkMode="car" relativePosition="1.0"  />
<event time="24751.0" type="left link" link="98615" vehicle="_saev95"  />
<event time="24751.0" type="entered link" link="98616" vehicle="_saev95"  />
[...]

PR #1130, opened 2020-08-28
Fix emissions calculation if vehicle uses the same link twice within a trip.

kt86:
see #1127 for more details.


kt86:
Solves the issue of the user. Jenkins and travis are happy as well. -> merge (without additional test).

PR #1131, opened 2020-08-28
ZonalDemandEstimator

tschlenther:
1) Renamed ZonalDemandAggregator to ZonalDemandEstimator and all implementations accordingly
2) Rewrote TimeDependentActivityBasedZonalDemandAggregator such that activity ends are now scaled down by computing the share of total activity ends in the given zone of all activity ends within the given time frame and multiplied with the fleet size. Thus, the resulting value is now in vehicle units (same as for 2 out of 3 other ZonalDemandEstimators). 
3) renamed enum values of ZonalDemandEstimatorType such that the unit of the return value hopefully is clarified. It is either rides/requests or vehicles.


tschlenther:
also put a @Nullable annotation into DrtZonalSystem along with some documentation in order to clarify how to use it.

tschlenther:
see also discussion in #1129

Issue #1132, opened 2020-08-29
DrtStayTaskEndTimeCalculator cannot handle EDrtChargingTask in matsim 13


mfrawley:
Hi there,
I am encountering a new bug after updating to Matsim 13. We are using the EDrtChargingTask in an edrt simulation and saw an exception which I traced back to DrtStayTaskEndTimeCalculator.java, in the switch there is a call to DrtTaskBaseType.getBaseType which takes a generic Task but "force-unwraps" the optional baseType property (which in the case of an EDrtChargingTask is null).

I would say in general the usage of the optional in getBaseType is unsafe, it takes as input a generic type and so it's return value should be optional. In the calculator this could then be checked and at the least throw a more descriptive error such as "to calculate end-times for non-standard task types please implement a subclass". I managed to resolve this by implementing a CustomCustomDrtStayTaskEndTimeCalculator to safely handle the charging tasks.

One additional note, to reproduce this you need to set idleVehiclesReturnToDepots to false in DrtConfigGroup.

michalmac:
Thanks for reporting this bug.

In general, all tasks that do not have a base task type are not supported by the core DRT (or taxi) functionality. They need implementing additional handling to prevent potential unexpected behaviour (otherwise an exception is thrown).

Having types and base types is a relatively new feature, a response to seeing that DRT/taxi are extended is so many different ways. We wanted to simplify extensibility but at the same time protect against unexpected weird behaviour (assuming that throwing an exception is better than silently producing wrong results).

mfrawley:
@michalmac Thank you for the explanation. Still I think it would be prudent to check the null value and throw a nicer error, or your default case remains unused. I see the attached EDrtStayTaskEndTimeCalculator which is more elegant than what I proposed. I will switch to using that once I can. We can resolve this issue if you disagree about the null handling, alternatively I could create a PR.

PR #1133, opened 2020-08-31
EDrtStayTaskEndTimeCalculator

tschlenther:
Possibly the answer to #1132.
Basically it is copied and adopted from `ETaxiStayTaskEndTimeCalculator`.
Was not sure about whether to the check of TaskType is right or if the base type shoult be checked instead:
`if (task.getTaskType().equals(EDrtChargingTask.TYPE))`


michalmac (approved):
Nice. Thanks for fixing it!

PR #1138, opened 2020-09-03
Add links to drt zones

tschlenther:
DrtZones have now information on all links inside of them.

@michalmac as you decided to include the field `Map<Id<Link>,Link>`, and as we already can think of at least 2 ways of setting a target link for a zone, i thought maybe we want to have a flexible way of determining a target link for each zone.
That way, we do not need a `targetLink` field inside `DrtZone`.
Instead, we have the interface `DrtZoneTargetLinkSelector` which is currently only used for rebalancing but could also be used for other purposes,
I implemented 2 variants: 

1. choose a random link (each time you query the selector)
2. always choose the most central link

default is 2.

I put the binding and the interface as well as both implementations into the zonal package.
If you think that this rather belongs into optimizer\rebalancing\mincostflow i am happy to move it (the binding would be done in the `DrtModeMinCostFlowRebalancingModule` instead of `DrtModeZonalSystemModule` then)

Please let me know if you think that putting the target link selection behind an interface is premature generalization.

### DIFF
--- a/contribs/drt/src/main/java/org/matsim/contrib/drt/analysis/zonal/MostCentralDrtZoneTargetLinkSelector.java
+++ b/contribs/drt/src/main/java/org/matsim/contrib/drt/analysis/zonal/MostCentralDrtZoneTargetLinkSelector.java
@@ -1,0 +1,51 @@
+/* *********************************************************************** *
+ * project: org.matsim.*
+ * Controler.java
+ *                                                                         *
+ * *********************************************************************** *
+ *                                                                         *
+ * copyright       : (C) 2007 by the members listed in the COPYING,        *
+ *                   LICENSE and WARRANTY file.                            *
+ * email           : info at matsim dot org                                *
+ *                                                                         *
+ * *********************************************************************** *
+ *                                                                         *
+ *   This program is free software; you can redistribute it and/or modify  *
+ *   it under the terms of the GNU General Public License as published by  *
+ *   the Free Software Foundation; either version 2 of the License, or     *
+ *   (at your option) any later version.                                   *
+ *   See also COPYING, LICENSE and WARRANTY file                           *
+ *                                                                         *
+ * *********************************************************************** */
+
+package org.matsim.contrib.drt.analysis.zonal;
+
+import org.matsim.api.core.v01.network.Link;
+import org.matsim.core.network.NetworkUtils;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class MostCentralDrtZoneTargetLinkSelector implements DrtZoneTargetLinkSelector{
+
+	Map<DrtZone,Link> targetLinks = new HashMap<>();
+
+	@Override
+	public Link selectTargetLinkFor(DrtZone zone) {
+		if(this.targetLinks.containsKey(zone)) return this.targetLinks.get(zone);
+
+		Double minDistance = Double.MAX_VALUE;
+		Link closestLink = null;
+		for(Link link : zone.getLinks().values()){
+			// vehicle will be standing at the toNode, this is why we use toNode rather than getCoord
+			double dist = NetworkUtils.getEuclideanDistance(zone.getCentroid(), link.getToNode().getCoord());
+			if(dist < minDistance){
+				minDistance = dist;
+				closestLink = link;
+			}
+		}
+		if(closestLink == null) throw new RuntimeException("could not determin most central link for zone " + zone);
+		this.targetLinks.put(zone, closestLink);
+		return closestLink;
+	}
+}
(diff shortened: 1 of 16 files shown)


tschlenther:
should we consider a 'consistency-checking' event handler that checks whether any DrtStayTask is started outside the DrtZonalSystem?

For sure, this could be implemented as unit test, but maybe we want it as a running thing. Question would be how to handle exceptions - throwing in error seems to be pretty restrictive. Some logging or maybe analysis output could be an option

contribs/drt/src/main/java/org/matsim/contrib/drt/analysis/zonal/MostCentralDrtZoneTargetLinkSelector.java
```diff
@@ -0,0 +1,51 @@
...
+
+import org.matsim.api.core.v01.network.Link;
+import org.matsim.core.network.NetworkUtils;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class MostCentralDrtZoneTargetLinkSelector implements DrtZoneTargetLinkSelector{
+
+	Map<DrtZone,Link> targetLinks = new HashMap<>();
```
michalmac, line 31: private final?

contribs/drt/src/main/java/org/matsim/contrib/drt/analysis/zonal/RandomDrtZoneTargetLinkSelector.java
```diff
@@ -0,0 +1,38 @@
...
+import java.util.ArrayList;
+import java.util.Random;
+
+public class RandomDrtZoneTargetLinkSelector implements DrtZoneTargetLinkSelector{
+
+	private final Random random = MatsimRandom.getLocalInstance();
+
+	@Override
+	public Link selectTargetLinkFor(DrtZone zone) {
+		return new ArrayList<>(zone.getLinks().values()).get(random.nextInt(zone.getLinks().size()));
```
michalmac, line 35: This may be quite expensive. Iterating would be cheaper. But I suggest storing that list (map of lists).

We could also switch from a map of links to a list of links (in DrtZone).

contribs/drt/src/main/java/org/matsim/contrib/drt/analysis/zonal/MostCentralDrtZoneTargetLinkSelector.java
```diff
@@ -0,0 +1,51 @@
...
+
+	@Override
+	public Link selectTargetLinkFor(DrtZone zone) {
+		if(this.targetLinks.containsKey(zone)) return this.targetLinks.get(zone);
```
michalmac, line 35: since we do not have so many zones, it's simpler to fill the map eagerly (and less error-prone).

michalmac (commented):
Nice. I like the idea of `DrtZoneTargetLinkSelector`. I added a few comments to both of its implementations.

contribs/drt/src/main/java/org/matsim/contrib/drt/analysis/zonal/MostCentralDrtZoneTargetLinkSelector.java
tschlenther: I put the map filling into the constructor - hope this is what you meant.

contribs/drt/src/main/java/org/matsim/contrib/drt/analysis/zonal/MostCentralDrtZoneTargetLinkSelector.java
tschlenther: oops

contribs/drt/src/main/java/org/matsim/contrib/drt/analysis/zonal/MostCentralDrtZoneTargetLinkSelector.java
```diff
@@ -0,0 +1,51 @@
...
+		Link closestLink = null;
+		for(Link link : zone.getLinks().values()){
+			// vehicle will be standing at the toNode, this is why we use toNode rather than getCoord
+			double dist = NetworkUtils.getEuclideanDistance(zone.getCentroid(), link.getToNode().getCoord());
+			if(dist < minDistance){
+				minDistance = dist;
+				closestLink = link;
+			}
+		}
+		if(closestLink == null) throw new RuntimeException("could not determin most central link for zone " + zone);
```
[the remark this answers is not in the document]
tschlenther: check

contribs/drt/src/main/java/org/matsim/contrib/drt/analysis/zonal/RandomDrtZoneTargetLinkSelector.java
tschlenther: My first though was:
`Let's stick to the map, we might come up with 'contains' queries later, which could be invoked on the key set then. `

But as we have a map in DrtZonalSystem and the method `getZoneForLinkId`, we have infrastructure for these kind of queries. 

I will change the fiel in DrtZone to a list

michalmac (approved):
🚀

PR #1140, opened 2020-09-06
Support dummy zones & refactor demand estimators

michalmac:
- support creating dummy zones (for including vehicles being outside the zonal system into the rebalancing procedure). The idea: when running rebalancing we could treat each of them as a dummy zone and even assume a reduced (negative??) cost of moving them to the actual DRT zones.
- remove special handling of `expectedDemand==0` from `LinearRebalancingTargetCalculator` (It was used in the past, when grid zones were generated also for areas without any links. Now such zones are filtered out, also for non-grid zonal systems)
- make `ZonalDemandEstimator` return `ToDoubleFunction` (instead of `ToIntFunction`). Demand predictions should be precise, not necessarily integer values (for a given area and time bin). It's the goal of the target calculator to come up with an integer number of vehicles per zone.

### DIFF
--- a/contribs/drt/src/main/java/org/matsim/contrib/drt/analysis/zonal/DrtZone.java
+++ b/contribs/drt/src/main/java/org/matsim/contrib/drt/analysis/zonal/DrtZone.java
@@ -22,6 +22,8 @@ package org.matsim.contrib.drt.analysis.zonal;
 
 import java.util.List;
 
+import javax.annotation.Nullable;
+
 import org.locationtech.jts.geom.prep.PreparedGeometry;
 import org.matsim.api.core.v01.Coord;
 import org.matsim.api.core.v01.network.Link;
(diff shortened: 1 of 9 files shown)


michalmac:
> make ZonalDemandEstimator return ToDoubleFunction (instead of ToIntFunction). Demand predictions should be precise, not necessarily integer values (for a given area and time bin). It's the goal of the target calculator to come up with an integer number of vehicles per zone.

@tschlenther , @vsp-gleich , @luchengqi7 
As a follow up, I would like to pass a time period as an argument to the demand estimator. Depending whether it is 15 min or 1 hour, a different result should be returned. (This makes the demand estimator different to the target calculator, which gives the desired number of vehicles for a given rebalancing round).

tschlenther:
Just to make sure i understand:
The reason to have a different time periond length for estimates than for target value calculation is to avoid overreacting to sudden peaks in estimates?
This would not be needed for all demand estimators (`EqualVehicleDensityZonalDemandEstimator` and `FleetSizeWeightedByPopulationShareDemandEstimator` work on a daily basis anyway). If we decide to implement that, time period width for estimates should always be greater than or equal to rebalancing time interval width.

michalmac:
> The reason to have a different time periond length for estimates than for target value calculation is to avoid overreacting to sudden peaks in estimates?

Right now, it is not clear what demand means. Looking at the existing demand estimator implementations, there are different meanings of it. I would like it to express gross demand in absolute numbers for a given period of time.

> This would not be needed for all demand estimators (`EqualVehicleDensityZonalDemandEstimator` and `FleetSizeWeightedByPopulationShareDemandEstimator` work on a daily basis anyway).

We should convert them into target calculators as this is what they actually compute.

> If we decide to implement that, time period width for estimates should always be greater than or equal to rebalancing time interval width.

It is up to the rebalancing strategy. After switching to doubles (from integers) we are quite flexible with the length of the time period (and also with the zone size). I guess a common case would be to run rebalancing every few minutes and use demand estimates for anything between 5 to  60 min (depending on the system dynamics, e.g. large vs small operating area).

PR #1142, opened 2020-09-08
Refactor target claculators

michalmac:
- remove FleetSizeWeightedByActivityEndsDemandEstimator (similar to PreviousIterationDRTDemandEstimator)
- convert EqualVehicleDensity and EqualVehiclesToPopulationRatio from demand estimators to rebalancing target calculators
- rename LinearRebalancingTargetCalculator to DemandEstimatorAsTargetCalculator
- add consistency checks to MinCostFlowRebalancingStrategyParams
- split ZonalDemandEstimatorWithoutServiceAreaTest into smaller test classes (one per tested class)

Notes:
- As for the alpha/beta params - they are used only in `DemandEstimatorAsTargetCalculator`. However, I would like to apply them to all target calculators (assuming alpha=1 and beta=0 as defaults) to have more configurability at the user level without touching the code. (That would mean that the target calculators would return `double` instead of `int`, and then `alpha * target + beta` would be converted into `int`)
- `EqualRebalancableVehicleDistributionTargetCalculator` and `EqualVehicleDensityTargetCalculator` are very similar. We could generalise them into one.


tschlenther:
You replaced ZonalDemandEstimatorWithoutServiceAreaTest with several new tests. I have a student which is currently working on a version of this test with a service area (on branch DRT-Hugo-Test). The new test structure does not need running simulations, which is nice.

Do you think we should test the results with all the different demand estimators / target calculators with a service area 
or rather test the DrtZonalSystem creation when service area is enabled (or maybe both) ?
Because, the way you now test the demand estimators (more vehicles than zones / less vehicles than zones...) i feel like adding a service area does not really add anything any more..

@hmarciales sorry

michalmac:
> Do you think we should test the results with all the different demand estimators / target calculators with a service area
> or rather test the DrtZonalSystem creation when service area is enabled (or maybe both) ?
> Because, the way you now test the demand estimators (more vehicles than zones / less vehicles than zones...) i feel like adding a service area does not really add anything any more..

Sorry for interfering w some other ongoing work...

IMO a unit test should test one well-defined particular function at a time (e.g. an implementation of an interface method). Possibly with all corner cases (each addressed in a separate test), at least in theory:-)

So I would suggest:
- writing separate tests classes for each target calculator implementation to tests if it works the way we want it (under different conditions)
- writing a separate test class to tests the zonal system and all assumptions behind it (e.g. no zones for geometries without links etc.)

If we test bigger chunks of functionality, e.g. by running a controller, then it is much harder to come up with a reasonable set of test cases, and later on - harder to maintain them (too many things going on inside).

tschlenther:
Thanks. I'll give the student the task to test the DrtZonalSystem generation with a service area (within DrtZonalSystemTest), instead.

PR #1143, opened 2020-09-08
Refine previous iteration DRT demand estimator

michalmac:
- clean up
- remove limit on max time
- re-do unit tests (no need to run a controller)


michalmac:
> * re-do unit tests (no need to run a controller)

@tschlenther Regarding your latest question, this is how I would approach testing demand estimators.

PR #1146, opened 2020-09-09
add main mode in trips csv

vsp-gleich:


tschlenther:
i just merged #1145 into the master, you might need to update again

PR #1151, opened 2020-09-15
use target alpha in EqualVehicleDensityTargetCalculator

tschlenther:
alpha controls the share of the fleet used for rebalancing. in some experiments we saw that it can be beneficial not to attempt to spread the entire fleet at each rebalancing time but rather redistribute half of it more frequently.


tschlenther:
same statement is valid for EqualVehiclesToPopulationRatioTargetCalculator

michalmac:
@tschlenther 
I am now working on applying both `alpha` and `beta` for all target calculators (and outside them). So first target calculators will compute the `target` value (a double) and then the actual target would be modified with `alpha * target + beta`

As for the rounding (and deciding whether it is up/down/nearest), I would leave it to the algorithm... perhaps....

tschlenther:
Oh sorry that i interfered that. I did not mean to. Just wanted to restart some experimental runs, but that can also wait for the final structure. I was not sure whether the current state is finsihed, becaue alpha was incorporated in one of the target calculators but not in all of them.
So we just delete this PR ?

michalmac:
No worries. Please keep this PR open, I will maybe steal some test code from you :-)

tschlenther:
@michalmac please go ahead and close this as soon as you don't need it any more :)

michalmac:
> @michalmac please go ahead and close this as soon as you don't need it any more :)

Thanks. Closing it... :-)

PR #1156, opened 2020-09-16
accessibility calculation speed up

nkuehnel:
two minor edits that speed up accessibility calculation (~10 times faster on my local machine with 16 processors)

### DIFF
--- a/contribs/accessibility/src/main/java/org/matsim/contrib/accessibility/AccessibilityAggregator.java
+++ b/contribs/accessibility/src/main/java/org/matsim/contrib/accessibility/AccessibilityAggregator.java
@@ -25,6 +25,7 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.TreeMap;
+import java.util.concurrent.ConcurrentHashMap;
 
 import org.apache.log4j.Logger;
 import org.matsim.api.core.v01.Coord;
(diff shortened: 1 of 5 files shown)


nkuehnel:
@dziemke please review... is replacing the extended least cost path tree ok?

nkuehnel:
oops, seems like this pull request also concludes some commits from the noise branch. sorry

dziemke:
> @dziemke please review... is replacing the extended least cost path tree ok?

The "...Extended" is very old code and was needed for backwards compatibility. Seems like it is not required anymore and it is time to abandon it. Would not have expected such a significant increase though. Did you test this (speed up by factor 10) extensively?

dziemke (approved):

nkuehnel:
@dziemke most speedup came from putting the getNearestLinkExactly() method call for the origin node outside the destination loop. The method contains recursive parts and is now called n times instead of n*m

After fixing that, my profiler showed that the extensive map operations in the additional computations of the extended tree takes up 30% of the computation time. 

After replacing that, what took 10 minutes before now took 1 minute in my use case, that's where x10 comes from :)
Now the actual routing takes up 90% of computation time according to the profiler, which should be the expected bottleneck.

dziemke:
Ok, that 10fold increase on the exchange of the router would have surprised me.

Thanks for finding the unnecessary repeated recompilation of the nearest link. That must have been a leftover from the latest previous refactoring!

PR #1157, opened 2020-09-16
Add teleporting passenger engine

michalmac:
This enables simulating DVRP without actually calling the optimiser 

A potential application: speeding up DRT, see some discussion here: https://github.com/matsim-org/matsim-libs/pull/1057#issuecomment-653741087


michalmac:
@vsp-gleich @ikaddoura In one of the next steps I will (try to) simplify the integration of https://github.com/matsim-vsp/drt-speed-up with DRT

PR #1158, opened 2020-09-16
Drt zonal system test

hmarciales:
New test for contrib DRT, the correct functionality of the functionality with a defined ServiceArea is tested


michalmac:
Thanks for discovering and fixing a bug in the grid utils.

Since I am responsible for maintaining DRT, two suggestions from my side:
- there are some comments in German - would be better to have them in English
- maybe squash this PR instead of merging it to master (it contains many commits with partial changes to the same file) - just to keep the git log smaller

PR #1165, opened 2020-09-18
provide AbstractMobsimModule with iteration number

michalmac:
Background:
- Some parameters in configs define iteration-specific behaviour of MATSim (e.g. do something only in certain iterations)
- mobsim (guice) modules are installed anew while instantiating a mobsim in each iteration, the scope of injection is one mobsim simulation

Providing the current iteration number to mobsim modules will add a context for interpreting the config (which is already available to mobsim modules) and simplify implementing a conditional behaviour, like switching on/off certain mobsim components/features depending on the iteration number.

Possible first application: switching off the vrp optimiser to speedup the mobsim (passengers are teleported instead). Currently this behaviour is hacked by replacing drt legs in plans before mobsim starts (https://github.com/matsim-vsp/drt-speed-up)

### DIFF
--- a/contribs/drt/src/main/java/org/matsim/contrib/drt/optimizer/DrtModeOptimizerQSimModule.java
+++ b/contribs/drt/src/main/java/org/matsim/contrib/drt/optimizer/DrtModeOptimizerQSimModule.java
@@ -1,0 +1,160 @@
+/*
+ * *********************************************************************** *
+ * project: org.matsim.*
+ * *********************************************************************** *
+ *                                                                         *
+ * copyright       : (C) 2018 by the members listed in the COPYING,        *
+ *                   LICENSE and WARRANTY file.                            *
+ * email           : info at matsim dot org                                *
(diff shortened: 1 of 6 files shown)


sebhoerl (approved):

sebhoerl:
So basically this would also allow to simplify a couple of other things down the line, like adding / removing listeners for writing statistics etc, I suppose?

michalmac:
> So basically this would also allow to simplify a couple of other things down the line, like adding / removing listeners for writing statistics etc, I suppose?

Exactly. However they need to be instantiated per mobsim. In the past having mobsim-scope event listeners was harder to code, so probably most of the existing ones have the controller scope. Now it is much easier with mobsim modules (`MobsimScopeEventHandling`).

kainagel (approved):
fine with me.  I think that we need to be a bit careful to not pull to many things to "before" the guice scope, but iteration number in addition to config makes sense to me.

PR #1166, opened 2020-09-20
Add passenger handler

michalmac:
Main changes:
- extract a superinterface `PassengerHandler` (picking up and dropping of passengers) from `PassengerEngine` (passenger-related actions (should) have access to the passenger handler only)
- add passenger pickedUp/droppedOff events (more handy than person enters/leaves vehicle events; they also have a reference to the corresponding request ID - important if more than one 
- re-design passenger events parsing


tschlenther:
👍

PR #1180, opened 2020-09-27
Clean up internal handling of event (handlers) in EventsManagerImpl

michalmac:
- remove obsolete searching for handlers of Event interfaces: after changing `Event` from an interface to a class (quite a few years ago), the class hierarchy of events is now a tree, so scanning the parent classes is enough
- increase type-safety of event handling in EventsManagerImpl (references to event classes are now of the type `Class<? extends Event>`)

Side effect:
- registering an event handler with a method `handleEvent(X event)` where `X` is neither `Event` nor its subclass will throw an exception (having such a method is either a bug or a really bad code smell given our design decisions wrt registering handlers)


mrieser (approved):
Thanks, looks fine.

PR #1185, opened 2020-09-29
Update FeedforwardRebalancingStrategy.java

luchengqi7:
Bug fixing: The previous code may add same vehicles to the relocation list twice and that will cause some problem. The bug is now fixed.

### DIFF
--- a/contribs/drt/src/main/java/org/matsim/contrib/drt/optimizer/rebalancing/Feedforward/FeedforwardRebalancingStrategy.java
+++ b/contribs/drt/src/main/java/org/matsim/contrib/drt/optimizer/rebalancing/Feedforward/FeedforwardRebalancingStrategy.java
@@ -92,9 +92,9 @@ public class FeedforwardRebalancingStrategy implements RebalancingStrategy {
 	public List<Relocation> calcRelocations(Stream<? extends DvrpVehicle> rebalancableVehicles, double time) {
 		List<Relocation> relocationList = new ArrayList<>();
 		double timeBin = Math.floor((time + feedforwardSignalLead) / timeBinSize);
-
 		Map<DrtZone, List<DvrpVehicle>> rebalancableVehiclesPerZone = RebalancingUtils
 				.groupRebalancableVehicles(zonalSystem, generalParams, rebalancableVehicles, time);
+		Map<DrtZone, List<DvrpVehicle>> actualRebalancableVehiclesPerZone = new HashMap<>();
 
 		// Feedback part
 		if (feedbackSwitch) {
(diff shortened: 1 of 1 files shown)


contribs/drt/src/main/java/org/matsim/contrib/drt/optimizer/rebalancing/Feedforward/FeedforwardRebalancingStrategy.java
```diff
@@ -124,8 +125,10 @@ public FeedforwardRebalancingStrategy(DrtZonalSystem zonalSystem, Fleet fleet, R
 				// vehileToSend after scaling.
 				int numAvailableVehiclesInZone = 0;
 				if (rebalancableVehiclesPerZone.get(departureZone) != null) {
-					numAvailableVehiclesInZone = (int) rebalancableVehiclesPerZone.get(departureZone).stream()
-							.filter(v -> !relocatedVehicles.contains(v)).count();
+					List<DvrpVehicle> actualRebalancableVehicles = rebalancableVehiclesPerZone.get(departureZone)
+							.stream().filter(v -> !relocatedVehicles.contains(v)).collect(Collectors.toList());
```
michalmac, line 129: The map with lists of rebalancable vehicles should be initialised only once, i.e. before the `for` loop starts.

michalmac: In a broader context (beyond the lines modified in this PR):
- call `rebalancableVehiclesPerZone.get(departureZone)` only once inside the loop and store in a local variable
- removing vehicles from the beginning of an array list is a bit costly, consider removing the last one from the list (line 141-142)

michalmac (commented):
Good spot!

Added a few comments to address before merging.

contribs/drt/src/main/java/org/matsim/contrib/drt/optimizer/rebalancing/Feedforward/FeedforwardRebalancingStrategy.java
```diff
@@ -109,8 +113,15 @@ public FeedforwardRebalancingStrategy(DrtZonalSystem zonalSystem, Fleet fleet, R
...
 		}
+
+		// Connection between feedback and feedforward part
 		Set<DvrpVehicle> relocatedVehicles = relocationList.stream().map(relocation -> relocation.vehicle)
 				.collect(toSet());
+		Map<DrtZone, List<DvrpVehicle>> actualRebalancableVehiclesPerZone = new HashMap<>();
+		for (DrtZone zone : rebalancableVehiclesPerZone.keySet()) {
+			actualRebalancableVehiclesPerZone.put(zone, rebalancableVehiclesPerZone.get(zone).stream()
+					.filter(v -> !relocatedVehicles.contains(v)).collect(Collectors.toList()));
+		}
```
michalmac, line 124: I think this whole section can be moved to inside the `if` structure, but outside the `for` loop, as we need it only if the feedforward is enabled.

michalmac (approved):
Nice! Just one comment from my side.

PR #1187, opened 2020-10-01
SUMO Contrib

rakow:
This PR adds a new contrib named "sumo" for integrating information from microscopic simulation [sumo](https://www.eclipse.org/sumo/)

This first version only provides a network converter allowing to convert the sumo network format.
The benefit is that it makes use of detailed lane information, which is currently not available in the matsim OSM converter.

More details can be found in the README.


tschlenther (approved):
seems to be fine 👍 thank you!

jfbischoff:
Also don't forget to add the contrib to the travis.yml 😉

PR #1191, opened 2020-10-05
move taxi fare params into taxi config

michalmac:
As agreed to during discussions at VSP. Same will be done for DRT fares

PR #1192, opened 2020-10-05
Move drt fares to drt

michalmac:
Like #1191, but this time for DRT.


michalmac:
This PR makes changes to the config file. The `drtfares` group is removed and the `drtfare` parameter sets are moved from there to  the `drt` parameter sets of the corresponding mode.

A few examples from the matsim-maas project of what should be changed: https://github.com/matsim-org/matsim-maas/commit/d056f22981f26c93e3796ab8e45926b29f8710d0

PR #1194, opened 2020-10-08
remove maven source plugin from freight

luchengqi7:
Hello Michal, 

I had a build error (see my previous email) with the matsim-lib (the freight module, which I have not touched before). I tried different methods to solve it. The build error was finally solved after I deleted one line in the pom file (see the commit). I guess that line is the reason for the build error, but I'm not sure. Could you have a look? Thanks!

### DIFF
--- a/contribs/drt/src/main/java/org/matsim/contrib/drt/optimizer/rebalancing/demandestimator/PreviousIterationDRTDemandEstimator.java
+++ b/contribs/drt/src/main/java/org/matsim/contrib/drt/optimizer/rebalancing/demandestimator/PreviousIterationDRTDemandEstimator.java
@@ -35,6 +35,8 @@ import org.matsim.contrib.drt.analysis.zonal.DrtZonalSystem;
 import org.matsim.contrib.drt.analysis.zonal.DrtZone;
 import org.matsim.contrib.drt.run.DrtConfigGroup;
 
+import com.google.common.base.Preconditions;
+
 /**
  * Aggregates PersonDepartureEvents per iteration for the given mode and returns the numbers from the previous iteration
  * as expected demand for the current iteration.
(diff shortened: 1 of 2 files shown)


michalmac (approved):
👍 

I would even remove the whole maven-source-plugin section (+ some old commented lines 19-31) to use the "contribs defaults" and thus keep things simple.

luchengqi7:
Hi Michal,

I have found another thing that we may need to make some changes. 

Perhaps it is a good idea to split the time bin size with the rebalance interval. For Adaptive real time rebalancing strategy, the time bin size and the rebalance interval are unrelated. If the time bin size is too small, there can be some problem. As a  result. a temporary solution of adding a "Math.max" operation is used here. We can have a disucssion for a permanant solution.

luchengqi7:
Sorry, I should split these two changes into two different pull request. But the checking for the first one has not yet finished, so I just add another commit to this pull requests.

michalmac (changes requested):
The main problem with `Math.max(drtCfg.getRebalancingParams().get().getInterval(), 1800);` is that it is a hidden workaround. Some algorithms will be aware of that, some not. I suggest to do it at the API level. Here is my suggestion:
- make the time bin independent on the rebalancing interval, as you pointed out (BTW. it does not need to go into a config group now)
- change `ZonalDemandEstimator.getExpectedDemandForTimeBin(double time)` to `getExpectedDemand(double fromTime, double toTime)` so that the rebalancing algorithm (1) knows what it gets, and (2) can get what it actually needs (different algorithms may work with different planning horizons)

So my suggestion is to drop the last commit and improve the `ZonalDemandEstimator` API in a separate PR.

luchengqi7:
Remove the maven source plug in. The plug in was causing build error in Windows operating system. Removal of the source plug in solve the build error. The other changes in this commit does not belong to this topic and therefore will be moved to other pull requests.

tschlenther:
> 
> 
> Hi,
> what is the reason to have demand aggregate time bin size to at least half an hour?
> In our study, we saw that w larger aggregates perform better in most cases, but i do not really see why we would require a minimum other than runtime reasons..

oh sorry, i saw the commit isolated before i saw this PR. Michal has presented a clean way to handle that, i support it. Never mind my comment above. Sorry

PR #1196, opened 2020-10-09
Refine demand estimator interface

michalmac:
Replace `ToDoubleFunction<DrtZone> getExpectedDemandForTimeBin(double time)` with `ToDoubleFunction<DrtZone> getExpectedDemand(double fromTime, double estimationPeriod)`

`demandEstimationPeriod` parameter was added to `MinCostFlowRebalancingParams`. Not all rebalancing strategies need to know that. Some actually do not use any demand estimation. So I decided to not have it one level above - i.e. in `RebalancingParams`.

### DIFF
--- a/contribs/drt/src/main/java/org/matsim/contrib/drt/optimizer/rebalancing/demandestimator/PreviousIterationDRTDemandEstimator.java
+++ b/contribs/drt/src/main/java/org/matsim/contrib/drt/optimizer/rebalancing/demandestimator/PreviousIterationDRTDemandEstimator.java
@@ -51,14 +51,15 @@ public final class PreviousIterationDRTDemandEstimator implements ZonalDemandEst
 	private final String mode;
 	private final String drtSpeedUpMode;
 	private final int timeBinSize;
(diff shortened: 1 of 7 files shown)


michalmac:
@luchengqi7 I hope this PR will solve the issues raised in https://github.com/matsim-org/matsim-libs/pull/1194 (regarding splitting  rebalancing interval and time bins used for demand estimation).

Later on we can improve `PreviousIterationDRTDemandEstimator` with using weighted averages if the estimation period touches 2 time bins.

luchengqi7 (approved):
Yes, I think this is a good updates. Thanks a lot!

I was adding the timeBinSize to the rebalancingParams. As some strategies will use time bin size and some will not use it. Perhaps it is better to have some repetition than having some unsed parameter. So I will remove my changes in the other branch and use this strucure instead.

michalmac:
> Perhaps it is better to have some repetition than having some unsed parameter.

Yes. My intuition says that it is better to keep these params possibly (or reasonably) lowest in the config-group/param-set tree. Otherwise we will be infested with lots of optional/vague params at higher levels.

PR #1198, opened 2020-10-12
take capacity of transit vehicles into account in transit routing

mrieser:


michalmac:
@mrieser This PR brakes build on appveyor. I guess this may be related to the order test execution. Here is the failure:
```
FlowCapacityTest.testFlowCapacityEfficiencyFactorWithLowValueAndDownscaling:365 expected:<201.0> but was:<199.0>
```

vsp-gleich:
Out of curiosity: How does this work? Is it the router who keeps track of how many agents plan to use which transit vehicle or does it refer to the last iteration's transit vehicle occupancy? If the best route for multiple agents involves the very same almost full vehicle which agent wins and is routed to use that vehicle (closing it for the others)?

mrieser:
@vsp-gleich: An EventHandler (managed by the SwissRailRaptor module) keeps track of the occupancy of the transit vehicles between stops during the simulation.  The router (SwissRailRaptor) can then use this collected data to calculate adapted costs when routing. So it works similar to the car/network-router, where only the state of the last iteration is taken into account. The router **does not** keep track of how many re-planned agents have been newly assigned to a route to update its internal data. If multiple agents are assigned to a nearly-full route, then the agent arriving first at the stop wins, as it is the case currently as well.

Note that the capacity-dependent routing is off by default and must be manually enabled by binding `CapacityDependentInVehicleCostCalculator` to `RaptorInVehicleCostCalculator`, as the default binding just uses the regular costs.


@michalmac: strange, especially as the merged-in feature is disabled by default, and the tests should also have no side-effects. And the `FlowCapacityTest` is actually from hermes which isn't really used by my MR. I'll have a look at what's going one, it might be some index-thingy as Hermes is involved.

vsp-gleich:
@mrieser Thank you for the explanation!

mrieser:
@vsp-gleich: As a note of interest: The changes in this PR also detects overfull transit vehicles where boarding is no longer possible at some stops and tries to find alternative routes for agents in such a cases. To my understanding, this is what the (basically undocumented?) "eventsbasedPTRouter" contrib also tries to solve, although for the older PT router. This feature can be independently enabled using a config-switch (`useCapacityConstraints` in module `swissRailRaptor`).

vsp-gleich:
@mrieser Yes, I also thought that this partly solves what eventsbasedPTRouter tries to solve. 
However, eventsbasedPTRouter also tries to solve deviations from schedule, i.e. it counts departures per time bin and does some frequency-based routing based on that. 
Maybe some day raptor could route on the last iteration's executed schedule (including delays) to take deviations from schedule into account (if someone considers that useful enough to spend resources on that).
Great to have capacity-constrains in raptor now!

mrieser:
@michalmac It should be fixed now, see #1200.

michalmac:
@mrieser Thanks for looking into it!

PR #1204, opened 2020-10-14
Test new drt-speed-up

michalmac:
As mentioned here: https://github.com/matsim-org/matsim-libs/pull/1193#issuecomment-707842382, this PR adds unit tests for drt-speed-up, where instead of running actual matsim, each iteration is replaced with:
1. stub DrtRequestAnalyzer.getPerformedRequestSequences() (to fake simulation)
2. call iter end event (to trigger drt-speed-up postprocessing)
3. assert currentAvgWaitingTime and currentAvgInVehicleBeelineSpeed

@vsp-gleich @ikaddoura I left a few notes inside `DrtSpeedUp`.

The most important one: I expect potential race conditions between drt-speed-up and opt-drt at the end of an iteration (the former using the fleet size to build stats about the current iteration, while the latter adapting the fleet size for the next iteration). My suggestion: move opt-drt adaptations to the iteration start step. Analogy: scoring is done after qsim, but re-planning is carried out at the beginning of next iteration.

### DIFF
--- a/contribs/drt/src/main/java/org/matsim/contrib/drt/analysis/DrtRequestAnalyzer.java
+++ b/contribs/drt/src/main/java/org/matsim/contrib/drt/analysis/DrtRequestAnalyzer.java
@@ -24,6 +24,7 @@ package org.matsim.contrib.drt.analysis;
 
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 
 import javax.annotation.Nullable;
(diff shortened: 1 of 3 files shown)


michalmac:
One more note: Since we crate "performedRequestSequences" (instead of running qsim), we know exactly what is the input for the drt-speed-up, so we can be explicit about what we expect as the output. Additionally, it's faster and less fragile to changes in other parts of matsim.

michalmac:
@vsp-gleich @tschlenther I just updated all projects (I am aware of) that use drt-speedup to use the new (integrated) one.

@ikaddoura I set the matsim dependency in the old (standalone) drt-speed-up project to the latest matsim weekly release before the new one was added there.

michalmac:
> @vsp-gleich @tschlenther I just updated all projects (I am aware of) that use drt-speedup to use the new (integrated) one.

it's not a weekly release though...

vsp-gleich:
> The most important one: I expect potential race conditions between drt-speed-up and opt-drt at the end of an iteration (the former using the fleet size to build stats about the current iteration, while the latter adapting the fleet size for the next iteration). 

Thanks for the test and for letting us know of the potential race condition! Having unit tests is great (and better then the score tests). But maybe we should consider having a small integration test on top? E.g. modify one of the RunDrtExampleIT tests to run until iteration 2 with a speed-up iteration in between. Or is it better if I write a separate test class for the speed-up integration test?

michalmac:
Sure. I suggest maybe adding the speedup to one of the existing ITs? I do not want to make the build much longer.

PR #1205, opened 2020-10-15
LZ4 Improvements

rakow:
Short PR to allow the EventReader to read "xml.lz4" files.

Also, this changes the lz4 format from block to frame format:
As described in this [thread](https://groups.google.com/forum/#!topic/lz4c/1g4cK5Yz_uU), the frame format can be used with the lz4 command line tools and has better interoperability.

I also did some benchmarks and noticed a quite significant regression in reading performance in MATSm 13.

Code snipped run on file with roughly 10 mio events
```java
EventsManager manager = EventsUtils.createEventsManager();
manager.initProcessing();
EventsUtils.readEvents(manager, XML_INPUT + compression);
```

|Benchmark|Mode|Samples|Score|Score Error (99,9%)|Unit|Param: compression|
|-----------|------|---------|-----|-------------------|-----|---------------------|
|MATSIm 12.0|avgt|5|18851,167160|900,754755|ms/op|
|MATSIm 12.0|avgt|5|21035,960020|2579,980226|ms/op|.gz
|MATSIm 12.0|avgt|5|19887,665060|2013,520721|ms/op|.zst

|Benchmark|Mode|Samples|Score|Score Error (99,9%)|Unit|Param: compression|
|-----------|------|---------|-----|-------------------|-----|---------------------|
|MATSIm 13 Snapshot|avgt|5|26010,928020|1946,984640|ms/op|
|MATSIm 13 Snapshot|avgt|5|30828,416980|1151,445063|ms/op|.gz
|MATSIm 13 Snapshot|avgt|5|28895,901280|2273,936140|ms/op|.lz4
|MATSIm 13 Snapshot|avgt|5|29983,215680|1537,772526|ms/op|.zst

I suspect this might be due to the parallel EventHandler causing additional overhead, but could not check in detail yet.

### DIFF
--- a/matsim/src/main/java/org/matsim/core/events/MatsimEventsReader.java
+++ b/matsim/src/main/java/org/matsim/core/events/MatsimEventsReader.java
@@ -75,7 +75,7 @@ public final class MatsimEventsReader implements MatsimReader {
 	@Override
 	public void readFile(final String filename) {
 		String lcFilename = filename.toLowerCase(Locale.ROOT);
-		if (lcFilename.endsWith(".xml") || lcFilename.endsWith(".xml.gz") || lcFilename.endsWith(".xml.zst")) {
+		if (lcFilename.endsWith(".xml") || lcFilename.endsWith(".xml.gz") || lcFilename.endsWith(".xml.zst") || lcFilename.endsWith(".xml.lz4")) {
 			new XmlEventsReader(this.events, this.customEventMappers).readFile(filename );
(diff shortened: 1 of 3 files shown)


mrieser (approved):
I don't know about frame vs. block. From the linked thread it sounds reasonable to change to Frame. It might result in incompatible files if somebody already used the lz4-compression provided by MATSim, as likely such files now can no longer be loaded by MATSim. On the other hand, considering that at least for Events lz4 never worked because of the missing if-statement, I think it's save to assume that lz4 was not yet in widespread use, so it's okay for me to change.

PR #1209, opened 2020-10-18
Switch to node-based beeline distances in DRT (and related)

michalmac:
No change at the API/interface level, only the implementations are slightly modified whenever we estimate beeline distance for vehicular traffic (as cars move basically between nodes).

Additionally:
- refine `MostCentralDrtZoneTargetLinkSelector`
- replace `LinkProvider` (introduced before we moved to Java 8) with `Function`
- switch from link to coord providers in `StraightLineKnnFinder`


michalmac:
@tschlenther Now `MostCentralDrtZoneTargetLinkSelector` does not rely on the order of links so much. As a result, the target link is even "more central". Expect some (rather tiny) differences in the output.

tschlenther:
👍

PR #1211, opened 2020-10-20
enable drt speed up in service area based integration test

vsp-gleich:


michalmac:
Thanks, @vsp-gleich !

PR #1215, opened 2020-11-03
Update travel time validator (based on HERE API)

luchengqi7:
The data structure and the request method of data from HERE API has changed. Therefore, some updates are necessary. In addition, there are also some changes within the MATSim universe. 

@michalmac and @cdobler 
In the "matsim/src/main/java/org/matsim/core/events/ParallelEventsManager.java", the event queue size is set to 65536 by default. With that value, reading the output event file from some scenario (e.g. dusseldorf 10pct) will lead to "queue is full" error. 

Therefore I have used the originally commented out queue size and double that value (the queue was still full with the original value). Then the event reader can read the output event file. However, it seems that there is a build error in the test (Travis CI) due to out of memoery. May I know if there is a better way to decide the size of the queue? Thank you very much!

@ikaddoura @jfbischoff 
It seems that the HERE API has gone through some changes. Therefore, we also need to update the way we request data from the HERE API. With some trial and error, I finally managed to create some validation results again. Perhaps you may want to have a look at the modification to the code? If there is any issue, please feel free to let me know! Thank you very much!


contribs/analysis/src/main/java/org/matsim/contrib/analysis/vsp/traveltimedistance/HereMapsRouteValidator.java
```diff
@@ -101,10 +93,10 @@ public HereMapsRouteValidator(String outputFolder, String appId, String appCode,
 		DecimalFormat df = (DecimalFormat) NumberFormat.getNumberInstance(locale);
 		df.applyPattern(pattern);
 
-		String urlString = "http://route.cit.api.here.com/routing/7.2/calculateroute.json?app_id=" + appId
-				+ "&app_code=" + appCode + "&waypoint0=geo!" + df.format(from.getY()) + "," + df.format(from.getX())
-				+ "&waypoint1=geo!" + df.format(to.getY()) + "," + df.format(to.getX()) + "&departure=" + date + "T"
-				+ Time.writeTime(trip.getDepartureTime()) + "&mode=fastest;car;traffic:enabled";
+		String urlString = "https://router.hereapi.com/v8/routes?" + "&apiKey=" + apiAcessKey + "&transportmode=car&origin="
```
michalmac, line 96: I suggest using https://developer.here.com/documentation/routing/dev_guide/topics/resource-calculate-matrix.html for this type of analysis.

For a set of nodes (zone centroids) and selected departure times, we could compute:
 - skim matrices based on matsim runs (e.g. using matsim-sbb-extensions)
 - matrices with HERE Matrix API

HERE Matrix API is limited to 15x100 matrices per call. Let's say we have 200 zones, then we need only 28 calls to fill up the whole matrix for a given departure time. It should take only a few seconds when making all 28 calls in parallel.

matsim/src/main/java/org/matsim/core/events/ParallelEventsManager.java
```diff
@@ -82,7 +82,8 @@ public ParallelEventsManager(final boolean syncOnTimeSteps, final int numOfThrea
 		this.eventsHandlers = new ArrayList<EventHandler>();
 		this.eventsArraySize = syncOnTimeSteps ? 512 : 32768;
 		this.eventsQueueSize = eventsQueueSize;
-		this.eventQueue = new ArrayBlockingQueue<>(eventsQueueSize);
+//		this.eventQueue = new ArrayBlockingQueue<>(eventsQueueSize);
+		this.eventQueue = new ArrayBlockingQueue<>(1048576 * 64);
```
michalmac, line 86: provide the queue size via a constructor - this should solve the problem without hardcoding the queue size.

michalmac (changes requested):
A few comments from my side.

BTW. I suggest dropping (instead of reverting) the changes in `ParallelEventsManager` to keep the history possibly clean.

mrieser:
> In the "matsim/src/main/java/org/matsim/core/events/ParallelEventsManager.java", the event queue size is set to 65536 by default. With that value, reading the output event file from some scenario (e.g. dusseldorf 10pct) will lead to "queue is full" error.

I think with higher values, there had been problems with tests requiring too much memory.
An alternative might be to use a different add-method for the queue: instead of using `ArrayBlockingQueue.add()` which produces an Exception when the queue is full, one could use `ArrayBlockingQueue.put()` which will wait for space to become available. I don't know the internals on how this code is used, there might be chance that a deadlock occurs instead (e.g. when some other code is waiting for all events being added to the queue, but this cannot happen, as the queue gets full), so it should be tested as well. If that does not happen, the `put` might be the better alternative to the `add`.

contribs/analysis/src/main/java/org/matsim/contrib/analysis/vsp/traveltimedistance/HereMapsRouteValidator.java
michalmac: @luchengqi7 I can help you with it if we decide to go that way (and if we want to invest time on that).

luchengqi7: Thank you very much for the feedback!

Yes. I think it will be a good idea to drop the change in the ParallelEventManager. It is just a tempeory fix. I will revert that commit. 

As for the HERE API, this is in another project. This validator compares the travel time and distance of a trip in our simulation to the same trip based on HERE database. The validator is ran seperately without MATSim iteration process. It intake the event file and extract trips from the event file. Then the trip will be compared to the data acquired from the HERE API. I think we will need the exact departure point for that validation. So I think it is a little bit different from the zonal based things. But that can be a good idea when deciding the cost for traveling between zones at different time of the day. Perhaps that can be a future implementation in the rebalancing/matching strategy.

By the way, do we also use that validator for other purposes (e.g. during or between MATSim iteration)? Because it seems that the original run code does not initialize the CarTripExtractor (an implemntation of event handler).

luchengqi7: Uh oh, I think I failed to drop that comit as the comit also covers other files. I will stash the commit at the end.

luchengqi7:
> which will wait for spa

Hello Marcel Rieser, thank you very much for the information! For this pull request, I will just follow the advice from Michal and construct the Paralleleventhandler with queue size as input parameter, as this pull request is about updating the Travel time and distance validation. We can try to update that setup in a later pull request.

michalmac (approved):
Looks good to me.

contribs/analysis/src/main/java/org/matsim/contrib/analysis/vsp/traveltimedistance/HereMapsRouteValidator.java
michalmac: Thanks for your explanations.

Anyway, if we would like to validate matsim TTs against HERE API in the future, I am happy to join.

PR #1217, opened 2020-11-06
Replace beeline TT with TT matrix in DRT

michalmac:
Compute travel time matrices (using auto-generated grid zones) in order to:
1. stop using beeline distances and TTs in DRT and taxi (insertion, relocation etc.)
2. get more accurate results for `SelectiveInsertionSearch` to minimise rejections.

**Changes in configs**
1. `DvrpTravelTimeMatrixParams` is added to the DVRP config. If not provided this is assumed as default (200 m as the cell size):
```
        <parameterset type="travelTimeMatrix">
            <param name="cellSize" value="200"/>
        </parameterset>
```
Currently there is only 1 global free-speed TT matrix with one exception: if drt/taxi uses a mode-filtered network, a separate matrix is created for such a mode.

2. `estimatedDrtSpeed` and `estimatedBeelineDistanceFactor` are removed from the DRT config.

**Performance**
Since currently only the free-speed TT matrix is available, its computation is done once per matsim run. Usually computing travel time matrices should not take a lot of time (compared to the DRT optimisation) or memory. If the RAM or CPU usage is too high, there are 3 options:
- reduce DVRP routing network to the operation area (also beneficial for the DRT optimiser)
- increase the cell size
- provide a custom matrix (overriding the default guice bindings).

### DIFF
--- a/contribs/drt/src/main/java/org/matsim/contrib/drt/optimizer/insertion/DetourTimeEstimator.java
+++ b/contribs/drt/src/main/java/org/matsim/contrib/drt/optimizer/insertion/DetourTimeEstimator.java
@@ -20,6 +20,7 @@ package org.matsim.contrib.drt.optimizer.insertion;
 
 import org.matsim.api.core.v01.network.Link;
 import org.matsim.contrib.util.distance.DistanceUtils;
+import org.matsim.contrib.zone.skims.DvrpTravelTimeMatrix;
 
 /**
  * @author michalm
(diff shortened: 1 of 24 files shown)


jfbischoff:
nice work 👍

neuma:
@michalmac Had you considered using a cache like [guava](https://github.com/google/guava/wiki/CachesExplained) instead of the Matrix? This would get rid of the size limitations. The cache can start empty and will only add entries once the requests come in. The cache can still be limited by the number of entries or its memory footprint - if this is indeed a problem. Dropping the entries with the fewest requests first would still allow for a maximum of performance gain.

michalmac:
Thanks for the hint!

The initial motivation was to not have a too big matrix, because one idea was about computing such matrices for a few selected times over a day. Then we could use the (faster) selective insertion search without risking so many rejections (due to having more accurate TTs) or even teleporting DRT vehicles (instead of passengers as is the case in drt-speed-up, similar to the teleported PT).

But if this is limiting us too often (I have already heard of a few cases) then maybe the limit should be removed. Anyone interested in removing it, please give a thumb up :-)

michalmac:
> But if this is limiting us too often (I have already heard of a few cases) then maybe the limit should be removed. Anyone interested in removing it, please give a thumb up :-)

Done in #1321

PR #1220, opened 2020-11-07
remove Thread as the parent class from QSim

michalmac:
I do not see it used anywhere.


mrieser:
Have you checked OTFVis? Out of my head, this might be the only reason why QSim was a Thread, so it can be controlled from OTFVis. But I'm not sure, as it would probably make more sense to wrap QSim as Runnable in a Thread if this is indeed necessary, than making QSim extend Thread.

michalmac:
Looking at the code, `QSim` is started in a typical way (i.e. with the `run()` method) in `OTFVis` (I guess in most cases from the main thread). And then `PlayPauseSimulationControl` is used as a way to hack into QSim to provide some sort of control over it.

PR #1235, opened 2020-11-13
Drt parameter test for pooling

jakobrehmann:
Adds tests that examine how the pooling behavior varies based on parameters in the drt config group: maxWaitTime, maxTravelTimeAlpha and maxTravelTimeBeta.


michalmac:
@jakobrehmann Thanks for adding tests. I just added myself as a reviewer. I would like to give my feedback before it gets merged.

michalmac (approved):
Thanks a lot for writing the tests! I left 2 comments. Please address (2). It would be great to also address (1), but that can be done later in a separate PR.

Comments:
1. I strongly suggest having unit test as close to the tested code as possible (let's call them "small unit tests" here). Spinning up the whole controller to test hard constraint makes the test too far away from the tested code. The main reasons against "big tests":
   - hard to maintain such test cases (e.g. the recent change in `DetourTimeProvider` already caused some recently added tests in this PR to fail, so they are now commented out, because it is hard to fix such "too big" tests).
   - hard to to tests all possible execution paths
   - long execution times (-> long build times)
I am aware that we already have too many such "too big" tests. I am slowly converting some of them to smaller ones that are closer to the tested production code. After all, it would be cheaper to write them small from the very beginning :-(

2. The PR changes only one file, bit consists from 15 interim commits. This is a good candidate for squashing (there is an option **Squash and merge**). Since we do not have lots of documentation, looking at the git log helps us understanding why certain decision were made. This is why the log on the master branch should be clean. More on this topic:
   - https://cloudfour.com/thinks/squashing-your-pull-requests/#:~:text=As%20a%20general%20rule%2C%20when,for%20the%20team%20to%20read.
   - https://softwareengineering.stackexchange.com/questions/263164/why-squash-git-commits-for-pull-requests/263172#263172

vsp-gleich:
Hi Michał, 
Jakob left for an internship last week, but I commissioned this test, so I will take care of this pull request.

1. Jakob will be back in January, so he could do this when he is back. I see the point that "small" tests are faster and easier to fix in code restructurings because changes in the expected values are easier to understand (because there is less code interfering with the test results). Still I'm not sure having only "small" tests is sufficient. The internals of drt change rather frequently and it is nice that this test does not touch any of the internals (only config and start from controler). From the drt user perspective such an integration test seems easier to understand and safer versus changes in the drt internals and how they play together with the rest of matsim. 

2. I will do squash and merge.

jakobrehmann:
@michalmac, thank you for the feedback and suggestions. I now see that squashing makes a lot of sense to maintain a clean history. Also, as @vsp-gleich said, I will be glad to start working on small tests when I return to VSP.

PR #1237, opened 2020-11-16
income-dependent marginal utility of money in PersonScoringParameters

tschlenther:
This PR introduces a new implementation of `ScoringParametersForPerson` which accounts for agent-specific values for marginalUtilityOfMoney.
This functionality relies on person attributes.

The resulting marginal utility of money is computed as:

mgnUtilityOfMoney = averageIncome / personalIncome

Note that the global average income is dynamically computed based on the population. However, e.g. in cases of heterogenous spatial distribution, it is possible to provide an agent-specific relevant income attribute. The value of this attribute then replaces global value for average income in the equation formulated above.

Be aware that all agents that have the personal income attribute is taken into account. If no such attribute is found for a person, the subpopulation's scoring parameters are taken (as usual).


ikaddoura (approved):
thanks!!

PR #1245, opened 2020-11-18
Improve usage of TerminationCriterion

sebhoerl:
For a current paper I was implementing convergence criteria (stabilizing of mode use, stabilizing of routed vs. observed travel times), and I noticed a couple of flaws with the `TerminationCriterion` interface, which was, up to now, not as useful as it could be. I noticed that it is easy to implement the interface and stop the simulation with it, yet many of the expected outputs are not generated (e.g. output events), which are usually generated by default in the last iteration, if they are configured to be written at all. The first objective of this PR is therefore to enable this functionality: By letting downstream `ControlerListener`s know that the current iteration is the last, they can set up all the necessary components (events writing, ...) to prepare for generating the output.

There is a second goal of this PR. Convergence criteria are usually based on the last iteration outcome, or the outcome of previous iterations. A convergence criterion would, therefore, decide after the outcome of the last iteration whether the simulation should stop. Unfortunately, at this point, we may not have tracked all the output (e.g. events), because this would have needed to be configured at the beginning of the iteration. Hence, we would need to run one more iteration to produce events output. Yet, in that iteration things may change (or even go terribly wrong) so the convergence criterion is not fulfilled any more. In this case, we may not want to make use of this output and actually run a couple more iterations to avoid stopping at some "outlier state". Therefore, I extended the convergence criterion to make two decisions:

- *Before* the iteration, `TerminationCriterion.mayTerminateAfterIteration` decides whether this *could* be the last iteration. This notifies all the downstream components to write the output, etc.
- *After* the iteration and *if* it was decided that the iteration may be the last, `TerminationCriterion.doTerminate` decides whether the simulation should actually be stopped. In case that out of stochasticity the last written state is an "outlier", here we can decide to continue the simulation.

Sumarry of major changes:
- Changes `TerminationCriterion` interface
- Changed iteration logic in `AbstractController`
- Added `AbstractIterationEvent` deriving from `ControlerEvent` to unify inclusion of `iteration` and `isLastIteration` in a couple of `ControlerEvent`s (e.g. `IterationStartsEvent`). This is useful for letting, e.g. `EventsHandling` know that it should prepare events writing for the current iteration. `ShutdownEvent` now includes the `iteration` to inform `DumpDataAtEnd` where to search for the last output files

### DIFF
--- a/contribs/accessibility/src/main/java/org/matsim/contrib/accessibility/AccessibilityFromEvents.java
+++ b/contribs/accessibility/src/main/java/org/matsim/contrib/accessibility/AccessibilityFromEvents.java
@@ -95,7 +95,7 @@ public final class AccessibilityFromEvents{
 			if ( controlerListener instanceof ShutdownListener ) {
 				MatsimServices controler = null ;
 				boolean unexpected = false ;
-				ShutdownEvent shutdownEvent = new ShutdownEvent( controler, unexpected ) ;
+				ShutdownEvent shutdownEvent = new ShutdownEvent( controler, unexpected, 0 ) ;
 				((ShutdownListener) controlerListener).notifyShutdown( shutdownEvent );
 			}
 		}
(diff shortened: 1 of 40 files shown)


jfbischoff:
Nice work, thanks!

sebhoerl:
@kainagel Any comments on this, or do you think it is fine to push?

kainagel (approved):
Sure, thanks a lot.

PR #1238, opened 2020-11-16
Added request retry handling

steffenaxer:
A rejected request leads normally directly to a stuckAndAbort event. This pull request allows to collect rejected requests and reinsert them into the passenger engine after waiting for a certain time interval (retryInveral). If the maximum age of a request is exceeded (maxRequestAge), the request is thrown away and causes a  stuckAndAbort event. Functionality is controlled by the help of DvrpRequestRetryParams

PR #1247, opened 2020-11-18
Added request retry handling

steffenaxer:
Added request retry handling #1238

### DIFF
--- a/contribs/drt/src/main/java/org/matsim/contrib/drt/analysis/DrtRequestAnalyzer.java
+++ b/contribs/drt/src/main/java/org/matsim/contrib/drt/analysis/DrtRequestAnalyzer.java
@@ -155,6 +155,11 @@ public class DrtRequestAnalyzer implements PassengerRequestRejectedEventHandler,
 		if (event.getMode().equals(mode)) {
 			performedRequestSequences.put(event.getRequestId(),
 					new PerformedRequestEventSequence(requestSubmissions.get(event.getRequestId()), event));
+			
+			// Remove rejections that could be handled due to PassengerEngineWithRetryLogic
+			if (rejectedRequestSequences.containsKey(event.getRequestId())) {
+				rejectedRequestSequences.remove(event.getRequestId());
+			}
 		}
 	}
 
--- a/contribs/drt/src/main/java/org/matsim/contrib/drt/run/DrtConfigGroup.java
+++ b/contribs/drt/src/main/java/org/matsim/contrib/drt/run/DrtConfigGroup.java
@@ -111,6 +111,11 @@ public final class DrtConfigGroup extends ReflectiveConfigGroupWithConfigurableP
 	public static final String IDLE_VEHICLES_RETURN_TO_DEPOTS = "idleVehiclesReturnToDepots";
 	static final String IDLE_VEHICLES_RETURN_TO_DEPOTS_EXP = "Idle vehicles return to the nearest of all start links. See: DvrpVehicle.getStartLink()";
 
+	
+	public static final String RETRY_REQUEST_HANDLING = "retryRequestHandling";
+	static final String RETRY_REQUEST_HANDLING_EXP = "Enable request repetition for rejected request. See: DvrpRequestRetryParams for available parameters";
+
+	
 	public static final String OPERATIONAL_SCHEME = "operationalScheme";
 	static final String OPERATIONAL_SCHEME_EXP = "Operational Scheme, either of door2door, stopbased or serviceAreaBased. door2door by default";
 
@@ -164,6 +169,8 @@ public final class DrtConfigGroup extends ReflectiveConfigGroupWithConfigurableP
 	private boolean changeStartLinkToLastLinkInSchedule = false;
 
 	private boolean idleVehiclesReturnToDepots = false;
+	
+	private boolean retryRequestHandling = false;
 
 	@NotNull
 	private OperationalScheme operationalScheme = OperationalScheme.door2door;
@@ -206,6 +213,7 @@ public final class DrtConfigGroup extends ReflectiveConfigGroupWithConfigurableP
 
 	@Nullable
 	private DrtSpeedUpParams drtSpeedUpParams;
+	
 
 	public DrtConfigGroup() {
 		super(GROUP_NAME);
@@ -236,6 +244,8 @@ public final class DrtConfigGroup extends ReflectiveConfigGroupWithConfigurableP
 		//drt speedup
 		addDefinition(DrtSpeedUpParams.SET_NAME, DrtSpeedUpParams::new, () -> drtSpeedUpParams,
 				params -> drtSpeedUpParams = (DrtSpeedUpParams)params);
+		
+
 	}
 
 	@Override
@@ -310,6 +320,7 @@ public final class DrtConfigGroup extends ReflectiveConfigGroupWithConfigurableP
 		map.put(REJECT_REQUEST_IF_MAX_WAIT_OR_TRAVEL_TIME_VIOLATED,
 				REJECT_REQUEST_IF_MAX_WAIT_OR_TRAVEL_TIME_VIOLATED_EXP);
 		map.put(DRT_SERVICE_AREA_SHAPE_FILE, DRT_SERVICE_AREA_SHAPE_FILE_EXP);
+		map.put(RETRY_REQUEST_HANDLING,RETRY_REQUEST_HANDLING_EXP);
 		return map;
 	}
 
@@ -497,6 +508,23 @@ public final class DrtConfigGroup extends ReflectiveConfigGroupWithConfigurableP
 	}
 
 	/**
+	 * @return -- {@value #RETRY_REQUEST_HANDLING_EXP}}
+	 */
+	@StringGetter(RETRY_REQUEST_HANDLING)
+	public boolean getRetryRequestHandling() {
+		return retryRequestHandling;
+	}
+
+	/**
+	 * @param retryRequestHandling -- {@value #RETRY_REQUEST_HANDLING_EXP}
+	 */
+	@StringSetter(RETRY_REQUEST_HANDLING)
+	public DrtConfigGroup setRetryRequestHandling(boolean retryRequestHandling) {
+		this.retryRequestHandling = retryRequestHandling;
+		return this;
+	}
+	
+	/**
 	 * @return -- {@value #IDLE_VEHICLES_RETURN_TO_DEPOTS_EXP}}
 	 */
 	@StringGetter(IDLE_VEHICLES_RETURN_TO_DEPOTS)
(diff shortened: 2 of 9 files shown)


steffenaxer:
I read that it might not be allowed to create a pull request from forked master branch to original master branch. Thus, I created a new branch and re-created the pull request. If this doesn't help, I would like to request again the support of more experienced MATSim devs ;-)

michalmac:
@steffenaxer thanks for your PR.

I quickly looked through the code, but before I start reviewing it, I would like to know what is the main motivation for this PR? Is it minimising the impact of rejections? Or maybe you want to simulate some more complex behaviour wrt rejection + resubmission?

sebhoerl:
@mrieser I'll continue the disucssion here as the other PR is closed: As far as I understood the problem is that we require that the "Branch" test passes, but we should enable the "PR" test on Github, then the tests would be run for every PR, isn't it that way? And for Jenkins, I think there seems to be options to build PRs: https://stackoverflow.com/questions/41310720/build-pull-request-from-forked-origins-in-jenkins

Wouldn't it make sense to set this up in a way so people can contribute from a fork, it seems like this is happening more and more as more people like to add their code, and this would be the natural way of doing so? Would it maybe be possible (for Jenkins, if we don't want to overload the server) to require at least one review from a main contributor before the tests are run?

steffenaxer:
@michalmac the reason for this change is mainly motivated by the fact, that MATSim till now has an issue within the mode innovation with modes that has a strong stochastic component that decides whether an agent gets stuck or not. Each single DRT trip can has a relative high likelihood for a rejection, which causes a stuck event.

This effect is highly problematic because a well scored drt person at iteration (i) might receive in iteration (i+1) a stuck event which triggers (with the worst plan selector) that this plan will be eliminated. Thus, evolving plans with various drt trips is something which is pretty hard to reach especially in scenarios in which you simulate extremely undersupplied drt scenarios (small fleets).

Thus, the main intention of this commit is to reduce strongly the likelihood of a rejection which allows a more realistic usage of drt in simulations.

mrieser:
@sebhoerl : 
> the problem is that we require that the "Branch" test passes, but we should enable the "PR" test on Github [...]
> Wouldn't it make sense to set this up in a way so people can contribute from a fork

Yes, this sounds fine. But I'm really no expert in Travis and don't know how to configure this. Looking at the configuration options confuses me more than it helps. Have you experience with this? I've assigned you more rights on the repository, so you should be able to see and change the project settings.

michalmac:
AFAIR we decided to not run travis for PRs, because this slows down our builds. And we also decided to prefer building branches rather than PRs.

michalmac:
Maybe there is a way to configure travis to filter PR builds based on whether it is a fork or not? Then we could run it for "forked PRs"

sebhoerl:
I found the previous discussion: https://github.com/matsim-org/matsim-libs/pull/1026

@michalmac But if your argument is that it takes to long, I'd say we disable the branch builds and just enable PR builds, I think this is also what was the conclusion back then. This would be the same run time?

sebhoerl:
Plus there would also be the option to avoid building anything for branches except `master`

michalmac:
I somehow recall that at some point after that discussion (#1026) we decided the opposite, because it is handier to build branches, no need to create (draft) PRs.

sebhoerl:
Hm, ok, not sure how to decide. From my side, I try to build everything locally before I push anything to not block the queues. So just making PRs build would just enforce this ;-)

michalmac:
I just found out that we probably can do conditional builds in order to build PRs for forks only. 
https://docs.travis-ci.com/user/conditions-v1

Something like `type == push OR (fork == true AND type == pull_request)`

sebhoerl:
Ah ok, yes, sure. But the problem is how to define the push requirement on Github. We can either choose that the Travis "Branch" must pass, or the Travis "PR", or both (but not "any of both"). So if we require "push", it is as before, it will not be possible to accept PRs from forks; if we require "pr", then in your example we would only be able to merge PRs from a fork :)

michalmac:
I assume that some of the code owners (of "matsim main" or of a specific contrib) need to be engaged in the process of reviewing and merging. So if we see that the PR is fully ready, reviewed and its CI build is passing, then the code owner can simply push that branch to `origin`. It's not much work and we do not have so many PRs coming from forks.

sebhoerl:
So this means we need to enable merging without passed checks. Because at the moment even though you are contributor, you cannot push without having the checks passed.... which is a higher risk I would say, because this also measn you can easily do some `git push origin master` by accident locally, which will then be merged without complaint because as a contributor you can always push

michalmac:
> then the code owner can simply push that branch to origin

I meant pushing to a **non-master** branch.

sebhoerl:
What about we require the "PR" build to succeed when something should be merged to `master`, but we configure Travis to always run the branch *except* when the target is `master`. Then only the PR should be build by Travis. Maybe that would work? Would allow people to get feedback on their branches and make the fork / merge process as before.

sebhoerl:
Hm, it is tricky. Because in the way you propose (like here), we don't really see if the code passes the tests. So we would create a new non-master branch and push the code there, then we would see that there are bugs, so we would need to go back to the initial PR to let somebody do the chanegs, etc.... While there would be a simple solution just basing the requirement in pull requests ;) 

I still don't see why the Branch tests are more important, are there many people who really use Travis to test their changes instead of running the tests locally?

vsp-gleich:
> Thus, the main intention of this commit is to reduce strongly the likelihood of a rejection which allows a more realistic usage of drt in simulations.

@steffenaxer Have you tried with rejections switched off? We use that for most of our vsp projects. Do you see an advantage in rejection and re-submission in comparison to no rejection in the first place?

michalmac:
> Hm, it is tricky. Because in the way you propose (like here), we don't really see if the code passes the tests. So we would create a new non-master branch and push the code there, then we would see that there are bugs, so we would need to go back to the initial PR to let somebody do the chanegs, etc.... While there would be a simple solution just basing the requirement in pull requests ;)

I think if we enable builds for PRs coming from forks then we will see if they are green or red. The author of the PR can then continue on his fork and adjust it in order to make it green and also to comply with the code owners's requests. When everyone is happy (travis, author and code owner), the branch gets pushed to our repo and a corresponding PR is created and then merged.

steffenaxer:
@vsp-gleich that has other negative side effects:

1. Very inefficient trips are accepted, that might be far away from hotspot demand areas --> not good for productivity
2. Other customers already part of a scheduled drt might get also affected, because they now get much higher travel times, because max travel time constraint is deactivated. --> not good for service quality
3. Allowing a "customer" to retry a request, while preserving hard constraints, is a more reasonable approach than disabling or soften important constraints

jfbischoff:
I see a good point in what @steffenaxer mentions here. Retrying probably also mimics people's real-life behavior. Rejections are really handy unless you have a service obligation (which is probably more what @vsp-gleich is researching about). 

The trouble with the rejections in general is that they pop-up at very random locations (also centrally located ones, where there is otherwise a good level of service).
Shooting another attempt some minutes later is likely to smoothen this out quite a bit.

michalmac:
> @michalmac the reason for this change is mainly motivated by the fact, that MATSim till now has an issue within the mode innovation with modes that has a strong stochastic component that decides whether an agent gets stuck or not. Each single DRT trip can has a relative high likelihood for a rejection, which causes a stuck event.
> 
> This effect is highly problematic because a well scored drt person at iteration (i) might receive in iteration (i+1) a stuck event which triggers (with the worst plan selector) that this plan will be eliminated. Thus, evolving plans with various drt trips is something which is pretty hard to reach especially in scenarios in which you simulate extremely undersupplied drt scenarios (small fleets).
> 
> Thus, the main intention of this commit is to reduce strongly the likelihood of a rejection which allows a more realistic usage of drt in simulations.

@steffenaxer I am sorry this PR thread got hi-jacked by our side-discussions on CI. Probably we should move the discussion to a separate issue.

What you proposed is really interesting, but has a few drawbacks. The main one is that it requires manipulating (revoking) already emitted rejection events, which may lead to errors in event handling (I see you have adapted `DrtRequestAnalyzer`, but this is only one event handler of many possible). My second concern: if we do not want to have a zero-rejection policy, we need to change the optimiser, so that it either relaxes some conditions checks or buffers requests if the initial scheduling attempt failed (but without sending rejection events).

I see there is a growing need for having zero-rejection DRT optimisation, so I will start working on relaxing the conditions check. In the meantime, you can continue using your approach locally (it's safe since you know about the event manipulation). 

BTW. The simplest way of using your `PassengerEngineWithRetryLogic` is by creating your customised `DrtModeQSimModuleWithRetryRequestHandling` and register it with `Controler.addOverridingQSimModule()`. This way we often use experimental/special features that are not available officially.

michalmac:
> 2. Other customers already part of a scheduled drt might get also affected, because they now get much higher travel times, because max travel time constraint is deactivated. --> not good for service quality

For the currently scheduled passengers, the agreed max wait/travel time constraints is not changed. At this point only traffic jams may cause some time violations.

steffenaxer:
@michalmac "DrtModeQSimModuleWithRetryRequestHandling" I know, and already have it that way. But it sounds (@jfbischoff ) that there is need of this kind of an approach. I want to make clear, that I don't call for a zero-rejection policy. A rejection is a natural phenomena you need to deal with, when thinking on DRT system. But, in combination with MATSims Scoring approach rejections might cause very crucial things, that make it hard to predict a reasonable demand for a certain DRT scenario. 

1st. concern:
Till now, I did not encountered any hard errors. But for sure, it might happen under the curtain. Maybe we can design a more "robust" queue based approach that handles things in a more confident way. 

2nd. concern:
I would not soften the constraints, they are good and they are required as they are. Agents need to have the possibility to retry if the system state is right now not allowing to insert this request. But 30 seconds later, things might look totally different.

sebhoerl:
(@michalmac Ok, I got your point now, we enable PR only for forks, so we can see if they pass, and then we can manually create a new PR to merge it. I think its a good compromise.)

steffenaxer:
On my point of view the suggested approach is the handiest way to do it. The passenger engine listens to the person departure event, thus we are not able to re-initialize a departure in case of a rejection. This would upset much more eventHandlers that are already listening to those kind of events in their correct sequence (act, leg, act), scoring for example...

The passenger related request and rejection events are happening at a very clear defined part of MATsim. From the point of view of the passenger engine, this retry is a completely new story, because the optimizer has already rejected the previous try. I don't see any kind of a problem here.

michalmac:
@steffenaxer thanks for your explanations. I see there are a few votes for having the re-submission feature. Let me have a detailed look at it, and I will come back with some detailed suggestions from my side.

steffenaxer:
@michalmac Thank you for that offer!

michalmac:
Closing this one as a modified version was merged in #1269

PR #1250, opened 2020-11-19
Enable building PRs from forks in Travis

sebhoerl:
As discussed in https://github.com/matsim-org/matsim-libs/pull/1247

The idea is that Travis always tests updated branches, but also pull requests if they come from a fork. We do not want to let Travis test every PR, because this would lead to doubly running the test for the branch and for the PR in most cases.


michalmac:
Seems it works - PR build has started for https://github.com/matsim-org/matsim-libs/pull/1247

PR #1252, opened 2020-11-20
Update the error message of the zone systems (when there are too many zones)

luchengqi7:
Hello Michal, 

I have a suggestion on changing the descrption of the Error message. The original error message may cause some confusion between the DRT zonal system and the DVRP travel time matrix zones. This happened to me these days and I have spent quite some time "debugging" a non-exsiting bug in DRT zonal system generation...

Perhaps it is a good idea to seperate it into two different error information (i.e. one for DRT zonal system and one for the DVRP travel time zones)?

Or perhaps shall we also make the DVRP travel time matrix also depends on shapefile? For example, for the Vulkaneifel scenario, the network of the whole northwestern Germany is used. The default cell size of 200 for the DVRP travel time matrix will yield more than 60,000 zones, which is above limit.

Thanks!

### DIFF
--- a/contribs/dvrp/src/main/java/org/matsim/contrib/zone/skims/Matrix.java
+++ b/contribs/dvrp/src/main/java/org/matsim/contrib/zone/skims/Matrix.java
@@ -49,7 +49,8 @@ public final class Matrix {
 
 	Matrix(Set<Zone> zones) {
 		size = zones.size();
-		checkArgument(size <= MAX_ZONE_COUNT, "Too many zones. Try using bigger cell sizes (when using grid zones)");
+		checkArgument(size <= MAX_ZONE_COUNT, "Too many zones for computing the DVRP travel time matrix."
+				+ " Try using bigger cell sizes (when using grid zones)");
 
 		//to make sure we do not refer to zones added later
 		Arrays.fill(zoneIndex2localIndex, -1);


michalmac:
Thanks. I slightly updated the error message.

> Or perhaps shall we also make the DVRP travel time matrix also depends on shapefile? For example, for the Vulkaneifel scenario, the network of the whole northwestern Germany is used. The default cell size of 200 for the DVRP travel time matrix will yield more than 60,000 zones, which is above limit.

That is possible, but:
- DVRP TT matrix is some sort of "optimisation" that the user should not be too much concerned about
- zones used for the matrix computation are likely different to the zones used for in the DRT for analysis and rebalancing.

Zones in the matrix computation need to cover all links that DRT vehicles can drive through, so often it is a bigger area than the actual service area (to which pickups and dropoffs are restricted). Probably the easiest way to solve this issue (and also speed up simultation) is to use mode filtering for the DVRP routing network.

I would be happy to have more feedback on this from other users and then we can come up with some improvements.

tschlenther:
Hi,
i stumbled over the same problem last week.
After all, it totally makes sense to me that we need (at least?) two different zone systems 
(1. optimizing/TT computation/dispatch and 2. analysis and possibly rebalancing

I do not think that the user should have functionality to set the dvrp travel time zone system per shape file. Instead, we should push users towards defining drt as mode in the network and use a mode-filtered subnetwork. 
Maybe this hint could be included in the error message?

Some additional discussion points:

1. Maybe we should rename DrtZonalSystem? (Possibly into DrtRebalancingZoneSystem)
2. Should it be (generally) be possible to decouple analysis zones from rebalancing zones? For example, by guice (and bind it by default to the same zone system as for rebalancing).

Could be, that a user is rather interested to lean analysis rather on the TT zone system than on the rebalancing zone system...

PR #1260, opened 2020-11-25
Minor emission changes

JWJoubert:
Introduced new traffic situation from HBEFA 4.1; made utils methods public.

### DIFF
--- a/contribs/emissions/src/main/java/org/matsim/contrib/emissions/EmissionModule.java
+++ b/contribs/emissions/src/main/java/org/matsim/contrib/emissions/EmissionModule.java
@@ -375,6 +375,7 @@ public final class EmissionModule {
 			else if(string.endsWith("Heavy")) hbefaTrafficSituation = HbefaTrafficSituation.HEAVY;
 			else if(string.endsWith("Satur.")) hbefaTrafficSituation = HbefaTrafficSituation.SATURATED;
 			else if(string.endsWith("St+Go")) hbefaTrafficSituation = HbefaTrafficSituation.STOPANDGO;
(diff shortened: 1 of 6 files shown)


kt86:
Dear @JWJoubert ,

thanks for this PR!
Sorry for the late reaction. I was off for some days. So it was good, to merge it and not to wait any longer!

Even if this is already merged, I would like to add my comment, because I think that I found the problem with the failing test mentioned here: https://github.com/matsim-org/matsim-libs/blob/30fd15c8ff9a14b157013ef3d3f80680b87a9966/contribs/emissions/src/main/java/org/matsim/contrib/emissions/WarmEmissionAnalysisModule.java#L584-L590.

You should change the value `5`  here https://github.com/matsim-org/matsim-libs/blob/30fd15c8ff9a14b157013ef3d3f80680b87a9966/contribs/emissions/src/test/java/org/matsim/contrib/emissions/TestWarmEmissionAnalysisModuleTrafficSituations.java#L226
to a value below <5. 
Reason: With the speed table defined in https://github.com/matsim-org/matsim-libs/blob/30fd15c8ff9a14b157013ef3d3f80680b87a9966/contribs/emissions/src/test/java/org/matsim/contrib/emissions/TestWarmEmissionAnalysisModuleTrafficSituations.java#L73
this test will calculate for the SG-Speed (10 kmph) now with the (new) SG_Heavy (speed = 5 kmph) value, because 10-5 = 5. And this leads of cause to a wrong result.

_An alternative for that is maybe to set the speed for the new defined SG_Heavy category to a value below 5._ 

---
This is maybe an good point to add a few lines more to the code to test the new calculation as well?

Best regards,
Kai

PR #1261, opened 2020-11-26
verify pushes to branches with GH Actions

michalmac:
Enable a GitHub Actions workflow for pushes to branches.

If the diff between the pushed commit and the master branch`(*)` is limited to contribs, only they are built (less than 15 min. built time). Otherwise also matsim is built (less than 30 min).
`(*)` For pushes to master the diff is done against the most recent commit before that push.

I will reconfigure the checks and make the new workflow required (instead of the travis one) for merging to master.

Sorry for not putting this into a discussion. The ongoing development is stalled, so it needs to be fixed asap. We can later revise some decisions.

==

In the next step I would like to configure the workflow for merging to master. Verifying such pushes is not time critical, so I was thinking about running them using a (very) limited number of parallel jobs, so it does not stall other builds. Something to discuss...

==

During the test period, we could continue using travis for testing PRs coming from external forks (instead of scraping it out completely). Again, something to discuss...

### DIFF
--- a/.github/workflows/verify-push.yaml
+++ b/.github/workflows/verify-push.yaml
@@ -1,0 +1,92 @@
+name: verify-push
+
+on: push
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        module:
+          - matsim
+          # sorted from longest to shortest (to minimise the overall test stage duration)
+          # (used in travis; not respected in GitHub Action workflows)
+          - contribs/vsp
+          - contribs/common
+          - contribs/taxi
+          - contribs/minibus
+          - contribs/signals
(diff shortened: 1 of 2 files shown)


kainagel:
Thanks.

You could also, to make things work again, just switch the acceptance tests from travis to jenkins.

Or not?

michalmac:
> You could also, to make things work again, just switch the acceptance tests from travis to jenkins.

Yes, that would also work as a quick fix. So maybe my excuse was a bit cheap.

Anyway, I am personally a bit sceptical about Jenkins:
- it does not scale with the size of projects and the number of projects (at least the current setup)
- not sure if this is the right setup for open source (most of the contributors, including me, do not have access to it and cannot configure/adjust it)

PR #1262, opened 2020-11-26
run travis only for PRs coming from forks

michalmac:
I suggest we only run travis for external PRs. Or maybe there are some other ideas?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,9 @@ jdk: openjdk11
 # https://docs.travis-ci.com/user/customizing-the-build/#Skipping-the-Installation-Step
 install: true
 
-if: type == push OR (fork == true AND type == pull_request)
+# if: type == push OR (fork == true AND type == pull_request)
+# due to the current quotas, run only for PRs coming from forks
+if: fork == true AND type == pull_request
 
 # For the following, see
 # https://blog.travis-ci.com/2017-09-12-build-stages-order-and-conditions.
@@ -30,7 +32,7 @@ if: type == push OR (fork == true AND type == pull_request)
 
 # define the order of the stages:
 stages:
-  # - package
+   # - package
   - test
 
 # yyyy It would be good to install matsim-main in the package stage


michalmac:
Not really necessary after this change: https://github.com/matsim-org/matsim-libs/pull/1266#issuecomment-734835692

mrieser:
Thanks Michał for all your work with the build systems!

Considering Travis provides a limited number of minutes (1000), and GitHub actions are also limited (I think 2000 minutes per month), would it make sense to distribute the builds, e.g. building branches on GitHub Actions, building PRs on Travis?

Just a thought...

michalmac:
Yes, I was thinking about maintaining travis is a smaller scale (e.g. builds for external PRs or pushes to master, or pushes to master executed on Windows as appveyor is very limited and does not run ITs). Also to have a backup solution.

I am not 100% sure, but to me it looks like the free 2000 min/month is only for private repos. Open-source is fully free (at least now). Maybe I am wrong? Looking at my personal billing I see 0 min used so far. How does it look for the matsim organisation? (I do not have access there:-/ )

mrieser:
Not sure how this "Free for public repos" should be interpreted. We'll see... I gave you additional rights, so you should be able to see it as well now.

michalmac:
> I gave you additional rights, so you should be able to see it as well now.

🎉
Thanks a lot!

kainagel:
Pls let me re-iterate that I read the travis blog such that we can apply, as an open source project, for more hours.  They also say there that one could apply for a number of regularly granted hours (e.g. per month).  I am not against splitting the work so we have a fallback option, but if "running out of build hours" is the main driver, then applying for more hours should be attempted.

michalmac:
Something that actually convinced me to try out GH Actions was the advertised 20 parallel jobs (5 parallel jobs was a bottleneck from time to time). Otherwise I would also opt for applying for more credits.

michalmac:
The situation with travis is a bit messy. Currently, the builds are not performed. According to https://devclass.com/2020/11/25/travis-ci-open-source-engagement/:
> Though renewable amounts are mentioned, the process looked a bit tedious as it has to be repeated every time the credit is exhausted."

Co I am closing this PR for the time being.

PR #1266, opened 2020-11-27
bind LinkBasedRelocationCalculator as singleton

michalmac:
Let's test the new CI for PRs coming from forks.

### DIFF
--- a/.github/workflows/verify-push.yaml
+++ b/.github/workflows/verify-push.yaml
@@ -1,9 +1,12 @@
 name: verify-push
 
-on: push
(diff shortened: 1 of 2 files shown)


michalmac:
Just enabled running PRs from forks (only). Exactly as we recently did for travis.

Works even smoother than with travis: the checks are identified by workflow jobs and not by the trigger type (i.e. push or PR -- as in travis). So no need to push an external PR to our repo (as a branch) to trigger a CI build for the required "build on push" check.

PR #1269, opened 2020-11-29
add DRT request insertion retry functionality

michalmac:
Inspired by https://github.com/matsim-org/matsim-libs/pull/1247

Main differences:
- minimal in terms of required changes in the existing code: the request retry is handled within the `insertion` package (insertion procedure), and so it does not involve changes in the event processing or adding another passenger engine
- request rejection events are not triggered until the final attempt
- the next retry time is computed individually for each unsuccessful request and the requests are queued accordingly (instead of running every N seconds the retry procedure for all unsuccessful requests)

### DIFF
--- a/contribs/drt/src/main/java/org/matsim/contrib/drt/optimizer/insertion/DefaultUnplannedRequestInserter.java
+++ b/contribs/drt/src/main/java/org/matsim/contrib/drt/optimizer/insertion/DefaultUnplannedRequestInserter.java
@@ -20,7 +20,7 @@
 package org.matsim.contrib.drt.optimizer.insertion;
 
 import java.util.Collection;
-import java.util.Iterator;
+import java.util.List;
 import java.util.Optional;
 import java.util.concurrent.ForkJoinPool;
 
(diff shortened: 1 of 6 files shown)


steffenaxer:
Thank you Michal for this new PR proposal. I'm fine with this code, but have one additional remark. In my version it was possible to react within the scoring on all thrown rejection events for a customer in terms of an additional scoring penalty. Thus, I would like to ask whether we could add an special event type, that has the personId of the customer in it (similar to the rejection event).  Maybe something like an intermediateRejectionEvent? Being able to score this intermediateRejectionEvents is an important feature...

michalmac:
Thanks for looking into this PR!

> In my version it was possible to react within the scoring on all thrown rejection events for a customer in terms of an additional scoring penalty.

What if you use the following events: request submitted, passenger picked up and passenger dropped off? Based on these three, you could assess the actual wait/travel times and assign the agent some penalty if the initially assumed max wait/travel times are violated.

steffenaxer:
Well, in principle, additional time consumption for actual wait/travel times for those components can be penalized this way. But I would request for an explicit penalty of retried request. One could measure the time between request and passengerScheduled, but I would have only an implicit number (time) of the total number of retries.

michalmac:
> One could measure the time between request and passengerScheduled, but I would have only an implicit number (time) of the total number of retries

Yes. It's implicit, but very accurate in this case.

We are also (slowly) working on a standardised pre-booking mechanism, where you will get this (or similar) type of events. I hope with the new functionality you will be able to address the raised issue.

steffenaxer:
OK, let's go for this way. Can't wait to see this merged. ;-)

Issue #1248, opened 2020-11-18
HERMES issues and improvements


mrieser:
I came across **some issues** with HERMES that I plan to fix soon when I find some time. Still, I wanted to document them here so others are aware of them:

1. schedules containing stops that are not used in any TransitRoute results in IndexOutOfBoundsExceptions, because Hermes sizes some ArrayList based on the actual number of used stops, instead of the highest `Id<TransitStop>.index()`. 
2. Hermes wrongly assumes that TransitRoute-Ids are globally unique. But in reality, TransitRoute-Ids only have to be unique per TransitLine, but not for the whole schedule. So there could be two or more transit routes with the same Id, if they are in different transit lines. This can be observed in the following (and probably other) places:  
    - `ScenarioImporter#initRoutesStations()` adds an ArrayList for each route, but then accesses this in several places with `transitRoute.getId().index()`, which leads to collisions (transit routes of different transit lines accessing the same data).
    - `ScenarioImporter#initRoutesStations()` contains code like `line_of_route = new int[Id.getNumberOfIds(TransitRoute.class)];` which will lead to problems in such cases.

While detecting these issues and looking at the source code of Hermes, I found some places where probably some **performance optimizations** could be performed:

3. ScenarioImporter:271

    ```java
    if (stopIds.contains(sid)) {
    } else {
      stopIds.add(sid);
    }
    ```
    Simplify to just `.add(sid)`, there's no need to call `.contains` on a set. It just leads to unnecessarily accessing the data-structure twice in some cases.

4. ScenarioImporter:599ff  
There are multiple (at least 3) calls to `stop_ids.get(stopidx)` which could be cached instead of doing multiple calls.

5. ScenarioImporter:557

    ```java
    ArrayList<Integer> stop_ids = this.route_stops_by_index.get(rid);
    ```
    This could probably use `int[]` instead of `ArrayList<Integer>`, which would be much more efficient (storing `int` instead of `Integer` uses much less memory, access might be a tiny bit faster due to less (un)boxing.

6. ScenarioImporter / Realm.agent_stops:

    ```java
	// Agents waiting in pt stations. Should be used as follows:
	// agent_stops.get(curr station id).get(line id).get(dst station id) -> queue of agents
	protected ArrayList<ArrayList<Map<Integer, ArrayDeque<Agent>>>> agent_stops;
    ```
    - ArrayList, 1 entry for each stop
    - Each entry is one ArrayList for the potential TransitLineIds
[...]

PR #1274, opened 2020-12-02
Hermes fixes and improvements

mrieser:
Various changes and fixes for Hermes, see #1248. Mostly related to correctly support rarely used features of the transit schedule, e.g. support multiple stops on the same link, support circular routes, support additional links before the first or after the last stop in a transit route. For this, I had to rewrite parts of the ScenarioImporter. But I've also added several unit tests, so I'm hopeful I didn't break anything.

With all this work, I've also cleaned up the code a bit and optimized it in some places, e.g. using `double[]` instead of `ArrayList<Double>` at some place, using `int[]` instead of `ArrayList<Integer>` in some other place, reducing the number of instantiated ArrayList and HashMaps in various places. In general, the memory consumption of Hermes should now be considerably lower than before.

I did some limited performance testing with a small scenario (20k agents, randomly distributed over the area of the city of Zurich, PT for the area of the city of Zurich). I did 3 runs each with the state of the current master branch, and the state of this branch, and measured the time required for hermes in the first 5 iterations. Then I averaged all the numbers (3 runs x 5 iterations), which resulted in the following:

| | master | hermes_fixes | change |
|---|---:|---:|---:|
| time hermes [s] | 16.38 | 14.11 | -15% |
| min. RAM [GB] | 4.2 | 1.9 | -50% |

Clearly, this is only a single measurement, running on a laptop. Larger scenarios running on different machines might behave differently. But the memory savings are real and performance should be at least as good, if not better, than before.

### DIFF
--- a/matsim/src/main/java/org/matsim/api/core/v01/IdMap.java
+++ b/matsim/src/main/java/org/matsim/api/core/v01/IdMap.java
@@ -126,7 +126,7 @@ public class IdMap<T, V> implements Map<Id<T>, V>, Iterable<V> {
 		return null;
 	}
 
-	V get(int index) {
+	public V get(int index) {
 		if (index < this.data.length) {
 			return (V) this.data[index];
 		}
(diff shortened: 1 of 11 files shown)


jfbischoff (approved):
Thanks for the improvements and adding all these tests!

mrieser:
Fixes #1248

PR #1275, opened 2020-12-03
Rename org.matsim.run.Controler to RunMatsim

### LINKED ISSUE #1273: Core has two Controler Classes - Very Confusing
The matsim core has two controler classes. The 'real' one under ```org.matsim.core.controler.Controler``` and some other under ```org.matsim.run.Controler```. 

I find this very confusing especially because the 'run' package name suggests it could also include the Matsim-Controler Class to start a matsim run. 

This class was created by @mrieser . Maybe you can give some insight whether this could be re-named or removed?

Janekdererste:
We currently have two Controler classes. One is the actual Controler used to control the program flow of the simulation and another one which was supposed to be outward facing infracstucture to make it easy to start matsim. 

This double naming is very confusing, therfore I renamed ```org.matsim.run.Controler``` to ```org.matsim.run.RunMatsim``` and everybody should be happy

fixes #1273


jfbischoff:
Are we still doing this? Having two controler classes keeps on confusing me from time to time :-)

Janekdererste:
Right. I still have this in my personal todo list, but it has never made it to the top. I will try to resolve this before the end of the year!

PR #1281, opened 2020-12-08
Change default freespeed- and capacity factor in OSM reader

rakow:
At the moment the OSM-Reader is reducing the freespeed of urban links by a factor of 0.5 to account for reduced speed due to traffic lights, crossings, etc.

### This PR changes the following in the OSM-Reader:
- Increase the default freespeed factor to 0.9
- Option to set this factor
- Option to configure the length of links that are adjusted

### Reasoning:

We (@Mengzhuoxiao @luchengqi7 @ikaddoura) compared uncongested travel times with routing data from HERE api and noticed a significant difference in both the Düsseldorf and Hamburg scenario.
Travel distances match pretty well, so the freeflow speed reduction done at the moment must be too large. From our experiment we concluded that 0.9 is a more reasonable value, but that might depend on the scenario. Therefore, an option to disable this behavior should also exist.


Janekdererste (commented):
This looks all good. 

I have a general suggestion. Maybe it would make sense to make the urban speed factor directly settable in the builder instead of setting a flag for whether the speed should be adjusted or not. This way we would also cover the case that someone wants to have adjusted speeds, but with another factor. 

Instead of the ```adjustSpeed``` flag and a ```DEFAULT_FREESPEED_FACTOR``` we would only have an ```urbanSpeedFactor``` property in the Reader. The builder could supply 0.9 as default value if nothing is set.

contribs/osm/src/test/java/org/matsim/contrib/osm/networkReader/SupersonicOsmNetworkReaderIT.java
```diff
@@ -27,6 +27,8 @@ public void test_andorra() {
 				.build()
 				.read(Paths.get(utils.getPackageInputDirectory()).resolve("andorra-latest.osm.pbf"));
 
+		NetworkUtils.writeNetwork(network, "expected.xml.gz");
```
Janekdererste, line 30: This probably has to be removed before merging

kainagel:
Good idea.  Try to avoid booleans in configuration except if for all future it is clear that there will only be two plausible options for the setting.

rakow:
We can certainly make this configurable, which might makes sense in this case.
But I void rather like to avoid putting too specific options for such post-processing into the reader.  I would prefer using the functional API, which already exists and can also be composed now:

https://github.com/matsim-org/matsim-libs/blob/a990fa87a07f2cd9f29187357ea01be0fe60fcfc/contribs/osm/src/test/java/org/matsim/contrib/osm/networkReader/SupersonicOsmNetworkReaderIT.java#L38-L44

Janekdererste:
> 
> 
> We can certainly make this configurable, which might makes sense in this case.
> But I void rather like to avoid putting too specific options for such post-processing into the reader. I would prefer using the functional API, which already exists and can also be composed now:
> 
> https://github.com/matsim-org/matsim-libs/blob/a990fa87a07f2cd9f29187357ea01be0fe60fcfc/contribs/osm/src/test/java/org/matsim/contrib/osm/networkReader/SupersonicOsmNetworkReaderIT.java#L38-L44

This approach is fine with me. The way things work at the moment is that all adjustments used by vsp are applied by default. See comments in code.

contribs/osm/src/main/java/org/matsim/contrib/osm/networkReader/LinkProperties.java
```diff
@@ -139,7 +147,7 @@ public static double getLaneCapacity(double linkLength, LinkProperties propertie
      * Calculate free speed of a link based on heuristic if it is an urban link.
      */
 	public static double calculateSpeedIfSpeedTag(double maxSpeed) {
-        double urbanSpeedFactor = maxSpeed < 51 / 3.6 ? 0.5 : 1.0; // assume for links with max speed lower than 51km/h to be in urban areas. Reduce speed to reflect traffic lights and suc
+        double urbanSpeedFactor = maxSpeed < 51 / 3.6 ? DEFAULT_FREESPEED_FACTOR : 1.0;
```
Janekdererste, line 150: if the urbanSpeed factor is supposed to be applied in ```afterLinkCreated``` then it shouldn't be applied here.

contribs/osm/src/main/java/org/matsim/contrib/osm/networkReader/OsmBicycleReader.java
```diff
@@ -34,8 +34,8 @@
 	public OsmBicycleReader(OsmNetworkParser parser,
 							Predicate<Long> preserveNodeWithId,
 							BiPredicate<Coord, Integer> includeLinkAtCoordWithHierarchy,
-							SupersonicOsmNetworkReader.AfterLinkCreated afterLinkCreated) {
-		super(parser, preserveNodeWithId, includeLinkAtCoordWithHierarchy, (link, tags, direction) -> handleLink(link, tags, direction, afterLinkCreated));
+							AfterLinkCreated afterLinkCreated, boolean adjustFreeSpeed) {
+		super(parser, preserveNodeWithId, includeLinkAtCoordWithHierarchy, (link, tags, direction) -> handleLink(link, tags, direction, afterLinkCreated), adjustFreeSpeed);
```
Janekdererste, line 38: the ```adjustFreeSpeed``` flag is not needed

contribs/osm/src/main/java/org/matsim/contrib/osm/networkReader/OsmSignalsReader.java
```diff
@@ -23,12 +23,12 @@
 	OsmSignalsReader(OsmSignalsParser parser,
 					 Predicate<Long> preserveNodeWithId,
 					 BiPredicate<Coord, Integer> includeLinkAtCoordWithHierarchy,
-					 AfterLinkCreated afterLinkCreated) {
+					 AfterLinkCreated afterLinkCreated, boolean adjustFreeSpeed) {
```
Janekdererste, line 27: the ```adjustFreeSpeed``` flag is not needed

contribs/osm/src/main/java/org/matsim/contrib/osm/networkReader/SupersonicOsmNetworkReader.java
```diff
@@ -40,6 +41,7 @@
 
     private final Predicate<Long> preserveNodeWithId;
     private final AfterLinkCreated afterLinkCreated;
+    private final boolean adjustFreeSpeed;
     private final BiPredicate<Coord, Integer> includeLinkAtCoordWithHierarchy;
```
Janekdererste, line 45: the ```adjustFreeSpeed``` flag is not needed

contribs/osm/src/main/java/org/matsim/contrib/osm/networkReader/SupersonicOsmNetworkReader.java
```diff
@@ -48,11 +50,26 @@
...
-                               AfterLinkCreated afterLinkCreated) {
+                               AfterLinkCreated afterLinkCreated,
+                               boolean adjustFreeSpeed) {
+        this.parser = parser;
+        this.preserveNodeWithId = preserveNodeWithId;
         this.includeLinkAtCoordWithHierarchy = includeLinkAtCoordWithHierarchy;
         this.afterLinkCreated = afterLinkCreated;
-        this.preserveNodeWithId = preserveNodeWithId;
-        this.parser = parser;
+        this.adjustFreeSpeed = adjustFreeSpeed;
```
Janekdererste, line 59: the ```adjustFreeSpeed``` flag is not needed

contribs/osm/src/main/java/org/matsim/contrib/osm/networkReader/SupersonicOsmNetworkReader.java
```diff
@@ -235,7 +252,8 @@ boolean isOnewayReverse(Map<String, String> tags) {
 
     private double getFreespeed(Map<String, String> tags, double linkLength, LinkProperties properties) {
         if (tags.containsKey(OsmTags.MAXSPEED)) {
-            return LinkProperties.calculateSpeedIfSpeedTag(parseSpeedTag(tags.get(OsmTags.MAXSPEED), properties));
+            double maxSpeed = parseSpeedTag(tags.get(OsmTags.MAXSPEED), properties);
+            return adjustFreeSpeed ? LinkProperties.calculateSpeedIfSpeedTag(maxSpeed) : maxSpeed;
```
Janekdererste, line 256: This call is no longer necessary if ```afterLinkCreated``` is usded to apply urban speed factor

contribs/osm/src/main/java/org/matsim/contrib/osm/networkReader/SupersonicOsmNetworkReader.java
```diff
@@ -302,16 +320,29 @@ private synchronized void addLinkToNetwork(Link link) {
...
     }
 
     public static abstract class AbstractBuilder<T> {
 
         Map<String, LinkProperties> linkProperties = LinkProperties.createLinkProperties();
         BiPredicate<Coord, Integer> includeLinkAtCoordWithHierarchy = (coord, level) -> true;
         Predicate<Long> preserveNodeWithId = id -> false;
-        AfterLinkCreated afterLinkCreated = (link, tags, isReverse) -> {
-        };
+        AfterLinkCreated afterLinkCreated = (link, tags, isReverse) -> { };
```
Janekdererste, line 343: This would set the default ```afterLinkCreated``` to ```SupersonicOsmNetworkReader.adjustFreespeed(0.9)```

contribs/osm/src/main/java/org/matsim/contrib/osm/networkReader/SupersonicOsmNetworkReader.java
```diff
@@ -391,6 +422,16 @@ private synchronized void addLinkToNetwork(Link link) {
             return this;
         }
 
+        /**
+         * This sets whether the free speed will be adjusted for urban links.
+         *
+         * @see LinkProperties#DEFAULT_FREESPEED_FACTOR
+         */
+        public AbstractBuilder<T> setAdjustFreeSpeed(boolean adjustFreeSpeed) {
```
Janekdererste, line 430: This is not necessary, or if one still wants to have the shortcut version for turning urban speed factor on and off could exchange the default ```afterLinkCreated``` with 
```
this.afterLinkCreated = ajdustFreeSpeed ? SupersonicOsmNetworkReader.adjustFreespeed(0.9) : SupersonicOsmNetworkReader.adjustFreespeed(01.0);
```
Though I'm not sure how to treat cases where the user has already attached an external ```afterLinkCreated``` Consumer

contribs/osm/src/main/java/org/matsim/contrib/osm/networkReader/SupersonicOsmNetworkReader.java
```diff
@@ -391,6 +422,16 @@ private synchronized void addLinkToNetwork(Link link) {
             return this;
```
Janekdererste, line 422: This would have to be changed to: 
```
this.afterLinkCreated.andThen(afterLinkCreated);
return this;
```

contribs/osm/src/main/java/org/matsim/contrib/osm/networkReader/SupersonicOsmNetworkReader.java
```diff
@@ -428,7 +469,7 @@ SupersonicOsmNetworkReader createInstance() {
             return new SupersonicOsmNetworkReader(
                     parser, preserveNodeWithId,
                     includeLinkAtCoordWithHierarchy,
-                    afterLinkCreated
+                    afterLinkCreated, adjustFreeSpeed
             );
```
Janekdererste, line 473: The flag can go away

Janekdererste (commented):
I left comments what would change if we remove the boolean flag from the reader and rely on the afterLinkCreated hook entirely. I think combining those is not a good idea, since one could end up in a situation where the factor is applied twice accidentally

Janekdererste (approved):
This looks good. 

I still don't understand why we need both options as in ```SupersonicOsmNetworkReaderIT(L:25-46)```. I thought the idea was to provide ```SupersonicOsmNetworkReader.adjustFreespeed``` to avoid passing all those explicit parameters to the reader and the corresponding methods. 

If we do it anyway, we could also remove the ```adjustFreespeed``` default implementation. But I'm fine to have it both ways if you think this is a good idea.

rakow:
Now it is the old way, as you described to apply these adjustments as defaults. It makes it easier to create new and standardized networks for future scenarios.  Using this approach it is needed to put all these explicit parameters in the builder I think.

Applying no defaults would require to put them elsewhere for our scenario creation, so we kept them in there at the moment.

PR #1282, opened 2020-12-08
Major Bug fixing for Feedforward Rebalancing Strategy

luchengqi7:
**Major Bug fixing** for Feedforward Rebalancing Strategy:
1. No rebalance in the second iteration
Preoblem: The reset function in the demand estimator (which update the demand database based on previous iteration) is called after the calculation of feedforward signal. Therefore, the feedforward signal for 2nd iteration (i.e. Iteration #1) is always emtpy.
Fix: Add an update function in the demand estimator and it is called before the calculation of the feedforward signal.
2. Wrong surplus calculation
Problem: In the feedforward signal handler, the surplus calculation is wrong. It should be -1 * net departure. 
Fix: List<DrtZoneVehicleSurplus> vehicleSurpluses = zonalSystem.getZones().values().stream()
					.map(z -> new DrtZoneVehicleSurplus(z, **(int) netDepartureInputFunction.applyAsDouble(z) * -1))**
					.collect(toList());
3. Change the requirement of the feedforward signal strength from @positive to @non-negative. This enables the possbility to totally turn off the feedforward part.

Due to the different auto formattor used, some of the other files are also shown as "changed". But in fact there is no changes.

### DIFF
--- a/contribs/drt/src/main/java/org/matsim/contrib/drt/analysis/zonal/DrtZonalWaitTimesAnalyzer.java
+++ b/contribs/drt/src/main/java/org/matsim/contrib/drt/analysis/zonal/DrtZonalWaitTimesAnalyzer.java
@@ -24,7 +24,13 @@ import java.io.BufferedWriter;
 import java.io.IOException;
 import java.text.DecimalFormat;
 import java.text.DecimalFormatSymbols;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Locale;
+import java.util.Map;
+import java.util.SortedSet;
+import java.util.TreeSet;
 
 import org.apache.commons.math3.stat.descriptive.DescriptiveStatistics;
 import org.apache.log4j.Logger;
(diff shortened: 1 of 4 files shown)


luchengqi7:
One additional point to this update (failed test):
The get zone for link ID has been updated. I think it is better to use the same way as it is used to determine wheter the link is within the service area. Otherwise, there will be some request "within" the service area but outside the zonal system.

 For the DRT servcie area, the coordinate of the "toNode" of the link is used (DrtModeModule, line 229), therefore I update the get zone for link ID function. And this cause the test to fail.

michalmac (approved):
LGTM.

As we discussed today, I reduced the changes to bugfixing only

michalmac:
> One additional point to this update (failed test):
> The get zone for link ID has been updated. I think it is better to use the same way as it is used to determine wheter the link is within the service area. Otherwise, there will be some request "within" the service area but outside the zonal system.
> 
> For the DRT servcie area, the coordinate of the "toNode" of the link is used (DrtModeModule, line 229), therefore I update the get zone for link ID function. And this cause the test to fail.

Let's work on this in a separate PR, especially that it (1) requires changes tests, (2) probably needs adaptations in other parts of code.

PR #1283, opened 2020-12-08
Flow efficiency calculation

tschlenther:
The interface FlowEfficiencyCalculator is extended by additional parameters.
This allows for flow efficiency computation depending on situational factors such as vehicle order, time gap between vehicles, road type etc.
This is especially interesting/important in case of AV, where capacity effects can be investigated.
In the vsp contrib, a new implementation of FlowEfficiencyCalculator is provided that represents a hierarchical computation process where situational impacts (such as bunching depending on vehicle type, turn restrictions etc.) can get accounted for.
A situational impact can hereby neglect all other impacts below in the hierarchy (for example if road type is rural no bunching and no turn restriction impact are allowed).

Technically, we need to store the flow capacity consumption of each vehicle in the buffer, as it is needed twice (at buffer entry time and later - when consumption of all vehicles that are still in the buffer since previous time steps is computed) and depends on the time gap of the corresponding vehicle to it's predecessor.


matsim/src/main/java/org/matsim/core/mobsim/qsim/qnetsimengine/flow_efficiency/FlowEfficiencyCalculator.java
```diff
@@ -17,10 +17,10 @@
      *
      * @param qVehicle         the vehicle that consumes efficiency
      * @param previousQVehicle the vehicle previously traveling on this link before {@code qVehicle}, may be null
-     * @param previousTimeDiff time difference of the previous vehicle traveling on this link. Can be null if two vehicles are in the buffer
-     *                         simultaneously.
+     * @param previousTimeDiff time gap to the previous vehicle traveling on this link. Can be null if two vehicles are in the buffer
```
rakow, line 20: Is this still correct? The only time when it will be null is when there is no previous veh, or are there still other cases?

matsim/src/main/java/org/matsim/core/mobsim/qsim/qnetsimengine/flow_efficiency/DefaultFlowEfficiencyCalculator.java
```diff
@@ -1,11 +1,13 @@
...
 import org.matsim.api.core.v01.network.Link;
-import org.matsim.vehicles.Vehicle;
+import org.matsim.core.mobsim.qsim.qnetsimengine.QVehicle;
+import org.matsim.lanes.Lane;
 
 public class DefaultFlowEfficiencyCalculator implements FlowEfficiencyCalculator {
 	@Override
-	public double calculateFlowEfficiency(Vehicle vehicle, Link link) {
-		return vehicle.getType().getFlowEfficiencyFactor();
+    public double calculateFlowEfficiency(QVehicle qVehicle, QVehicle previousQVehicle, Double previousTime, Link link, Id<Lane> laneId) {
```
rakow, line 10: previousTime should be previousTimeDiff  now

matsim/src/main/java/org/matsim/core/mobsim/qsim/qnetsimengine/flow_efficiency/FlowEfficiencyCalculator.java
tschlenther: No you are right. In fact, it will only be null for the first vehicle ever passing the link..

matsim/src/main/java/org/matsim/core/mobsim/qsim/qnetsimengine/flow_efficiency/DefaultFlowEfficiencyCalculator.java
tschlenther: Or maybe timeGapToPreviousVehicle? Or timeDiffToPreviousVehicle?

rakow: I like `timeGapToPreviousVehicle`. But initially I just wanted to point out that the implemented class doesn't use the same name as in the interface. If you rename it, then it should be consistent, as others might use these as examples.

PR #1287, opened 2020-12-11
Get zone for link ID: use to node of the link

luchengqi7:
**This Pull request is open for discussion**

When we determine wheter a link (i.e. DRT stops)  is within the service area, the coordinate of the "toNode" of the link is used (see DrtModeModule, line 229), Therefore, I think it is better to also use "toNode" of the link when determine the zone a link belongs to. If we use the link's coordinate, there are likely to be some requests within the service area but outside any of the zone. This will add some unwanted effect to the analysis. 

Alternatively, we can also change the way we determine whether a link is within the service area (i.e. change the code DrtModeModule).

Attention: the test will fail because this change will change the zone belonging of some links, which will make the test fail. If we want to implment this pull request, we need to update the test as well.

### DIFF
--- a/contribs/drt/src/main/java/org/matsim/contrib/drt/analysis/DrtAnalysisControlerListener.java
+++ b/contribs/drt/src/main/java/org/matsim/contrib/drt/analysis/DrtAnalysisControlerListener.java
@@ -106,8 +106,8 @@ public class DrtAnalysisControlerListener implements IterationEndsListener {
 				filename(event, "drt_rejections", ".csv"),
 				String.join(";", "time", "personId", "fromLinkId", "toLinkId", "fromX", "fromY", "toX", "toY"), seq -> {
 					DrtRequestSubmittedEvent submission = seq.getSubmitted();
-					Coord fromCoord = network.getLinks().get(submission.getFromLinkId()).getCoord();
-					Coord toCoord = network.getLinks().get(submission.getToLinkId()).getCoord();
+					Coord fromCoord = network.getLinks().get(submission.getFromLinkId()).getToNode().getCoord();
+					Coord toCoord = network.getLinks().get(submission.getToLinkId()).getToNode().getCoord();
 					return String.join(";", submission.getTime() + "",//
 							submission.getPersonId() + "",//
 							submission.getFromLinkId() + "",//
--- a/contribs/drt/src/main/java/org/matsim/contrib/drt/analysis/DrtTrip.java
+++ b/contribs/drt/src/main/java/org/matsim/contrib/drt/analysis/DrtTrip.java
@@ -57,9 +57,9 @@ final class DrtTrip {
 		this.person = submittedEvent.getPersonId();
 		this.vehicle = pickedUpEvent.getVehicleId();
 		this.fromLinkId = submittedEvent.getFromLinkId();
-		this.fromCoord = linkProvider.apply(fromLinkId).getCoord();
+		this.fromCoord = linkProvider.apply(fromLinkId).getToNode().getCoord();
 		this.toLink = submittedEvent.getToLinkId();
-		this.toCoord = linkProvider.apply(toLink).getCoord();
+		this.toCoord = linkProvider.apply(toLink).getToNode().getCoord();
 		this.waitTime = pickedUpEvent.getTime() - submittedEvent.getTime();
 		this.unsharedDistanceEstimate_m = submittedEvent.getUnsharedRideDistance();
 		this.unsharedTimeEstimate_m = submittedEvent.getUnsharedRideTime();
--- a/contribs/drt/src/main/java/org/matsim/contrib/drt/analysis/zonal/DrtZonalSystem.java
+++ b/contribs/drt/src/main/java/org/matsim/contrib/drt/analysis/zonal/DrtZonalSystem.java
@@ -74,7 +74,8 @@ public class DrtZonalSystem {
 	@Nullable
 	private static String getGeometryIdForLink(Link link, Map<String, PreparedGeometry> geometries) {
 		//TODO use endNode.getCoord() ?
-		Point linkCoord = MGC.coord2Point(link.getCoord());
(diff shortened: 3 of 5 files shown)


vsp-gleich (approved):
Drt stops created with the service area based approach are located at the toNode's coordinate, so the toNode's coordinate seems the better choice then. With door2door and stops from a TransitSchedule file it's less clear.

michalmac (approved):
👍 

For vehicle-movement related stuff, we have recently fully switched to `link.getToNode().getCoord()`. This was probably the last place when we use `link.getCoord()` as it is also used for computing zonal demand stats (therefore the `TODO` left in the code). But I think it is good if we keep it simple and use `link.getToNode().getCoord()` everywhere.

tschlenther:
> Alternatively, we can also change the way we determine whether a link is within the service area (i.e. change the code DrtModeModule).

For requests, the toNode is probably taken because with standard MATSim, this is where vehicles/persons enter traffic. So this creates some consistency.
I like the idea to stretch that consistency also to rebalancing and drt operation in general! Please move forward :)

> Attention: the test will fail because this change will change the zone belonging of some links, which will make the test fail. If we want to implment this pull request, we need to update the test as well.

Please also update the tests :)

michalmac:
@luchengqi7 I think there are still some places where we use `link.coord`, especially in the DRT analysis package, like this:

https://github.com/matsim-org/matsim-libs/blob/98eee1442a4f69978e5a0f8ee091d2588f79a515/contribs/drt/src/main/java/org/matsim/contrib/drt/analysis/DrtAnalysisControlerListener.java#L109-L110

So probably this was something that we marked it with `TODO`. So I am wondering if we should switch to `link.toNode.coord` there? Any opinions?

vsp-gleich:
> So probably this was the reason for marking the changed line with `TODO`. So I am wondering if we should switch to `link.toNode.coord` there? Any opinions?

I think the current code makes perfect sense for door2door. For stop-based the stop coordinate would be best and for service area based the stop coordinate is the toNode. So I'm not sure. For consistency with the rest of code we might change to the toNode's coord, but it is hard to make it fully consistent as this would make it less consistent with door2door.

michalmac:
> I think the current code makes perfect sense for door2door. For stop-based the stop coordinate would be best and for service area based the stop coordinate is the toNode. So I'm not sure. For consistency with the rest of code we might change to the toNode's coord, but it is hard to make it fully consistent as this would make it less consistent with door2door.

There is an option to use different zones for zonal analysis and different for rebalancing? Just a thought...

michalmac:
> I think the current code makes perfect sense for door2door. For stop-based the stop coordinate would be best and for service area based the stop coordinate is the toNode.

But isn't door2door close to the service-area approach? We just deliberately decided to use `link.toNode.coord` for the service area, but we could also use `link.coord`.

BTW. We are even not super consistent there:
https://github.com/matsim-org/matsim-libs/blob/d4fc4b7ceaeaed8ab30b8abf3d1116c3c50fb30f/contribs/drt/src/main/java/org/matsim/contrib/drt/run/DrtModeModule.java#L229
vs
https://github.com/matsim-org/matsim-libs/blob/9c24cb66643b88de156085f9d18b92bc40cac509/contribs/drt/src/main/java/org/matsim/contrib/drt/routing/DrtStopFacilityImpl.java#L41

(both lines are executed from `createDrtStopNetworkFromServiceArea()`)

vsp-gleich:
> BTW. We are even not super consistent there:
> 
> https://github.com/matsim-org/matsim-libs/blob/d4fc4b7ceaeaed8ab30b8abf3d1116c3c50fb30f/contribs/drt/src/main/java/org/matsim/contrib/drt/run/DrtModeModule.java#L229
> 
> vs
> 
> https://github.com/matsim-org/matsim-libs/blob/9c24cb66643b88de156085f9d18b92bc40cac509/contribs/drt/src/main/java/org/matsim/contrib/drt/routing/DrtStopFacilityImpl.java#L41

Oh dear, I was not aware of that. That has to be fixed. I think we got into the service area based approach when we had difficulties to implement service areas correctly with door2door and therefore maybe thought of it as an area restricted door2door approach where link.getCoord() has sense. But actually it's never real door2door with the service based approach (agents always walk to the center of the link, not the closest location on the link). The systems we model mostly have virtual stops and in real world it makes sense to place those stops at the intersections, i.e. the toNode. So I would move from link.getCoord() to link.getToNode().getCoord().

michalmac:
> So I would move from link.getCoord() to link.getToNode().getCoord().

So we fully switch to `link.getToNode().getCoord()`? Also zonal analysis (to avoid any inconsistencies)? I am fine with it 👍

vsp-gleich:
> So we fully switch to `link.getToNode().getCoord()`? Also zonal analysis (to avoid any inconsistencies)? I am fine with it +1

Yes, I would do so. It's trading consistency with door2door for consistency with service area based, but overall I would say it's an improvement because code would become fully consistent `link.getToNode().getCoord()`.

luchengqi7:
> > So I would move from link.getCoord() to link.getToNode().getCoord().
> 
> So we fully switch to `link.getToNode().getCoord()`? Also zonal analysis (to avoid any inconsistencies)? I am fine with it 👍

I also think it is a good idea👍

jfbischoff:
I think the original idea behind using the center of the link was probably that some links could have two DRT stops and using the link center would treat them in a more equal way.
Given how the whole project has evolved over the years, using toNode throughout the project makes probably more sense.

luchengqi7:
Hello everyone,

I have changed the line 41 of the DrtModeModule.java as well as the testing. Perhaps you can take a look if there is anything that is still missing or inconsistent. Then  I will merge this pull request to the master branch. Thank you very much!


https://github.com/matsim-org/matsim-libs/blob/9c24cb66643b88de156085f9d18b92bc40cac509/contribs/drt/src/main/java/org/matsim/contrib/drt/routing/DrtStopFacilityImpl.java#L41

PR #1291, opened 2020-12-15
Contrib dependency management

rakow:
This PR tries to improve situations in which version conflicts can arise, especially with dvrp contrib and `guava`.

MATSim core is implicitly importing `guava`  due to `guice`, but dvrp is relying on a newer version of `guava`.

If now `dvrp` is imported transitively and not directly, maven will use the older `guava` version of matsim, and thus dvrp will not work correctly. Currently, the developer would manually have to take care of this conflict.

To resolve this, I fixated `guava` and also some other common libraries that might cause problems.
I also pulled most of the duplicated maven config and dependency management into the top level pom.


contribs/pom.xml
```diff
@@ -249,6 +249,32 @@
 		</dependency>
 	</dependencies>
 
+	<!-- Some commonly used dependencies with fixed versions, please try to avoid replicating the versions in contrib poms -->
+	<dependencyManagement>
```
michalmac, line 253: Should we maybe have this section I level up so that everything is kept in sync?

michalmac: One more thought: Maybe we should also do that for other dependencies used in many places (esp. geotools)?

michalmac (approved):
Thanks for cleaning this up. 👍

contribs/pom.xml
rakow: It is not possible to put it higher up, because there is no common parent pom for contribs, core, etc. But maybe that is something that should be changed...

Geotools is only imported directly by the core, so putting it there won't have any effect. I think one needs to define a BOM that can be imported by other projects, in order to fixate such dependencies that might cause problems.

michalmac: Here's the parent pom: https://github.com/matsim-org/matsim-libs/blob/master/pom.xml.

Some geotools dependencies are imported by specific contribs:
https://github.com/matsim-org/matsim-libs/blob/f2ef30cd8b688737e0c71e636e3dc69d5f3b7918/contribs/accessibility/pom.xml#L92-L94
But I agree that without BOM it will be too much of manual work.

michalmac:
> If now `dvrp` is imported transitively and not directly, maven will use the older `guava` version of matsim, and thus drvp will not work correctly. Currently, the developer would manually have to take care of this conflict.

Maybe we should use maven enforcer to detect such issues in the future as they happen from time to time and it's hard to detect version conflicts manually? Just an idea...

contribs/pom.xml
rakow: The pom in contrib does not declare any parent, so I think changes in the mentioned file does not affect it, but I need to double-check.

rakow:
> Maybe we should use maven enforcer to detect such issues in the future as they happen from time to time and it's hard to detect version conflicts manually? Just an idea...

Yes, that would probably be a good idea.

contribs/pom.xml
michalmac: Oh, you're right. I am so much used to always having parents in mvn submodules, but not the case here...

matsim/pom.xml
```diff
@@ -318,6 +262,13 @@
 			<groupId>com.google.inject</groupId>
 			<artifactId>guice</artifactId>
 			<version>${guice.version}</version>
+			<!-- Use newer guava version -->
+			<exclusions>
```
michalmac, line 266: do we need this exclusion given we have now have the guava version specified in `dependencyManagement`?

pom.xml
```diff
@@ -1,22 +1,309 @@
...
+                <version>${geotools.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.geotools.jdbc</groupId>
+                <artifactId>gt-jdbc-postgis</artifactId>
+                <version>${geotools.version}</version>
+            </dependency>
+
+            <dependency>
+                <groupId>com.fasterxml.jackson.core</groupId>
```
michalmac, line 156: I think they also provide BOM, so maybe we could use it instead here?

michalmac (approved):
Nice! So much cleaner POMs 🚀

pom.xml
rakow: Yes, good Idea. I changed it.

rakow:
I'm not quite sure if this work on jenkins correctly. It seems that one build failed. Maybe @mrieser can help to look at that.

contribs/dvrp/pom.xml
```diff
@@ -66,6 +62,13 @@
 			<groupId>ch.sbb</groupId>
 			<artifactId>matsim-sbb-extensions</artifactId>
```
mrieser, line 63: Do you still need that? SwissRailRaptor was migrated to MATSim core. Do you need any of the other parts of the sub-extensions?

mrieser:
> I'm not quite sure if this work on jenkins correctly. It seems that one build failed. Maybe @mrieser can help to look at that.

Can you be more specific to which build you refer?

In general: Not having the parent-relationship in various poms was a concise design decision a while ago. If somebody checks out the MATSim repository and only want's to use the MATSim core, they could just change into that directory and run Maven normally, as it was basically a self-contained Maven project. Now, this no longer works. If you would now do a `mvn install` inside the `matsim` directory, it would likely complain that it can't find the parent pom, as that was never install. So people now first have to do a `mvn install` on the very top level, to get all the required dependencies, before they can start building the sub-projects.

Or in other words: Now, people really have to do a `mvn install` on the whole monolithic repository first. Before, they could just do a `mvn install` on that part that they actually needed.

I don't know how many users are affected by this. Considering that the build server differentiates between "core" and "contrib" and probably never does an install at the top level (or at least not during regular builds, maybe some nightly job does it) this might lead to problems on the build server by having the parent pom's missing or at least outdated.

benchmark/pom.xml
```diff
@@ -34,26 +28,10 @@
 			</plugin>
 			<plugin>
 				<artifactId>maven-assembly-plugin</artifactId>
-				<configuration>
```
mrieser, line 37: Why was the assembly configuration removed?
How do you create a release of the Benchmark now?

rakow: You are right, this was a mistake. I restored the original config.

michalmac:
Just wondering how many people do build matsim only. Maybe reducing the development/maintenance effort is worth giving up the "install only matsim" option. Something to discuss...

BTW. Matsim (main) depends on matsim-examples (test scope), so building it alone is only partially operational.

michalmac:
Oh, I think you can still build only matsim with:
`mvn install --also-make --projects matsim`

This will also build the matsim-examples.

contribs/dvrp/pom.xml
michalmac: We are using `LeastCostPathTree` from SBB to compute trees and matrices.

michalmac:
> I don't know how many users are affected by this. Considering that the build server differentiates between "core" and "contrib" and probably never does an install at the top level (or at least not during regular builds, maybe some nightly job does it) this might lead to problems on the build server by having the parent pom's missing or at least outdated.

If this PR may break builds on Jenkins (snapshot builds) then maybe we should merge it after the holiday season?

rakow:
@mrieser thank you for the comments. It is good to know the reasoning behind it.
In general, I agree with michaz, that it is not worth the effort to keep it separated, if there is another way to build only some subprojects as he suggested.

But let's not break anything for now. For me, it can wait until after the holidays.

rakow:
Is this good to merge now? The last build did work on jenkins without any problems. It also looks like that the CI scripts that are in the repo should already handle this correctly.

michalmac:
I think http://ci.matsim.org:8080/job/MATSim_M2/ and http://ci.matsim.org:8080/job/MATSim_contrib_M2/ may need to be adapted, so that we can still deploy snapshots. Not sure about other build jobs on Jenkins. So I think we should coordinate it with @mrieser

mrieser:
Yes, the jobs `MATSim_M2` and `MATSim_contrib_M2` need to be updated on Jenkins, but likely also the jobs for code coverage. But that should be doable.

I think we could just merge this, and then start fixing the Jenkins jobs one by one.

rakow:
I don't have access to jenkins unfortunately. But please go ahead when you have the time to look into it.

PR #1292, opened 2020-12-16
Adding satisfaction rate to the drt analysis

luchengqi7:
As we have discussed before, it may be a good idea to add in this satisfaction rate into our analysis (See attached picture). 

Definition of satisfaction rate:
SR(x) = N(request with waiting time below x minutes)/N(total requests)

Here we have included SR(10) and SR(15), which are commonly used in different studies.

SatisfactionRate

### DIFF
--- a/contribs/drt/src/main/java/org/matsim/contrib/drt/analysis/DrtTripsAnalyser.java
+++ b/contribs/drt/src/main/java/org/matsim/contrib/drt/analysis/DrtTripsAnalyser.java
@@ -144,8 +144,8 @@ public class DrtTripsAnalyser {
 			directDistanceStats.addValue(trip.unsharedDistanceEstimate_m);
 			traveltimes.addValue(trip.arrivalTime - trip.departureTime);
 		}
-		double satisfactionRate600 = getSatisfactionRate(waitStats, 600);
-		double satisfactionRate900 = getSatisfactionRate(waitStats, 900);
+		double percentageWaitTimeBelow600 = getPercentageWaitTimeBelow(600, waitStats);
+		double percentageWaitTimeBelow900 = getPercentageWaitTimeBelow(900, waitStats);
 		
 		return String.join(delimiter, format.format(waitStats.getValues().length) + "",//
 				format.format(waitStats.getMean()) + "",//
@@ -153,8 +153,8 @@ public class DrtTripsAnalyser {
 				format.format(waitStats.getPercentile(95)) + "",//
 				format.format(waitStats.getPercentile(75)) + "",//
 				format.format(waitStats.getPercentile(50)) + "",//
-				format.format(satisfactionRate600) + "",//
-				format.format(satisfactionRate900) + "",//
+				format.format(percentageWaitTimeBelow600) + "",//
+				format.format(percentageWaitTimeBelow900) + "",//
 				format.format(rideStats.getMean()) + "",//
 				format.format(distanceStats.getMean()) + "",//
 				format.format(directDistanceStats.getMean()) + "",//
@@ -478,7 +478,7 @@ public class DrtTripsAnalyser {
 		return result.toString();
 	}
 	
-	public static double getSatisfactionRate(DescriptiveStatistics stats, int timeCriteria) {
+	public static double getPercentageWaitTimeBelow(int timeCriteria, DescriptiveStatistics stats) {
 		double[] waitingTimes = stats.getValues();
 		
 		if (waitingTimes.length == 0) {
--- a/contribs/drt/src/main/java/org/matsim/contrib/drt/analysis/DrtTripsAnalyser.java
+++ b/contribs/drt/src/main/java/org/matsim/contrib/drt/analysis/DrtTripsAnalyser.java
@@ -485,7 +485,7 @@ public class DrtTripsAnalyser {
 			return 1; // When there is no request in zone, we assume "everyone" is satisfied
 		}
 		
-		int count = 0;
+		double count = 0;
 		for (int i = 0; i < waitingTimes.length; i++) {
 			if (waitingTimes[i] - timeCriteria < 0) {
 				count += 1;
--- a/contribs/drt/src/main/java/org/matsim/contrib/drt/analysis/DrtTripsAnalyser.java
+++ b/contribs/drt/src/main/java/org/matsim/contrib/drt/analysis/DrtTripsAnalyser.java
@@ -479,15 +479,20 @@ public class DrtTripsAnalyser {
 	}
 	
 	public static double getSatisfactionRate(DescriptiveStatistics stats, int timeCriteria) {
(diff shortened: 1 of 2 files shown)


luchengqi7:
@michalmac, as you mentioned, a CDF plot is also very helpful (see attached photo). Do we want Java to plot that CDF or do we want to use other data processing tools (e.g. excel, R, python...) to make that plot based on the output data (in that case, we can put CDFs of several different runs on one plot)? 

CDF

luchengqi7:
By the way, as can be seen in the first plot, the fluctuation of the performance no longer exisits! In this case, the flow and storage capacity of the network are set to very large (i.e. no congestion). And the simple unit capcity matching strategy I have written is used. So now we only need to focus on the fluctuation of the results by extensive insertion serach algorithm between different runs.

tschlenther:
> 
> 
> @michalmac, as you mentioned, a CDF plot is also very helpful (see attached photo). Do we want Java to plot that CDF or do we want to use other data processing tools (e.g. excel, R, python...) to make that plot based on the output data (in that case, we can put CDFs of several different runs on one plot)?
> 
> CDF

i could add this to the home work for my matsim advanced student :) that would mean he has at least 2 plots to create ..

luchengqi7:
> > @michalmac, as you mentioned, a CDF plot is also very helpful (see attached photo). Do we want Java to plot that CDF or do we want to use other data processing tools (e.g. excel, R, python...) to make that plot based on the output data (in that case, we can put CDFs of several different runs on one plot)?
> > CDF
> 
> i could add this to the home work for my matsim advanced student :) that would mean he has at least 2 plots to create ..

Good idea 👍 :)

contribs/drt/src/main/java/org/matsim/contrib/drt/analysis/DrtTripsAnalyser.java
```diff
@@ -472,4 +477,17 @@ public static String summarizeDetailedOccupancyStats(
...
+	
+	public static double getSatisfactionRate(DescriptiveStatistics stats, int timeCriteria) {
+		double[] sortedWaitingTime = stats.getSortedValues();
+		for (int i = 0; i < sortedWaitingTime.length; i++) {
+			if (sortedWaitingTime[i] > timeCriteria) {
+				double numOfRequestsInZone = sortedWaitingTime.length;
+				double output = i / numOfRequestsInZone;
+				return output;
+			}
+		}
```
michalmac, line 489: since data is sorted, we could use a binary search https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Arrays.html#binarySearch(double%5B%5D,double)

michalmac (commented):
Nice!

A few questions to discuss before merging:
- Do we want to have 10 and 15 hardcoded? Also in column names? Max acceptable wait time may be different in different simulations, so 10 and 15 may not always be the best thresholds. Maybe it should be calculated for 50% and 75% of max wait time?
- maybe rename the `satisfaction_rate_` into something which directly says about wait time. We could also have satisfaction rates wrt other criteria.

As for the chart, I think we can do it in another PR.

vsp-gleich:
>     * Do we want to have 10 and 15 hardcoded? Also in column names? Max acceptable wait time may be different in different simulations, so 10 and 15 may not always be the best thresholds. Maybe it should be calculated for 50% and 75% (or some other %) of max wait time
I agree that hardcoded values are not the best option. Still I don't think  50% and 75% (or some other %) of max wait time would be better. There is too much interaction with alpha and beta for the maxWait parameter to be really decisive. We even had runs where we deliberately set a max wait time much lower than the wait time we were aiming for in order to reduce the number of vehicles as far as possible but still make the algorithm serve people at low wait times when the small fleet is underused outside peak hours. If enlarging the config with another parameter is not an option I actually find 10min good for a hardcoded value.

>     * maybe rename the `satisfaction_rate_` into something which directly says about wait time. We could also have satisfaction rates wrt other criteria.
I agree, satisfaction rate is too generic for a name. Maybe _shareTripsBelowXXs_?

luchengqi7:
Thanks for the feedback! Yes, I will change the name to shareTripsBelowXXs, or perhaps shareOfWaitingTimeBelowX. 

The term of satisfaction comes from the idea when designing a good mobility service for the public. Perhaps the criteria used for PT design (e.g. bus, trian frequency) may be agood reference. I have read from some places that 10 minutes is kind of a threshold. If a PT service has an interval lower or equal to 10 mintues, people tends to just pop up at bus/train stop spontanuously without refereing to the timetable. And this can be a good criteria for the DRT service. Meanwhile, the 15 minutes criteria can also be found in some study related to DRT. But it would be also nice to not have those hard coded criteria.

As there is more and more happening for the DRT analysis (I am also trying to playing new criteria, such as availability), maybe we can consider introduce a DRT analysis Config Group? Then we can put those as default values. What do you think?

contribs/drt/src/main/java/org/matsim/contrib/drt/analysis/DrtTripsAnalyser.java
luchengqi7: 👍 I will try with that. Thanks!

michalmac:
> Thanks for the feedback! Yes, I will change the name to shareTripsBelowXXs, or perhaps shareOfWaitingTimeBelowX.

> The term of satisfaction comes from the idea when designing a good mobility service for the public. Perhaps the criteria used for PT design (e.g. bus, trian frequency) may be agood reference. I have read from some places that 10 minutes is kind of a threshold. If a PT service has an interval lower or equal to 10 mintues, people tends to just pop up at bus/train stop spontanuously without refereing to the timetable. And this can be a good criteria for the DRT service. Meanwhile, the 15 minutes criteria can also be found in some study related to DRT. But it would be also nice to not have those hard coded criteria.

For me, "satisfaction rate" is good. I only suggested to include "wait time" there so that we know what criterion it is about. (Potentially we could have a satisfaction rate wrt total travel time, detour, crowdedness ...)

> 
> As there is more and more happening for the DRT analysis (I am also trying to playing new criteria, such as availability), maybe we can consider introduce a DRT analysis Config Group? Then we can put those as default values. What do you think?

If 10 min and 15 min is something quite stable we can start with fixed values.

As for having a DRT analysis config - yes, would be good in the long term. I think we already have some parameters at the top level, so aggregating them into a single group would be a nice cleanup. If you decide to add such a config, I suggest doing that in a separate PR, so we can discuss it.

contribs/drt/src/main/java/org/matsim/contrib/drt/analysis/DrtTripsAnalyser.java
michalmac: Oh, I just looked into `DescriptiveStatistics.getSortedValues()`. It performs sorting every time it is called. So my final recommendation is to:
- call `stats.getValues()`
- loop though the whole array and count the positives

PR #1293, opened 2020-12-16
Write seconds in DrtRoute description 

tschlenther:
Currently, the maxWaitTime and directRideTime are dumped out in the format OptionalTime[%s]. This can not be parsed (with FloatingDecimal.parseDouble), which means, we can currently not read plans with drt routes.

This PR proposes to dump out only the numeric values of maxWaitTime and directRideTime (this is how it used to be) and throw an Exception is one of those is undefined. As legs are either routed or they aren't, it should be okay to throw an exception, shouldn't it?


michalmac (approved):
Thanks for fixing it!

PR #1294, opened 2020-12-16
Add AMD version of patched JOGL version

michaz:


vsp-gleich:
Hi Michael,
if I understand correctly this solves it only for Windows on AMD machines. Could you also add Linux on AMD? I would do it myself if just adding those lines to pom.xml like in this pull requets would be sufficient.

michaz:
Ah. Oops. Yeah.

Now I uploaded the linux-amd64 files. Now it should suffice to add them to the POM like I did.

vsp-gleich:
Thank you @michaz :+1:

PR #1303, opened 2020-12-19
Emissions from positions and re-factored Hbefa-Keys and Values

Janekdererste:
Refactored HbefaEmissionFactorKeys, so that loading can be a little easier. 
Replaced equals/hashCode methods with autogenerated versions to adhere to 'effective java' standards
Changed naming of hbefa-Keys and Values to be slightly shorter (this causes so many changes)

Added First idea of a position based Emissions calculation, so far it only does warm emissions.


kt86:
As discussed. I removed my review request here. Please go ahead.

PR #1311, opened 2020-12-22
Feature/add sbb-extensions contrib

jfbischoff:
This moves the matsim-sbb-extensions from its own repository to a matsim-contrib. We have had that on our to-do list for some time, also due to recent changes to travis building and the overall effort it takes to maintain release for a relatively small repository.

Considering that the extensions are now used in DVRP and given the rapid speed @michalmac is developing things :-) this should be beneficial for refactoring.


michalmac:
Xmas 🎁 from SBB 😄

mrieser (approved):

PR #1323, opened 2021-01-04
add typed setters for inner config groups

kainagel:
Issue I came across while teaching the advanced class. I can see that these typed setters are additional work on the programming side, but they make life easier for users who write the config in code (rather than in xml).


contribs/drt/src/main/java/org/matsim/contrib/drt/run/DrtConfigGroup.java
```diff
@@ -646,4 +648,13 @@ public DrtInsertionSearchParams getDrtInsertionSearchParams() {
 	public Optional<DrtRequestInsertionRetryParams> getDrtRequestInsertionRetryParams() {
 		return Optional.ofNullable(drtRequestInsertionRetryParams);
 	}
+
+	/**
+	 * Convenience method that brings syntax closer to syntax in, e.g., {@link PlansCalcRouteConfigGroup} or {@link PlanCalcScoreConfigGroup}
+	 */
+	public final void addDrtInsertionSearchParams(final DrtInsertionSearchParams pars) {
+		testForLocked() ;
```
michalmac, line 656: Maybe `testForLocked()` could be inlined into `addParameterSet()` (at least in `ReflectiveConfigGroupWithConfigurableParameterSets` if not `ReflectiveConfigGroup`) ?

michalmac (approved):
Looks good to me, but I only reviewed the first commit (related to DVRP). Added one suggestion (as a comment).

contribs/drt/src/main/java/org/matsim/contrib/drt/run/DrtConfigGroup.java
kainagel: I moved it up one level.  I would be fine with moving it up more than that, but don't have time to check the fallout.  In general, this is a device that I introduced so that certain params that cannot change after the scenario was generated can be locked.  But it is not used systematically beyond that.

PR #1331, opened 2021-01-06
Remove event throwing from ColdEmissionAnalysisModule. 

Janekdererste:


kt86 (approved):
Thanks for finding and fixing this bug of throwing the cold emission events twice. We will fix the now ignored tests during our ongoing "cleanup emission tests work"

PR #1341, opened 2021-01-12
Tinker with events manager

Janekdererste:
The default EventsManager for the Qsim didn't provide the functionality I needed for concurrently handling events after the call to 'afterSimStep'. The changed implementation awaits the processing of such events before events of the next time step are accepted. 

Since I had trouble understanding the logic of the present implementation, I tried to simplify the code by using an executor service which receives a task for every event that is processed. 

I tested the speed of the Task based implementation against the Thread based one by running the berlin-scenario (branch 5.5.x) and comparing overall runtimes. The run times with the new implementation were always slightly faster than the old one. (The only goal being not to slow things down by using higher level concurrency)

Since this is quite an important part of matsim could you @rakow maybe have a look. That I didn't do anything fundamentally wrong?


rakow:
Thank you janek, it looks really good! I like the idea of using an executor service.
Such synchronization mechanisms are always inherently complex.
I think there might be a race condition, but these are sometimes nearly impossible to spot without specific tools.

matsim/src/main/java/org/matsim/core/events/SimStepParallelEventsManagerImpl.java
```diff
@@ -181,252 +157,115 @@ public void initProcessing() {
...
-		
-		/*
-		 * After the simulation Events are processed in
-		 * the Main Thread.
-		 */
-		this.parallelMode = false;
 
-		if (throwable != null) {
-			throw new RuntimeException("Exception while processing events. Cannot guarantee that all events have been fully processed.", throwable);
+		log.info("finishProcessing: Before awaiting all event processes");
```
rakow, line 161: In the comment it says that processEvent(..) might be called after finishProcessing().
If this is really done from a different thread than I think a new event could be queued after the phaser advanced, but before init was set to false.

`isInitialized` is an atomic, but never any atomic operations like compareAndSwap are used, which is bit suspicious.

Issue #1343, opened 2021-01-15
Failed builds on jenkins after declaring matsim-all as parent for matsim and contribs


michalmac:
This is related to merging this PR: https://github.com/matsim-org/matsim-libs/pull/1291. The decision was to merge the PR and then fix build issues.

The list of failing jobs (as of now):
- http://ci.matsim.org:8080/job/MATSim_M2/
- http://ci.matsim.org:8080/job/MATSim_contrib_M2/

A few more jobs that are executed periodically (weekly) may also fail. Let's keep an eye on them and update the list.

michalmac:
@marcel I hope you do not mind that I have assigned it to you?

michalmac:
@mrieser I hope you do not mind that I have assigned it to you?

tschlenther:
On a matsim-berlin-branch, i point to matsim 13.0-2021w02-SNAPSHOT. Travis [fails to build it](https://travis-ci.org/github/matsim-scenarios/matsim-berlin/builds/754610512). I understand from the error message that the vsp contrib points to the signals contrib which points to matsim-examples in version 13.0-SNAPSHOT which again points to  matsim-all 13.0-SNAPSHOT which can not be found.
Could this be related to this issue?

rakow:
@tschlenther yes. But is there a reason why the vsp contrib points to a snapshot? I think that should be changed anyway.

michalmac:
I guess this is related to specifying dependency versions using a fixed property:
```xml
<version>${matsim.version}</version>
```
instead of
```xml
<version>13.0-SNAPSHOT</version>
```

rakow:
I was just wondering if there is a reason for it. Probably it was not intended and ${project.version} should have been used.

michalmac:
@tschlenther Not sure if that will temporarily fix your issues, but you could try adding the unresolved dependencies to your POM. I would start with `signals`
```xml
		<dependency>
			<groupId>org.matsim.contrib</groupId>
			<artifactId>signals</artifactId>
			<version>${matsim.version}</version>
			<scope>compile</scope>
		</dependency>
```
This way you should directly point to the weekly release of signals (not the snapshots).

mrieser:
I have reconfigured `MATSim_M2` and `MATSim_contrib_M2` on Jenkins to the best of my knowledge. Let's see...

rakow:
[...]

PR #1344, opened 2021-01-15
use 13.0-SNAPSHOT instead of matsim.version var as version in POMs

michalmac:
a potential fix for https://github.com/matsim-org/matsim-libs/issues/1343#issuecomment-760895531

### DIFF
--- a/contribs/freight/pom.xml
+++ b/contribs/freight/pom.xml
@@ -10,7 +10,6 @@
 	<name>freight</name>
 
 	<properties>
-		<matsim.version>13.0-SNAPSHOT</matsim.version>
 		<jsprit.version>1.8</jsprit.version>
 	</properties>
 
@@ -81,13 +80,13 @@
 		<dependency>
 			<groupId>org.matsim.contrib</groupId>
 			<artifactId>roadpricing</artifactId>
-			<version>${matsim.version}</version>
+			<version>13.0-SNAPSHOT</version>
 		</dependency>
 
 		<dependency>
 			<groupId>org.matsim.contrib</groupId>
 			<artifactId>otfvis</artifactId>
-			<version>${matsim.version}</version>
+			<version>13.0-SNAPSHOT</version>
 		</dependency>
 
 		<dependency>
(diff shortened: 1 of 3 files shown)


michalmac:
> [ERROR] Failed to execute goal on project matsim-berlin: Could not resolve dependencies for project com.github.matsim-scenarios:matsim-berlin:jar:5.5.1: Failed to collect dependencies at org.matsim.contrib:vsp:jar:13.0-2021w02-SNAPSHOT -> org.matsim.contrib:signals:jar:13.0-SNAPSHOT -> org.matsim:matsim-examples:jar:13.0-SNAPSHOT: Failed to read artifact descriptor for org.matsim:matsim-examples:jar:13.0-SNAPSHOT:

It seems that `org.matsim.contrib:vsp:jar:13.0-2021w02-SNAPSHOT` depends on `org.matsim.contrib:signals:jar:13.0-SNAPSHOT` etc. which should not be the case. It should depend on `org.matsim.contrib:signals:jar:13.0-2021w02-SNAPSHOT`.

The problem is that `<version>${matsim.version}</version>` is not being updated while making a release, it still points to `<version>13.0-SNAPSHOT</version>`

However, if we define the version as `<version>13.0-SNAPSHOT</version>` it should get updated by maven.

rakow (approved):

michalmac:
To picture the underlying problem (fixed in this PR), here are the matsim-related dependencies for matsim-berlin that is based on `13.0-2020w51-SNAPSHOT`:

image

Whenever there is `13.0-SNAPSHOT` instead, it means the version of the dependency was not bumped to `13.0-2020w51-SNAPSHOT` during the release process. This release should fix that.

**Potential issues when using past weekly releases:**

Past matsim weekly releases are potentially impacted by this bug. So some unexpected behaviour may happen of different sorts line not-reproducible results or compile errors (in particular, when depending on vsp, freight or integration).

**How to fix it?**

Safest/easiest: try to switch to the upcoming matsim weekly release.

Alternative: It is also possible to fix it by overriding the SNAPSHOT dependencies by explicitly declaring the dependencies. But that is a bit of manual work, and a bit error prone. Suggested in cases when switching to the upcoming weekly release is not possible.

PR #1356, opened 2021-01-28
add constructor which allows for marginal utility of money

ikaddoura:
Logic similar to the money scoring function; allow to plug together a scoring context in which the person specific marginal utility of money is used. mainly meant as a fall back in case the person specific scoring parameters approach uses up too much memory...


tschlenther:
For some reason the roadpricing-contrib build fails, that is also the case in #1355. 
Besides that:

Thank you for pushing this!
I will try to use that with matsim-berlin and observe whether it leads to the same results and also observe runtime.

PR #1366, opened 2021-02-02
rename (misleading) performance indicator from revenue distance to pa…

tschlenther:
…ssenger distance.

The term revenue distance indicates that it can be multiplied with the distance fee in order to obtain the distance-dependent-component of the total operator income. That is not correct, as the indicator also is increased for detours with passengers on board and passengers only pay for direct distances.

Please let me know if i missed places where to adjust the naming or if i misunderstood the code. Thank you :)

### DIFF
--- a/contribs/drt/src/main/java/org/matsim/contrib/drt/analysis/DrtTripsAnalyser.java
+++ b/contribs/drt/src/main/java/org/matsim/contrib/drt/analysis/DrtTripsAnalyser.java
@@ -368,7 +368,7 @@ public class DrtTripsAnalyser {
 	 */
 	public static void writeVehicleDistances(Map<Id<Vehicle>, DrtVehicleDistanceStats.VehicleState> vehicleDistances,
 			String iterationFilename) {
-		String header = "vehicleId;drivenDistance_m;occupiedDistance_m;emptyDistance_m;passengerDistance_pm";
+		String header = "vehicleId;drivenDistance_m;occupiedDistance_m;emptyDistance_m;passengerDistanceTraveled_pm";
 		BufferedWriter bw = IOUtils.getBufferedWriter(iterationFilename);
 		DecimalFormat format = new DecimalFormat();
 		format.setDecimalFormatSymbols(new DecimalFormatSymbols(Locale.US));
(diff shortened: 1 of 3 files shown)


contribs/drt/src/main/java/org/matsim/contrib/drt/analysis/DrtTripsAnalyser.java
```diff
@@ -385,7 +385,7 @@ public static void writeVehicleDistances(Map<Id<Vehicle>, DrtVehicleDistanceStat
 						+ format.format(vehicleState.totalDistance) + ";"//
 						+ format.format(vehicleState.totalOccupiedDistance) + ";"//
 						+ format.format(vehicleState.totalDistanceByOccupancy[0]) + ";"//
-						+ format.format(vehicleState.totalRevenueDistance));
+						+ format.format(vehicleState.totalPassengerMeter));
```
michalmac, line 388: `totalPassengerDistance`? So that it follows the column names ("totalPassengerDistance" and "averagePassengerDistance")

contribs/drt/src/main/java/org/matsim/contrib/drt/analysis/DrtTripsAnalyser.java
```diff
@@ -410,27 +410,26 @@ public static String summarizeVehicles(Map<Id<Vehicle>, DrtVehicleDistanceStats.
...
 		DescriptiveStatistics driven = new DescriptiveStatistics();
-		DescriptiveStatistics revenue = new DescriptiveStatistics();
+		DescriptiveStatistics passengerMeter = new DescriptiveStatistics();
 		DescriptiveStatistics occupied = new DescriptiveStatistics();
 		DescriptiveStatistics empty = new DescriptiveStatistics();
 
 		for (DrtVehicleDistanceStats.VehicleState state : vehicleDistances.values()) {
 			driven.addValue(state.totalDistance);
-			revenue.addValue(state.totalRevenueDistance);
+			passengerMeter.addValue(state.totalPassengerMeter);
```
michalmac, line 419: As above: maybe `passengerDistance`? (consistency)

I am not against adding "meters" at the end, but then we need to update names of other variables/fields

michalmac (approved):
Thanks for fixing this bug!

Not sure about the naming (see comments)

contribs/drt/src/main/java/org/matsim/contrib/drt/analysis/DrtTripsAnalyser.java
michalmac: or "passengerTraveledDistance"

contribs/drt/src/main/java/org/matsim/contrib/drt/analysis/DrtTripsAnalyser.java
michalmac: or `totalPassengerTraveledDistance`?

PR #1369, opened 2021-02-04
Adding lower bound to the storage capcity

luchengqi7:
I think it make sense to set 1.0 as the lower bound for the storage capcity. With that, a link will be able to accommodate at least 1 vehicle. Without that lower bound, short link will have storage capacity less than 1 (especially in the scaled down simulation), and vehicle will get stuck in "traffic" even if there is no traffic at all.

Any idea on this? 

This pull request is based on this Q&A entry: https://github.com/matsim-org/matsim-code-examples/issues/522#issue-801253381


rakow (changes requested):
Please add a comment to the  code that explains why we added this lower bound

rakow (approved):

michalmac:
According to https://github.com/matsim-org/matsim-libs/blob/6a2dcbf2ee8432ccdc070738ed6b8e9d91665d93/matsim/src/main/java/org/matsim/core/mobsim/qsim/qnetsimengine/QueueWithBuffer.java#L661 storage capacity only needs to be above 0. If there is no vehicle at a link (`usedStorageCapacity` is 0), an incoming one will get accepted (if other criteria are also met). 

`0` is a universal reference point. It works regardless of how we scale up/down vehicles. It is not the case with `1`, so it should not be used as a fixed lower bound.

michalmac:
@luchengqi7 
If we are experiencing some issues with `QueueWithBuffer`, I suggest we split it into three different implementations (one for each `TrafficDynamics` option (i.e. `queue`, `withHoles`, and `kinematicWaves`). The current all-in-one queue implementation is very "iffy" (so it's super easy to have some bugs in it).

PR #1370, opened 2021-02-04
Road pricing maintenance

kainagel:


kt86:
Beside some maintenance work, in this PR the calculation of `cordon` toll is removed, since it is too brittle (wants all links inside cordon marked as tolled; if re-entering a tolling zone during the day, no additional toll is calculated). Please use instead the `link`pricing implementation instead and only use the links of the cordon -> where the tolling takes place.

Issue #1349, opened 2021-01-19
How to incorporate trip-level attributes for routing?


sebhoerl:
In a project we are currently working with trip-level attributes. As one example, we are looking at defining the maximum waiting time for DRT trips individually for each trip, so it can be dependent on the person, on the time of day, on the area, etc. This is actually not so easy, because generally not a lot of information is passed down in the whole routing infrastructure.

For instance, `RoutingModule.calcRoute` only gets an origin facility, destination facility, departure time, and the person. A sensible way of defining additional attributes for a trip, is to put them in the preceding activity. Currently, we use the `Person` object with the selected plan and the `departureTime` to reconstruct the `Activity` before the trip that is to be routed in `calcRoute` based on the `departureTime`. However, this is a bit "hacky" so I am wondering what would be the clean way to pass down additional information. Some options would be:

- Pass the preceding (or maybe even both surrounding activities) to the `calcRoute` method. This is how things are mostly done in the DMC extension, but maybe in the context of `RoutingModule` / `TripRouter` this would have any unintended effects (or maybe there are good reasons to not expose the activity to the routing infrastrucutre) ?
- Pass the attributes of the preceding activity to `calcRoute`. Alternatively, have something like `<tripAttributes>` defined for each activity?
- Any other ideas?

In both cases, the relevant methods in `TripRouter` and `RoutingModule` could be extended to have additional parameters. In the activity-based case it may be wise to have two methods and deprecate the current one without the activities. The attributes-based version is less intrusive as we could have two methods with a sensible default for the current case (an empty map without any specific attributes). Still, it may be good to deprecate the current version in this case.

What does the community think? :)

kainagel:
Still had this in my email inbox because I wanted to comment.  Looks like you have moved on in the meantime ... which is completely understandable.   I just wanted to say that we (Gregor Leich and me) had the same problem with the `routingMode` attribute that we wanted to register for each trip.  We ended up putting it into all the legs, and having routines that make this consistent between legs.  This is not great, but it had the advantage to not make a decision on the issue you are facing.

PR #1383, opened 2021-02-10
Pass preceding activity attributes as trip attributes to routing

sebhoerl:
Following up on #1349, this is a proposal of having trip-specific attributes, which can be passed on to the routing modules. This allows routers to take into account trip-specific criteria, e.g. the use of certain service requirements for a DRT or taxi trip, or the use of special connections in an adapted public transport router.

The implementation here passes the attributes of the *preceding activity* on to the `RoutingModule` so they can be treated as "trip attributes". Later on, we could potentially make this explicit by introducing `<tripAttributes></tripAttributes>` in the definition of an activity.

Let me know what you think.

### DIFF
--- a/contribs/accessibility/src/main/java/org/matsim/contrib/accessibility/SwissRailRaptorAccessibilityContributionCalculator.java
+++ b/contribs/accessibility/src/main/java/org/matsim/contrib/accessibility/SwissRailRaptorAccessibilityContributionCalculator.java
@@ -31,6 +31,7 @@ import org.matsim.facilities.*;
 import org.matsim.pt.transitSchedule.TransitStopFacilityImpl;
 import org.matsim.pt.transitSchedule.api.TransitSchedule;
 import org.matsim.pt.transitSchedule.api.TransitStopFacility;
+import org.matsim.utils.objectattributes.attributable.Attributes;
 
 import java.util.*;
 import java.util.concurrent.ConcurrentHashMap;
@@ -139,7 +140,7 @@ class SwissRailRaptorAccessibilityContributionCalculator implements Accessibilit
             Map<Id<? extends BasicLocation>, AggregationObject> aggregatedOpportunities, Double departureTime) {
         double expSum = 0.;
 
-        final Map<Id<TransitStopFacility>, SwissRailRaptorCore.TravelInfo> idTravelInfoMap = raptor.calcTree(origin, departureTime, null);
+        final Map<Id<TransitStopFacility>, SwissRailRaptorCore.TravelInfo> idTravelInfoMap = raptor.calcTree(origin, departureTime, null, new Attributes());
 
         for (final AggregationObject destination : aggregatedOpportunities.values()) {
             //compute direct walk costs
--- a/contribs/accessibility/src/main/java/org/matsim/contrib/accessibility/TripRouterAccessibilityContributionCalculator.java
+++ b/contribs/accessibility/src/main/java/org/matsim/contrib/accessibility/TripRouterAccessibilityContributionCalculator.java
@@ -49,6 +49,7 @@ import org.matsim.facilities.ActivityFacilitiesFactory;
 import org.matsim.facilities.ActivityFacilitiesFactoryImpl;
 import org.matsim.facilities.ActivityFacility;
 import org.matsim.facilities.ActivityFacilityImpl;
+import org.matsim.utils.objectattributes.attributable.Attributes;
 
 /**
  * @author nagel, dziemke
@@ -132,7 +133,7 @@ class TripRouterAccessibilityContributionCalculator implements AccessibilityCont
 			ActivityFacility destinationFacility = activityFacilitiesFactory.createActivityFacility(null, destination.getNearestBasicLocation().getCoord());
 
 			Gbl.assertNotNull(tripRouter);
-			List<? extends PlanElement> plan = tripRouter.calcRoute(mode, origin, destinationFacility, departureTime, null);
+			List<? extends PlanElement> plan = tripRouter.calcRoute(mode, origin, destinationFacility, departureTime, null, new Attributes());
 
 			double utility = 0.;
 			List<Leg> legs = TripStructureUtils.getLegs(plan);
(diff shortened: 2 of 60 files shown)


mrieser:
Adding the "Attributes" to all the calls to `calcRoute` must have been quite some work. Still, I wonder if there would not have been a less-intrusive way...

Some thinking from my side:
- To accommodate the additional "trip attributes", the API of `TripRouter.calcRoute(...)` and `RoutingModule.calcRoute(...)` are extended by an additional argument `Attributes`.
- In the population, the trip attributes cannot be stored on a "trip", as this is not existing as element. Instead, based on the code modifications, the attributes are taken from the "from-activity". I guess this is a reasonable solution, but not obvious / not really documented. (what I want to say: users need to know this, it's hard to figure this out just from the API).
- as the attributes are taken from the from-activity, why not pass them along with the from-facility already passed to the methods?
- `ActivityFacility` already is attributable, `TransitStopFacility` is attributable, `DrtStopFacility` [recently became attributable](https://github.com/matsim-org/matsim-libs/pull/1384). But `Facility` is not yet Attributable. Why not?

As a suggestion:
- Why not making `Facility` attributable, then the trip-attributes could just be passed along the from-facility to all the router-calls, without a need for so many code changes.
- It would even allow to also pass along the attributes on the "to-facility", allowing routing modules e.g. to take infrastructure at the origin or the destination into account.
- Having the attributes in the "from-" and "to-facilities" of `calcRoute(...)` seems to better relate to the attributes in the activities, making it probably also easier to understand for other users.

What do you think about this?

In the suggested case, in most places `FacilitiesUtils.toFacility(...)` is called to get a facility from an activity to be passed to the router. So only this helper-method would have to be adapted to make sure to return a facility object that includes the activities attributes, and that's more or less all the necessary code changes (well, besides adding `extends Attributable` to the `Facility` interface).

sebhoerl:
Hi Marcel, there is a time and a person component about it. The router should not only be responsive to the location of start and end, but also different people may have different preferences when leaving from a certain place. Also, one person may have different preferences whether they leave from home to go to work or to a leisure activity. They may decide for instance for DRT which service to use (e.g. have stricter requirements on the accepted detour etc.). But maybe I got your idea wrong about the facilities? 

Or you want to add the attributes of the activity kind of automatically to the attributes of the facility when it is passed on, kind of like a delegate? I think this would also be confusing as the relevant facilities usually have their own attributes (DRT stops, transit stops, ...) which may have their own meaning in the router.

Don't worry, most of the refactoring was automatic, I'm happy to go with a different solution if we find one :)

mrieser:
> Or you want to add the attributes of the activity kind of automatically to the attributes of the facility when it is passed on, kind of like a delegate?

exactly. Currently, in a lot of cases, a custom Facility is created based on the Activity just for the routing, and I thought it would be easy to just also integrate the attributes there.

So while it is a facility, for the router it's not a "static infrastructure facility", but a "person-dependent facility based on an activity".

sebhoerl:
I agree, also sounds like a viable approach. The only problem I could think of is that somebody was implementing a `RoutingModule` and expects a certain type of facility to which the argument is then cast like `(TransitStopFacility)`. But one could argue that this wasn't the cleanest approach than before. I can check how it works in the code if we have some `RoutingFacility` decorator that is wrapped around the current facilities that are passed to the RoutingModule.

michalmac:
> But Facility is not yet Attributable. Why not?

Maybe we could provide a default method at the level of Attributable (or at least Facility) that returns an empty-unmodifiable attributes. This should help with the transition.

> Adding Attributes tripAttriutes to RoutingModule::calcRoute

If we decide to go in this direction (at some point in the future), I would consider providing trip attributes via `Attributable` instead of passing in `Attributes` directly. Attributes are flexible, but this flexibility comes at a cost (speed and safety). Just an idea...

sebhoerl:
Okay, I want to pick this up again. Re-reading the comments I'm not sure anymore, if we were thinking in the right direction. So initially what I'm after is to have attributes being defined per trip. Now having the attributes attached to a facility seems strange, because facilities are usually the "locations" where trips start and end, while I'm rather looking for user preferences. So there can be one facility as the origin of a trip, but two agents may have different "trip attributes" starting at the same facility. Adding those trip-attributes to an automatically generated facility seems quite like a hack to me.

If you look, for instance, at `FacilitiesUtils.toFacility(Activity activity, ActivityFacilities facilities)` there is the following logic:
- If the facility referenced in the activity exists, return the *actual* `ActivityFacility` (from the facilities file)
- If the facility does not exists, return a `ActivityWrapperFacility`

Now the actual `ActivityFacility` is already attributable, so people may define attributes in the facilities file already. Now what was suggested above is to "merge" the person-specific attributes from the activity into the newly generated activity, which would then not be an `ActivityFacility` anymore, but in any case some wrapper. Do you see my concern?

So I would still say that a sensible option is to have `RoutingModule::calcRoute(..., Attributes tripAttributes)` where the `tripAttributes` come from the preceding activity.

An alternative option could be to somehow give the routing infrastructure a chance to "look up" the necessary information. Like a routing module could have the `Population` and as it already has the `Id<Person>` it could look up the activity if also activites had an ID or something in that direction. But all of that seems more complicated than the option above. What do you think?

sebhoerl:
Maybe another more adventurous approach could be to clean up things in general, so we don't have such a long argument list for all the routing cases:

```
interface RoutingModule {
	public List<? extends PlanElement> calcRoute(RoutingRequest request);
}
```

```
interface RoutingRequest extends Attributable {
	Facility getFromFacility();
	Facility getToFacility();
	double getDepartureTime();
	Optional<Person> getPerson();
}
```

kainagel:
Re "where to attach the info": For the "routingMode" in the end we decided to attach it to every leg in the trip, and to have code that makes this consistent across the legs in a trip.   (I.e. the matsim code actually knows about trips, it is just the xml data model that does not.)

An alternative might be to finally decide that "trip" becomes a bona fide data object, and encode that in xml.

kainagel:
(I would also be ok with a more general Request object.)

PR #1384, opened 2021-02-10
Make DRT stops attributable

sebhoerl:


contribs/drt/src/main/java/org/matsim/contrib/drt/routing/DrtStopFacilityImpl.java
```diff
@@ -27,28 +27,32 @@
...
 
 /**
  * @author Michal Maciejewski (michalm)
  */
 public class DrtStopFacilityImpl implements DrtStopFacility {
-	public static <F extends Identifiable<?> & Facility> DrtStopFacility createFromIdentifiableFacility(F facility) {
+	public static <F extends Identifiable<?> & Facility & Attributable> DrtStopFacility createFromIdentifiableFacility(F facility) {
 		return new DrtStopFacilityImpl(Id.create(facility.getId(), DrtStopFacility.class), facility.getLinkId(),
-				facility.getCoord());
+				facility.getCoord(), facility.getAttributes());
```
michalmac, line 39: Are facilities (practically) always Attributable? If that is not the case, maybe do `if facility instanceof Attributable` then `facility.getAttributes()` else `new Attributes()` ?

michalmac (commented):
LGTM

Before merging: I have a question on whether we should assume facilities are attributable.

contribs/drt/src/main/java/org/matsim/contrib/drt/routing/DrtStopFacilityImpl.java
sebhoerl: Good point, the `TransitStopFacilities` are attributable, so when reading from the transit schedules file I was expecting that the `DrtStopFacility` takes over the attributes from the file. What other use cases are there? Also, the case you mention will not happen because the template signature of the function requires `Attributable`. If somebody tries to call `createFromIdentifiableFacility` with something that is now `Attributable`, it will be a syntax error. But maybe the function could be renamed to "IdentifiableAttributable" or just "createFromFacility" ?

michalmac: > What other use cases are there? 

No idea. I do not work a lot on building scenarios. That's why I am asking :-)

> Also, the case you mention will not happen because the template signature of the function requires Attributable.

Yes, but that's only after your change. Before it wasn't required.

So I would say if it is very unlikely that facilities will not be `Attributable`, let's merge. Otherwise, rather than requiring `Attributable`, take the alternative approach with `instanceof Attributable`.

michalmac: As for the method name, `createFromFacility` sounds good.

michalmac (approved):
LGTM.

PR #1386, opened 2021-02-10
quick fix for SingleInsertionDetourPathCalculator

helgehr:
I noticed that there are Problems with `SelectiveInsertionSearch`. When no insertion is found with `ExtensiveInsertionSearch` the `MultiInsertionDetourPathCalculator.calculatePaths` function returns `DetourData` with empty HashMaps. However, the `SingleInsertionDetourPathCalculator.java.calculatePaths` function checks first if `filteredInsertions` has size one - when no insertion is viable this leads to an error.

This fix just checks first if `filteredInsertions` is empty and then returns new `DetourData` with empty HashMaps manually, that is probably not the most elegant solution but it works.


michalmac:
Thanks a lot for spotting this issue! I was too quick with refactoring in #1385 and so haven't added tests. Should be fixed via #1387.

PR #1393, opened 2021-02-15
remove deprecated freightUtils.runJsprit(scenario, fcg) and only use …

kt86:
…runJsprit(scenario) instead.


kt86:
I decided to remove the old (now deprecated) part, because it was not too long out there and therefore is probably not used outside of our group. 

@rewertvsp : Maybe we have used it in zeroCUTS and you have to update it there when updating to a new MATSim weekly release. I do not see, that we are using it at any other places.

tschlenther:
Hey, i actually used it last week. Any recommendation for or a utility method instead? Or are users supposed to set everything up themselves again?

kt86:
Perfect. Yes, it is not necessary to extract the freight config group before handing it over, because it is accessible via the scenario.

Sorry, that I forgot you, @tschlenther :(

PR #1394, opened 2021-02-16
Applied fix for parameter explanation

helgehr:
I noticed that the parameter explanation of `maxTravelTimeAlpha` and `maxTravelTimeBeta` is misleading. As the maximum Travel Time is actually calculated in `org.matsim.contrib.drt.routing.DrtRouteCreator#getMaxTravelTime` with the formula:
```
drtCfg.getMaxTravelTimeAlpha() * unsharedRideTime + drtCfg.getMaxTravelTimeBeta()
```
instead of the `estimated_drt_travel_time` as written in the current parameter explanation.


helgehr:
Oh, I think actually the `org.matsim.contrib.drt.routing.DrtRouteCreator#getMaxTravelTime` should be replaced by estimatedDrtTravelTime or something similar because the Travel Time is calculated in `org.matsim.contrib.dvrp.path.VrpPaths#createPath` and this is only equal to the unshared ride time if `travelTimeEstimationAlpha` and `travelTimeEstimationBeta` are zero.

michalmac (approved):
Thanks. Looks good to me.

michalmac:
> Oh, I think actually the `org.matsim.contrib.drt.routing.DrtRouteCreator#getMaxTravelTime` should be replaced by estimatedDrtTravelTime or something similar because the Travel Time is calculated in `org.matsim.contrib.dvrp.path.VrpPaths#createPath` and this is only equal to the unshared ride time if `travelTimeEstimationAlpha` and `travelTimeEstimationBeta` in the dvrp config group are zero.

Not sure if I understood. `DrtRouteCreator.getMaxTravelTime()` computes the max travel time. Depending on the setup, the DRT algorithm will either reject or penalise insertions where the max travel time is violated.

helgehr:
I meant that the actual link travel times which are used to estimate the `unsharedTravelTime` are estimated by the `DvrpOfflineTravelTimeEstimator`/`DvrpOnlineTravelTimeEstimator`. But I guess these travel times are still for unshared rides because the considered route is calculated directly from origin to destination. So nevermind my second comment.

michalmac:
@helgehr Please update your branch so that it can be merged.

helgehr:
Ok, it should be updated now.

PR #1396, opened 2021-02-17
ChargingWithAssignmentLogicInterface

nkuehnel:
allow more flexibility for DRT charging logic


nkuehnel:
We would like to have some more flexibility for charging tasks in an E-DRT project to allow vehicles to interrupt charging even though charging is not completed.

Would be helpful if ChargingWithAssignmentLogic would be an interface

michalmac (approved):
Looks good to me. Thanks!

sebhoerl (approved):

PR #1402, opened 2021-02-24
Fix lane order for certain lanes to link assignments

rakow:
The current lane implementation always expect one incoming lane. 
If one wants to make it of zero length and let the other lanes start immediately the code does not handle the order correcty. Occasionally it would work depending on the order the lanes have been added / written in the xml file.
This PR fixes the behavior and adds a test for this case.


kainagel:
A remark if this comes up again: The original design did not allow that two turn lanes deviate at exactly the same position.  Evidently, there should have been a consistency checker checking for that.  If it is now fixed, then it is fixed.  If this comes up again ... before this takes a lot of time, I would suggest to add that consistency checker and live with the limitation for the time being.

PR #1405, opened 2021-03-01
cleanup Link interface

michalmac:
While working on special DVRP links (for fixed-demand fixed-traffic repeatable simulations -- meant for e.g. benchmarking or debugging), I did some cleanups around `Link`:

The main changes:
- add `getCapacityPeriod()` to `Link`: since `Link` offers `getFlowCapacity()` and `getFlowCapacity(time)` that both return a capacity given a predefined capacity period, it makes sense to provide the information on what the capacity period is, so the API user can translate these values to some absolute numbers (e.g. per hour or per second).
- providing default implementation for `getFlowCapacityPerSec(time)` and `getFlowCapacityPerSec()` in the `Link` interface (using the introduced  `getCapacityPeriod()` method), so they do not need to be implemented in each `Link`-extending class.

   Actually these two `getFlowCapacityPerSec` methods are meant for mobsims, so the next step could be actually removing them from `Link` and adding as util methods somewhere inside to the mobsim packages.

- some little cleanups in `LinkImpl`

### DIFF
--- a/matsim/src/main/java/org/matsim/core/network/LinkImpl.java
+++ b/matsim/src/main/java/org/matsim/core/network/LinkImpl.java
@@ -174,17 +174,7 @@ import org.matsim.utils.objectattributes.attributable.Attributes;
 		return this.capacity;
 	}
 
-
-	@Override
-	public double getFlowCapacityPerSec() {
-		return this.getCapacity() / network.getCapacityPeriod();
-	}
-	@Override
-	public double getFlowCapacityPerSec(@SuppressWarnings("unused") final double time) {
-		return this.getCapacity(time) / network.getCapacityPeriod();
-	}
-	
-	double getCapacityPeriod() {
+	public double getCapacityPeriod() {
 		// since the link has a back pointer to network, we can as well provide this here (????)
 		// TimeVariantLinkImpl needs this ... but why?
 		return network.getCapacityPeriod() ;
--- a/matsim/src/main/java/org/matsim/core/network/LinkImpl.java
+++ b/matsim/src/main/java/org/matsim/core/network/LinkImpl.java
@@ -20,10 +20,6 @@
 
 package org.matsim.core.network;
 
-import java.io.IOException;
-import java.io.ObjectInputStream;
-import java.util.Collections;
-import java.util.HashSet;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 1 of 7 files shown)


matsim/src/main/java/org/matsim/core/network/LinkImpl.java
```diff
@@ -283,16 +241,6 @@ public final void setAllowedModes(final Set<String> modes) {
...
 
-//	/*deliberately package*/ final void setOrigId2(final String id) {
-//		this.origid = id;
-//	}
-	
-	private void readObject(ObjectInputStream ois) throws ClassNotFoundException, IOException {
-		ois.defaultReadObject();
-		this.from.addOutLink(this);
-		this.to.addInLink(this);
-	}
```
mrieser, line 294: Isn't this method used by Serialization, e.g. in OTFVis?

mrieser:
Thanks for the cleanup of the code!

I guess at some point we need a discussion on how close our classes should resemble the data structure in XML, or if the classes are actual domain objects. Until now, my interpretation was that the classes followed closely the XML structure (there were some exceptions from this rule, e.g. the `Person.setSelectedPlan()` despite `selected` being an attribute of a `Plan` and not a `Person`), but the additional `Link.getCapacityPeriod()` clearly has no counterpart on the link-level in XML, but on the network-level.

I can see that it helps making the current implementation cleaner, and considering that are already those `perSec()`-methods that don't directly reflect anything in the XML as well, the change is okay for me.

I have only the one thing I noted above about the `readObject` method where I'm unsure if its removal breaks OTFVis or not, but otherwise I'm fine with the change.

matsim/src/main/java/org/matsim/core/network/LinkImpl.java
mrieser: thanks for checking!

mrieser (approved):

michalmac:
@mrieser Thanks for your quick reply!

This PR was mostly focused on cleaning up and reducing code duplication. I think that having both `getFlowCapacity()` and `getFlowCapacityPerSec()` is probably too much. So something to discuss in the future.

BTW. I just found a related commit: https://github.com/matsim-org/matsim-libs/commit/51dea280f7a85607d8c9f7efb3824f671ce04372 that describes the reason behind adding `getFlowCapacityPerSec()`,  a good input for further discussions.

PR #1409, opened 2021-03-06
improve performance of SRR intermodal routing

balacmi:
We have a situation where every time intermodal stops are searched where
specific stop attribute is used as a filter a new quad tree is
generated with those stops. This is making intermodal router to
considerably underperform.

This commit ensures that the quad tree is created only if it does not
exist in the method prepareStopFilterQuadTreeIfNotExistent

In our set-up (KNR, PNR, walk) this leads to speedup of factor ~15 for
intermodal routing


matsim/src/main/java/ch/sbb/matsim/routing/pt/raptor/SwissRailRaptorData.java
```diff
@@ -42,547 +42,603 @@
...
+			List<TransitRoute> transitRoutes = new ArrayList<>(line.getRoutes().values());
+			transitRoutes.sort(Comparator.comparingDouble(tr -> getEarliestDeparture(tr).getDepartureTime())); // sort
+																												// routes
+																												// by
+																												// earliest
+																												// departure
+																												// for
+																												// additional
+																												// performance
+																												// gains
```
mrieser, line 139: what is this for?

matsim/src/main/java/ch/sbb/matsim/routing/pt/raptor/SwissRailRaptorData.java
```diff
@@ -42,547 +42,603 @@
...
+		int indexRoutes = 0;
+		int indexRouteStops = 0;
+		int indexDeparture = 0;
+
+		// enumerate TransitStopFacilities along their usage in transit routes to
+		// (hopefully) achieve a better memory locality
+		// well, I'm not even sure how often we'll need the transit stop facilities,
+		// likely we'll use RouteStops more often
+		Map<TransitStopFacility, Integer> stopFacilityIndices = new HashMap<>(
+				(int) (schedule.getFacilities().size() * 1.5));
```
mrieser, line 125: for the future: If you have a need to reformat, please commit this separately. This way, it's very hard to see what has actually changed in the file.

matsim/src/main/java/ch/sbb/matsim/routing/pt/raptor/SwissRailRaptorData.java
```diff
@@ -42,547 +42,603 @@
...
+		this.occupancyData = occupancyData;
+	}
+
+	public static SwissRailRaptorData create(TransitSchedule schedule, @Nullable Vehicles transitVehicles,
+			RaptorStaticConfig staticConfig, Network network, OccupancyData occupancyData) {
+		log.info("Preparing data for SwissRailRaptor...");
+		long startMillis = System.currentTimeMillis();
+
+		Map<Id<Vehicle>, Vehicle> vehicles = transitVehicles == null ? Collections.emptyMap()
+				: transitVehicles.getVehicles();
```
mrieser, line 90: please set a wider line limit, this is actually much more difficult to read than when it's all on one line

matsim/src/main/java/ch/sbb/matsim/routing/pt/raptor/SwissRailRaptorData.java
```diff
@@ -42,547 +42,603 @@
...
+		}
+		// if the stop facilities are different, and the destination stop is part
+		// of the current route, it does not make sense to transfer here
+		if (toStopIsPartOfRouteButNotSame(fromRouteStop, toRouteStop)) {
+			return false;
+		}
+		// assuming vehicles serving the exact same stop sequence do not overtake each
+		// other,
+		// it does not make sense to transfer to another route that serves the exact
+		// same upcoming stops
```
mrieser, line 352: seriously, do I have to look over all thees changes? It's not even nicer formatted with single words on a line. How should that "auto-format" improve legibility?

mrieser (changes requested):
Sorry, there are too many irrelevant code changes to do a proper code-review. I don't see where the actual change you want feedback on is located.

matsim/src/main/java/ch/sbb/matsim/routing/pt/raptor/SwissRailRaptorData.java
balacmi: Seriously, sorry. will make another commit with only the change there. The formatting came on its own...

mrieser: thanks, it looks now much more concise!

matsim/src/main/java/ch/sbb/matsim/routing/pt/raptor/SwissRailRaptorData.java
```diff
@@ -546,9 +546,17 @@ public CachingTransferProvider getTransferProvider(RTransfer transfer, CachingTr
...
+		// we do not have to do anything
+		if (stopFilterAttribute2Value2StopsQT.containsKey(stopFilterAttribute)
+			&& stopFilterAttribute2Value2StopsQT.get(stopFilterAttribute).containsKey(stopFilterValue))
+			return;
+		// otherwise we have to add this map
+		// only in case there is no attribute key
 		if (!stopFilterAttribute2Value2StopsQT.containsKey(stopFilterAttribute)) {
 			stopFilterAttribute2Value2StopsQT.put(stopFilterAttribute, new HashMap<>());
 		}
+
```
mrieser, line 559: nice catch! This indeed looks like it was an oversight before, that the QuadTree was generated over and over.

The PR is okay for me, but as we're talking about performance, some additional suggestions to make it even faster:

At the moment, there are at least 2 accesses to the map, in some cases even 4: 1) `...containsKey(..)` on line 551, 2) `...get(..)` (line 552), 3) `!...containsKey(..)` (line 556) and 4) `...get(..)` at the bottom of the method. This could be replaced by a single `computeIfAbsent()` like this:

```java
// if stopFilterAttribute/stopFilterValue combination exists we do not have to do anything
Map<String, QuadTree<TransitStopFacility>> filteredQTs = 
        this.stopFilterAttribute2Value2StopsQT.computeIfAbsent(stopFilterAttribute, key -> new HashMap<>());
if (filteredQTs.containsKey(stopFilterValue))
    return;

....

filteredQTs.put(stopFilterValue, stopsQTFiltered);
```

This way, there are less map-access operations, and less if-statements which should lead to an additional (albeit small) speed up.

mrieser (approved):

jfbischoff (approved):
Thanks for finding :)

matsim/src/main/java/ch/sbb/matsim/routing/pt/raptor/SwissRailRaptorData.java
balacmi: Great, thanks, will include it.

PR #1422, opened 2021-03-18
Early warning mechanisms for drastic changes in drt performance w/ respect to customer experience

jakobrehmann:
RunDrtExampleIT is modified to read drt_customer_stats output file and confirm that various values are not more than +-20% than the values from commit ee6b608 (March 2021). The following customer parameters are checked: rejectionRate, rejections, waitAverage, inVehicleTravelTimeMean, & totalTravelTimeMean


vsp-gleich:
This test is thought to make us aware of issues like the sudden wait time increase we had from 13.0-2020w51-SNAPSHOT to later versions in matsim-berlin runs (with everything identical besides the matsim-libs version). Unfortunately, after implementation of this test we saw that in the small mielec scenario run in this IT test, we did not see that increase in wait time, so it would not have prevented last December's performance drop. However, we cannot run the big berlin scenario as an integration test and the small existing mielec tests might still help in other occasions.

vsp-gleich (approved):

michalmac (approved):
I really like the idea! ⭐ 

I added 2 commits to this PR:
- little refactorings (mainly: switch to assertj, which is used in DVRP/DRT/... tests; adding `Stats` class)
- reduced the percentage from 20% to 1% to really have it as an early warning system. Maybe 1% is too strict. Maybe different stats would need different "acceptance ranges". Let's reiterate on it later if we experience too many failures.

jakobrehmann:
@michalmac Thanks for the helpful feedback and refactoring. 

I traveled back in time through the git log to see how the 1% allowable difference holds up: somewhere around PR #1364 (Feb 1, 2021), testRunDrtExampleWithRequestRetry begins to fail. Sometime around PR #1350 (Jan 19, 2021), all three tests start to fail. This is just meant as some additional Info; I agree that we should stick with 1% and see when and why the early warning tests fail in the future.

michalmac:
> I traveled back in time through the git log to see how the 1% allowable difference holds up: somewhere around PR #1364 (Feb 1, 2021), testRunDrtExampleWithRequestRetry begins to fail. Sometime around PR #1350 (Jan 19, 2021), all three tests start to fail.

Interesting. The PRs are not DRT-related. But I guess there were some drt changes in between. However, if we observe tests failing due to changes in other parts of matsim, maybe we need to run them as so-called DVRP benchmarks (i.e. traffic and demand are fixed) to narrow the scope.

PR #1428, opened 2021-03-24
New very fast network router: SpeedyALT

mrieser:
This PR provides a new very fast car/network router: SpeedyALT.

It is an implementation of the ALT algorithm (also known as A*-Search with Landmarks and Triangle Inequality), but highly optimized. Still, the code is rather compact and hopefully easy to understand and maintain. (it should definitely be easier to understand than the current FastAStarLandmarks, which is distributed over 4 hierarchies of inheritance, with additional delegation etc).

In my testing, routing several thousand realistic trips across Switzerland on a detailed navigation network, the speed-up over `FastAStarLandmarks` was around 3 (in detail: 2.97 single-threaded and 3.44 multi-threaded). Speed-up over `AStarLandmarks` (the current default router in MATSim) is around 6, and against `Dijkstra` the speed-up is over 20. Clearly, those values may depend on the actual scenario used, but in any case they should be substantial.  Initialization time was halfed compared to `(Fast)AStarLandmarks`, saving additional time in each replanning phase.

### DIFF
--- a/contribs/drt/src/main/java/org/matsim/contrib/drt/optimizer/insertion/MultiInsertionDetourPathCalculator.java
+++ b/contribs/drt/src/main/java/org/matsim/contrib/drt/optimizer/insertion/MultiInsertionDetourPathCalculator.java
@@ -43,12 +43,11 @@ import org.matsim.contrib.dvrp.path.OneToManyPathSearch.PathData;
 import org.matsim.contrib.dvrp.trafficmonitoring.DvrpTravelTimeModule;
 import org.matsim.core.mobsim.framework.events.MobsimBeforeCleanupEvent;
 import org.matsim.core.mobsim.framework.listeners.MobsimBeforeCleanupListener;
+import org.matsim.core.router.speedy.SpeedyGraph;
 import org.matsim.core.router.util.TravelDisutility;
 import org.matsim.core.router.util.TravelTime;
 
 import com.google.common.annotations.VisibleForTesting;
-
-import ch.sbb.matsim.routing.graph.Graph;
 
 /**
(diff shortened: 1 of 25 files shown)


rakow:
Maybe this should become the default implementation. I don't see a reason to keep and maintain all the other different LeastPathCalculators that offer no advantage in any use-case.

rakow (approved):

sebhoerl:
Sounds great! I'm just testing with a large-scale instance of Rhônes-Alpes to see what it brings in terms of performance :) 

However, I also just tested with a small instance of Île-de-France locally, unfortunately I am getting an error during (discrete_mode_choice) routing:

```
oops 141714   -1   
```

In `DAryMinHeap` line 63. Any idea what it might be? It is also possible that something in the input data / plans is wrong that before did not lead to an error. It runs fine with `FastAStarLandmarks`.

mrieser:
`oops`, haha, I'll have a look at it :-)

mrieser:
I'll need to improve that error message. It is generated when an element in the heap gets its cost decreased, but the element is no longer in the heap. How small is the scenario, and is it available? I could give it a try myself and see what causes it.

sebhoerl:
I'll try to isolate the error... when running routing the specific relation (Node, Node, Departure time) directly after loading the network the error does not happen. So I assume maybe it has to do something with travel times as this happens in the routing after the first iteration. I'll keep you posted!

sebhoerl:
I'll send you an email.

mrieser:
bugfix: see https://github.com/matsim-org/matsim-libs/pull/1430

PR #1429, opened 2021-03-24
Make CarrierAgentTracker an interface

tschlenther:
`CarrierAgentTracker` was `@Provided` by the `CarrierModule` before and now is made an interface to use injection.
As `CarrierAgentTracker` is also needed to trigger scoring, it does not have mobsim-scope. This is the reason why we can not put it behind it's own proper provider, because the `CarrierControlerListener` recreates the CarrierAgentTracker object outside of the mobsim.

Is there a specific reason why the AgentTracker or the  `CarrierScoringFunction` need to be reconstructed every iteration? If not, we could further simplify things by having one stable CarrierAgentTracker object which is a ControlerListener itself (so no need for CarrierControlerListener to trigger scoring). Even if the `CarrierAgent` should continue to get reconstructed every iteration, the `CarrierAgentTracker` could do that in that setup. 

I did these changes when i came across the JointDemand package, which ideally should use the `FreightAgentSource` instead of defining it's infrastructure that converts `CarrierPlans` into `MobsimAgents`.
In the end, i concluded that only a few lines in the `createPlans()` and `handleEvent(ActivityStartEvent)` methods of the new interface `CarrierAgentTracker` would have to be added. So this PR is a preparation to clean that up up and synchronize the packages a bit.
This would also lead to consistent handling of freight plans, in other words it would also adress issues mentioned in #1427, which deals with TimeWindowHandling. The freight contrib has it's logic to handle TimeWindows.

Additionally, i set the leg mode to `CarrierUtils.getCarrierMode(carrier)` in the `createPlans()` method, which returns `TransportMode.car` by default (Tests run though). That means, it should now be easy to let freight agents use other network modes while using the freight contrib.

### DIFF
--- a/contribs/freight/src/main/java/org/matsim/contrib/freight/controler/CarrierAgent.java
+++ b/contribs/freight/src/main/java/org/matsim/contrib/freight/controler/CarrierAgent.java
@@ -17,10 +17,7 @@ import org.matsim.api.core.v01.population.Leg;
 import org.matsim.api.core.v01.population.Person;
 import org.matsim.api.core.v01.population.Plan;
 import org.matsim.api.core.v01.population.Route;
-import org.matsim.contrib.freight.carrier.Carrier;
-import org.matsim.contrib.freight.carrier.CarrierVehicle;
-import org.matsim.contrib.freight.carrier.FreightConstants;
-import org.matsim.contrib.freight.carrier.ScheduledTour;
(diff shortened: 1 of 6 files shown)


tschlenther:
i'll close this PR since it was conducted as an intermediate step in the process of closing the gap between the jointDemand package and the freight contrib. This seems not be feasible in the near project, due to time issues :/

PR #1436, opened 2021-04-06
Emissions: new consitency check for tables with different vehicle categories

### LINKED ISSUE #1226: Inappropriate validation check in emissions code
At the start of a simulation, the `emissions` module builds the necessary cold and warm emission tables. In the [`WarmEmissionsAnalysisModule`](https://github.com/matsim-org/matsim-libs/blob/e22e41a01dcf6109138efd08ca0ea7800ed5d69c/contribs/emissions/src/main/java/org/matsim/contrib/emissions/WarmEmissionAnalysisModule.java#L99) there is a _consistency check_ that was changed from `while`-loops to `for`-loops. @kainagel commented (line 99-101) that the purpose of the code is to act as a consistency check.

It essentially checks that there is an entry in the emissions table for all possible combinations of road category, `HbefaTrafficSituation`, `HbefaVehicelCategory`, `HbefaVehicleAttribute` and `Pollutant`. In principle this is fine, but practically there is now an inconsistency, which results in the `RuntimeException` being thrown at [line 127](https://github.com/matsim-org/matsim-libs/blob/e22e41a01dcf6109138efd08ca0ea7800ed5d69c/contribs/emissions/src/main/java/org/matsim/contrib/emissions/WarmEmissionAnalysisModule.java#L127).

The problem is that the `HbefaVehicleCatergory` and the `HbefaVehicleAttribute`  are _aligned_. That is, you may have `PASSENGER_CAR` as a vehicle category, and then its associated `PC P Euro-3` in the vehicle attributes for the emissions concept. The `PC` at the start of the attribute is aligned with the vehicle category. Similarly, you may have a `HEAVY_GOODS_VEHICLE` vehicle category with its `HGV D Euro-III` emissions concept in the attributes.

But because the code now tests for _all possible combinations_, there will not (and should not) be an entry for a `PASSENGER_CAR` with an attribute `HGV D Euro-III`... so the code aborts.

The solution seems that one must find a way to convert the first portion of the attributes, like `PC` or `HGV`, to a vehicle category and then first check if it is the same/aligned with the `HbefaVehicleCategory`. But I am not convinced I have a clean way of doing that right now, especially because there are (currently) seven different [`HbefaVehicleCategory`](https://github.com/matsim-org/matsim-libs/blob/master/contribs/emissions/src/main/java/org/matsim/contrib/emissions/HbefaVehicleCategory.java)s to consider. We currently only use passenger cars and heavy goods vehicles. The code will work fine if you only work with a **_single vehicle category_**.

kt86:
As mentioned in #1226 there is a problem with the existing consistency checks, if the lookup table includes data for different vehicle types.

As solution there will be added a new level for the consistency check, that will check the tables consistency for each `VehicleCategory` independently: `HbefaTableConsistencyCheckingLevel.perVehCat` Thanks to @Jonas116 for doing the work during the MATSim advanced class.

Before we can merge it, I have to look into it and clean up a bit.

closes #1226 

Open points remaining: 
- We should also have a similar test for the cold emissions calculation
- Cleaning up and reduce duplicated lines of code inside the consistency check


jfbischoff:
@kt86 are you still working on this PR?

kt86:
More or less yes. It is still on my desk and I should bring it to a qualified end :) But still do not know when :(

Issue #1378, opened 2021-02-09
BinTray is closing!


mrieser:
We currently use Bintray to host our jar files. But Bintray is shutting down on May 1st 2021: https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter/ . Additional info can be found in this article: https://www.infoq.com/news/2021/02/jfrog-jcenter-bintray-closure/ .

So we will have to decide where to host our jar-files (including the weekly snapshots) in the future and migrate everything.

(Update: oss.jfrog.org might not be affected, i.e. snapshots including the weekly releases might not be affected)

## Tasks
- [ ] decide on alternative Maven repository for hosting artifacts
- [ ] adapt deployment in pom.xml for:
    - [ ] matsim-libs
    - [ ] weekly snapshots
    - [ ] changes on Jenkins?
- [ ] migrate repository content to new repository
- [ ] change configured repository for dependencies in pom.xml for:
    - [ ] matsim-libs
    - [ ] matsim-example-project
    - [ ] others?
- [ ] Make sure that MATSim does not use any third-party dependencies hosted on Bintray, e.g. start a clean maven build with an empty local repository
    - [ ] Update affected third-party dependency repositories where possible
    - [ ] Keep a copy of required dependencies and self-host them if no other solution is possible.
- [ ] announce change in a MATSim news blog post.

## Alternatives
Potential alternatives for Maven repositories (feel free to add, I just wrote down what came to my mind)
- GitHub packages
- self-hosted
- simunto-hosted
- JFrog Platform
- Maven Central (only releases, no snapshots)
- OSSRH (Sonatype OSS Repository hosting, includes snapshots, allows publishing to Central)
- other?

We should try to figure out advantages and disadvantages for each alternative, especially regarding the hosting of the weekly-releases which quickly fill up any potential quota.

## Alternatives evaluation

|   | GitHub | Self-Hosted | Simunto-Hosted | JFrog | Maven Central | OSSRH |
|---|---|---|---|---|---|---|
| URL | https://github.com/features/packages | | | https://jfrog.com/pricing/#sass | | https://central.sonatype.org/pages/ossrh-guide.html |
| Storage-Quota,<br> Pricing | "free for public repositories", but no quota provided | . | . | 2 GB Storage / free <br> more storage / very expensive | | |
| Advantages | likely high-availability | . | . | likely high-availability | | |
| Disadvantages | . | own responsibility for availability | | | | |

[...]

PR #1440, opened 2021-04-09
deploy to GitHub Packages; add snapshot and release deployment workflows

michalmac:
First steps addressing issues in #1378:
- migrate to GitHub packages
- add workflows that automatically deploy snapshots and (GitHub) releases

I tested this setup on a toy repo that resembles the matsim repo structure: https://github.com/michalmac/try-out-github-packages.

I was able to deploy to GitHub packages and then use the deployed jars in another local project. One caveat though: a public read token needs to be used (this is a current limitation of GitHub Packages). This is the currently suggested approach by GitHub according to https://github.community/t/partial-support-for-maven-snapshots/115628/12:
```xml
<repository>
  <id>github-public</id>
  <url>https://public:xxxxxxxxxxxxxxxxx@maven.pkg.github.com/michalmac/try-out-github-packages/</url>
</repository>
```

GitHub offers some control over published packages (incl. deleting them to some extend).

It is still unclear about the available storage limit. In my billing, I can see some usage reported (in contrast to Actions which do not give any minutes reported in the billing).

### DIFF
--- a/.github/workflows/deploy-on-release-created.yaml
+++ b/.github/workflows/deploy-on-release-created.yaml
@@ -1,0 +1,42 @@
+name: deploy-on-release-created
+
+on:
+  release:
+    types:
+      - created
+
+jobs:
+  deploy-release:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout git repo
+        uses: actions/checkout@v2
+
+      - name: Cache local Maven repository
+        uses: actions/cache@v2
+        with:
+          path: ~/.m2/repository
+          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+          restore-keys: |
+            ${{ runner.os }}-maven-
+
+      - name: Setup Java
+        uses: actions/setup-java@v1
+        with:
+          java-version: 11
+
+      - name: Set MATSim version
+        run: mvn versions:set --batch-mode -DnewVersion=${{ github.event.release.tag_name }} -DgenerateBackupPoms=false
+
+      # Build and publish are separated so we start deploying only after all jars are built successfully
+      - name: Build jars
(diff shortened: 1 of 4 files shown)


.github/workflows/deploy-on-release-created.yaml
```diff
@@ -0,0 +1,42 @@
...
+          restore-keys: |
+            ${{ runner.os }}-maven-
+
+      - name: Setup Java
+        uses: actions/setup-java@v1
+        with:
+          java-version: 11
+
+      - name: Set MATSim version
+        run: mvn versions:set --batch-mode -DnewVersion=${{ github.event.release.tag_name }} -DgenerateBackupPoms=false
```
michalmac, line 30: `-DnewVersion=${{ github.event.release.tag_name }}` assumes that the release (and tag) are just numbers (e.g. `13.0`). So far we have used `matsim-x.x` as a tag, but would suggest switching to the (more standard) `x.x` format.

mrieser (approved):
Having no practical experience with GitHub Actions or GitHub Packages, my review might be a bit limited in meaning. Looking at the commit, it all makes sense, and the changes in the pom.xml are also fine.

What troubles me a bit is the fact the one needs a token to access the maven repository at GitHub. This is not really practical for wide-spread distribution, and I was not aware of that before.  I guess for the moment it is okay, but we might want to continue discussing this issue in #1378 and probably migrate to a better solution later.

michalmac:
Yes, this PAT is a bit of a downside. I was also not aware of it beforehand.

I will create a machine user (in `matsim-org`) and create a read token for packages for that user. Then instead of adding this line into POMs of matsim-dependent projects:
```xml
<url>https://maven.pkg.github.com/michalmac/try-out-github-packages/</url>
```
we would simply add something like this:
```xml
<url>https://public:xxxxxxxxxxxxxxxxx@maven.pkg.github.com/michalmac/try-out-github-packages/</url>
```
Slightly longer, less readable, but should fix the issue in a reasonable way.

PR #1441, opened 2021-04-10
fix: change artifact ids to lowercase

michalmac:
According to https://maven.apache.org/guides/mini/guide-naming-conventions.html we should use lowercase latter in artifact ids. We are violating this rule in: `cadytsIntegration`, `commercialTrafficApplications` and `eventsBasedPTRouter`.

Not obeying this rule causes problems with uploading artifacts to GitHub Packages. Also some mvn plugins may not work.

Since we are just about to bump to `14.0-SNAPSHOT`, I think this is the best moment to switch to lowercase (as we will avoid issues like `cadytsIntegration-13.0-SNAPSHOT` being different than `cadyts-integration-13.0-SNAPSHOT`, which would likely lead to some hard to find bugs).

### DIFF
--- a/contribs/cadytsIntegration/pom.xml
+++ b/contribs/cadytsIntegration/pom.xml
@@ -6,7 +6,7 @@
   </parent>
   <modelVersion>4.0.0</modelVersion>
   <groupId>org.matsim.contrib</groupId>
-  <artifactId>cadytsIntegration</artifactId>
+  <artifactId>cadyts-integration</artifactId>
   <name>cadytsIntegration</name>
     <build>
         <plugins>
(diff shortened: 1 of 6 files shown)


michalmac:
This is the problem we are currently experiencing with uppercase letters: https://github.com/matsim-org/matsim-libs/runs/2309175633?check_suite_focus=true

mrieser (approved):
yes, makes sense to follow the standard a bit more closely.

michalmac:
I just enabled the auto-merge option for the matsim repo: https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/automatically-merging-a-pull-request. Should work similar to the two auto-merge bots we have. Let's see how it works.

PR #1452, opened 2021-04-12
deploy PR-labelled versions

michalmac:
See discussion here: https://github.com/matsim-org/matsim-libs/issues/1378#issuecomment-804845481

The PR-labelled version (e.g. `14.0-PR1534`) is created and deployed on merging PR to master (and only master). This is how it looks on the test repo: https://github.com/michalmac/try-out-github-packages/packages/718303

The PR-labelled releases are meant to replace the weekly releases (and potentially even the standard snapshots). We could also think of publishing snapshots to GitHub Packages (for the developers who know all the risks behind) and PR-labelled versions to the matsim maven repo (for the general public)...

Happy to hear your thoughts :-)

### DIFF
--- a/.github/workflows/deploy-on-pr-merge.yaml
+++ b/.github/workflows/deploy-on-pr-merge.yaml
@@ -3,10 +3,16 @@ name: deploy-on-pr-merge
 on:
   push:
     branches:
-      - master
+      - master # snapshot deployment
+  pull_request:
+    types:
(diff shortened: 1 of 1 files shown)


michalmac:
I think I may add auto-tagging of merge commits to this workflow. This way anyone can easily check out a given PR-labelled version and build it (could be useful for old PR-labelled versions if we are not able to keep all of them in the mvn repo due to limited storage).

PR #1460, opened 2021-04-18
Workflow improvements

michalmac:
A series of little improvements in the `deploy-***` workflows:
- bump the mvn assemby plugin so `assembly.skipAssembly` flag is respected
- prevent unnecessary re-compilation when deploying already created jars (unnecessary copying resources was the trouble maker)
- stop installing jars to local mvn repo (and later caching them) in GH Actions when deploying to the remote maven repo

### DIFF
--- a/.github/workflows/deploy-on-pr-merge.yaml
+++ b/.github/workflows/deploy-on-pr-merge.yaml
@@ -46,7 +46,7 @@ jobs:
 
       - name: Publish jars to matsim maven repo
         # fail at end to deploy as many jars as possible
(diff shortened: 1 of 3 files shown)


michalmac:
After merging this PR, snapshots are build in only ~6 min and PR-labelled versions in ~10 min.

The latter takes longer, because downloading the mvn cache fails when the `deploy-on-pr-merge` workflow is triggered in response to the "PR closed" event. Seems it's not only affecting us: https://github.com/actions/cache/issues/478

PR #1465, opened 2021-04-21
New contrib: MATSim Application

rakow:
This PR introduces a new contrib for streamlining the creation of MATSim scenario applications.
Usually a scenario does not only consist of config and input files, but also specific preparation and analysis code.

This contrib allows to build a single executable class that can used for running, preparing and analyzing a scenario with the use of "subcommands" and the new `MATSimApplication` class:

```java
@CommandLine.Command(header = ":: Open Leipzig Scenario ::", version = RunLeipzigScenario.VERSION)
@MATSimApplication.Prepare({
        CreateNetworkFromSumo.class, CreateTransitScheduleFromGtfs.class, TrajectoryToPlans.class, 
        MergePopulations.class, ExtractRelevantFreightTrips.class, DownSamplePopulation.class,
        CreateLandUseShp.class, ResolveGridCoordinates.class, GenerateShortDistanceTrips.class,
})
@MATSimApplication.Analysis({
        AnalysisSummary.class, TravelTimeAnalysis.class
})
public final class RunLeipzigScenario extends MATSimApplication {

    static final String VERSION = "1.0";

    public RunLeipzigScenario() {
        super(String.format("scenarios/input/leipzig-v%s-25pct.config.xml", VERSION));
    }
    public static void main(String[] args) {
        MATSimApplication.run(RunLeipzigScenario.class, args);
    }

}
````

All functionality is exposed via a user-friendly command line interface. This shows the output of the example when running without any arguments:

grafik

With a command line interface all necessary steps to build a scenario can be automated more easily. (e.g. with any scripting language or build tools such as `Make`)

This contrib already includes some often needed preparation and analysis steps used in vsp scenarios. (e.g. network creation, demand preprocessing, gtfs integration, etc.)


mrieser:
Wow, just based on the description and code example, this looks great!

Any reason why you chose to adopt this annotation-based approach for specifying all the build steps, as opposed to a more "traditional" style, e.g. using arrays or Lists to pass the classes?

In my opinion, annotations are okay when writing the code, but it makes it more difficult to reuse functionality programmatically. E.g. a GUI where the user can select which steps should be performed (e.g. checkboxes for "convert network", "convert GTFS") would be difficult to implement using this annotation-approach, as the GUI would probably have to create a java-file to "configure" the run. (Maybe there is an alternative option to use it, I haven't yet checked the code in full detail).

rakow:
In my opinion annotations are useful to add additional information to classes, fields, etc. without the need to define it at a separate place, either programmatically or perhaps at with config files. I think this makes it more concise overall and better integrated.
Some years ago [spring](https://spring.io/) was configured with xml files, but this shifted completely towards annotation (successfully).

In this contrib the annotations at the class only indicate the presence of a step. The steps itself have additional options and need to be run with the intended parameters. This can be either from command-line or from code. Hard-coding this also with annotations would make it indeed not very re-usable. So there still needs to be a layer that executes the steps, which I didn't want to force to be in java.

It would be no problem to access the metadata of commands and options at runtime and show an input mask to the user.
The library can parse all annotations and generate a `Model.CommandSpec` that would contain all needed information.
The `execute` method on `MATSimApplication` can be used to execute any step from code.  So I think this use-case can be implemented without problems and the need to generate anything.

PR #1466, opened 2021-04-22
Position emission module while using it

Janekdererste:


Janekdererste:
@vsp-gleich I will merge this so I can use it in the depending project. Any reason why I shouldn't?

vsp-gleich:
> @vsp-gleich I will merge this so I can use it in the depending project. Any reason why I shouldn't?

This pull request redistributes the positionEvents along the links such that the second lost between LinkLeaveEvent and LinkEnterEvent for crossing the intersection is instead already used to progress on the following link. This is what @kainagel  and I came up with to remove that one second without emissions during link change. That might be the only objection. Is it ok to merge @kainagel ?

PR #1467, opened 2021-04-22
Add coord to activity end event

vsp-gleich:
We stumbled across trips in the trips.csv where the home coordinate for the start of the first trip and the home coordinate for the end of the return trip home are different. The reason is apparently that trips.csv takes all its information from the experienced plans, the experienced plans are entirely based on events and only ActivityStartEvents have a Coord, ActivityEndEvents do not. Consequently, for the first activity there is only an ActivityEndEvent without Coord, so the Coord is null in the experienced plans and TripsAndLegsCSVWriter falls back onto the linkId and substitutes the missing Coord with the position of the link which is typically not at the same place as the activity coord. All later actvities (and trips) are not affected, because there is an ActivityStartEvents with a Coord.

As far as I remember, when ActivityStartEvents got a coord attribute we were concerned with events files increasing in size due to the additional coord field. So I created a branch addCoordToActivityEndEvent to evaluate the increase in events file size. For the open berlin model 1% sample the events file increased from 188MB to 192MB (RunBerlinScenario). I find this reasonable for the gain in accuracy (and simpler to read experienced plans). Are there any objections to adding the Coord to ActivityEndEvents?


jfbischoff:
Go for it!

jfbischoff (approved):

kainagel:
There are (I think in PopulationUtils) centralized methods that compute coordinates.  If they are used, the results should always be the same.

vsp-gleich:
> There are (I think in PopulationUtils) centralized methods that compute coordinates. If they are used, the results should always be the same.

@kainagel  I think the issue here is different. PopulationUtils.decideOnCoordForActivity is probably what you meant. That method suffers from the same problem that plans.xml typically has activities with coords. But the ActivityEndEvent in events.xml does not have any coord. Experienced plans are written by EventsToActivities whcih is no idea of the plans.xml, for the first activity of the day it has only the information passed via the ActivityEndEvent which currently does not include the coord. PopulationUtils.decideOnCoordForActivity does the same kind of educated guess what position to assume if there is no facility id and no coord and falls back to the link position just as TripsAndLegsCSVWriter. So it will also give the position of the link, but not the coord found in the plans.xml.

PR #1468, opened 2021-04-23
Freight improvements

mrieser:
After looking a bit into the freight contrib, I found some issues I decided to tackle.

Some are just minor performance improvements, others just code cleanup (renaming constants, fixing typos).
The probably biggest change is to make FreightConfigGroup actually usable and accepting parameters from a config.xml-file.

If you review this PR, it might be easier to look at the individual commits to get some context of the performed changes.


kt86:
Thank you very much @mrieser for doing the improvements! 
I try to have a look  at it in the next week. If I am absent then, maybe @rewertvsp or @tschlenther can jump in ...

contribs/freight/src/main/java/org/matsim/contrib/freight/carrier/CarrierPlanWriter.java
```diff
@@ -23,7 +23,9 @@
  * 
  * @author sschroeder
  *
+ * @deprecated Use {@link CarrierPlanXmlWriterV2} instead which writes the newer format
```
kt86, line 26: just one idea triggered by this change: 
What do you think if we use the `CarrierPlanWriter` as default and call from here the `CarrierPlanXmlWriterV2`. That would have the advantage, that one could introduce a WriterV3 (internally) at one point in time, and everyone using the more general `CarrierPlanWriter`  will write out the data with the most current version. (Similar to the approach in https://github.com/matsim-org/matsim-libs/blob/master/matsim/src/main/java/org/matsim/vehicles/MatsimVehicleWriter.java).

The stuff here could be moved out to a "V1" version _or_ we could maybe use something like "MatsimCarrierPlanWriter" as "general" class and delegate from there.

I think due to the public usage of the V2 writer we cannot reduce its visibility yet, but we could deprecate it and send the user to the more general one.

kt86 (approved):
Hello @mrieser: Thankyou a lot for all the work. It looks totally fine for me! Thank you also for correcting all the typos, language errors,... besides the improvements!

I have just added one comment regarding the carrierPlanWriter. That idea was only triggered by your commit. It is no reason to not merge it now.

Unfortunately I am now out for a month, so I have no time to do the change myself. It is also totally fine to open an issue from that comment/idea and assign it to me. Then I can do it when I am back in office.

contribs/freight/src/main/java/org/matsim/contrib/freight/carrier/CarrierPlanWriter.java
mrieser: I renamed the old writer and added a new one, similar to the VehiclesWriter and other writers in MATSim.

JWJoubert (commented):
A bit of a late response, but we're happy with the improvements. Thanks!

PR #1472, opened 2021-04-26
reduce memory consumption in ScoringParameters

tschlenther:
We had faced problems of high memory consumption and run times when using the rather new `IncomeDependentUtilityOfMoneyPersonScoringParameters` implementation. We figured out that the `ScoringParameters.Builder` makes defensive copies of all activity and mode parameters. We stated that activity parameters (at least the way we use them - in matsim-berlin for example) are really not something that needs to be configurable on an agent basis but rather on a subpopulation basis. So there should be a way of circumventing (heaps of) defensive copies of activity parameters when using other scoring parameters on an individual basis.
This is why i implemented (yet) another constructor for `ScoringParameters`.
While doing so, i realized that this class is acutally only taking builders of the `ActivityUtilityParameters` and `ModeUtilityParameters` and that the `build` method of those is only called in the `ScoringParameters.build()` method (lazy initialization). This however, prevented to achieve what i tried to achieve: not building activity parameters everytime an (individual) `ScoringParameters` set is built. So i changed the getters and setters, too. This might possibly lead to more memory usage upfront and also break some existing code bits but, at least to me, makes the instantiation process of `ScoringParameters` more clear and straight forward. Please let me know, what you think @kainagel and whether we can merge this.

I tested the results with matsim-berlin in a 1 percent sample, concerning run time.
Left is the old version, right is the new one

grafik


tschlenther:
a note: the runs were not run on the same machine, but with the same heap size and the same number of cores....
the run on the right hand side was run on my local machine, maybe there were some interferences with other processes.... still it lookes better to me

mrieser:
We had very high memory usage in the scoring function at SBB about two years ago, mostly due to a very high number of activity types. We partially fixed it by only copying those activity types that are actually used by the plan (plus all the *-interaction activity types as they can appear due to modeChoice-replanning), and by potentially re-using objects between agents. See [this commit in the sbb code](https://github.com/SchweizerischeBundesbahnen/matsim-sbb/commit/e523319931673c2162e633ee152de2ce3da83917) for details.

tschlenther:
seems like you came up with a similar solution for the same problem @mrieser. With this approach, it is actually up to the 'user' whether to use the same objects for activity parameters throughout all agents or to provide only a corresponding sample of those. Thanks for sharing this.

kainagel (approved):
I have talked to Tilmann and it seems ok.

PR #1477, opened 2021-04-29
Again trying to fix kinematic waves qsim

ikaddoura:
This is a pull request is created to discuss some issues in the kinematicWaves behavior. In a recent pull request (https://github.com/matsim-org/matsim-libs/pull/1414), the consideration of the number of lanes was added to the computation of 'maxFlowFromFdiag'. Together with this change, also the logic of the accumulated inflow capacity was changed from
`this.accumulatedInflowCap = Math.min(accumulatedInflowCap + maxFlowFromFdiag, maxFlowFromFdiag)`
to
`this.accumulatedInflowCap = Math.min(accumulatedInflowCap + maxFlowUsedInQsim, maxFlowUsedInQsim)`.

After introducing this change, we observed a major increase in car travel times in the open Berlin scenario, which is why we again looked into the code and now tend to change this back to the old behavior. I am not convinced that the old approach is the correct one but it solves our congestion problem. I tested it with RunBerlinScenarioTest.eTest1pctUntilIteration1() in matsim-berlin, branch 5.6.

Yet, I see the following issues with the old approach:

- In some cases 'maxFlowFromFdiag' can  be much higher compared to the actual flow capacity (which is specified in the network) which means that the inflow is much different than the outflow!
- In the current version, there are two options to deal with the situation 'maxFlowFromFdiag < flowCapacityPerTimeStep': (a) reduce the flow capacity and (b) increase the number of lanes. In the (b) case `this.accumulatedInflowCap = Math.min(accumulatedInflowCap + maxFlowFromFdiag, maxFlowFromFdiag)` does not really make sense...

@tschlenther @vsp-gleich @rakow @michalmac

### DIFF
--- a/matsim/src/main/java/org/matsim/core/mobsim/qsim/qnetsimengine/QueueWithBuffer.java
+++ b/matsim/src/main/java/org/matsim/core/mobsim/qsim/qnetsimengine/QueueWithBuffer.java
@@ -191,6 +191,7 @@ final class QueueWithBuffer implements QLaneI, SignalizeableItem {
 	private double accumulatedInflowCap = 1. ;
 	
 	private final FlowEfficiencyCalculator flowEfficiencyCalculator;
+	private double maxFlowFromFdiag;
 
 	private QueueWithBuffer(AbstractQLink.QLinkInternalInterface qlink, final VehicleQ<QVehicle> vehicleQueue, Id<Lane> laneId,
 							double length, double effectiveNumberOfLanes, double flowCapacity_s, final NetsimEngineContext context,
@@ -399,7 +400,7 @@ final class QueueWithBuffer implements QLaneI, SignalizeableItem {
 				// yyyyyy this should possibly be getFreespeed(now). But if that's the case, then maxFlowFromFdiag would
 				// also have to be re-computed with each freespeed change. kai, feb'18
 				
-				final double maxFlowFromFdiag = (this.effectiveNumberOfLanes/context.effectiveCellSize) / ( 1./(HOLE_SPEED_KM_H/3.6) + 1/this.qLink.getFreespeed() ) ;
+				this.maxFlowFromFdiag = (this.effectiveNumberOfLanes/context.effectiveCellSize) / ( 1./(HOLE_SPEED_KM_H/3.6) + 1/this.qLink.getFreespeed() ) ;
 				final double minimumNumberOfLanesFromFdiag = this.flowCapacityPerTimeStep * context.effectiveCellSize * ( 1./(HOLE_SPEED_KM_H/3.6) + 1/this.qLink.getFreespeed() ); 
 						
 				if ( maxFlowFromFdiag < flowCapacityPerTimeStep ) {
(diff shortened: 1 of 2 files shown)


tschlenther:
I do think that #1414 leads to a correct computation of the maximum capacity according to the fundamental diagram `maxFlowFromFdiag`. But this is only set for the inflow, not for the outflow. The outflow capacity is taken from the link attribute, if i am not mistaking. As @ikaddoura mentioned, this value can be set arbitrarily and thus differ enormously from `maxFlowFromFdiag`.

Before #1414, the inflow capacity was **always** set to the maximum capacity **for one lane**. That means, it was probably overestimated for all links with one lane and probably underestimated for links with more than one lane.

With #1414, the inflow capacity is now set according to the actual number of lanes and is the minimum of the links' capacity attribute and `maxFlowFromFdiag`. 

So, I think, actually #1414 did the right things and we should not really revert it. To me, the question rather is how the inflow capacity should be correctly set? Right now, we allow the outflow to differ from the fundamental diagram, but not the inflow.

Should those values be equal?

tschlenther:
another question is:

looking into commit e24e6eb40132789918e16514e3804991f131a99b by @kainagel which introduced the inflow capacity constraint, it clearly had the assumption of one lane per link. Thus, it also stated that the accumulated inflow capacity can be 1 at max. That is not valid any more, is it? So we should also change the initialisation value for the accumulated inflow capacity (is still 1).

tschlenther:
> Now we have for the inflow: Use the set-by-user link-attribute for inflow, but if it is too high use "maxFlowFromFdiag".
> Maybe we just change the logic back to always use "maxFlowFromFdiag" for the inflow, and then deal with the new travel times (which will probably mean re-calibration)

i just looked at a problematic run (problem is most obvious in iter 0 before all car-users can run away) with VIA. it really looks like we moved the bottleneck from the end of a link to it's start or the end of the upstream link. I'll try running the same thing with the inflow set to (corrected) `maxFlowFromFdiag` and will report back

kainagel:
Some remarks:
* `this.maxFlowUsedInQsim` almost certainly is correct.  `maxFlowFromFdiag` was a local variable, and was probably meant as such (helping to explain intermediate results of the local computation, but not meant to be elevated to non-local useage).
* The `REDUCE_FLOW_CAPACITY` approach indeed had inflow capacity smaller than outflow capacity.  Since this is just the backwards compatible behavior, I would just leave it at that.  Maybe it ignored the number of lanes; if so, it would make sense to change this (or add a third alternative).
* In general, the new setting should lead to _higher_ flow capacities, and thus to shorter travel times.

tschlenther:
> inflow capacity was changed from
> `this.accumulatedInflowCap = Math.min(accumulatedInflowCap + maxFlowFromFdiag, maxFlowFromFdiag)`
> to
> `this.accumulatedInflowCap = Math.min(accumulatedInflowCap + maxFlowUsedInQsim, maxFlowUsedInQsim)`.
> 
I checked again and indeed it is that exact change which causes the car travel times to be 3 times higher (!) than before in matsim-berlin iteration 0. The reason why we see `higher` travel times (even though we now respect the correct number of lanes) is: 
We had always set the inflow to the maximum capacity of one lane, according to the fundamental diagram. Now it might be less than that. And that will actually be the case for many links with one lane. 

....

after thinking quite a while about that:

Unfortunately, the `REDUCE_FLOW_CAPACITY` approach is `not` backwards-compatible. But it at least consistend with the number of lanes and the flow diagram, so i would hol on to it as it is. 
If we want to provide backwards-compatibility, although this was a major bug in the past, we would have to provide a third alternative that works outside of the `maxFlowFromFdiag < flowCapacityPerTimeStep` check and sets `this.maxFlowUsedInQsim` with a minimum value (of the max capacity of one lane). I am not sure, if we should do that.

tschlenther:
closing this PR, as it does lead to backwards-compatiblity and would revert some improving changes. Will soon merge #1479 instead

PR #1479, opened 2021-04-30
add backwards-compatibility for kinematic waves

tschlenther:
This is an alternative to #1477.
As mentioned there, we used to set the inflow of a qlink alwas to the maximum capacity of one lane according to the fundamental diagram. That was before #1414. With #1414, the inflow capacity is now set to the actual user-defined flow capacity, as long this does not exceed the laws of the fundamental diagram. But if it does, either the number of lanes or the flow capacity of the qlink will be adjusted. This is done correctly and not based on the assumption of a single lane.
However, this might lead to dramatic changes in capacities and thus in travel times.

#1477 suggests to rewind problem-causing changes of #1414. But actually, #1414 seems to be correct and consistent, so we should keep everything.

As people might need backwards-compatibility (like @vsp-gleich), this PR suggests another `TrafficDynamicsCorrectionApproach`.
This raises the question: **Should we allow people to use old and structurally wrong functionality?** Maybe we should do so at least for an intermediate period, so people can adjust and re-calibrate their scenarios (matsim-berlin would need that e.g.).

As `TrafficDynamicsCorrectionApproach.REDUCE_FLOW_CAPACITY` was considered to be backwards-compatible, which it actually is not, i changed the default value to `TrafficDynamicsCorrectionApproach.INCREASE_NUMBER_OF_LANES`, which assumes that the user has put more attention into defining capacities than into number of lanes. This may be another point of discussion.

@kainagel @vsp-gleich @ikaddoura @rakow

### DIFF
--- a/contribs/application/pom.xml
+++ b/contribs/application/pom.xml
@@ -86,6 +86,11 @@
 		</dependency>
 
 		<dependency>
+			<groupId>com.fasterxml.jackson.dataformat</groupId>
+			<artifactId>jackson-dataformat-yaml</artifactId>
+		</dependency>
+
+		<dependency>
 			<groupId>org.assertj</groupId>
 			<artifactId>assertj-core</artifactId>
 			<version>3.18.1</version>
--- a/contribs/application/src/main/java/org/matsim/application/MATSimApplication.java
+++ b/contribs/application/src/main/java/org/matsim/application/MATSimApplication.java
@@ -1,6 +1,12 @@
 package org.matsim.application;
 
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.dataformat.yaml.YAMLFactory;
+import com.fasterxml.jackson.dataformat.yaml.YAMLGenerator;
 import com.google.common.collect.Lists;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 import org.matsim.api.core.v01.Scenario;
 import org.matsim.application.commands.RunScenario;
 import org.matsim.application.commands.ShowGUI;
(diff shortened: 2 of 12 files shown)


tschlenther:
I do not really have a clue, why the cadyts test now fails. all links have the same capacities and numbers of lanes there...
maybe @Janekdererste can help?

tschlenther:
i still do not understand why, but cadyts seems not to work with TrafficDynamicsCorrectionApproach.INCREASE_NUMBER_OF_LANES
i will set the default back to TrafficDynamicsCorrectionApproach.REDUCE_FLOW_CAPACITY which is not really backwards-compatible but consistent in terms of the fundamental diagram

kainagel:
Thank you for working on this.  But what was the old approach (even before the lanes-increasing approach was introduced), and why was it wrong?  Thanks ...

kainagel:
To summarize:
### `Before` Ihab's changes: 
inflow cap = max of fundamental diagram (= ok) but only for 1 lane (not ok)

### After Ihab's changes there were two (config) settings: REDUCE_FLOW_CAPACITY and INCREASE_NUMBER_OF_LANES.

* REDUCE_FLOW_CAPACITY
inflow cap = min(  max of fundamental diagram with correct number of lanes ((ok)) , ~outflowCap ((not ok))~ ) 

* INCREASE_NUMBER_OF_LANES
if necessary, increase number of lanes so such that flow cap can be served.

### Tilmann: some intermediate state

### Suggestion

* REDUCE_INFLOW_CAPACITY_TO_ONE_LANE: inflow cap = max of fundamental diagram *with one lane*

* REDUCE_INFLOW_CAPACITY: inflow cap = max of fdiag *with correct nr of lanes*

* INCREASE_NUMBER_OF_LANES: leave as is

PR #1481, opened 2021-05-03
SwissRailRaptor with trip length awareness 

balacmi:
We have been working with the intermodal feature of the SwissRailRaptor extensively lately. Through this work we have encountered few places where SRR could be improved. 

One of the biggest challenges with intermodal router is the computation time. Imagine a situation where you have a trip that is 1km long, but your search radius for a particular access/egress service is 5 km. It is hard to imagine that a person would search for a stop further than the length of the trip itself. Therefore, it would be good to have a settable parameter than can be used to limit the search radius depending on the length of the trip. Currently, however, this cannot easily be added, as the [RaptorStopFinder](https://github.com/matsim-org/matsim-libs/blob/1cd0cd7bb65e7af9fd7abfec0bf8e61cc90acf46/matsim/src/main/java/ch/sbb/matsim/routing/pt/raptor/RaptorStopFinder.java#L18) does not take both start and end facility as arguments. This pull request:

- adds both starting and ending facility to the stop finder
- adds the parameter that can be used to adjust the search radius dependent on the trip crowfly distance.
- adapts intermodal and config tests

### DIFF
--- a/matsim/src/main/java/ch/sbb/matsim/config/SwissRailRaptorConfigGroup.java
+++ b/matsim/src/main/java/ch/sbb/matsim/config/SwissRailRaptorConfigGroup.java
@@ -420,7 +420,7 @@ public class SwissRailRaptorConfigGroup extends ReflectiveConfigGroup {
         private static final String PARAM_PERSON_FILTER_VALUE = "personFilterValue";
         private static final String PARAM_STOP_FILTER_ATTRIBUTE = "stopFilterAttribute";
         private static final String PARAM_STOP_FILTER_VALUE = "stopFilterValue";
-        private static final String PARAM_PERCENT_TRIP_SEARCH_RADIUS = "percentTripSearchRadius";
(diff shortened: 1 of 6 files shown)


jfbischoff:
Seems like a legit addition to me.

mrieser (approved):
Thanks. I'm not the expert on intermodal access/egress, so I trust @jbischoff's verdict.

Code looks good and clean, thanks for adding a test case!

vsp-gleich:
Seems like an interesting addition, thanks! It might give problems when beeline distance differs a lot from what walk as fallback can offer, e.g. if walk is routed over the network and the trip goes from one side of a river to the other.

balacmi:
I am not sure how this can be a problem?

vsp-gleich:
You might want to take pt and some access/egress modes to go from side of the river to the other. Re-reading the code it seems you still use the serach extension radius to look for other stops if you did not find any stop. Then it should not be a problem.

balacmi:
That is what I thought ;)

PR #1485, opened 2021-05-04
Improve land use filter

luchengqi7:
For the region without any industrial/commercial areas, we will choose a random link for thsoe area. Originally, those regions will be ignored when the land use filter is applied.

Additionaly, there is a minor fix in the Freight Trip extraction code. Now the departure time will always be before 24:00:00.

The added commented-out line in the pom file (hybridsim and protobuf) is for the Apple m1 chip. The automatic detection will not work properly on Apple m1 processor. Therefore, manual setup is necessary. To use the manual setup, simply replace the original line with the commented-out line below.


rakow:
Will close this one and create a new one with additional fixes.

PR #1486, opened 2021-05-04
Migrated ModeChoiceCoverage analysis tool from matsim-berlin to matsim-libs

jakobrehmann:
ModeChoiceCoverage is a ControlerListener that does the following calculation in each iteration: the percentage of trips have used a certain mode at least once (or 5x, 10x, …) in all previous iterations. In essence, this tool can be used to check whether enough iterations have been completed for agents to make full use of the mode choice innovation strategy. The results of this analysis are printed as outputs in text-based and graphical forms.


rakow:
Thank you, looks very useful!

vsp-gleich (approved):

PR #1490, opened 2021-05-06
Modestatsplot

mdziakowski:
is doing what it should, creats a modestats xyline chart and a stackt bar chart


tschlenther:
I think you have mixed in other things here @mdziakowski , could that be?
Normally, only `ModeStatsControlerListener` should be modified...

mdziakowski:
two branches were mixed together

PR #1491, opened 2021-05-06
Remove distribution module; mvn assembly only in matsim (core)

michalmac:
This PR simplifies the "maven structure" of matsim-libs and is motivated by:
- the recent changes in the mvn project layout
- the decision about producing assembly (zip) only for the matsim (core) while releasing matsim

Actions taken:
- the `distribution` module is removed
- all unnecessary assembly definitions are removed
- all unnecessary assembly-plugin declarations are removed (the only remaining one is in `matsim\pom.xml` and is available in the `release` profile)
- GitHub workflows are adapted

Besides having less modules and files to maintain:
- calling `mvn package` etc will not trigger creating assemblies (no need to specify `-Dassembly.skipAssembly=true`)
- calling mvn in workflows/build scripts is more compact
- instead of skipping assemblies (`-Dassembly.skipAssembly=true`) in almost all workflows, enable assemblies (with `-P release`) in selected workflows (enabling special build features on demand seems cleaner than enabling all features by default and then trying out to disable them in most of the cases)


michalmac:
This PR requires removing `-pl -distribution` from mvn arguments in some of the build scripts on Jenkins. I will take care of it (once this PR is accepted).

I am also planning to move the javadoc generation to the release profile (in a follow-up PR).

mrieser (approved):
Thanks a lot, that's great. I think it's closer to what I originally also had, but at some point somebody (not sure anymore who it was, maybe mzilske) preferred to have the assemblies built during the package phase. I guess there is no right or wrong on that, but from a performance/resource perspective, it clearly makes sense to build those artifacts only for releases. So it's all "go" from me.

michalmac:
@mrieser Thanks a lot for reviewing both PRs.

For the record-keeping purposes: I adjusted configs of the following builds on Jenkins:
- removed `!distribution` from the list of projects to build in "contrib" and "contrib-code-cov" (this PR)
- removed "skipping javadoc" from the github build, now skipped by default (PR #1495)

Nothing had to be changed in the weekly build script.

michalmac:
We are having some red builds on Jenkins due to the recently merged in changes in  #1491, #1495 and #1497. I am looking into them.

michalmac:
I fixed the failing Jenkins builds:
- Added `-P hudson` to matsim (core) and matsim-examples builds, so all 3 builds use the same profile (previously it was only the contribs). This fixed the failing matsim (core) build.
- upgraded `JavaAPIforKml` from 2.2.0 to 2.2.1 which (unexplainably; see #1501) fixed the issue with running enforcer on `contribs/vsp` (fixed for both matsim-contribs and matsim-github builds).

PR #1493, opened 2021-05-07
Temporary solution for m1 mac users

luchengqi7:
To solve for the build error on m1 MacBook, the following line needs to be changed. Simply replace the original line with the line in the comment.

@michalmac : May I know if there is any other better solution for this issue? For now, I will need to un-do the changes in the pom file before I can create any PR...


michalmac:
The issue should get be solved via #1494

PR #1499, opened 2021-05-12
Extension to freight generation

luchengqi7:
The international trips can now be included. The generated file is available on SVN: https://svn.vsp.tu-berlin.de/repos/shared-svn/projects/german-wide-freight/v1.2

The boundary links are now generated manually. Currelty, 15 links around german border are set to be the enter/exit point. More enter/exit points can be added.


contribs/application/src/main/java/org/matsim/application/prepare/freight/GenerateGermanWideFreightTrips.java
```diff
@@ -163,6 +168,19 @@ public Integer call() throws Exception {
 
         MutableInt totalGeneratedPerson = new MutableInt();
 
+        //analysis cross broader trips
+        boolean includeInternationalTrips = false;
+        List<Id<Link>> boundaryLinkIds = new ArrayList<>();
+        if (pathToBoundaryLinks != null) {
+            includeInternationalTrips = true;
+            BufferedReader csvReader = new BufferedReader(new FileReader(pathToBoundaryLinks.toString()));
```
rakow, line 176: The reader is not closed and also uses old style syntax. Use try-with-resource block and nio api.
```try (BufferedReader csvReader = Files.newBufferedReader(pathToBoundaryLinks)) {...}```

rakow (changes requested):

rakow (approved):

PR #1515, opened 2021-05-20
Add service highway type to osm reader

rakow:
Service roads can be quite important in a fine-grained network. i.e, small streets to parking lots, possibly even with bus stops.


rakow:
As discussed with @Janekdererste, we will not change the default for now and see if there are further scenarios that need to have service roads included.

Issue #1482, opened 2021-05-03
Extension search radius in SRR defaults to negative infinity


balacmi:
Currently the extension search radius in the intermodal swiss rail router defaults to negative infinity. 

In cases where this is not set leads to an interesting behavior when there are less than 2 stops found for access/egress:

https://github.com/matsim-org/matsim-libs/blob/1cd0cd7bb65e7af9fd7abfec0bf8e61cc90acf46/matsim/src/main/java/ch/sbb/matsim/routing/pt/raptor/DefaultRaptorStopFinder.java#L166

In this case the search radius will be negative infinity and no stops will be found, even though, previously there might have been 1 stop.

I would suggest setting the default extension search radius to 0.0. What others think?

mrieser:
The commit that added those parameters and set them to negative infinity has the following commit message:

> force users to set values in intermodalAccessEgress by config consistency checker and default at negative infinity

So it seems to have been a deliberate choice by @vsp-gleich. Maybe he can comment on this issue and explain a bit more of the reasoning, why not a useful default value was chosen.

balacmi:
I see... 
I noticed this behavior while working on the test case for the SRR. As consistency checker is only used when there is a controller, this check was not picked up. 

Still putting the default value to 0.0 would ensure that even outside of the controller the above mentioned situation does not occur and the number of stops is maintained at 1. Consistency checker nicely checks if this value is larger than 0 so it will still work as previously intended.

vsp-gleich:
Sorry for the late answer, our TU email infrastructure was down for 2 weeks. If I remember correctly I talked about this with @jfbischoff at the time and we had a tendency to force people to explicitly set a value here, because it is hard to come up with a value good for all access modes in all possible contexts (I implemented those search extension radii and thought of e.g. walk for access rather 300m and for car access rather several km because some stop filter applies). Obviously there is now a tendency back towards a "reasonable" default. Now it's open to discussion what a reasonable default would be here, I guess definitely something larger than 0. Using 0 instead of negative infinity avoids finding no stops, but then it's not clear if this is an advantage (works out of the box) or a disadvantage (it's still wrong and nobody will notice).
So maybe 0 is as wrong as negative infinity here, so I'm not against it. Maybe knowing the use case which bypasses the config consistency checker helps to understand what is best.

vsp-gleich:
[...]

PR #1519, opened 2021-05-26
default value of the extension search radius is now 0.0

balacmi:
This pull request:

- changes the default value of the extension search radius in SRR to 0 from negative infinity
- adds the default value in the config parameter description
- changes the config consistency check, to allow 0.0 value, and requires the value to be non-negative instead of larger than zero.


vsp-gleich:
Thanks for this pull request to resolve #1482 . The new default value of 0 is what we accepted in #1482.

I am not sure why you changed the config comment and the config consistency checker though. For most people search extension radius should be > 0. For those few who want to limit the search radius to the initial search radius for whatever reason there is still the option to do this using maxRadius. I would rather suggest keeping the config consistency checker as is and setting a config comment such as "It is strongly reccomended to use a search extension radius larger than 0, e.g. for walk something like 300m".

balacmi:
In my opinion the default value should ensure that the other behavior is maintained i.e., that at least one transit stop is provided to the router. This is ensured with the value of 0.0. 

It is then up to the user to define a non-negative extension search radius if they think that more than one stop should be found within a larger search area. Otherwise, the default behavior does not exist and the user is forced to provide a value.

vsp-gleich:
Obviously nobody else has an opinion on this. Our long term default on search extension radius can be found in [TransitRouterConfig](https://github.com/matsim-org/matsim-libs/blob/fc97a437f52e4c23d50935e8ea6790e2bfbe10e8/matsim/src/main/java/org/matsim/pt/router/TransitRouterConfig.java#L50-L60). I think we have to enforce a value > 0 using the config consistency checker, since 0 has no sense and can even be emulated using maxRadius. The decision is then whether we want a default value which somehow works for all modes such as 200m or 300m or whether we force the user to think about this value and set a default value not accepted by the config consistency checker (0m). A 300m default is less bad than 0 and for modes which work better with a higher extension radius it is maybe more of a performance loss (look for stops in 1000m, 1300m, 1600m, and then 1900m instead of 1000m, then 2000m) than an unacceptable change in result if we assume finding at least 2 stops is all we want (we might find only 2 stops in 1900m, but 4 stops in 2000m).

balacmi:
Fine, I can revert the consistency checker to 0 and add a default value of 200m. Would this be ok for you? I am not so much concerned about this as it is really a minor thing.

vsp-gleich:
Yes, I am ok with > 0 enforced by consistency checker and a default of 200m

PR #1520, opened 2021-06-02
Srr intermodal

balacmi:
This pull request enables intermodal SRR to keep the interaction activities generated by the access/egress modes. 

The following is done:
- TransitRouter can now return a sequence of PlanElement-s instead of
just legs. THis is especially important for intermodal SRR
- The route that it is returned by the SRR now keeps the interaction
activities contained in the route and RouterUtils.fillWithActivities()
adds pt interaction activities where appropriate (i.e., where other
interaction activities do not exist
- The current mechanism assumes that the access/egress trips when
containing multiple segments they are structured with alternating legs
and activities (i.e., leg-act-leg)
- Adaption of all the tests affected
- Added two intermodal tests (1) testing keeping the non-pt interaction
activity and (2) testing adding pt interaction activities for intermodal
trips
- Added activity interaction time to the travel time calculation in the DefaultRaptorIntermodalAccessEgress

Actual code has very little changes. However, the tests required a lot of adaptations due to previous assumption that SRR can only contain legs, which is not the case anymore.

Any comments or suggestions for further improvements are very welcomed.


vsp-gleich (approved):
Looks good to me from reading the code. Thank you!

PR #1534, opened 2021-06-07
update workflow to lastest version from master

mrieser:

PR #1536, opened 2021-06-07
remove distribution module

mrieser:


michalmac:
@mrieser I added a few comments to https://github.com/matsim-org/matsim-libs/pull/1534. Reverting 2 lines to the previous state should fix the issue (without the need to remove `distribution` module from `pom.xml`)

mrieser:
Thanks @michalmac! It looks as if it worked now, and as the distribution module is not needed anymore anyway, I think it doesn't matter if it is removed or if the two lines are adapted to get it running. The distribution module was also removed in the master, so 13.x is now more similar to master.

mrieser:
@kainagel: version 13.1-PR1356 is now available in the maven repo, so hopefully it works now.

PR #1537, opened 2021-06-08
Make accessegresstype settable from the config file. 

Janekdererste:
As this was marked to be included into the 13.0 release but was forgotten.

I wanted to use that config switch and was wondering why I couldn't set things from the xml config. I will keep this PR until the end of the week in case anybody has had a reason that this should still not be settable from the config.


jfbischoff:
Thanks for doing this and sorry I forgot about it :-)

PR #1540, opened 2021-06-10
Ci improvements

michalmac:
Mostly meant to simplify building and releasing matsim on many long-lived branches (master, 13.x, etc). A short description commit by commit:
1. Actions workflows are executed using the definition from the current branch, so it is enough to refer to the actual base branch (like `master` or `13.x` etc.). Same changes will be applied to configs in 13.x (via a separate PR).

2. By adding `verify-all-jobs-successful` we can easily define one required check (in branch protection) for each branch. Right now we have to create a long list of all required checks (currently almost 40 !!!) and keep it updated (which means lots of manual edits and high chances of making it wrong). Additionally this list may be different for master, 13.x, 14.x and so on:
image

3. `benchmark` module was never included into the builds, now it's added.

### DIFF
--- a/.github/workflows/deploy-on-pr-merge.yaml
+++ b/.github/workflows/deploy-on-pr-merge.yaml
@@ -4,15 +4,14 @@ on:
   push:
     branches:
       - master # snapshot deployment
-      - 13.x
   pull_request:
     types:
-      - closed # pr-labelled deployment (only if PR is merged to master or 13.x)
+      - closed # pr-labelled deployment (only if PR is merged to master)
 
 jobs:
   deploy-snapshot:
     name: deploy snapshot/PR-labelled version
-    if: github.event_name == 'push' || (github.event.pull_request.merged == true && (github.event.pull_request.base.ref == 'master' || github.event.pull_request.base.ref == '13.x'))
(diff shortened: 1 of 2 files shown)


michalmac:
Here is the visualisation of this new workflow setup: https://github.com/matsim-org/matsim-libs/actions/runs/925350112

Once `Matrix: build` is completed, `verify-all-jobs-successful` is run. Merging will be possible only if the latter does not fail.

michalmac:
After changing the protection rules, all open PRs will show something like:
```verify-all-jobs-successful Expected — Waiting for status to be reported```
Please merge `origin/master` into your PR, so the changed workflow on your PR branches will be updated.

kainagel:
Yes I noticed that.  Thanks a lot ...

michalmac:
There is a little race condition right now: `verify-all-jobs-successful` (for external pull-request) appears immediately after PR is created (and is shown as skipped - which is correct), whereas the other one (for pushes to matsim-libs) will appear only after the job matrix is built.

The problem: once the first `verify-all-jobs-successful` appears, the PR is mergeable (as GitHib considers skipped checks/jobs as "OK"). I will fix it later today.

PR #1560, opened 2021-06-22
Person money event analysis

vsp-gleich:
Adds aggregated personMoneyEvent table (PersonMoneyEventsAggregator) at each iteration and writes out all PersonMoneyEvents of the last iteration (PersonMoneyEventsCollector).

By default not installed, can be switched by installing PersonMoneyEventsAnalysisModule.


vsp-gleich:
Will have tests added when we have ressources to write tests.

PR #1572, opened 2021-06-29
Drt add idle vehicle rate to vehicle stats

vsp-gleich:


contribs/drt/src/main/java/org/matsim/contrib/drt/util/stats/DrtVehicleOccupancyProfileCalculator.java
```diff
@@ -98,16 +104,19 @@ public DrtVehicleOccupancyProfileCalculator(String dvrpMode, FleetSpecification
...
-		}
-		vehicleStates.clear();
+			vehicleStates.clear();
 
-		nonPassengerServingTaskProfiles.values().forEach(this::normalizeProfile);
-		vehicleOccupancyProfiles.forEach(this::normalizeProfile);
+			nonPassengerServingTaskProfiles.values().forEach(this::normalizeProfile);
+			vehicleOccupancyProfiles.forEach(this::normalizeProfile);
+			wasConsolidatedInThisIteration = true;
+		}
```
michalmac, line 119: Nice that now the client code is not need to worry about calling the consolidation. However, it still needs to be aware that the consolidation is run once per iteration (after a qsim ends), so calling get...() in the middle of a qsim will lead to errors.

Maybe adding some comments would be helpful.

Another option: require `!wasConsolidatedInThisIteration` inside all `handleEvent()` methods.

michalmac (approved):
Nice!

Perhaps to be fully on the safe side, we should protect against calling `get...()` in the middle of a simulation (see the comments).

contribs/drt/src/main/java/org/matsim/contrib/drt/analysis/DrtAnalysisControlerListener.java
```diff
@@ -60,24 +64,29 @@
...
 	private final DrtRequestAnalyzer drtRequestAnalyzer;
+	private final DrtVehicleOccupancyProfileCalculator drtVehicleOccupancyProfileCalculator;
 	private final DrtConfigGroup drtCfg;
 	private final QSimConfigGroup qSimCfg;
 	private boolean headerWritten = false;
 	private boolean vheaderWritten = false;
 	private final String runId;
 	private final DecimalFormat format = new DecimalFormat();
 	private final int maxcap;
+	private final String notAvailableString = "NA";
```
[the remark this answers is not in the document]
vsp-gleich: fixed

contribs/drt/src/main/java/org/matsim/contrib/drt/util/stats/DrtVehicleOccupancyProfileCalculator.java
vsp-gleich: I solved this now using an `AfterMobsimListener`, so consolidate() checks whether the `AfterMobsimEvent` has already been thrown and fails otherwise. Those `get...()` methods all call consolidate() in the end, so checking at consolidate() should be sufficient. This seems more computation time efficient than checking at each `handleEvent()` method. There should not be any ordering issues since the output analysis is called with the `IterationEndsEvent` which occurs after the `AfterMobsimEvent`.

vsp-gleich:
@michalmac Thanks for your comments!

contribs/drt/src/main/java/org/matsim/contrib/drt/util/stats/DrtVehicleOccupancyProfileCalculator.java
michalmac: Looks really nice! Thanks!

Issue #1566, opened 2021-06-25
DrtRoute should provide the unshared ride path


steffenaxer:
@michalmac, in order to implement different spatial pricing mechanisms for DRT, I would like to request the link sequence of the unshared ride path, which is internally available (VrpPathWithTravelData), but not used within the DrtRouteCreator.

sebhoerl:
Addition: The idea could be to encode the content of the DrtRoute like it is now done for the DefaultTransitPassengerRoute, i.e. a JSON object with supplementary information. This could also be optional depending on a flag (whether to provide the link sequence in the route, or not), and maybe this could also be useful for the emulation of DRT trips.

steffenaxer:
I provided a proposal for this issue, see #1579

steffenaxer:
PR has been merged.

PR #1579, opened 2021-07-02
DrtRoute should provide the unshared ride path

steffenaxer:
This PR is a proposal for #1566


sebhoerl:
Just as a comment: This now replicates the same set-up as we have in `DefaultPassengerTransitRoute` for public transport, which is nice as things become more streamlined and extensible.

One remark: Maybe we should have a fallback for old-format route descriptions, e.g. either
- Testing the description with a regex like `[0-9]+ [0-9]+`
- Testing whether description starts with `{`

And creating a deprecation warning if old-format route is used (and then to be removed in later version). But let's first see what everybody else says :)

steffenaxer:
> Just as a comment: This now replicates the same set-up as we have in `DefaultPassengerTransitRoute` for public transport, which is nice as things become more streamlined and extensible.
> 
> One remark: Maybe we should have a fallback for old-format route descriptions, e.g. either
> 
>     * Testing the description with a regex like `[0-9.]+ [0-9.]+`
> 
>     * Testing whether description starts with `{`
> 
> 
> And creating a deprecation warning if old-format route is used (and then to be removed in later version). But let's first see what everybody else says :)

@sebhoerl, I can't find code that reads directly from the RouteDescription. Do you have any code, related to DMC that parses the RouteDescription? Maybe you can point me to a part, where the information from the RouteDescription is parsed and used. In this case, I will try to implement a retrofit version.

contribs/drt/src/main/java/org/matsim/contrib/drt/routing/DrtRoute.java
```diff
@@ -39,44 +47,60 @@
...
-		return maxWaitTime.seconds() + " " + directRideTime.seconds();
+		try {
+			return new ObjectMapper().writeValueAsString(routeDescription);
+		} catch (JsonProcessingException e) {
+			throw new RuntimeException(e);
+		}
 	}
 
 	@Override
 	public void setRouteDescription(String routeDescription) {
```
sebhoerl, line 98: @steffenaxer This is called when the route is read from a plans file. So here could be a condition to fall back to the old format in `routeDescription`.

steffenaxer: Ok, got it. I will provide a retrofit version

michalmac (approved):
⭐ 

Re original request in #1566: I do not see any reason for not providing the unshared path :-)

steffenaxer:
I need to test if things are working now when reading an old plans file. There is currently no test that reads in plans that already contains routed drt plans.

sebhoerl:
I think it will be enough to have a test for both formats on `setRouteDescription`

steffenaxer:
> I think it will be enough to have a test for both formats on `setRouteDescription`, even better without a static plans file (whose format may change in the future).

I added a commit that contains a handling for old and new format. @sebhoerl maybe you could have a look?

sebhoerl (approved):

sebhoerl:
Nice, but I think the test is not committed yet, is it?

steffenaxer:
Commit dee1acf92b56459a4133bb6b137e98fb2848bdc1 contains your proposed implementation. I haven't implemented a warning. Do you think we need something?

sebhoerl:
Ah, sorry. I meant to add a unit test for `DrtRoute::setRouteDescription`.

steffenaxer:
Ok, I assumed it, but was not 100% sure. I will add a unit test tomorrow. Thanks for your help!

steffenaxer:
@sebhoerl junit test is implemented. Please have a look.

PR #1580, opened 2021-07-05
Add routingMode to PersonDepartureEvent

### LINKED ISSUE #1577: Communicate routingMode in events
I was just experimenting with some feeder simulations by chaining the DRT routing module and PT routing module. The reuslting "routing mode" is called "feeder". What I noticed is that we can nicely deduct the mode of each leg from the events (through the departure events) but there is not really a way to deduct the "trip mode", which would be the "routingMode" internally of MATSim. So effectively, "feeder" never appears anywhere in the events.

Some options:
- Introduce a new event (e.g. `PersonTripEvent`) that is sent by the QSim when an agent departs from a non-interaction activity
- Add a `routingMode` attribute to the `PersonDepartureEvent`

What do you think? (I'd find the second option less intrusive)

sebhoerl:
This PR adds the `routingMode` to `PersonDepartureEvent`.

Most important changes:
- Event readers for `PersonDepartureEvent`
- `PersonDepartureEvent` itself
-  `QSim::arrangeAgentDeparture` : Here, we need to perform a check and then cast for `PlanAgent`, because `MobsimAgent` does not know about the `routingMode`. Maybe not the nicest way from an architecture point of view, but I'm not sure how to do it better without majro refactoring. Ideas welcome :)
- Otherwise mostly fixing unit tests
- Added test in `NetworkRoutingInclAccessEgressModuleTest::routingModeInEvents`

Closes #1577.


sebhoerl:
Disclaimer: If didn't update those unit tests yet where we need to update the reference events files. I guess is something that is worth doing once we have agreed if the PR should be merged or not ;-)

Issue #1570, opened 2021-06-29
TrajectoryToPlans typical duration is capped at 4 hours


Janekdererste:
A student in the matsim class noticed that the duration of activities and the assigned activity type with duraction (e.g. "home_14400") sometimes do not match. Looking at the code it seems like the duration we use for the different activity types is capped at 4 hours. Is this intended behaviour, and if so, why do we do it like this?

The code can be found here:
https://github.com/matsim-org/matsim-libs/blob/master/contribs/application/src/main/java/org/matsim/application/prepare/population/TrajectoryToPlans.java#L130

jingjunL:
Thanks Janek for posting this out :) Below is a screenshot of a typical agent plan with the "capping 4 hours" setting. He did a home activity over the entire simulation day, of which the typical duration would be 14400 (so 4 hours in total), meaning he should stop this activity at around 4 am in the morning. However, the end_time parameter is 12:58 (far more than 4 hours after midnight).

<img width="558" alt="1623240055(1)" src="https://user-images.githubusercontent.com/63421908/123768179-6422e780-d8c8-11eb-9e82-e98092c4c629.png">

rakow:
This code did come from the OpenBerlinScenario, It seems the duration is capped at 24 bins, which is 14400 in this case.
To me, it looks a bit like the bin size was 1h at some point and the maximum capped to 1 day. Probably the max bin number was not updated.
  
@ikaddoura please let us know if there is indeed a reason to cap the maximal bin size at 14400 seconds.

Janekdererste:
Thanks @rakow , I was suspecting something like this as well. I will create a PR with a fix and invite @ikaddoura as a reviewer, in case we are wrong

ikaddoura:
Such a time cap does not make sense. I just looked into one of the input files of the OpenBerlinScenario and everything seems to be okay here. There are several activities without this weird time cap, e.g. home_58800 or home_33000.

See for example here: https://svn.vsp.tu-berlin.de/repos/public-svn/matsim/scenarios/countries/de/berlin/berlin-v5.5-1pct/input/berlin-v5.5-1pct.plans.xml.gz

PR #1582, opened 2021-07-06
expand number of categories to fill the entier day.

Janekdererste:
Have enough activity categories to span a whole day.

Fixes https://github.com/matsim-org/matsim-libs/issues/1570

Does it make sense to have this capped at 24 hours? or should we just have as many activity times as there are durations?


rakow:
I think it would be great if this maxBin number could be configured via an `@CommandLine.Option`

rakow (approved):

ikaddoura (approved):
looks fine

Janekdererste:
I have to fix test cases. But I haven't forgotten about this.

PR #1584, opened 2021-07-07
Fix snapshot positions

Janekdererste:
Make the snapshot positions align with physics, so that emission calculation from snapshot positions is possible


Janekdererste:
@vsp-gleich assigned you as a review because you were involved. Have a look if you'd like

vsp-gleich:
@Janekdererste I have a project deadline on Friday, so if you need to proceed earlier just merge and don't wait.

PR #1592, opened 2021-07-13
Add filtering to snapshots, to reduce amount of snapshots

Janekdererste:


contribs/application/src/main/java/org/matsim/application/analysis/CheckPopulation.java
```diff
@@ -37,14 +37,14 @@
 	@CommandLine.Parameters(arity = "1", paramLabel = "INPUT", description = "Path to population")
 	private Path input;
 
-	@CommandLine.Option(names = "--attribute", arity = "0..*", description = "Print full distribution for selected attributes", required = false)
-	private List<String> queryAttr = new ArrayList<>();
+	@CommandLine.Option(names = "--attribute", arity = "0..*", description = "Print full distribution for selected attributes")
```
Janekdererste, line 40: @rakow IntelliJ automatically inserted final here. Will the command line utility work with the assigned default values or will it try to replace these with another object? 

I'll happily take them out again, just curious

rakow: It might work for mutable options, like in this case. But I think it is better to add a test case to make sure. CheckPopulation doesn't have one yet.

rakow: Looking at your tests, many are already failing because of the final added in different places.

Janekdererste: ja, ist mir schon aufgefallen. Ich muss dieses Setting wohl mal wieder ausstellen 🙄.

Janekdererste:
Setting this to automerge, if you don't have any objectsion @rakow

rakow (approved):

PR #1594, opened 2021-07-15
App contrib extensions

rakow:
Added functionality to the application contrib:
*  Make trips to legs optional in clean population, as it removes routes
*  Fix parsing of MATSim args when no = is used as separator
*  Added command to insert home activities at the end of agent plans 
*  Added spatial air pollution aggregation


tschlenther:
>     * Added command to insert home activities at the end of agent plans
does this work like it was done for the Kelheim scenario?

rakow:
> 
> 
> > ```
> > * Added command to insert home activities at the end of agent plans
> > ```
> 
> does this work like it was done for the Kelheim scenario?

It is similar, but a bit simplified, because it does not make use of routing information (only beeline distance).

Issue #1581, opened 2021-07-06
Minibus contrib: escalating money events


JWJoubert:
Hi @neuma, @vsp-gleich 

I am trying to escalate a money event (like a fine, or toll) from the minibus vehicle, all the way back to the association (the agent deciding on how to change the routes and/or vehicles). The [code example](https://github.com/matsim-org/matsim-libs/blob/01ef87abd7e3faca035ead42c3ba27c75d98adca/contribs/minibus/src/main/java/org/matsim/contrib/minibus/RunConfigurableMinibusExample.java) is no longer working.

A `PersonMoneyEvent` seems to only hit the commuter's pocket; not the person driving the minibus ;-)

I then tried doing this using a `LinkEnterEvent` listener:
- from the event I can get the vehicle `Id`;
- I then have to iterative search through the `POperators` and match the vehicle's `Id` with the `POperator`'s `Id`;
- I then reduce the `POperator`'s budget with whatever the fine amount is.

Currently, I do not seem to get the behavioural response I'm expecting.

I do not know if there is an intuitively more accurate way of doing it? Especially using existing `minibus` infrastructure. I was hoping to use something like the [`StageContainer2AgentMoneyEvent`](https://github.com/matsim-org/matsim-libs/blob/01ef87abd7e3faca035ead42c3ba27c75d98adca/contribs/minibus/src/main/java/org/matsim/contrib/minibus/scoring/StageContainer2AgentMoneyEvent.java) or the [`TicketMachineI`](https://github.com/matsim-org/matsim-libs/blob/01ef87abd7e3faca035ead42c3ba27c75d98adca/contribs/minibus/src/main/java/org/matsim/contrib/minibus/fare/TicketMachineI.java).

Do you have any advice, please? Or can you point me to code where I can navigate through?

JWJoubert:
Hm, any help @neuma or @vsp-gleich? I just need a pointer on how to (programmatically) get a "money event" to a driver/operator's `Stagecontainer` maybe?

vsp-gleich:
Hi @JWJoubert , 

according to the git history the code example does not work since January 2017. I cannot give any better advice than what it is saying, copy and adapt your own `PModule`. That does not seem very hard at first glance, since this class only plugs other classes together and you only have to add or replace one with your own implementation. That does not mean you necessarily have to use a different `PModule`, there might be other options (start on a branch / configure via PConfig) if this is to end up in matsim-libs.

[...]

PR #1600, opened 2021-07-20
Adding arbitrary cost to minibus vehicle

JWJoubert:
The goal is to give (public) access to a minibus vehicle’s `CostContainer` so that arbitrary costs/expenses can be added. For example, toll, fines, or other daily expenses.


JWJoubert:
@vsp-gleich, I wanted to make some progress on simplifying the different cost collectors/handlers/containers. But after many UML printouts of the classes, I gave up. There are very many cross-references and handlers being added to multiple containers. I did try and create a bit of a mind map... but for now, I just need a way to add arbitrary costs to a vehicle from _outside_ the contrib *without* having to integrate the `roadpricing` contrib and create too many interdependencies.

I hope this is an easy enough (and acceptable) workaround for you.

contribs/minibus/src/main/java/org/matsim/contrib/minibus/hook/PBox.java
```diff
@@ -263,6 +260,10 @@ TransitSchedule getpTransitSchedule() {
 		return operators;
 	}
 
+	public OperatorCostCollectorHandler getOperatorCostCollectorHandler(){
+		return operatorCostCollectorHandler;
+	}
+
```
vsp-gleich, line 266: This is potentially dangerous

contribs/minibus/src/main/java/org/matsim/contrib/minibus/scoring/OperatorCostCollectorHandler.java
```diff
@@ -165,4 +166,8 @@ public void notifyAfterMobsim(AfterMobsimEvent event) {
 			}
 		}
 	}
+
+	public Map<Id<Vehicle>, OperatorCostContainer> getVehicleIdToOperatorCostContainerMap(){
+		return this.vehId2OperatorCostContainer;
+	}
```
vsp-gleich, line 172: This is potentially dangerous

contribs/minibus/src/main/java/org/matsim/contrib/minibus/scoring/OperatorCostContainer.java
```diff
@@ -72,6 +73,15 @@ public void addDistanceTravelled(double meterTravelled1) {
...
 
+	/**
+	 * Adds any additional expense that the vehicle may incur, for example toll
+	 * or fines.
+	 */
+	public void addArbitraryExpense(double expense){
+		this.otherExpensesIncurred += expense;
+	}
+
+
```
vsp-gleich, line 84: this seems ok since the other methods affecting cost are public too

vsp-gleich:
This pull request is supposed to solve #1581

vsp-gleich (commented):
Opening up the OperatingCostContainer and allowing public access to it is a potentially dangerous step since it allows for arbitrary modifications of the cost calculated (including the meters and seconds travelled). This is something we tend to avoid or even build back. I somehow liked the idea of processing PersonMoneyEvents more (using the driver id to associate them to the right vehicle, see #1581) although that obviously processes all PersonMoneyEvents with the driver id which is nice if the roadpricing contrib throws PersonMoneyEvents and so it might work out of the box without having them depend on each over. This is not to say I am completely against, but maybe it is better to wait for @andreas77 opinion.

JWJoubert:
> I somehow liked the idea of processing PersonMoneyEvents more (using the driver id to associate them to the right vehicle, see #1581) although that obviously processes all PersonMoneyEvents with the driver id which is nice if the roadpricing contrib throws PersonMoneyEvents and so it might work out of the box without having them depend on each over. This is not to say I am completely against, but maybe it is better to wait for @andreas77 opinion.

I did try it,  @vsp-gleich, but have not found a way to get the driver's Id from the vehicle.
```
 Vehicle2DriverEventHandler vehicle2DriverEventHandler = new Vehicle2DriverEventHandler();
 Id<Person> driverId = vehicle2DriverEventHandler.getDriverOfVehicle(linkEnterEvent.getVehicleId());
```
keeps returning a `null` for the `driverId`. Maybe I'm using the `Vehicle2DriverEventHandler` incorrectly?

vsp-gleich:
> keeps returning a `null` for the `driverId`. Maybe I'm using the `Vehicle2DriverEventHandler` incorrectly?

I think you have to register the Vehicle2DriverEventHandler like the other event handlers.

JWJoubert:
> keeps returning a `null` for the `driverId`. Maybe I'm using the `Vehicle2DriverEventHandler` incorrectly?

Yes, thanks, @vsp-gleich ... I got that sorted now. This means I might have a shot at rather implementing it with your proposed `PersonMoneyEvent`. I'll try that now.

kainagel:
I think that the correct design direction is already written down in lines 62-65 of `OperatorCostContainer`:
```
	// yyyy my intuition is that this class should be very much reduced, just
	// returning "cost", and handling arbitrary events, and figure
	// out the rest internally. That would make it much more flexible with respect
	// to replacement. kai, jan'17
```
If this is too much design change for the time being, the alternative would be add a "handleMoneyEvent" method, and then modify `OperatorCostCollectorHandler` such that the correct money events are caught there and passed to the operator.

Or am I overlooking something?

JWJoubert:
Indeed, @kainagel @vsp-gleich, that is what I implemented in the revised pull request. Or rather, tried.

I agree with your intuition, @kainagel (in the comment) - I just do not know where to start. There seems to be very many crossreferences and containers and multiple stage/handlers being added. And after a few mind-maps and many pages of UML printouts to try and map the relations... I gave up  (for now) ;-)

JWJoubert:
Happy (or, happier) with _this_ implementation, @vsp-gleich & @kainagel?

kainagel:
To me, this is looking ok now.  I took the opportunity to combine three packages (move `operator` and `scoring` into `hook`), which reduces public footprint.  In particular, the details of what we are changing here are no longer exposed, meaning that we can change them later.  If the test pass, and the code is still useful after my reducing public footprint, as far as I am concerned it can be merged.  Maybe @vsp-gleich has further comments ...

contribs/minibus/src/main/java/org/matsim/contrib/minibus/hook/OperatorCostCollectorHandler.java
```diff
@@ -165,4 +157,33 @@ public void notifyAfterMobsim(AfterMobsimEvent event) {
...
+	public Map<Id<Vehicle>, OperatorCostContainer> getVehicleIdToOperatorCostContainerMap(){
+		return Collections.unmodifiableMap( this.vehId2OperatorCostContainer );
+	}
+
+	@Override
+	public void handleEvent(PersonMoneyEvent event) {
+		// yyyy what comes below is not so beautiful since it depends on conventions: If some string contains something of some other string,
+		// then ... It looks like this was the original design of the minibus contrib, so in principle it should be operated out at some
+		// point, but probably not now.  kai, jul'21
+
```
vsp-gleich, line 170: I agree to Kai's comment here

contribs/minibus/src/main/java/org/matsim/contrib/minibus/hook/OperatorCostCollectorHandler.java
```diff
@@ -165,4 +157,33 @@ public void notifyAfterMobsim(AfterMobsimEvent event) {
...
+		if(event.getPersonId().toString().contains(this.pIdentifier)){
+			/* It is a minibus driver. Now find the correct vehicle this person is driving. */
+			for(Id<Vehicle> vehicleId : vehId2OperatorCostContainer.keySet()){
+				/* The driver Id contains the vehicle Id. For example, if the
+				vehicle Id is para_1_2 then you will find that the driver's Id
+				is something like pt_para_1_2_para_. I do not know where this
+				is set/created, but it may change in the future (JWJ '21) */
```
vsp-gleich, line 177: We can find out, but it does not seem useful to invest much time into that given Kai's comment above. Looping over all vehicles does not seem very efficient, but for a temporary solution it should be ok unless you report performance issues having a serious impact on you, @JWJoubert .

vsp-gleich (approved):
For a temporary solution this seems ok. On the long run it would be nice to reduce the naming conventions (see Kai's comment) which would cause some huge changes to the code. I am currently calculating pt operating cost for a different project. Maybe the minibus calculation can be made more generic and re-usable for other projects in the far future.

JWJoubert:
> For a temporary solution this seems ok. On the long run it would be nice to reduce the naming conventions (see Kai's comment) which would cause some huge changes to the code.

I agree, @vsp-gleich. The naming conventions are often confusing: sometimes the `Operator*` refers to the vehicle (driver), and sometimes the vehicle _owner_.

PR #1605, opened 2021-07-22
3 ways of creating an EV fleet specification

michalmac:
Three options:
- from a separate EV file (EV config) -- this is the existing approach
- from the "standard" matsim vehicles (if `QSimConfigGroup.VehiclesSource.fromVehiclesData` is selected)
- via a custom binding (e.g. generated vehicles)

### DIFF
--- a/contribs/ev/src/main/java/org/matsim/contrib/ev/EvConfigGroup.java
+++ b/contribs/ev/src/main/java/org/matsim/contrib/ev/EvConfigGroup.java
@@ -19,12 +19,14 @@
 
 package org.matsim.contrib.ev;
 
-import org.matsim.core.config.Config;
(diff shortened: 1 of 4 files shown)


tschlenther (commented):
looks decent to me 👍  thank you :)

michalmac:
I forgot to mention the order of precedence:
```
if the EV file is provided in the config:
        read vehicles from that file
else if QSimConfigGroup.VehiclesSource.fromVehiclesData (i.e. the standard matsim vehicle file is provided):
        create EVs from the standard matsim vehicles (expecting custom EV attributes)
else
        do nothing (assume there is a custom binding provided by the user of the EV contrib, if none exists - Guice will throw an exception)
```

PR #1607, opened 2021-07-23
Dvrp vehicles from matsim vehicles

michalmac:
The following options are allowed
- separate DVRP vehicles file
- from matsim vehicles
- custom binding (e.g. generated vehicles)

Different DVRP modes can choose different strategies.

This PR is about DVRP vehicles and is similar to #1605 (that was about EVs).


michalmac:
> - from matsim vehicles

For the second option to work, we need to add additional attributes into matsim vehicles (see `DvrpVehicleSpecificationWithMatsimVehicle`).

michalmac:
I forgot to mention the order of precedence:
```
if the taxi/drt/... vehicle file is provided in the taxi/drt/... config:
        read vehicles from that file
else if QSimConfigGroup.VehiclesSource.fromVehiclesData (i.e. the standard matsim vehicle file is provided):
        create DVRP vehicles from the standard matsim vehicles (expecting custom DVRP attributes)
else
        do nothing (assume there is a custom binding provided by the user of the DVRP contrib, if none exists - Guice will throw an exception)
```

PR #1618, opened 2021-08-01
Stop multi-node dijkstra after finding least-cost end node

michalmac:
Compared to the `allEndNodesReached`, this criterion detects cases when the first of the end nodes is reached. Each end node may be assigned some additional cost (that can be interpreted as an access or egress cost, depending on the search type: either forward or backward).


michalmac:
> Each end node may be assigned some additional cost (that can be interpreted as an access or egress cost, depending on the search type: either forward or backward).

This means that if a node can be reached before all other end nodes, but it has a relatively high additional cost, some other end node may be the selected as the least cost one.

PR #1623, opened 2021-08-12
add method addCarrierVehilceTypes to FreightUtils

rewertvsp:
I add this one method because it was missing. I hope this is ok for you

### DIFF
--- a/contribs/freight/src/main/java/org/matsim/contrib/freight/utils/FreightUtils.java
+++ b/contribs/freight/src/main/java/org/matsim/contrib/freight/utils/FreightUtils.java
@@ -209,6 +209,20 @@ public class FreightUtils {
 		}
 		return types;
 	}
+	
+	/** Adds the CarrierVehicleTypes to the scenario.
+	 * @param scenario
+	 * @param newTypes
+	 */
+	public static void addCarrierVehicleTypes(Scenario scenario, CarrierVehicleTypes newTypes) {
+		CarrierVehicleTypes existingTypestypes = (CarrierVehicleTypes) scenario.getScenarioElement(CARRIERVEHICLETYPES);
+		if (existingTypestypes == null) 
+			scenario.addScenarioElement(CARRIERVEHICLETYPES, newTypes);
+		else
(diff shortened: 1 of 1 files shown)


contribs/freight/src/main/java/org/matsim/contrib/freight/utils/FreightUtils.java
```diff
@@ -209,6 +209,20 @@ public static CarrierVehicleTypes getCarrierVehicleTypes(Scenario scenario) {
 		}
 		return types;
 	}
+	
+	/** Adds the CarrierVehicleTypes to the scenario.
+	 * @param scenario
+	 * @param newTypes
+	 */
+	public static void addCarrierVehicleTypes(Scenario scenario, CarrierVehicleTypes newTypes) {
+		CarrierVehicleTypes existingTypestypes = (CarrierVehicleTypes) scenario.getScenarioElement(CARRIERVEHICLETYPES);
```
kt86, line 218: `Typestypes`? Should it not be something like `existingCarrierTypes`?

contribs/freight/src/main/java/org/matsim/contrib/freight/utils/FreightUtils.java
```diff
@@ -209,6 +209,20 @@ public static CarrierVehicleTypes getCarrierVehicleTypes(Scenario scenario) {
 		}
 		return types;
 	}
+	
+	/** Adds the CarrierVehicleTypes to the scenario.
+	 * @param scenario
+	 * @param newTypes
+	 */
+	public static void addCarrierVehicleTypes(Scenario scenario, CarrierVehicleTypes newTypes) {
```
kt86, line 217: I would like to have a short test for this. to see, if the types are added correctly

kt86, line 217: I would prefer to have it more analog to the `addOrGetCarriers` method in line 182ff. --> not set the values directly; This seems to be closer to the MATSim standard way of doing it.

The types can be set then outside the FreightUtils

PR #1627, opened 2021-08-20
Consolidate departure time calculations along plans

sebhoerl:
Based on #929 this PR migrates the `TimeInterpreter` which was used in `DMC` to track activity durations and travel times along a plan. The idea is that based on this information, departure times of trips can be determined exactly when routing a plan trip by trip. This functionality is needed in many places in MATSim (PlanRouter, DRT, ...), but so far was scatter over many, partly redundant, static methods. 

In this PR, the `TimeInterpreter` interface is introduced, along with a number of implementations which correspond to the `ActivityDurationInterpretation` in the `plans` config group. Mainly, `minOfDurationAndEndTime` or `tryEndTimeThenDuration` can be chosen, and the corresponding instance will be bound via Guice (i.e. `EndTimeThenDurationInterpreter` and `MinimumEndTimeAndDurationInterpreter`).

The various static methods have been removed and replaced by `TimeInterpreter`. For that, it is injected where necessary.

An additional point that came up in #929 is that, so far, the `PlanRouter` would not advance departure times along the plan. If, for instance, an activity has an end time of 08:00:00, but the trip that is routed right before that activity ends at 08:10:00, the traditional version would still route the next trip as if it departs at 08:00:00. This remains the default behavior, but it can now be controlled by `RoutingDepartureTimeInterpration` in the `plansConfigGroup`. The options are `nominal` (default - just use the time that is written in the activity), and `cumulative` (new - track travel times along the plan). Technically, the flag controls whether activity end times may override already further advanced times in the `TimeInterpreter`.


sebhoerl:
Current implementation was a bit too optimistic as TimeInterpreter needs to be injected in a lot of places (although before, Scenario or Config needed to be passed to make use of the static methods). The code should be changed such that instead of having a "tracking along a plan" tool, we have a TimeInterpreter, which resembles the functionality of the existing static methods. This will also allow to smoothly consolidate the functionality (some of the static methods used have their own specificities):

- [ ] Revert most from above
- [ ] Introduce a TimeInterpreter *object* which provides the methods that were formerly static
- [ ] In a first step, simply replace the static calls with calls to the interpreter object
- [ ] In a second step, consolidate functions that are now part of the interpreter object

PR #1635, opened 2021-08-23
Adding noise analysis to the MATSim Application

luchengqi7:
The noise analysis is added to the MATSim analysis. The analyzed region can be set by the boundary coordinates or by shapefile. 

As there are some updates in the master branch. This pull request is based on the master branch. Necessary changes in other files are added in order to make the noise analysis runable.


contribs/application/src/main/java/org/matsim/application/analysis/noise/NoiseAnalysis.java
```diff
@@ -0,0 +1,134 @@
...
+            config.plans().setInputFile(runDirectory + "/" + runId + ".output_plans.xml.gz");
+        } else {
+            config.network().setInputFile(runDirectory + "/" + runId + "output_network.xml.gz");
+            config.plans().setInputFile(runDirectory + "/" + runId + "output_plans.xml.gz");
+        }
+        config.controler().setOutputDirectory(runDirectory);
+
+        // adjust the default noise parameters
+        NoiseConfigGroup noiseParameters = ConfigUtils.addOrGetModule(config, NoiseConfigGroup.class);
+        noiseParameters.setReceiverPointGap(250);
```
rakow, line 67: The receiver point gap might be an option configurable as command line option. There should be very few magic numbers in the command scripts.

PR #1640, opened 2021-08-24
Pass arguments in DRT route creation to allow trip-based requirements

sebhoerl:
These changes make it possible to implement a custom extension of the `DrtRouteCreator` which bases waiting time on user-specific or even trip-specific requirements.

### DIFF
--- a/contribs/drt/src/main/java/org/matsim/contrib/drt/routing/DefaultDrtRouteUpdater.java
+++ b/contribs/drt/src/main/java/org/matsim/contrib/drt/routing/DefaultDrtRouteUpdater.java
@@ -17,13 +17,16 @@
  * *********************************************************************** */
 package org.matsim.contrib.drt.routing;
 
+import java.util.LinkedList;
+import java.util.List;
+import java.util.concurrent.Future;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
-import java.util.stream.Stream;
 
 import org.matsim.api.core.v01.network.Link;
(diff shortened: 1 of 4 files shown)


contribs/drt/src/main/java/org/matsim/contrib/drt/routing/DefaultDrtRouteUpdater.java
```diff
@@ -67,19 +72,28 @@ public DefaultDrtRouteUpdater(DrtConfigGroup drtCfg, Network network,
...
+			for (Trip trip : TripStructureUtils.getTrips(person.getSelectedPlan())) {
+				for (Leg leg : trip.getLegsOnly()) {
+					if (leg.getMode().equals(drtCfg.getMode())) {
+						futures.add(executorService.submitRunnable(router -> updateDrtRoute(router, person, trip.getTripAttributes(), leg)));
+					}
+				}
+			}
+		}
+		
+		futures.forEach(Futures::getUnchecked);
```
sebhoerl, line 87: I tried a bit around, but I couldn't come up with a more elegant solution to pass a list of arguments from all levels of the person -> trip -> leg hierarchy to the router. Maybe there is a better way?

michalmac: Looks nice and readable 👍

michalmac (approved):
Looks good to me. Thanks!

PR #1641, opened 2021-08-25
Attribute-based stop finder in DRT

sebhoerl:
- Passes trip attributes on to the AccessEgressFacilityFinder so acces/egress facilities can be, for instance, selected by certain trip-level service requirements, e.g. standard vs. premium service
- Implementation of an attribute-based finder, which looks up DRT stops based on "stopNetworks" that they belong to (defined via their attributes) and which are requested by a "stopNetwork" trip-level attribute in the routing stage.


sebhoerl:
@michalmac Just looked into it and found the catch: `DrtStopFacility` is `Attributable`, but not the general `Facility`, that's why the class is living in `drt`. What do you propose?

michalmac:
@sebhoerl Ah, I see. Maybe then provide `parseStopNetworks()` as a `Function<Facility, Set<String>>` to the `create()` method? The rest of the code should be applicable to any type of facilities.

BTW. Not sure why only some subtypes of `Facility` are `Attributable` but not all? So alternatively, we could make `Facility` implement `Attributable`, but maybe this is too much given the purpose of this PR.

sebhoerl:
Okay, 

```
<T extends Facility & Attributable> AttributeBasedStopFinder create(...., Collection<T> facilities)
```
does the trick ;)

michalmac:
> Okay,
> 
> ```
> <T extends Facility & Attributable> AttributeBasedStopFinder create(...., Collection<T> facilities)
> ```
> 
> does the trick. Then one can do:
> 
> ```
> AttributeBasedStopFinder.create(..., drtStopNetwork.getStops().values());
> ```

Nice! I just took one step further here: #1676 (only refactoring, no functional changes).

PR #1648, opened 2021-08-30
Taxi: pre-compute origin-destination paths

michalmac:
In some configurations, the taxi optimiser may not know the destination until the passenger is picked up. But we can obtain this information already on request submission and use it to pre-compute the origin-destination path in the background.

If sufficiently many threads are used, all such paths are found before the pickup simulation ends. Consequently, QSim is not (directly) slowed down by computing these paths.

4-6 threads is usually enough. It's recommended to specify a higher number than that if possible (see the new param in the taxi config) - of course, threads will probably be not 100% busy. By default, it is set to the number of available cores.


michalmac:
This PR adds additional ~33% computation time reduction on top of PR #1646 (QSim duration for 10% Berlin AT case). So the overall speedup is above 2.

PR #1650, opened 2021-08-31
Emissions: No automagic convertion of Hbefa VehCatagory for cold emissions

kt86:
After the discussion in #1578 here a clean PR for solving this.
@JWJoubert, @tschlenther, @kainagel can you please comment on this? (@kainagel: we can also talk on this in our meeting on thursday)

Issue description:
- HBEFA does _not_ provide any **cold** start emissions for veh categories except `cars` and `light commercial vehicle (LCV)`
- The previous idea was to convert those missing vehicle types to `cars` and use that table. This does not work for (semi) detailed vehicle types if not providing average data for cars, because the entries in the other fields are not the same e.g. for HGV and cars. Now allowing falling-back and looking up average values does not seems to be useful, as we try to step away from that approach (see #960 #849 )

Solution: 
- remove the automagic convertion to cars (or any other vehicle category)
- This may break somebodies code, because you now need data for e.g. HGV in you HBEFA table -> make sure that you have a HBEFA table that corresponds to your vehicles
- If you want to recreate the existing results, you need to create new entries with the emissions values from cars
- motorcycle was treated differently: It just returned "0" -> This was also removed and an entry in the table is needed.
- for scenarios with only cars and/or LCV everything should be as it is

Further notes / discussion:

1.  I am wondering if we should introduce an enum and switch to introduce option for backward compatibility. (that is the reason why I let some extracted code fragments there) or if this is more confusing. If we decide _not_ to introduce such compatibility, we can clean up a bit more :)
1. For the existing tests I decided to copy the emission factors from cars to (new) HGVs in the table.
Since we will most probably use 0.0 in the future (see #1578), I am unsure if I should better use 0.0 here as well and "correct" the expected results of the junit tests
1. I plan to add some of the cold tables containing (only) the 0.0 factors to the public. So everyone who wants to add such lines to his/her table(s) can just copy/paste it. I would store it in [https://github.com/matsim-org/matsim-libs/tree/master/examples/scenarios/emissions-sampleScenario](https://github.com/matsim-org/matsim-libs/tree/master/examples/scenarios/emissions-sampleScenario) if there is no opposite voting here.


kainagel (approved):
Thank you so much for that "clean" approach.  I am very happy with the general direction.  I have no intuition re the additional questions, sorry.

kt86:
> Further notes / discussion:
> 
>     1. I am wondering if we should introduce an enum and switch to introduce option for backward compatibility. (that is the reason why I let some extracted code fragments there) or if this is more confusing. If we decide _not_ to introduce such compatibility, we can clean up a bit more :)
Have not done this. If needed, someone has to add this later :) For being able to do this more easily, I let some part of the code as it is now.

>     2. For the existing tests I decided to copy the emission factors from cars to (new) HGVs in the table.
>        Since we will most probably use 0.0 in the future (see ...), I am unsure if I should better use 0.0 here as well and >"correct" the expected results of the junit tests
I have adapted the tests for the 0.0 values

>     3.  I plan to add some of the cold tables containing (only) the 0.0 factors to the public. So everyone who wants to add such lines to his/her table(s) can just copy/paste it. I would store it in https://github.com/matsim-org/matsim-libs/tree/master/examples/scenarios/emissions-sampleScenario if there is no opposite voting here.
I have add the table (coldTableExcept_LCV_PassCar_AllZero.csv) and wrote a small readme.

kt86:
This branch should stay available.
@tschlenther, @GregorRyb : Please delete this branch, when you do not it any longer.

PR #1655, opened 2021-09-03
fix: stop re-using AgentSnapshotInfo builder to ensure thread safety

michalmac:
I have recently noticed that `TestPositionEmissionModule` randomly fails (at least on Window machines: https://ci.appveyor.com/project/michalmac/matsim-libs/builds/40571932)

The problem is that not always the same `positionEmission` events are produced (resulting in different emission stats). This, in turn, is caused by `position` events (of agents travelling inside vehicles) not always having `vehicleId` set (for such events, no `positionEmission` is generated). After digging in a bit, it turned out that there are some race conditions when building `AgentSnapshotInfo`s.


michalmac:
@Janekdererste I see you are working on this branch: https://github.com/matsim-org/matsim-libs/tree/fix_position_emission_module. Perhaps this PR already fixes the problem?

Janekdererste:
Ah, thanks for taking this on. I noticed the race condition and reverted back to non-parallel snapshot generation, since the bottleneck of my computation is elsewhere anyway. 

Still this problem should be very well paralizable, so great, that you are doing it.

PR #1656, opened 2021-09-06
Include stop task (pick up and drop off) in the drt stopping task writer

luchengqi7:
In the visualisation (VIA), the vehicle will disappear when pick up or drop off passenger. The DRT vehicle writer will now also write the stop task (pick up and drop off) and that can be used by the visualisation.


michalmac (commented):
LGTM.

The only things that comes to my mind is that maybe in future we want to filter by (dvrp) mode. But that can be added when needed.

PR #1669, opened 2021-09-16
Enable shp options to read shapefile from URL

luchengqi7:
Previously the shp option cannot read shapefile on SVN properly (because it is an URL instead of a file). With this new if statement, the URL shapefile should be able to read.


rakow:
This won't work. An `URL` can not be passed as a `Path` easily.

Issue #1521, opened 2021-06-02
matsim 13 and JDK 16


michalmac:
Since we are receiving many issues that relate to this problem, perhaps we should do something about it.

Potential actions to consider:
- fix-release Matsim 13.1 to support JDK 16 (cherry-picking #1498). Matsim 13 will live for a while and before we release Matsim 14, JDK 17 (and 18?) will become available (so the problem will grow).
- switch from Java LTS JDK (11, 17,...) to latest JDK in CI. Either for all CIs (but keeping Java version fixed at the lastest LTS) or selected CIs (like AppVeyor where we already test against the windows platform)
- increase the frequency of major releases (with GitHub Actions there are way fewer manual steps around releasing matsim).

Any opinions?

Janekdererste:
> fix-release Matsim 13.1 to support JDK 16 

Fixing support for the latest Java Version would be nice, since selecting the correct version other than the newest available is difficult for people who are new to Java. We usually release a x.1 Version at the end of the summer semester. We could include it there. Or would you like to make a release before that date?

> switch from Java LTS JDK (11, 17,...) to latest JDK in CI

I guess this would be benefitial to detect incompatibilities. On the other hand this will also block everyone from merging things as long as eventual incompatibilities persist when the latest SDK changes. So, I think setting the Build-SDK manually and fixing possible errors would be a good option instead of automatically using the latest available one.

michalmac:
> > fix-release Matsim 13.1 to support JDK 16
> 
> Fixing support for the latest Java Version would be nice, since selecting the correct version other than the newest available is difficult for people who are new to Java. We usually release a x.1 Version at the end of the summer semester. We could include it there. Or would you like to make a release before that date?

Nice. Nothing super urgent, so let's stick to the standard scheme for release 13.1.

> > switch from Java LTS JDK (11, 17,...) to latest JDK in CI
> 
> I guess this would be benefitial to detect incompatibilities. On the other hand this will also block everyone from merging things as long as eventual incompatibilities persist when the latest SDK changes. So, I think setting the Build-SDK manually and fixing possible errors would be a good option instead of automatically using the latest available one.

I was a bit imprecise. I am also in favour of manually bumping the JDK version in the build config. I was mainly suggesting switching from the LTS sequence (11, 17, ...) to the more fine-grained and frequent increments (11, 12, 13, 14,...).

Janekdererste:
[...]

PR #1671, opened 2021-09-16
CI: switch to JDK 17 (zulu distro)

michalmac:
Changing to `zulu` as `adopt` is not yet available. Let's see if we are (almost) ready for JDK 17.


michalmac:
Looks like only one test had to be fixed and we are JDK17 ready.
As for `zulu` - It's TCK compliant (while `adopt` is not). Azure uses Zulu for Java. So the change of the distro is fine.

I will merge this PR according to what we decided here: https://github.com/matsim-org/matsim-libs/issues/1521 (i.e. always use the latest JDK in GitHub Actions).

michalmac:
As for the Java language level  - we are still at 11. No changes here.

PR #1674, opened 2021-09-17
appveyor: change jdk from 11 to 16

michalmac:
(currently latest on appveyor)

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -10,7 +10,7 @@ clone_depth: 5
 
 environment:
   matrix:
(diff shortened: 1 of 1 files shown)


Janekdererste:
Just out of interest: Why didn't you change to jdk17?

michalmac:
They do not support 17, so 16 is the latest. I will bump to 17 one it's there.

Here is the list of supported JDKs: https://www.appveyor.com/docs/windows-images-software/#java

Janekdererste:
That makes sense

PR #1679, opened 2021-09-21
Update IncomeDependentUtilityOfMoneyPersonScoringParameters.java

luchengqi7:
There is some problem casting String into Double in the original way. I have made some changes. I don't know if there is other way to do this.

Also I have made some minor changes based on IntelliJ's suggestion. 

ClassCastException: class String cannot be cast to class Double (String and Double are in module java.base of loader 'bootstrap')

### DIFF
--- a/contribs/vsp/src/main/java/playground/vsp/scoring/IncomeDependentUtilityOfMoneyPersonScoringParameters.java
+++ b/contribs/vsp/src/main/java/playground/vsp/scoring/IncomeDependentUtilityOfMoneyPersonScoringParameters.java
@@ -25,6 +25,7 @@ import org.matsim.api.core.v01.population.Person;
 import org.matsim.api.core.v01.population.Population;
 import org.matsim.core.config.groups.PlanCalcScoreConfigGroup;
 import org.matsim.core.config.groups.ScenarioConfigGroup;
+import org.matsim.core.population.PersonUtils;
 import org.matsim.core.population.PopulationUtils;
 import org.matsim.core.scoring.functions.ActivityUtilityParameters;
 import org.matsim.core.scoring.functions.ScoringParameters;
@@ -77,7 +78,7 @@ public class IncomeDependentUtilityOfMoneyPersonScoringParameters implements Sco
 				"Income values <= 0 are ignored. Agents that have negative or 0 income will use the marginalUtilityOfMOney in their subpopulation's scoring params..");
 		OptionalDouble averageIncome =  population.getPersons().values().stream()
 				.filter(person -> person.getAttributes().getAttribute(PERSONAL_INCOME_ATTRIBUTE_NAME) != null) //consider only agents that have a specific income provided
-				.mapToDouble(person -> Double.parseDouble((String) Objects.requireNonNull(person.getAttributes().getAttribute(PERSONAL_INCOME_ATTRIBUTE_NAME))))
+				.mapToDouble(PersonUtils::getIncome)
 				.filter(dd -> dd > 0)
 				.average();
 
@@ -133,7 +134,7 @@ public class IncomeDependentUtilityOfMoneyPersonScoringParameters implements Sco
 
 			if (person.getAttributes().getAttribute(PERSONAL_INCOME_ATTRIBUTE_NAME) != null){
 				//here is where we put person-specific stuff
-				double personalIncome = Double.parseDouble((String) Objects.requireNonNull(person.getAttributes().getAttribute(PERSONAL_INCOME_ATTRIBUTE_NAME)));
+				double personalIncome = PersonUtils.getIncome(person);
 				if(personalIncome > 0){
 					builder.setMarginalUtilityOfMoney(subpopulationScoringParams.getMarginalUtilityOfMoney()  * globalAvgIncome / personalIncome);
 				} else {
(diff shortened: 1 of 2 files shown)


contribs/vsp/src/main/java/playground/vsp/scoring/IncomeDependentUtilityOfMoneyPersonScoringParameters.java
```diff
@@ -80,11 +77,11 @@ private double computeAvgIncome(Population population) {
 				"Income values <= 0 are ignored. Agents that have negative or 0 income will use the marginalUtilityOfMOney in their subpopulation's scoring params..");
 		OptionalDouble averageIncome =  population.getPersons().values().stream()
 				.filter(person -> person.getAttributes().getAttribute(PERSONAL_INCOME_ATTRIBUTE_NAME) != null) //consider only agents that have a specific income provided
-				.mapToDouble(person -> (double) person.getAttributes().getAttribute(PERSONAL_INCOME_ATTRIBUTE_NAME))
+				.mapToDouble(person -> Double.parseDouble((String) Objects.requireNonNull(person.getAttributes().getAttribute(PERSONAL_INCOME_ATTRIBUTE_NAME))))
```
michalmac, line 80: I think you do not need to check `Objects.requireNonNull` for the `PERSONAL_INCOME_ATTRIBUTE_NAME` attribute as one line above you are actually filtering out persons without this attribute.

contribs/vsp/src/main/java/playground/vsp/scoring/IncomeDependentUtilityOfMoneyPersonScoringParameters.java
```diff
@@ -136,7 +133,7 @@ public ScoringParameters getScoringParameters(Person person) {
 
 			if (person.getAttributes().getAttribute(PERSONAL_INCOME_ATTRIBUTE_NAME) != null){
 				//here is where we put person-specific stuff
-				double personalIncome = (double) person.getAttributes().getAttribute(PERSONAL_INCOME_ATTRIBUTE_NAME);
+				double personalIncome = Double.parseDouble((String) Objects.requireNonNull(person.getAttributes().getAttribute(PERSONAL_INCOME_ATTRIBUTE_NAME)));
```
michalmac, line 136: Ditto

tschlenther:
great thanks.

actually, this class sometimes led to problems when run on several threads (e.g. triggered by PersonPrepareForSim). This is because the map is not a concurrent map, i believe. Thus the call `this.activityParamsPerSubpopulation.computeIfAbsent` was problematic. As we are improving this class now, can you @rakow maybe help and make it thread-safe?
If you find no time, i think it is okay to put it on  a mid-term to do list. Thank you :)

rakow:
The tests are failing now:
```
java.lang.ClassCastException: class java.lang.Double cannot be cast to class java.lang.String (java.lang.Double and java.lang.String are in module java.base of loader 'bootstrap')
	at playground.vsp.scoring.IncomeDependentUtilityOfMoneyPersonScoringParametersTest.setUp(IncomeDependentUtilityOfMoneyPersonScoringParametersTest.java:91)
````

So this value can either be String or Double ?

**Comment by luchengqi7** [comment]
I just realized that this is Double in the Berlin scenario. It may be different in different scenario...

**Comment by luchengqi7** [comment]
For the Kelheim scenario, I think I can change the person attribute also to double so that this one does not need to be changed.

File: contribs/vsp/src/main/java/playground/vsp/scoring/IncomeDependentUtilityOfMoneyPersonScoringParameters.java (hunk 1 above)
Comment (luchengqi7) [reply contribs/vsp/src/main/java/playground/vsp/scoring/IncomeDependentUtilityOfMoneyPersonScoringParameters.java:80]: Yeah, it's suggested by the IntelliJ. I can remove that.

**Comment by kainagel** [comment]
Thank you for working on this.  Three things:

(1) Could you please, in such "standardized" cases, not use person.getAttributes() directly.  Instead, design two static functions, one getting and one for setting the attribute.  Look into PersonUtils for examples.

(2) The attribute is somewhat typed ... as one can see when looking into the plans file.  In consequence, if you say

  ... = (xyz) person.getAttributes().getAttribute(...)

and the attribute is not of type xyz in the plans file, you get a runtime exception.   Once you pass this through anything else before the cast (e.g. notNull), then this protection gets lost.

(3) I think that the personal income attribute is a double.  In consequence, the design should be

```
... class PersonUtils ... {

   ... getPersonalIncome( person ) {
      return (double) person.getAttributes().get( PERSONAL_INCOME_ATTRIBUTE_NAME );
   }
   ... setPersonalIncome( person, income ) {
      person.getAttributes().putAttribute( PERSONAL_INCOME_ATTRIBUTE_NAME, income );
   }
}
```

Could you please talk to Christian in case of questions?  If you cannot sort it out together, please talk to me.

Thanks ...

luchengqi7:
> Thank you for working on this. Three things: (1) Could you please, in such "standardized" cases, not use person.getAttributes() directly. Instead, design two static functions, one getting and one for setting the attribute. Look into PersonUtils for examples. (2) The attribute is somewhat typed ... as one can see when looking into the plans file. In consequence, if you say ... = (xyz) person.getAttributes().getAttribute(...) and the attribute is not of type xyz in the plans file, you get a runtime exception. Once you pass this through anything else before the cast (e.g. notNull), then this protection gets lost. (3) I think that the personal income attribute is a double. In consequence, the design should be ``` ... class PersonUtils ... { ... getPersonalIncome( person ) { return (double) person.getAttributes().get( PERSONAL_INCOME_ATTRIBUTE_NAME ); } ... setPersonalIncome( person, income ) { person.getAttributes().putAttribute( PERSONAL_INCOME_ATTRIBUTE_NAME, income ); } } ``` Could you please talk to Christian in case of questions? If you cannot sort it out together, please talk to me. Thanks ...

Okay, we will work on it. Thanks for the feedback!

PR #1682, opened 2021-09-22
Adding PT fare module

luchengqi7:
The PT fare module calculate PT fare in a more reasonalbe way than the current implementation. 

So far, the distance based PT fare calcualtion is implemented. This method will take the Euclideance distance between the departure PT stop and the arrival PT stop of **a whole PT leg** (i.e. the full leg between 2 real activities) to calculate the fare. Therefore, it is irrelevant to how many times a agent has change line in this leg and the extra distance caused by line transfer). This also make the fare calcualation for inter-modal trip (e.g. PT + DRT) easier. 

Besides, the upper bound mechanism is also introduced in this module. The upper bound mechanism can simulate the PT ticket subscription. 
- For an agent only make few trips, we calculate the fare based on each leg. 
- For agent who make many trips per day, we will assume that agent has a ticket subscription and we will use the longest trip that agent makes in the day to determine the price of the ticket subsription. We will still collect fare after each leg. But we will issue refund to that agent at the end of iteration so that he/she only pays for the daily price of the ticket subsription. 

The other fare calculation method can also be easily added. (e.g. zone based fare calculation in Stuttgart). Currently, it is not yet added to the config file. If everyone is happy with this structure, then we can move on to make it configuratble in the config file (just like those vsp experimental modules).


contribs/vsp/src/main/java/playground/vsp/pt/fare/DistanceBasedPtFareHandler.java
```diff
@@ -0,0 +1,86 @@
...
+    @Override
+    public void handleEvent(ActivityStartEvent event) {
+        if (!StageActivityTypeIdentifier.isStageActivity(event.getActType())){
+            Id<Person> personId = event.getPersonId();
+            if (personDepartureMap.containsKey(personId)){
+                if (!personArrivalMap.containsKey(personId)){
+                    throw new RuntimeException("A person has departed but not arrived. This should not happen! " +
+                            "Please make sure the event is read/handled in the normal sequence");
+                }
+                Link fromLink = network.getLinks().get(personDepartureMap.get(personId));
```
vsp-gleich, line 65: Thanks for your work on that topic! The current version here seems as if it essentially only considers the last pt leg. What we want is the pt trip, i.e. the first boarding stop and the last arrival stop. This avoids the problem I explained with direct train being cheaper than a route using a change of trains (the transfer station makes for one more leg) although passing over the exact same tracks and having the same fare.

contribs/vsp/src/main/java/playground/vsp/pt/fare/DistanceBasedPtFareHandler.java
```diff
@@ -0,0 +1,86 @@
...
+        this.distanceFare = params.getDistanceFare();
+        this.network = network;
+    }
+
+    @Override
+    public void handleEvent(PersonDepartureEvent event) {
+        if (event.getLegMode().equals(TransportMode.pt)){
+            if (!personDepartureMap.containsKey(event.getPersonId())){
+                personDepartureMap.put(event.getPersonId(), event.getLinkId()); // Departure place is fixed
+            }
```
vsp-gleich, line 45: Maybe it would be nicer to use the transit stop coord instead of link.getToNode() since the stop coord can differ from the link toNode. Same for the arrival stop. This could be done using the pt interaction activities which have the stop coord.

luchengqi7: May I know how the PT stop is recognized in the MATSim? Is it possible to have PT stop not on the link/Node of the network?

michalmac (approved):
From my side (DRT), everything looks good. Thanks for fixing the typo.

PR #1694, opened 2021-09-30
Sharing contrib

sebhoerl:
This PR migrates the sharing contrib from https://github.com/eqasim-org/matsim-sharing which allows to wrap any mode in MATSim into a sharing service, be it network- or teleportation-simulated modes. The contrib provides the tools to define stations and vehicles belonging to the service and support station-based and freefloating services. See the README file in the contrib folder for more information.

Not sure who has to review this :) 

@balacmi @tschlenther


tschlenther:
Thank you really much!

I quickly went over the changes and have no major change requests, as it seems to mirror the version in eqasim more or less. 
However, I have the following remarks:

1. One has to be careful when configuring the SharingModule _AND_ other dvrp modules, such as drt for example. The problem is that `SharingUtils.configureQSim()` overrides the binding of all dvrp modes, such that drt leads to an exception. If one does not take care, it can also happen that the sharing modes are overridden and they ae silently simulated as 'normal, underlying' modes w/o the sharing-logic, i.e. without the vehicle booking etc.
So it best to use `DvrpQSimComponents.activateModes()` and configure all dvrp modes oneself after binding all the modules (remember, this only is required in cases where the sharing contrib is combined with other dvrp-dependent contribs).
Maybe we should put this into the readme or provide another example.

2. It would be nice if the SharingServiceCOnfigGroups could be read from xml files. This way, one could read in the output config without problems (e.g. when chaining some runs).

I think that both 1. and 2. are no hard requirements that should stop this PR and could be met later.
I'd have time in November to contribute again. So please go ahead and merge when you think this is ready.

Thanks again :)

sebhoerl:
Waiting for another main branch update to merge it in again, I think setting the old "automerge" tag messed soemthing up with the status checks.

PR #1696, opened 2021-10-01
Make use of interfaces for NetworkBasedTransportCosts

steffenaxer:
This PR allows to replace NetworkBasedTransportCosts with custom implementations as they are now using an interface named `TransportCosts` which extends previous `VehicleRoutingTransportCosts` and provides three additional methods. Those three additional methods have been unsed internally in the previous NetworkBasedTransportCosts, but their implementation have neven been enforced by an interface.

### DIFF
--- a/contribs/freight/src/main/java/org/matsim/contrib/freight/jsprit/NetworkRouter.java
+++ b/contribs/freight/src/main/java/org/matsim/contrib/freight/jsprit/NetworkRouter.java
@@ -31,8 +31,8 @@ public class NetworkRouter {
 	 * @param {@link CarrierPlan}
 	 * @param {@link NetworkBasedTransportCosts}
 	 */
-	public static void routePlan(CarrierPlan plan, NetworkBasedTransportCosts netbasedTransportCosts){
-		new TimeAndSpacePlanRouter(netbasedTransportCosts.getRouter(), netbasedTransportCosts.getNetwork(), netbasedTransportCosts.getTravelTime()).run(plan);
+	public static void routePlan(CarrierPlan plan, TransportCosts freightTransportCosts){
+		new TimeAndSpacePlanRouter(freightTransportCosts.getRouter(), freightTransportCosts.getNetwork(), freightTransportCosts.getTravelTime()).run(plan);
 	}
 
 }
--- a/contribs/freight/src/main/java/org/matsim/contrib/freight/jsprit/TransportCosts.java
+++ b/contribs/freight/src/main/java/org/matsim/contrib/freight/jsprit/TransportCosts.java
@@ -1,0 +1,15 @@
+package org.matsim.contrib.freight.jsprit;
+
+import com.graphhopper.jsprit.core.problem.cost.VehicleRoutingTransportCosts;
+import org.matsim.api.core.v01.network.Network;
+import org.matsim.core.router.util.LeastCostPathCalculator;
+import org.matsim.core.router.util.TravelTime;
+
+/**
+ * @author: Steffen Axer
+ */
+public interface TransportCosts extends VehicleRoutingTransportCosts {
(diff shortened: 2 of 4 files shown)


tschlenther:
looks okay to me on a first glance

contribs/freight/src/main/java/org/matsim/contrib/freight/jsprit/NetworkRouter.java
```diff
@@ -31,8 +31,8 @@
 	 * @param {@link CarrierPlan}
 	 * @param {@link NetworkBasedTransportCosts}
 	 */
-	public static void routePlan(CarrierPlan plan, NetworkBasedTransportCosts netbasedTransportCosts){
-		new TimeAndSpacePlanRouter(netbasedTransportCosts.getRouter(), netbasedTransportCosts.getNetwork(), netbasedTransportCosts.getTravelTime()).run(plan);
+	public static void routePlan(CarrierPlan plan, TransportCosts freightTransportCosts){
+		new TimeAndSpacePlanRouter(freightTransportCosts.getRouter(), freightTransportCosts.getNetwork(), freightTransportCosts.getTravelTime()).run(plan);
```
kt86, line 35: This is currently the only place where the (package-private) `.getRouter()` is used.
`getNetwork() `and `getTravelTime()` could at least inside MATSim be package-private as well since this here is also the signle location, where it is used. but I do not know if someone outside will used it.

contribs/freight/src/main/java/org/matsim/contrib/freight/jsprit/TransportCosts.java
```diff
@@ -0,0 +1,15 @@
...
+
+import com.graphhopper.jsprit.core.problem.cost.VehicleRoutingTransportCosts;
+import org.matsim.api.core.v01.network.Network;
+import org.matsim.core.router.util.LeastCostPathCalculator;
+import org.matsim.core.router.util.TravelTime;
+
+/**
+ * @author: Steffen Axer
+ */
+public interface TransportCosts extends VehicleRoutingTransportCosts {
```
kt86, line 11: Do we have a better name for it? Please have a look to my comment in the PR conversation.

kt86:
Hello @steffenaxer ,

I like the idea, to open the `NetworkBasedTransportCosts` for other implementations! So thank you for working on it!

The issue is - as usual - that I do not have the time for diving in deeper.

Few quick comments / ideas : 
1. Do we find a better name for the interface? `TransportCosts` is for me a bit to general for being here somewhere in the freight contrib. --> The problem is, that the current class has somehow the nearly best name I have now in mind :( -> `VRPTransportCosts`?
2. I am a bit unsure, if we should put the previous internal methods into a (public) interface or if there is maybe the possibility to reduce / building them back. Jsprit does not need them, so I am not totally convinced if we really need them (enforced) in any of our (MATSim) implementation(s). -> see also my comment in the code.
Especially for the `getRouter(),` which is currently only package-private, one time used in an Construc and will redirect to some getOrCreate-like method. As far as I can see it would currently also be an option to make `getNetwork()` and `getTravelTime()` package-private. 
3. Another (further?) step would maybe, that one extracts the internal classes (e.g. the `TransportData`, `BaseVahicleTransportCosts`,..)  from the `NetBasedTransportCosts`. Then it would be a bit more obvious which one can be (re)used by other implementations. But it may depends on the part of your own NetworkBasedTransportCosts implementation and what parts of it you would like to (ex)change. :)

Nevertheless I will approve this PR, so the person who want's to merge  it can do it - hopefully after targeting at least 1.) and maybe thinking about 2.) - without waiting for me.

kt86 (approved):
Please have a look to my comments / ideas. 
I will nevertheless approve it for not blocking the process.

PR #1698, opened 2021-10-02
Change visibility to allow code reusage instead of code copy

steffenaxer:
This PR enables users to create custom JointDemandModule by reusing existing code instead of copying code. Switched some classes and static methods from package exclusive visibility to public. (@tschlenther, @mrieser, @kt86)

### DIFF
--- a/contribs/commercialTrafficApplications/src/main/java/org/matsim/contrib/commercialTrafficApplications/jointDemand/commercialJob/ChangeCommercialJobOperator.java
+++ b/contribs/commercialTrafficApplications/src/main/java/org/matsim/contrib/commercialTrafficApplications/jointDemand/commercialJob/ChangeCommercialJobOperator.java
@@ -45,7 +45,7 @@ public final class ChangeCommercialJobOperator extends AbstractMultithreadedModu
 
     private final Carriers carriers;
 
-	ChangeCommercialJobOperator(GlobalConfigGroup globalConfigGroup, Carriers carriers) {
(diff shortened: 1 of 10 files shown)


tschlenther:
This basically sets all classes in the commercialJob package to public.
I am not a 100% certain if that is what would be our preferred design. 
If you want to replace one or two bindings made by the JointDemandModule (e.g. custom CommercialJobGenerator), then you could do that in a custom controler module that is added to the controler as overriding module after the JointDemandModule.

Another question that pops up in my mind is whether we should rename the commercialJob package?

kt86:
I agree with @tschlenther that we should be careful in opening the accessibility to public. This is maybe not an issue in the short term, but for mid- and long-term maintenance it is often not helpful. So, if there are other ways of doing it, I would prefer that :)

Regarding the renaming: Yes, I agree, that the name is not very intuitive and that now is still a good time to change it. Does anyone of you has an idea?

tschlenther:
> I agree with @tschlenther that we should be careful in opening the accessibility to public. This is maybe not an issue in the short term, but for mid- and long-term maintenance it is often not helpful. So, if there are other ways of doing it, I would prefer that :)
>
we basically agreed to go the mild way and only put the interfaces public, see #1701. Thus, i will close this PR.

> Regarding the renaming: Yes, I agree, that the name is not very intuitive and that now is still a good time to change it. Does anyone of you has an idea?

we could move everything one package up into jointDemand package and instead put the run examples into a subpackage named run or examples. i will do that in a separate PR.

PR #1738, opened 2021-10-28
Include fine grained cold position emission events

Janekdererste:
Cold PositionEmissionEvents were taken from the emission module as is. This was not fine grained enough for the targeted resolution of PALM simulations.

We take the cold emission events from the emission module and distribute them over the PositionEmissionEvents for the cold start phase of a trip now.

Instead of issuing cold and warm PositionEmissionEvents we issue a combined event


contribs/emissions/src/main/java/org/matsim/contrib/emissions/PositionEmissionsModule.java
```diff
@@ -189,31 +195,33 @@ private void handlePositionEvent(PositionEvent event) {
...
+			if (distance > 2000) {
+				vehiclesEmittingColdEmissions.remove(vehicle.getId());
+				return emissionCalculator.getEmptyColdEmissions();
 			}
+
+			// HBEFA assumes a constantly decreasing ammount of cold emissions depending on the distance travelled
+			// the underlying emission module simplifies this into two steps. Between 0-1km and 1-2km. We use the same
+			// classes here because we don't want to rewrite all the stuff. The cold emission module computes emissions
+			// for 1000m. We take these as is and muliply with distanceToLastPosition / 1000. This way we have the fraction
+			// of cold emissions for the distance travelled during the last time step janek oct' 2021
```
kt86, line 217: Comment for the comment: HBEFA provides the data in these two bins: 01-km and 1-2 km. So it is not our decision to use these classes ;)

contribs/emissions/src/main/java/org/matsim/contrib/emissions/WarmEmissionAnalysisModule.java
```diff
@@ -307,22 +304,19 @@ void throwWarmEmissionEvent(double leaveTime, Id<Link> linkId, Id<Vehicle> vehic
...
 				// compute faction.  This cannot be done earlier since efkey.component is needed.
 				fractionStopGo = getFractionStopAndGo(freeVelocity_ms * 3.6, averageSpeed_kmh, vehicleInformationTuple, efkey);
-				logger.info("fractionStopGo is: " + fractionStopGo);
 
 				double efStopGo_gpkm = 0.;
 				if (fractionStopGo > 0) {
 					// compute emissions from stop-go fraction:
 					efkey.setTrafficSituation(STOPANDGO);
 					efStopGo_gpkm = getEf(vehicleInformationTuple, efkey).getFactor();
-					logger.warn("pollutant=" + warmPollutant + "; efStopGo=" + efStopGo_gpkm);
```
kt86, line 317: I would prefer setting it to logger.debug(...) instead of removing it.
Then it is available if one would like to have it.

contribs/emissions/src/main/java/org/matsim/contrib/emissions/WarmEmissionAnalysisModule.java
```diff
@@ -307,22 +304,19 @@ void throwWarmEmissionEvent(double leaveTime, Id<Link> linkId, Id<Vehicle> vehic
...
 				}
 
 				double efFreeFlow_gpkm = 0. ;
 				if ( fractionStopGo<1.) {
 					// compute emissions for free-flow fraction:
 					efkey.setTrafficSituation(FREEFLOW);
 					efFreeFlow_gpkm = getEf(vehicleInformationTuple, efkey).getFactor();
-					logger.warn("pollutant=" + warmPollutant + "; efFreeFlow=" + efFreeFlow_gpkm);
```
kt86, line 325: Same here: I would prefer setting it to logger.debug(...) instead of removing it.

contribs/emissions/src/main/java/org/matsim/contrib/emissions/WarmEmissionAnalysisModule.java
```diff
@@ -105,38 +105,36 @@ public WarmEmissionAnalysisModule(
 					// If this causes too many problems, we could insert a switch (or attach it to the fallback behavior switch).  kai, feb'20
 					// Eventually vehicle category and vehicle attribute should be alligned in order to make the allCombinations setting useful
 					// see discussion in  https://github.com/matsim-org/matsim-libs/issues/1226 kturner, nov'20
-					if (detailedHbefaWarmTable != null) {
```
kt86, line 108: is it ensured before, that it is not null? If yes, it is ok for removing. If not I would let it there (Or do you have changed here anything else in this block, what I do not see in the github diff view?)

kt86 (approved):
Introduction of the more "general" EmissionEvents is fine for me. Thanks for doing it and reducing some duplicated code!

contribs/emissions/src/main/java/org/matsim/contrib/emissions/PositionEmissionsModule.java
Janekdererste: Nice. I read their documentation but obviously missed the part where they described the abstraction of two distance classes. I guess this is even better since one doesn't have to discuss this simplification since it happened upstream.

contribs/emissions/src/main/java/org/matsim/contrib/emissions/WarmEmissionAnalysisModule.java
Janekdererste: Okay, I can do that. I find the emission contrib's logging quite verbose overall anyway, though.

contribs/emissions/src/main/java/org/matsim/contrib/emissions/WarmEmissionAnalysisModule.java
Janekdererste: Will change it.

contribs/emissions/src/main/java/org/matsim/contrib/emissions/WarmEmissionAnalysisModule.java
Janekdererste: This is tested before already and was therefore superfluous.

PR #1742, opened 2021-10-29
implement proposed BatteryCharging method for energy estimation charged in a fixed period

nkuehnel:


contribs/ev/src/main/java/org/matsim/contrib/ev/charging/FastThenSlowCharging.java
```diff
@@ -54,6 +54,47 @@ public double calcChargingPower(double maxPower) {
 		}
 	}
 
+	@Override
+	public double calcEnergyCharge(ChargerSpecification charger, double chargePeriod) {
+		double remainingTime = chargePeriod;
+
+		double maxChargingPower = charger.getPlugPower();
+		double powerCharged = 0;
```
michalmac, line 62: I guess you mean `energyCharged`

contribs/ev/src/main/java/org/matsim/contrib/ev/charging/FastThenSlowCharging.java
```diff
@@ -54,6 +54,47 @@ public double calcChargingPower(double maxPower) {
 		}
 	}
 
+	@Override
+	public double calcEnergyCharge(ChargerSpecification charger, double chargePeriod) {
```
michalmac, line 58: This class does a bit more than a simple mathematical operation, so please add some tests for this methods (there are some tests for the other method, maybe you can base on them).

michalmac (approved):
Thanks! LGTM

Added a few comments to address.

contribs/ev/src/main/java/org/matsim/contrib/ev/charging/FastThenSlowCharging.java
nkuehnel: fixed in 4dc3a81

contribs/ev/src/main/java/org/matsim/contrib/ev/charging/FastThenSlowCharging.java
nkuehnel: You are right! I added a test (and indeed there were som ebugs ;) ) thanks! fixed in 672afbf

PR #1748, opened 2021-11-04
config: accept only "true"/"false" strings as boolean values

michalmac:
Otherwise all non-"true" values are considered "false".

Like here: https://github.com/matsim-scenarios/matsim-berlin/blob/5.5.x/scenarios/berlin-v5.5-1pct/input/drt/berlin-drt-v5.5-1pct.config.xml#L195. The intention was to actually apply the filtering to the "drt" mode (so the value was meant to be "true", but currently no filtering is done).

### DIFF
--- a/contribs/parking/src/main/resources/parkingsearch/config.xml
+++ b/contribs/parking/src/main/resources/parkingsearch/config.xml
@@ -31,7 +31,7 @@
 		<param name="flowCapacityFactor" value="1.0"/>
 		<param name="storageCapacityFactor" value="1.0"/>
 		<param name="stuckTime" value="10"/>
-		<param name="removeStuckVehicles" value="yes"/>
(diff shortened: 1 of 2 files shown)


contribs/parking/src/main/resources/parkingsearch/config.xml
```diff
@@ -31,7 +31,7 @@
 		<param name="flowCapacityFactor" value="1.0"/>
 		<param name="storageCapacityFactor" value="1.0"/>
 		<param name="stuckTime" value="10"/>
-		<param name="removeStuckVehicles" value="yes"/>
+		<param name="removeStuckVehicles" value="false"/>
```
michalmac, line 34: This is a different example: the aim was to have the removal enabled, but actually you cannot remove stuck vehicles in DVRP, so "yes" worked properly (i.e. as "false")

PR #1756, opened 2021-11-10
Configurable mode for travel time validation

rakow:
The TravelTimeValidator now accepts two mode options: One for the mode of the MATSim trips to validate and one for the routing mode to be used in the API.
E.g. this allows to also validate bike trips.

### DIFF
--- a/contribs/analysis/src/main/java/org/matsim/contrib/analysis/vsp/traveltimedistance/GoogleMapRouteValidator.java
+++ b/contribs/analysis/src/main/java/org/matsim/contrib/analysis/vsp/traveltimedistance/GoogleMapRouteValidator.java
@@ -11,7 +11,6 @@ import org.matsim.core.utils.collections.Tuple;
 import org.matsim.core.utils.geometry.CoordinateTransformation;
 
 import java.io.BufferedReader;
-import java.io.File;
 import java.io.IOException;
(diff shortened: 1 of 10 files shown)


luchengqi7 (approved):
Maybe we can change the default date of the HERE API to a working day? (TravelTimeAnalysis, line 174)

rakow:
Why is not using the configured date, but hardcoded ?

luchengqi7:
> Why is not using the configured date, but hardcoded ?

I mean the default value. When the user forget to input the data then default value will be used. And for HERE API it would be nice to have a working day perhaps? But it's not that important...

rakow:
Ok, I have changed it. But in the Line you mentioned it was not using the user input but a fixed value.

luchengqi7:
> Ok, I have changed it. But in the Line you mentioned it was not using the user input but a fixed value.

Oh, I see. Emm, it seems that something was changed... Because for Google Map, we can only use a date in the future (therefore, I remember I set the default date to the same day in the next week). For HERE API, we can only use a date in the past. So A fixed date in the past is used (but I think I set it to some other date instead of 20210101...) Perhapse we can change the default date to the same day in the last week... 

I can do that.

rakow:
Ok, I didn't know that. I changed the default now to depend on the API which was selected.

PR #1766, opened 2021-11-23
Added Congestion analysis

rakow:
Added the implementation of "A Traffic Congestion Assessment Method for Urban Road Networks Based on Speed Performance Index" from @luchengqi7, into the existing link stats analysis class.

### DIFF
--- a/contribs/application/src/main/java/org/matsim/application/analysis/LinkStats.java
+++ b/contribs/application/src/main/java/org/matsim/application/analysis/LinkStats.java
@@ -64,6 +64,10 @@ public class LinkStats implements MATSimAppCommand {
 	@CommandLine.Mixin
 	private CsvOptions csv = new CsvOptions();
 
+	public static void main(String[] args) {
+		new LinkStats().execute(args);
+	}
+
 	@Override
 	public Integer call() throws Exception {
 
--- a/contribs/application/src/main/java/org/matsim/application/analysis/LinkStats.java
+++ b/contribs/application/src/main/java/org/matsim/application/analysis/LinkStats.java
@@ -1,7 +1,10 @@
 package org.matsim.application.analysis;
 
 import com.beust.jcommander.internal.Lists;
-import org.apache.commons.csv.CSVFormat;
+import it.unimi.dsi.fastutil.doubles.DoubleArrayList;
+import it.unimi.dsi.fastutil.doubles.DoubleList;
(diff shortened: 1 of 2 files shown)


contribs/application/src/main/java/org/matsim/application/analysis/LinkStats.java
```diff
@@ -19,36 +23,50 @@
...
 	private Path network;
 
+	@CommandLine.Option(names = "--events", description = "Input event file", required = true)
+	private Path events;
+
 	@CommandLine.Option(names = "--output", description = "Path for output csv", required = true)
 	private Path output;
 
+	@CommandLine.Option(names = "--output-congestion", description = "Path for output csv for congestion", required = true)
+	private Path outputCongestion;
```
luchengqi7, line 47: Perhaps it will be a good idea to use the output folder instead, then we dont need to enter the output path one by one.

Then we can have: 
Path linkStatsOutputPath = Path.of(outputFolder.toString + "/link-stats.tsv");
Path congestionAnalysisOutputPath = Path.of(outputFolder.toString + "/congestion-analysis.tsv");

contribs/application/src/main/java/org/matsim/application/analysis/LinkStats.java
```diff
@@ -19,36 +23,50 @@
...
-	private Path events;
-
 	@CommandLine.Option(names = "--network", description = "Input network file", required = true)
 	private Path network;
 
+	@CommandLine.Option(names = "--events", description = "Input event file", required = true)
+	private Path events;
+
 	@CommandLine.Option(names = "--output", description = "Path for output csv", required = true)
 	private Path output;
```
luchengqi7, line 44: may be change to tsv?

contribs/application/src/main/java/org/matsim/application/analysis/LinkStats.java
rakow: That is a reasonable suggestion and would be a bit easier to use. But one would lose the flexibility to define the file names. Personally, I prefer to have these options that need to be set explicitly.

luchengqi7 (approved):

contribs/application/src/main/java/org/matsim/application/analysis/LinkStats.java
luchengqi7: Okay. That makes sense.

PR #1771, opened 2021-11-26
Fix timing issues in dvrp

### LINKED ISSUE #1743: Timing inconsistencies in VrpPaths (for diversion)
I'm currently porting the dispatcher from Alonso-Mora et al. to MATSim. When implementing the dispatcher, I noticed that some routing capabilities of `dvrp` provide inconsistent predictions of arrival times, even under strict free flow conditions and even when using the `QSimFreeFlowTravelTime` estimator. The code for the unit test can be found at https://github.com/matsim-org/matsim-libs/commit/3e4ad4a9f8829dec865dd5bd34af545215fb7deb.

The experiment in the unit test is as follows: I construct a very simple scenario with a straight line of 8 links with length 1000 m and an odd free flow speed, to make sure that the `QSimFreeFlowTravelTime` works properly and rounding does not have an influence. I then create one DVRP vehicle at L0 and add one task to the schedule, which makes it drive to L8. Every 10 seconds, I obtain the next diversion point for the vehicle from the `OnlineDriveTaskTracker` of the active drive task and perform an updated routing to the final link L8. During the simulation, I track the predicted arrival time when doing the initial routing, after every diversion routing, and I track the final arrival time from the events.

Ideally, all cases should produce the same arrival time, only then the predictions are correct. Currently, the following happens:

- Tracking the event-based arrival time gives an arrival time at 657s (reference value)
- The initial routing predicts an arrival time of 656s (wrong, too quick)
- At every diversion, the new routing gives an arrival time of 658s (wrong, too slow)
- Once the vehicle enters the final link, all remaining diversions give 657s (correct)

Findings for the time being:
- Different than stated in `VrpPaths.createPath`, it seems to take two seconds (one for "enter traffic", one for "leave link") to depart from an activity
- The change of prediction after diversion when entering the last link comes from the fallback in `VrpPaths.createPath` with `considerZeroLengthPath == true`
- The difference between the initial routing and the diversion routing comes from the fact that for the initial routing, `VrpPaths.createPath` considers the time to depart (rightfully so, but maybe faulty). However, when diverting, this buffer should not be considered as, there, the vehicle will just enter the next link fluently after traversing the current one. 

Now, to fix the issue and get correct predictions:
- I think it would make sense to either add a `boolean divertingPath` argument in `VrpPaths.*` which disables the use of the initial slack for departing. Alternatively, we could have a set of `VrpPaths.createPathForDiversion` methods. This fixes the difference between the initial and diversion routing.
- For the standard case, I have the feeling that the `VrpPaths.FIRST_LINK_TT` should be 2s instead of 1s? But here I am not sure, maybe I am missing something in my experimental setup?

sebhoerl:
This PR fixes some timing issues in `dvrp` as discussed in #1743 and tests them in `DiversionTest`. Fixes #1743.

Some tests in `drt` needed to be fixed. But it seems that in both cases waiting times actually decreased, so I guess this is fine @michalmac ?


sebhoerl:
Now I merged in master and the test were failing again, not sure what has changed in between, but I needed to adjust the DRT tests again.

michalmac:
> Now I merged in master and the test were failing again, not sure what has changed in between, but I needed to adjust the DRT tests again.

Nothing has happened in the meantime that could impact the results. The two tests you had to adjust after merging in master were added in #1750 that was after you branched out from master. As a result, they where not on your branch, so you could not have fixed them in the first round.

sebhoerl:
Are you sure? Have a look at the last two passed commits above. In the first of them, I was changing the reference values, but the ones that I changed from don't correspond at all to the example that you have cited above. Then, there was the merge and the tests didn't pass anymore. And if you look at the latest commit, I was changing the reference values again, here the initial ones are the ones shown in the PR that you have linked.

In any case, do you think the values are close enough or we should check even closer? In any case, the differences come from the way how travel times are calculated in terms of initial routing vs. detouring etc. in DRT, I guess. Another issue is that often VrpPaths.getLastLinkTT is used, which is only based on freeflow travel time. But in the actual calculation in VrpPaths I now use the time from the TravelTime object, so the routing is consistent.

michalmac:
> Are you sure? Have a look at the last two passed commits above. In the first of them, I was changing the reference values, but the ones that I changed from don't correspond at all to the example that you have cited above. Then, there was the merge and the tests didn't pass anymore. And if you look at the latest commit, I was changing the reference values again, here the initial ones are the ones shown in the PR that you have linked.

The two tests you had to fix in a27ca2e are: `testRunDrtExampleWithNoRejections_ExtensiveSearch()` and `testRunDrtExampleWithNoRejections_SelectiveSearch()`. Non of them were present in feb4aca. They got into your branch after you merged in master (the middle commit that failed: 6b92f0d)

sebhoerl:
Ah ok, yes, those two commits fix different tests. Though the question remains, you think this is fine, or we should dig deeper? Or otherwise put: It is expected that if we test for exact times, some things will change with the fixes in VrpPaths. Is this ok, or do we add some "backports" into DRT so that the tests pass without changing the numbers, but by biasing the results of VrpPaths, which should now be exact when using it with QSimFreeflowSpeed and using the correct non-diversion or diversion-based functions in VrpPaths :)

michalmac:
> In any case, do you think the values are close enough or we should check even closer? In any case, the differences come from the way how travel times are calculated in terms of initial routing vs. detouring etc. in DRT, I guess. Another issue is that often VrpPaths.getLastLinkTT is used, which is only based on freeflow travel time. But in the actual calculation in VrpPaths I now use the time from the TravelTime object, so the routing is consistent.

The differences in results are acceptable. The scenario used there has congestion, therefore the travel time estimates are not precise, so it's not the best case for analysing the impacts of changing path travel times by 1 second.

I do not see any changes on how the last link TT is calculated. But you said you changed something?

contribs/dvrp/src/main/java/org/matsim/contrib/dvrp/path/VrpPaths.java
```diff
@@ -98,18 +123,25 @@ public static VrpPathWithTravelData createPath(Link fromLink, Link toLink, doubl
 
 		// there is no extra time spent on queuing at the end of the last link
 		links[count + 1] = toLink;
-		linkTT = getLastLinkTT(toLink, currentTime);// as long as we cannot divert from the last link this is okay
```
sebhoerl, line 101: @michalmac Before getLastLinkTT (which only uses freeflow travel time) was used here, now I use the time coming from the TravelTime object

sebhoerl: But I left the getLastLinkTT function there, because otherwise I'd need to change a lot of other code. Basically, I would need to pass down the TravelTime object to the code that creates travel time graphs which involves changing many different classes. Not sure if it is worth it at this point, but we could also do that

sebhoerl: That's the point, if I want to do exactly the same thing, I need to pass `TravelTime` to getLastLinkTT. But this means I need to pass down `TravelTime` down to all the places where getLastLinkTT is used. This requires passing on `TravelTime` through the constructors of quite some classes, so I have the feeling that some resdesign of some classes that create the travel time graphs etc. may be cleaner (and some of the uses of getLastLinkTT were not completely clear to me, I think). So that's why I would deprecate getLastLinkTT for now and look into this in a later step. What do you say?

michalmac:
> Some tests in `drt` needed to be fixed. But it seems that in both cases waiting times actually decreased, so I guess this is fine @michalmac ?

Now, after adding the two test cases, they are a bit worse on average, but I would not draw any conclusions out of it. We would need a bigger free-float scenario. Next time I run Berlin I will check for potential changes, but I do not expect significant changes.

michalmac:
I experimented a bit and moved the diversion point from the start of the next link to the end of the previous link (which may be even the right way of modelling diversions) and adjusted the diversion start time by subtracting 1 second.

A few benefits:
- `createZeroLengthPath()` always creates paths with 0 second duration, whereas in your implementation it's either 0 s (departure) or -1 s (diversion -- a bit hacky)
- diversions and regular departures look more similar (the path always starts just before a node)
- if we simplify and assume we only need 1 second to depart, handling of diversion points and regular departures would be identical in `VrpPaths`. Then this extra 1 second could be addressed elsewhere (and only if needed - for instance in your pooling algorithm).

contribs/dvrp/src/main/java/org/matsim/contrib/dvrp/path/VrpPaths.java
```diff
@@ -38,31 +39,55 @@
...
-				router.calcLeastCostPath(fromLink.getToNode(), toLink.getFromNode(), departureTime + FIRST_LINK_TT,
+		Path path = fromLink == toLink ? null
+				: router.calcLeastCostPath(fromLink.getToNode(), toLink.getFromNode(), departureTime + FIRST_LINK_TT,
 						null, null);
 		return VrpPaths.createPath(fromLink, toLink, departureTime, path, travelTime);
 	}
 
-	public static VrpPathWithTravelData createZeroLengthPath(Link fromTolink, double departureTime) {
-		return new VrpPathWithTravelDataImpl(departureTime, 0, new Link[] { fromTolink }, new double[] { 0 });
+	public static VrpPathWithTravelData calcAndCreatePath(LinkTimePair diversionPoint, Link toLink,
```
michalmac, line 48: It is hard to differentiate which of method is meant for regular departures and which for diversion (the only difference the argument list: `LinkTimePair` vs `Link` and `double`)

Maybe rename to `calcAndCreatePathForDiversion()`?

contribs/dvrp/src/main/java/org/matsim/contrib/dvrp/path/VrpPaths.java
```diff
@@ -38,31 +39,55 @@
...
+	}
+
+	public static VrpPathWithTravelData createZeroLengthPath(Link fromTolink, double departureTime, boolean calculateForDiversion) {
+		// To make sure that arrival time stays correct although no special diversion
+		// will be implemented after calling OnlienDriveTaskTracker.divert
+		double travelTime = calculateForDiversion ? -NODE_TRANSITION_TIME : 0.0;
+		return new VrpPathWithTravelDataImpl(departureTime, travelTime, new Link[] { fromTolink }, new double[] { 0 });
+	}
+
+	public static VrpPathWithTravelData createZeroLengthPath(LinkTimePair diversionPoint) {
```
michalmac, line 63: Rename to `createZeroLengthPathForDiversion()`?

contribs/dvrp/src/main/java/org/matsim/contrib/dvrp/path/VrpPaths.java
```diff
@@ -38,31 +39,55 @@
...
+		return createPath(diversionPoint, toLink, pathData.getPath(), travelTime);
+	}
+
 	public static VrpPathWithTravelData createPath(Link fromLink, Link toLink, double departureTime, Path path,
 			TravelTime travelTime) {
-		return createPath(fromLink, toLink, departureTime, path, travelTime, true);
+		return createPath(fromLink, toLink, departureTime, path, travelTime, true, false);
+	}
+
+	public static VrpPathWithTravelData createPath(LinkTimePair diversionPoint, Link toLink, Path path,
```
michalmac, line 82: Rename to `createPathForDiversion()`?

contribs/dvrp/src/main/java/org/matsim/contrib/dvrp/path/VrpPaths.java
```diff
@@ -98,18 +123,25 @@ public static VrpPathWithTravelData createPath(Link fromLink, Link toLink, doubl
...
 	}
 
-	public static final double FIRST_LINK_TT = 1;
+	public static final double FIRST_LINK_TT = 2;
+	public static final double NODE_TRANSITION_TIME = 1;
 
+	/**
+	 * It would make sense to replace these calls with TravelTime.getTravelTime(...) whereever it is used /sh nov 21
+	 */
+	@Deprecated
```
michalmac, line 141: Maybe adapt this method to use `TravelTime` for the last link?

BTW. Using the travel time will lead to some overestimated travel times, but perhaps it is more accurate than assuming the free-speed times (underestimated travel times).

michalmac (approved):
LGTM. I added a few comments, nothing blocking.

I really like the test you added. This ensures that if there are some changes in QSim, we will get notified :-)

I am a bit worried about the additional complexity. Lots of new methods added, while the impact of +/- 1 second on results may not be significant. See discussion in #1743

contribs/dvrp/src/main/java/org/matsim/contrib/dvrp/path/VrpPaths.java
michalmac: Deprecating is also fine with me. This can be addressed anytime later.

PR #1779, opened 2021-12-02
Dynamic time constraints for DRT requests

sebhoerl:
Proposition for some functionality which will allow defining a custom logic on how the latest pick-up and drop-off times are calculated *during* the simulation (i.e. not at "routing time"). This allows, for instance, to react to delayed agents, which however, need to reach a train or activity on time.


michalmac (commented):
`DrtRequestCreator` is a very simple class (basically doing 2 things: send a "request submitted" event and return a request). The best way to change this behaviour is by providing a custom implementation of `PassengerRequestCreator`. That also means no need to introduce another binding.

sebhoerl:
Actually, that's true. Not sure why we wanted to add a new interface in the first place. Maybe it had to do with h#1778. I agree, just replacing PassengerRequestCreator is better.

PR #1783, opened 2021-12-06
Some DRT analysis improvements

sebhoerl:
This PR modifies some DRT analysis functionality:

- Analysis is corrected for prebooking cases (before waiting time was calculated from the DrtRequestSubmissionEvent to the PassengerPickupEvent, now it is calculated from the PersonDepartureEvent to the PassengerPickupEvent). The PersonDepartureEvent can either come after the DrtRequestSubmissionEvent (= prebooking) or before (= instant).
- Currently, analysis assumes that requests are either rejected or scheduled. Other optimizers may reject requests after they have initially been scheduled (which also happens in reality with ridesharing operators). This is now taken into account, a request that is rejected at any time is regarded as "rejected".
- So far, analysis plots only show the unsharedRideTime vs. actualRideTime with constraint boundaries, but no similar analysis for waiting times (except in an aggregated way) exist to check whether the waiting time constraint is fulfilled. The PR adds two new plots (constraints_drt_travel_time, constrains_drt_waiting_time), which allow for this analysis and also in the case where an implementation of `DrtRequestCreator` makes use of *individual* contraints per request.

@michalmac Let me know which parts are interesting to be commited. I think the prebooking / rejection analysis is quite important, the plots may be bonus.

0 constraints_drt_waiting_time
0 constraints_drt_travel_time

### DIFF
--- a/contribs/drt/src/main/java/org/matsim/contrib/drt/analysis/DrtEventSequenceCollector.java
+++ b/contribs/drt/src/main/java/org/matsim/contrib/drt/analysis/DrtEventSequenceCollector.java
@@ -76,7 +76,7 @@ public class DrtEventSequenceCollector
 		private PassengerRequestRejectedEvent rejected;
 		
 		@Nullable
-		private PersonDepartureEvent departed;
+		private PersonDepartureEvent departure;
 		@Nullable
 		private PassengerPickedUpEvent pickedUp;
 		@Nullable
@@ -93,7 +93,7 @@ public class DrtEventSequenceCollector
 				PassengerDroppedOffEvent droppedOff, List<PersonMoneyEvent> drtFares) {
 			this.submitted = Objects.requireNonNull(submitted);
 			this.scheduled = Objects.requireNonNull(scheduled);
-			this.departed = departed;
+			this.departure = departed;
 			this.pickedUp = pickedUp;
 			this.droppedOff = droppedOff;
 			this.drtFares = new ArrayList<>(drtFares);
@@ -112,7 +112,7 @@ public class DrtEventSequenceCollector
 		}
 		
 		public Optional<PersonDepartureEvent> getDeparted() {
-			return Optional.ofNullable(departed);
+			return Optional.ofNullable(departure);
 		}
 
 		public Optional<PassengerPickedUpEvent> getPickedUp() {
@@ -186,7 +186,7 @@ public class DrtEventSequenceCollector
 				// sequence and note down the person id to fill in the departure event later on.
 				sequencesWithoutDeparture.computeIfAbsent(event.getPersonId(), id -> new LinkedList<>()).add(sequence);
 			} else {
-				sequence.departed = departureEvent.get();
+				sequence.departure = departureEvent.get();
 			}
 		}
 	}
@@ -202,7 +202,7 @@ public class DrtEventSequenceCollector
 				// is down (usually right after).
 				departuresWithoutSequence.computeIfAbsent(event.getPersonId(), id -> new LinkedList<>()).add(event);
 			} else {
-				sequence.get().departed = event;
+				sequence.get().departure = event;
 			}
 		}
 	}
--- a/contribs/drt/src/main/java/org/matsim/contrib/drt/analysis/DrtEventSequenceCollector.java
+++ b/contribs/drt/src/main/java/org/matsim/contrib/drt/analysis/DrtEventSequenceCollector.java
@@ -85,14 +85,14 @@ public class DrtEventSequenceCollector
 		private List<PersonMoneyEvent> drtFares = new LinkedList<>();
 
 		EventSequence(DrtRequestSubmittedEvent submitted) {
-			this.submitted = submitted;
+			this.submitted = Objects.requireNonNull(submitted);
 		}
 		
 		public EventSequence(PersonDepartureEvent departed, DrtRequestSubmittedEvent submitted,
 				PassengerRequestScheduledEvent scheduled, PassengerPickedUpEvent pickedUp,
 				PassengerDroppedOffEvent droppedOff, List<PersonMoneyEvent> drtFares) {
 			this.submitted = Objects.requireNonNull(submitted);
-			this.scheduled = Objects.requireNonNull(scheduled);
+			this.scheduled = scheduled;
 			this.departure = departed;
 			this.pickedUp = pickedUp;
 			this.droppedOff = droppedOff;
@@ -103,12 +103,12 @@ public class DrtEventSequenceCollector
 			return submitted;
 		}
 
-		public PassengerRequestScheduledEvent getScheduled() {
-			return scheduled;
+		public Optional<PassengerRequestScheduledEvent> getScheduled() {
+			return Optional.ofNullable(scheduled);
 		}
 		
-		public PassengerRequestRejectedEvent getRejected() {
-			return rejected;
+		public Optional<PassengerRequestRejectedEvent> getRejected() {
+			return Optional.ofNullable(rejected);
 		}
 		
 		public Optional<PersonDepartureEvent> getDeparted() {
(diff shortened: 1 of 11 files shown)


contribs/drt/src/main/java/org/matsim/contrib/drt/analysis/DrtEventSequenceCollector.java
```diff
@@ -60,31 +65,35 @@
...
-		private final PassengerRequestScheduledEvent scheduled;
-		//pickedUp and droppedOff may be null if QSim ends before the request is actually handled
+		
+		@Nullable
+		private PassengerRequestScheduledEvent scheduled;
+		@Nullable
+		private PassengerRequestRejectedEvent rejected;
+		
+		@Nullable
+		private PersonDepartureEvent departed;
```
michalmac, line 79: Rename to `departure`? Otherwise, it sounds like we need to send a caravan for that person :-)

contribs/drt/src/main/java/org/matsim/contrib/drt/analysis/DrtEventSequenceCollector.java
```diff
@@ -97,6 +106,14 @@ public DrtRequestSubmittedEvent getSubmitted() {
 		public PassengerRequestScheduledEvent getScheduled() {
 			return scheduled;
 		}
+		
+		public PassengerRequestRejectedEvent getRejected() {
```
michalmac, line 110: Optional

contribs/drt/src/main/java/org/matsim/contrib/drt/analysis/DrtEventSequenceCollector.java
```diff
@@ -60,31 +65,35 @@
...
+			this.submitted = submitted;
 		}
-
-		public PerformedRequestEventSequence(DrtRequestSubmittedEvent submitted,
+		
+		public EventSequence(PersonDepartureEvent departed, DrtRequestSubmittedEvent submitted,
 				PassengerRequestScheduledEvent scheduled, PassengerPickedUpEvent pickedUp,
 				PassengerDroppedOffEvent droppedOff, List<PersonMoneyEvent> drtFares) {
 			this.submitted = Objects.requireNonNull(submitted);
 			this.scheduled = Objects.requireNonNull(scheduled);
```
michalmac, line 95: Do not require non-nullity on `scheduled`

contribs/drt/src/main/java/org/matsim/contrib/drt/analysis/DrtEventSequenceCollector.java
```diff
@@ -60,31 +65,35 @@
...
 		private PassengerDroppedOffEvent droppedOff;
 		@Nullable
-		private List<PersonMoneyEvent> drtFares;
+		private List<PersonMoneyEvent> drtFares = new LinkedList<>();
 
-		public PerformedRequestEventSequence(DrtRequestSubmittedEvent submitted,
-				PassengerRequestScheduledEvent scheduled) {
-			this(submitted, scheduled, null, null, List.of());
+		EventSequence(DrtRequestSubmittedEvent submitted) {
+			this.submitted = submitted;
```
michalmac, line 88: add `Objects.requireNonNull(submitted)`

contribs/drt/src/main/java/org/matsim/contrib/drt/analysis/DrtEventSequenceCollector.java
```diff
@@ -97,6 +106,14 @@ public DrtRequestSubmittedEvent getSubmitted() {
 		public PassengerRequestScheduledEvent getScheduled() {
```
michalmac, line 106: Now optional.

contribs/drt/src/main/java/org/matsim/contrib/drt/analysis/DrtEventSequenceCollector.java
```diff
@@ -217,11 +247,85 @@ public void handleEvent(PersonMoneyEvent event) {
...
+		boolean foundDeparture = false;
+
+		if (potentialDepartures != null) {
+			Iterator<PersonDepartureEvent> iterator = potentialDepartures.iterator();
+
+			while (iterator.hasNext()) {
+				PersonDepartureEvent departedEvent = iterator.next();
+
+				if (event.getFromLinkId().equals(departedEvent.getLinkId())) {
+					if (foundDeparture) {
```
michalmac, line 276: remove `foundDeparture` and use `result.isPresent()` instead

contribs/drt/src/main/java/org/matsim/contrib/drt/analysis/DrtEventSequenceCollector.java
```diff
@@ -217,11 +247,85 @@ public void handleEvent(PersonMoneyEvent event) {
...
+		boolean foundSequence = false;
+
+		if (potentialSequences != null) {
+			Iterator<EventSequence> iterator = potentialSequences.iterator();
+
+			while (iterator.hasNext()) {
+				EventSequence sequence = iterator.next();
+
+				if (sequence.submitted.getFromLinkId().equals(event.getLinkId())) {
+					if (foundSequence) {
```
michalmac, line 313: remove `foundSequence` and use `result.isPresent()` instead

contribs/drt/src/main/java/org/matsim/contrib/drt/analysis/DrtEventSequenceCollector.java
```diff
@@ -217,11 +247,85 @@ public void handleEvent(PersonMoneyEvent event) {
...
 			}
 		}
 	}
+
+	/*
+	 * This function is helper that finds a PersonDepartureEvent for a given
+	 * DrtRequestSubmittedEvent. This means that the departure has happened before
+	 * submission, which is the usually case for instant requests.
+	 */
+	private Optional<PersonDepartureEvent> popDepartureForSubmission(DrtRequestSubmittedEvent event) {
```
michalmac, line 263: Suggestion: return simply a (nullable) PersonDepartureEvent? `Optional` is intended for public API to make it more expressive, whereas here it's a private method and the results is consumed inside this class.

contribs/drt/src/test/java/org/matsim/contrib/drt/util/DrtEventsReadersTest.java
```diff
@@ -65,7 +65,7 @@
 
 	//standard dvrp events are tested in DvrpEventsReadersTest
 	private final List<Event> drtEvents = List.of(
-			new DrtRequestSubmittedEvent(0, mode, request, person, link1, link2, 111, 222),//
+			new DrtRequestSubmittedEvent(0, mode, request, person, link1, link2, 111, 222, Double.NaN, Double.NaN),//
```
michalmac, line 68: For testing event writing/reading I would use something else than `NaN` and perhaps different values for each field (to make sure they are not confused).

michalmac (approved):
Thanks! Looks good to me.

I left some comments (a few things to fix + a few suggestions).

Since now everything may be optional (except the submission event), the event sequences are infested with Optionals. I will try to address this issue later, but let's merge this PR first.

michalmac:
> * So far, analysis plots only show the unsharedRideTime vs. actualRideTime with constraint boundaries, but no similar analysis for waiting times (except in an aggregated way) exist to check whether the waiting time constraint is fulfilled. The PR adds two new plots (constraints_drt_travel_time, constrains_drt_waiting_time), which allow for this analysis and also in the case where an implementation of `DrtRequestCreator` makes use of _individual_ contraints per request.

Currently, the constraint boundaries are presented for all 3 plots:
- ride times (the lower black line)
06000_cap4 0 drt_detours_drt_rideTimePlot
- travel times (the lower black line)
06000_cap4 0 drt_detours_drt_travelTimePlot
- wait times (the vertical black line)
06000_cap4 0 waitTimeComparison_drt

However, the boundaries make sense only for the default constraint definitions.

PR #1786, opened 2021-12-08
DVRP: improved last link travel time computation

michalmac:
This PR replaces the deprecated last link TT calculation with an improved estimation method. This is a follow up to changes made in #1771.

### DIFF
--- a/contribs/drt/src/main/java/org/matsim/contrib/drt/optimizer/insertion/DetourTimeEstimator.java
+++ b/contribs/drt/src/main/java/org/matsim/contrib/drt/optimizer/insertion/DetourTimeEstimator.java
@@ -24,6 +24,7 @@ import org.matsim.api.core.v01.network.Link;
 import org.matsim.contrib.common.util.DistanceUtils;
 import org.matsim.contrib.dvrp.path.VrpPaths;
 import org.matsim.contrib.zone.skims.DvrpTravelTimeMatrix;
+import org.matsim.core.router.util.TravelTime;
 
 /**
  * @author michalm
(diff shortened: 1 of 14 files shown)


michalmac:
@sebhoerl I compared results obtained in `RunDrtExampleIT` from before #1771 and after #1786: https://github.com/matsim-org/matsim-libs/compare/38210b4...45bb65d/?diff=unified to see how both PRs acted together.

Looks like we are back to (almost) the same results as before for the cases where rejections are turned off.

- `testRunDrtExampleWithNoRejections_ExtensiveSearch()` gives the same results as before
- `testRunDrtExampleWithNoRejections_SelectiveSearch()` gives almost the same results, i.e.
```
				.waitAverage(293.47)
				.inVehicleTravelTimeMean(389.71)
				.totalTravelTimeMean(683.18)
```
vs
```				.waitAverage(293.63)
				.inVehicleTravelTimeMean(388.85)
				.totalTravelTimeMean(682.48)
```
For other three cases (all with rejections) there are differences when comparing "before vs after". Two out of these three differ in the number of rejections, which makes other stats (like wait, travel time etc.) incomparable.

michalmac:
The issue was that the old formula (i.e. before #1771)
```
Math.floor(lastLink.getLength() / lastLink.getFreespeed(time));
```
and the current one:
```
travelTime.getLinkTravelTime(lastLink, time, null, null) - NODE_TRANSITION_TIME;
```
are the same (given that `QSimFreeSpeedTravelTime` is used as `travelTime`, which adds `NODE_TRANSITION_TIME` to the free-speed travel time), whereas the interim formula:
```
Math.floor(lastLink.getLength() / lastLink.getFreespeed(time)) - NODE_TRANSITION_TIME;
```
underestimated the time by 1 second (i.e. `NODE_TRANSITION_TIME`).

===

I assume that the rest of differences between before #1771 and after #1786 are caused by the extra second added for time spent on the initial link (now: 2 seconds, before: 1 second). This is why I had to add 1 second (120 -> 121) here:
https://github.com/matsim-org/matsim-libs/blob/dffae7575527bdcc6ed849d21a45ef9672be7c6e/contribs/drt/src/test/java/org/matsim/contrib/drt/optimizer/insertion/DrtPoolingParameterTest.java#L85

(as now the additional 1 second on the initial link is not compensated by the "lost second" on the last link).

Issue #1760, opened 2021-11-16
Relax unoccupied custom task constraints for dvrp?


nkuehnel:
The VehicleOccupancyProfileCalculator assumes that tasks are either 
a) passenger serving tasks
b) non-passenger serving tasks

In consequence, passenger serving tasks will _always_ be labeled as 0 pax/1 pax/ etc.:
https://github.com/matsim-org/matsim-libs/blob/6b038c562bb17bd4a4cbd7c29bed797fd12b3458/contribs/dvrp/src/main/java/org/matsim/contrib/util/stats/VehicleOccupancyProfileCalculator.java#L155-L162

In the shifts extension, we added relocate for shift changeover / relocate for shift break tasks to operational facilities such as hubs. We would like them to show up individually in the occupancy profiles so we set them as non-passenger-serving tasks. However, in our implementation, drivers may still serve rides on their way to the hub if the detour is acceptable. In this case, the code quoted above fails with an exception since passengers are not allowed during such tasks. Our current "quick fix" is to set these tasks as passenger serving tasks in the DrtModeAnalysisModule - however, now we do not see them in the occupancy plot anymore:

https://github.com/matsim-org/matsim-libs/blob/6b038c562bb17bd4a4cbd7c29bed797fd12b3458/contribs/drt/src/main/java/org/matsim/contrib/drt/extension/shifts/run/MultiModeShiftDrtModule.java#L33-L41

Could we maybe relax this constraint and allow passengers to be on (user-definable) non-regular passenger serving tasks as well?

michalmac:
It's hard to mix vehicle occupancy with task types in a consistent way on a 2D plot as these are 2 separate dimensions. So either you lose some information on the vehicle occupancy or on the relocation tasks, which means the values on the plot are not reliable.

What if there was a separate chart that shows currently executed task types (instead of vehicle occupancy)?

sebhoerl:
Yes, somehow two different plots seems like a good idea?

nkuehnel:
Sometimes the simplest solutions are the best - you are absolutely right, it would be much cleaner. If no one is already working on such analysis I will try to add one in the next days

michalmac:
@nkuehnel I haven't started working on it, so please go ahead. It would be great if you could add it to dvrp (so it could be used both in drt and taxi).

[...]

PR #1788, opened 2021-12-09
Add DVRP vehicle task profiles 

nkuehnel:
see #1760


nkuehnel:
This is how it looks like for the Holzkirchen test scenario:

1 task_time_profiles_StackedArea_drt


So far this is fine. However, I wonder if we could make name() in DrtTaskType non-final?

https://github.com/matsim-org/matsim-libs/blob/45bb65d6de2eb68bb2210a72fae82955e52fbbf2/contribs/drt/src/main/java/org/matsim/contrib/drt/schedule/DrtTaskType.java#L57-L60

The reason is, our shift changeover and shift break tasks are also DrtStopTasks (extends) to allow passengers to board/alight. Now as you can see in the plot, these tasks do not show up but are subsumed in the STOP tasks. If we would allow to let extended classes to override the name() method we could show them separately. One could also check for instanceof in the profile calculator but not sure this would be better..

@michalmac  would that make sense?

michalmac (approved):
Thanks! LGTM

michalmac:
> The reason is, our shift changeover and shift break tasks are also DrtStopTasks (extends) to allow passengers to board/alight. Now as you can see in the plot, these tasks do not show up but are subsumed in the STOP tasks. If we would allow to let extended classes to override the name() method we could show them separately. One could also check for instanceof in the profile calculator but not sure this would be better..
> 
> @michalmac would that make sense?
> 
> A similar thing would be the "WaitForShiftStayTask" which is also just shown as "STAY" right now

I see your point. However, making the `name()` method non-final will not solve the issue. The main problem is that the profiles are grouped by task type. One quick solution is defining the grouping as a function (with the default `Task::getTaskType`) - that something you can add (in a separate PR)

However, in the long term I would avoid/limit extending tasks. It is a shortcut (which I also used in EDrt/ETaxi-tasks to attach attributes like energy consumptions), but once the task obtains additional functions, it's no more a regular stay or stop task. For this reason, we introduced the split between "base task type" and "task type". For instance, your shift-related tasks may still have `STOP` as the _base type_ but have something else than `DrtStopTask.TYPE` as the _type_ (*).

(*) it is also possible to specify the base type is null... You can find a bit more info on this here: https://github.com/matsim-org/matsim-libs/pull/1063

nkuehnel:
> The main problem is that the profiles are grouped by task type. 

Right. The name would only change the label in the figure but would not change the underlying grouping..makes sense.

> For this reason, we introduced the split between "base task type" and "task type". For instance, your shift-related tasks may still have STOP as the base type but have something else than DrtStopTask.TYPE as the type (*).

Not sure that wrk out of the box - the reason we extended DrtStopTask was that there are parts in the passenger serving logic that explicitly ask for DrtStopTasks, e.g.:
https://github.com/matsim-org/matsim-libs/blob/2a22ab0ea4bfd01912b51a54f1126036d4181b4f/contribs/drt/src/main/java/org/matsim/contrib/drt/optimizer/VehicleDataEntryFactoryImpl.java#L96-L109

If I understand correctly, the Waypoint.Stop class, which is later used for insertion generation, currently only works with DrtStopTask, so passenger serving tasks somehow have to be DrtStopTasks:
https://github.com/matsim-org/matsim-libs/blob/2a22ab0ea4bfd01912b51a54f1126036d4181b4f/contribs/drt/src/main/java/org/matsim/contrib/drt/optimizer/Waypoint.java#L122-L144

> One quick solution is defining the grouping as a function (with the default Task::getTaskType) - that is something you can add (in a separate PR)

Maybe we go ahead with this solution for now then, thanks @michalmac !

michalmac:
Good point. Looks like we reached the state that we need `DrtStopTask` to be an interface.

nkuehnel:
Alright, guess that's work for another PR then :)

michalmac:
> Alright, guess that's work for another PR then :)

I will take care of it as this is a generic DVRP issue/feature. I started some refactorings in that direction.

nkuehnel:
Great!

PR #1792, opened 2021-12-10
Introduce StayTaskEndTimeCalculator binding

sebhoerl:
Makes it possible to override StayTaskEndTimeCalculator more flexibly (which was a copy and paste for the shifts module now, and there would be another copy and paste for the upcoming Alonso-Mora algorithm code).

@nkuehnel While going through this, I noticed that there is no occurence of a `ShiftDrtStayTaskEndTimeCalculator` being wrapped around a `EDrtStayTaskEndTimeCalculator`. Shouldn't there be one for the `eshifts` extension?

### DIFF
--- a/contribs/drt/src/main/java/org/matsim/contrib/drt/extension/edrt/run/EDrtModeOptimizerQSimModule.java
+++ b/contribs/drt/src/main/java/org/matsim/contrib/drt/extension/edrt/run/EDrtModeOptimizerQSimModule.java
@@ -57,6 +57,7 @@ import org.matsim.contrib.dvrp.run.AbstractDvrpModeQSimModule;
 import org.matsim.contrib.dvrp.run.DvrpConfigGroup;
 import org.matsim.contrib.dvrp.run.DvrpModes;
 import org.matsim.contrib.dvrp.schedule.ScheduleTimingUpdater;
+import org.matsim.contrib.dvrp.schedule.ScheduleTimingUpdater.StayTaskEndTimeCalculator;
 import org.matsim.contrib.dvrp.vrpagent.VrpAgentLogic;
 import org.matsim.contrib.ev.infrastructure.ChargingInfrastructure;
 import org.matsim.contrib.ev.infrastructure.ChargingInfrastructures;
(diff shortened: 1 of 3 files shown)


nkuehnel:
@sebhoerl thnaks for pointing this out, however we did not want to wrap the EDrtStayTaskEndTimeCalculator because it assumes that Charging Activity durations are fixed and charging cannot be interrupted. However, in the shifts extension the operational tasks define the duration and charging may be interrupted.

michalmac (commented):
Similar to #1779, I would prefer overriding the `ScheduleTimingUpdater` binding rather than adding yet another binding (i.e. `StayTaskEndTimeCalculator`).

For instance, at one point we may need to refactor `ScheduleTimingUpdater` to an interface to provide more flexibility here. Then `StayTaskEndTimeCalculator` may be used only by one specific implementation (as it only refers to one narrow aspect of updating task times).

michalmac:
> Okay, I think in the current setup this holds. But, actually, this is a preparational step for pushing the code of the Alonso-Mora dispatcher. There, we make use of the StayTaskEndTimeCalculator when recreating vehicle schedules and to ensure that we are consistent with what standard DRT is doing. Furhermore, we can then easily port shifts, electric, ... things to the Alonso-Mora implementation, because it tells us how to integrate oepartional tasks, for instance. So in the current situation I follow you argument, because the StayTaskEndTimeCalculator is _only_ used by the ScheduleTImingUpdater, but after we'll have another use case that needs the very functionality that is nicely interfaced by StayTaskEndTimeCalculator. What do you say?

Hard to say without looking at the code. Do you have that code pushed somewhere?

My argument is: maybe someone would like to change the way how we predict the end of time, or change anything else in the updating procedure, then these changes will be easier to address having `ScheduleTimingUpdater` as an interface, rather than adding multiple but narrow-scope strategies.

===

But I am also fine with trading the `ScheduleTimingUpdater` binding (which is not an interface now) for the `StayTaskEndTimeCalculator` one (at least until we need more flexibility wrt task times updating). In the end, I want to avoid having too many bindings. (BTW. Would be great to have support for private guice modules, so we hide some implementation-specific bindings).

sebhoerl:
Ok, let's see, we can also let this PR sit, and we check again after I prepare the PR for the dispatcher. 
I'll also check whether it makes sense to use the ScheduleTimingUpdater instead of StayTaskEndTimeCalculator.

PR #1800, opened 2021-12-15
Extract StayTask and DrtStopTask interfaces

michalmac:
To allow more flexibility when implementing custom tasks (incl. mixing two different task types into one). See: https://github.com/matsim-org/matsim-libs/pull/1788#issuecomment-991355810


nkuehnel:
Very nice, thank you!

PR #1809, opened 2021-12-19
Variable stop durations for DRT

sebhoerl:
Currently, the stop duration in DRT is defined via configuration, and it is a constant value per stop. This PR introduces a StopDuration interface, which is inspired by the TravelTime interface. Based on the vehicle (e.g., special type) or the pickup and dropoff requests (which may have specific needs), the interface decides how long the stop will take. The default implementation is based on the configuration:

```bindModal(StopDuration.class).toInstance((vehicle, dropoffs, pickups) -> drtCfg.getStopDuration());```

There is one caveat: Sometimes, DRT will add new requests to *ongoing* stops. In that case, a custom implementation of `StopDuration` may modify the duration of the `DrtStopTask`, but this information would not be passed down to the active `BusStopActivity`. For now, a warning is produced if such a case happens.

### DIFF
--- a/contribs/drt/src/main/java/org/matsim/contrib/drt/extension/edrt/run/EDrtModeOptimizerQSimModule.java
+++ b/contribs/drt/src/main/java/org/matsim/contrib/drt/extension/edrt/run/EDrtModeOptimizerQSimModule.java
@@ -45,6 +45,7 @@ import org.matsim.contrib.drt.optimizer.insertion.UnplannedRequestInserter;
 import org.matsim.contrib.drt.optimizer.rebalancing.RebalancingStrategy;
 import org.matsim.contrib.drt.run.DrtConfigGroup;
 import org.matsim.contrib.drt.schedule.DrtTaskFactory;
+import org.matsim.contrib.drt.schedule.DrtStopTask.StopDuration;
 import org.matsim.contrib.drt.scheduler.DefaultRequestInsertionScheduler;
 import org.matsim.contrib.drt.scheduler.DrtScheduleInquiry;
 import org.matsim.contrib.drt.scheduler.EmptyVehicleRelocator;
@@ -133,7 +134,7 @@ public class EDrtModeOptimizerQSimModule extends AbstractDvrpModeQSimModule {
 						getter.getModal(QSimScopeForkJoinPoolHolder.class).getPool()))).asEagerSingleton();
 
 		bindModal(InsertionCostCalculator.InsertionCostCalculatorFactory.class).toProvider(modalProvider(
-				getter -> DefaultInsertionCostCalculator.createFactory(drtCfg, getter.get(MobsimTimer.class),
+				getter -> DefaultInsertionCostCalculator.createFactory(getter.getModal(StopDuration.class), getter.get(MobsimTimer.class),
 						getter.getModal(CostCalculationStrategy.class))));
 
 		install(DrtModeOptimizerQSimModule.getInsertionSearchQSimModule(drtCfg));
@@ -166,14 +167,14 @@ public class EDrtModeOptimizerQSimModule extends AbstractDvrpModeQSimModule {
 		bindModal(DrtScheduleInquiry.class).to(DrtScheduleInquiry.class).asEagerSingleton();
 
 		bindModal(RequestInsertionScheduler.class).toProvider(modalProvider(
-						getter -> new DefaultRequestInsertionScheduler(drtCfg, getter.getModal(Fleet.class),
+						getter -> new DefaultRequestInsertionScheduler(getter.getModal(StopDuration.class), getter.getModal(Fleet.class),
 								getter.get(MobsimTimer.class), getter.getModal(TravelTime.class),
 								getter.getModal(ScheduleTimingUpdater.class), getter.getModal(DrtTaskFactory.class))))
 				.asEagerSingleton();
 
 		bindModal(ScheduleTimingUpdater.class).toProvider(modalProvider(
 				getter -> new ScheduleTimingUpdater(getter.get(MobsimTimer.class),
-						new EDrtStayTaskEndTimeCalculator(drtCfg)))).asEagerSingleton();
+						new EDrtStayTaskEndTimeCalculator(getter.getModal(StopDuration.class))))).asEagerSingleton();
 
 		bindModal(VrpAgentLogic.DynActionCreator.class).toProvider(modalProvider(
 				getter -> new EDrtActionCreator(getter.getModal(PassengerHandler.class), getter.get(MobsimTimer.class),
--- a/contribs/drt/src/main/java/org/matsim/contrib/drt/extension/edrt/schedule/EDrtStayTaskEndTimeCalculator.java
+++ b/contribs/drt/src/main/java/org/matsim/contrib/drt/extension/edrt/schedule/EDrtStayTaskEndTimeCalculator.java
@@ -18,14 +18,14 @@
 
 package org.matsim.contrib.drt.extension.edrt.schedule;
 
-import org.matsim.contrib.drt.run.DrtConfigGroup;
 import org.matsim.contrib.drt.schedule.DrtStayTaskEndTimeCalculator;
+import org.matsim.contrib.drt.schedule.DrtStopTask.StopDuration;
 import org.matsim.contrib.dvrp.fleet.DvrpVehicle;
 import org.matsim.contrib.dvrp.schedule.StayTask;
 
 public class EDrtStayTaskEndTimeCalculator extends DrtStayTaskEndTimeCalculator {
-	public EDrtStayTaskEndTimeCalculator(DrtConfigGroup drtConfigGroup) {
-		super(drtConfigGroup);
+	public EDrtStayTaskEndTimeCalculator(StopDuration stopDuration) {
+		super(stopDuration);
 	}
 
 	@Override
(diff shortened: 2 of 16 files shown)

### CODE REVIEW

**Comment by michalmac** [comment]
@sebhoerl I wanted to review this PR, but before doing that I have a few questions:
- Do you have any specific `StopDuration` implementation in mind?
- Have you checked for potential performance regression for the selective insertion? I am mainly thinking about having additional computations inside `InsertionDetourTimeCalculator`.

**Comment by nkuehnel** [comment]
@michalmac one very recent use case is the assessment of wheelchair accessible vehicles (WAV) and wheelchair requests where we expect significantly higher stop durations due to additional tasks of the driver for (de-)boarding and securing the passenger, since only some of the vehicles in the fleet will be WAVs it would be helpful to represent different stop durations.

There is also evidence that these follow a log normal distribution which would be another stochastic eement that could be computed on the fly
(Garnier et al, 2020, Adjusting Dwell Time for Paratransit Services https://journals.sagepub.com/doi/abs/10.1177/0361198120931099)

**Comment by michalmac** [comment]
> @michalmac one very recent use case is the assessment of wheelchair accessible vehicles (WAV) and wheelchair requests where we expect significantly higher stop durations due to additional tasks of the driver for (de-)boarding and securing the passenger, since only some of the vehicles in the fleet will be WAVs it would be helpful to represent different stop durations.

Thanks. Let me then dive deeper into this :-)  Can we assume the stop duration is additive per request? So one pickup/delivery request extends the stop duration by the same amount of time regardless of how many other people enter/leave the vehicle? I am asking because there are many VRP solvers that make such assumptions (e.g. jsprit).

> There is also evidence that these durations follow a log normal distribution which would be another stochastic element that could be computed on the fly

Yes, but that can be modelled inside the bus activity and not at the level of the task/algorithm.

**Comment by nkuehnel** [comment]
> Can we assume the stop duration is additive per request? So one pickup/delivery request extends the stop duration by the same amount of time regardless of how many other people enter/leave the vehicle? I am asking because there are many VRP solvers that make such assumptions (e.g. jsprit).

In our case we (will) assume it is subadditive, i.e. picking up 6 passengers at once will always take less time than picking up six individually. Currently, we still use a fixed stop duration for all stops.

> Yes, but that can be modelled inside the bus activity and not at the level of the task/algorithm.

True, this is not required to be exactly known by the algorithm in advance :D
Still, the scheduler/algorithm should account for different expected values depending on the expected passenger (groups)

**Comment by michalmac** [comment]
> Have you checked for potential performance regression for the selective insertion? I am mainly thinking about having additional computations inside InsertionDetourTimeCalculator.

I ran the old Berlin DRT scenario with the fixed stop duration and comparing this PR vs master and see a roughly 5% increase of the QSim computation time (excluding the initial travel-time matrix computation). So this is not too bad, but maybe there is a way to structure the `StopDuration` interface in a way that the impact on computation time would almost be unnoticeable? This is also why I was asking the questions about additivity etc.

**Comment by michalmac** [comment]
> In our case we (will) assume it is subadditive, i.e. picking up 6 passengers at once will always take less time than picking up six individually.

But maybe there is a non-zero fixed time of a bus stop and request-based times that can be computed individually per each request and summed up?

**Comment by sebhoerl** [comment]
I had in mind making the stop duration dependent on the number of requests. So a simple implementation might be `(v, d, p) -> stopDurationPerPerson * (d.size() + p.size())`. This could be implemented much simpler compared to the interface solution, but then Nico came with the wheelchair use case, so I was putting it behind the interface.

**Comment by michalmac** [comment]
What if we modelled it as a sum of:
- `fixed_stop_duration(v)`
- sum of `pickup_duration(v,r)` for all `r` to be picked up
- sum of `dropoff_duration(v,r)` for all `r` to be dropped off

So all components depend on the vehicle (`v`) and both the pickup/dropoff-related components depend on the requests (`r`).

Would that be a good approximation of reality? (especially when compared to the accuracy of travel time approximations)

**Review by michalmac — COMMENTED** [review]
Nice work! We always missed this functionality.

Before approving the PR I would like to discuss the `StopDuration` interface.

**Comment by michalmac** [comment]
> There is one caveat: Sometimes, DRT will add new requests to ongoing stops. In that case, a custom implementation of StopDuration may modify the duration of the DrtStopTask, but this information would not be passed down to the active BusStopActivity. For now, a warning is produced if such a case happens.

That's fine. We can later adapt `BusStopActivity`.

**Comment by michalmac** [comment]
An alternative approach to discuss:

```java
	interface StopDuration {
		double calcDuration(DvrpVehicle vehicle, Collection<DrtRequest> dropoffRequests,
				Collection<DrtRequest> pickupRequests);
	}

	interface AdditionalStopDuration {
		//if stopTask is null - it's a new (potential) stop
		double calcForPickup(DvrpVehicle vehicle, DrtStopTask stopTask, DrtRequest pickupRequest);

		//if stopTask is null - it's a new (potential) stop
		double calcForDropoff(DvrpVehicle vehicle, DrtStopTask stopTask, DrtRequest dropoffRequest);
	}
```
In most of the places we would call `StopDuration.calcDuration`. When calling it we already know the stop task (either we are constructing it or we need to re-calculate its duration), so we could even pass the stop task (if we really need that).

However, `InsertionDetourTimeCalculator` needs to compute the increase in the stop duration due to inserting the pickup/dropoff of a new request. This is where using the `StopDuration` interface is a bit unhandy:
https://github.com/matsim-org/matsim-libs/blob/4780631fd72fcd4a3db7d13da602541322f99db4/contribs/drt/src/main/java/org/matsim/contrib/drt/optimizer/insertion/InsertionDetourTimeCalculator.java#L128-L130
For the fixed stop duration, the new interface already adds 5% to the qsim time (for drt-intense runs with the selective insertion). If a more complex formula is used this may become a bit of a hotspot. `AdditionalStopDuration` aims at reducing that impact. Moreover, `AdditionalStopDuration` may compute some rough estimates while `StopDuration` may provide more accurate times.

sebhoerl:
This looks fine to me. However, I would then propose to have an interface that implements all three functions, so that implementing one part forces you to implement the other.

```java
	interface StopDuration {
		double calcDuration(DvrpVehicle vehicle, Collection<DrtRequest> dropoffRequests,
				Collection<DrtRequest> pickupRequests);

		//if stopTask is null - it's a new (potential) stop
		double calcForPickup(DvrpVehicle vehicle, DrtStopTask stopTask, DrtRequest pickupRequest);

		//if stopTask is null - it's a new (potential) stop
		double calcForDropoff(DvrpVehicle vehicle, DrtStopTask stopTask, DrtRequest dropoffRequest);
	}
```

Or, it should probably be possible to avoid the first function call altogether and provide some helper/utility function that only makes use of the AdditionalStopDuration.

sebhoerl:
Reading again, maybe it makes sense to have it separate, basically you're saying that `AdditionalStopDuration` would rather be something living in the `drt.optimizer.insertion` space, while the `StopDuration` would live in `dvrp` related to scheduling, right?

michalmac:
> Reading again, maybe it makes sense to have it separate, basically you're saying that `AdditionalStopDuration` would rather be something living in the `drt.optimizer.insertion` space, while the `StopDuration` would live in `dvrp` related to scheduling, right?

Yes, that was my intention: `StopDuration` would be part of the "DRT model", while `AdditionalStopDuration` is meant for specific optimiser implementation only.

nkuehnel:
@sebhoerl @michalmac I think I got lost in the discussion - is there a preferred solution/update to this? :)

michalmac:
> @sebhoerl @michalmac I think I got lost in the discussion - is there a preferred solution/update to this? :)

I think we agreed to use `StopDuration` at the level of DRT, however, for the insertion-based optimiser, we will also need to provide `AdditionalStopDuration` (which is important as that optimiser is used for biggest scenarios and we should ensure the duration calculation is very fast).

sebhoerl:
Yes, that was the idea, unfortunately I don't find the time to update the PR lately. @nkuehnel Feel free in case you feel motivated ;)

neuma:
I like the idea of non-fixed stop durations. I understand that the proposed approach allows to set the stop duration dynamically during the mobsim based on the actual stop, the vehicle and the requests to handle.

What I currently do not understand is what happens if the requests change during a (somewhat longish) stop, e.g. a new pickup request gets assigned to the vehicle while it is still unloading.

Is the idea to return the total stop time once the vehicle reaches the stop or can the stop time be incremented by each request handled? The ComplexTransitStopHandler, for example, iterates over the requests again and again until it returns a stop time of 0 seconds. In each iteration, it handles as much requests as possible and returns an intermediate stop time specifying when it is able to handle the next requests. Meanwhile, new passengers/requests may arrive at the stop which may then also be handled.

sebhoerl:
Hi @neuma , there is two aspects to it. The first is the insertion heuristic, that needs to know what happens when a request is added. Currently, this is not so relevant, because the stop time is always the same and requests are always dropped off at the beginning of the stop, and picked up at the end. So if a request arrives that can be picked up before the current task is over, it is simply added, otherwise another one is created (if I'm not mistaken).

The other aspect is the simulation of the task, which is kind of independent. There the important part is the `BusStopActivity` which is currently initialized with a fixed duration and once the task is started, it cannot be changed anymore.

So the logic needs to be adapted in both places, and currently the insertion heuristic part replicates exactly what the simulation part does.

So let's say a new request arrives that is compatible with one that is currently being picked up (BusStopActivity already started), then, currently, we would need to add a new independent task in any case (also in the code that I presented above). This is an edge case, and currently the code above would just print a warning. So this is also something to consider for a clean implementation of all of this.

sebhoerl:
Superseeded by #1973.

PR #1810, opened 2021-12-19
Algorithm by Alonso-Mora et al. as a DRT extension

sebhoerl:
This PR adds the algorithm by Alonso-Mora et al. as an extension to the DRT package. It is compatible with shifts and electric vehicles. 

Some remarks:
- The standard version of AM solves a MILP to assign vehicles to requests. In this version, we use a simple best-response assignment. Alternatively, solvers using GLPK and Cbc can be used via configuration. Those expect the executables `glpsol` or `cbc` to be available on the system. Alternatively, we have implementations that use directly the Java interfaces for Gurobi and GLPK. However, we don't provide them here as (1) GLPK-JNI would need another external Maven repository, which we may want to avoid here (@michalmac ?), and the Gurobi interfaces are only available as a package that needs to be added explicitly to the classpath. Currently, the two implelentations are in a MOIA internal repository, but we will put the Maven packages online in some way (have to figure out how). 
- The core difference between AM is that the vehicle schedules are continuously rebuilt. Instead of only inserting requests in existing vehicle schedules, all potential combinations of requests and vehicles are re-evaluated in every decision step. This is a common scheme in many heuristics for fleet dispatching or horizon-based MILP formulations. Hence, those interfaces could be extracted again in a next step to set up a "AMoDeus 2.0" as a MATSim contrib with dynamic vehicle rescheduling.
- The changes in DRT itself are minimal: (1) Changing DefaultDrtOptimizer to its interface DrtOptimizer at one point, and (2) adjusting the relocation strategies such that they are able to work with multiple "PassengerRequestScheduledEvent" as this happens frequently with AM.


sebhoerl:
@nkuehnel Could you please check the conflicts?

nkuehnel:
> @nkuehnel Could you please check the conflicts?

@sebhoerl  Done. I also had to adjust some classes to the new task interfaces/implementations in drt, plz check

sebhoerl:
Thanks!

michalmac:
> * The standard version of AM solves a MILP to assign vehicles to requests. In this version, we use a simple best-response assignment.

Good idea!

> Alternatively, solvers using GLPK and Cbc can be used via configuration. Those expect the executables `glpsol` or `cbc` to be available on the system. Alternatively, we have implementations that use directly the Java interfaces for Gurobi and GLPK. However, we don't provide them here as (1) GLPK-JNI would need another external Maven repository, which we may want to avoid here (@michalmac ?), and the Gurobi interfaces are only available as a package that needs to be added explicitly to the classpath. Currently, the two implelentations are in a MOIA internal repository, but we will put the Maven packages online in some way (have to figure out how).

It would be good to maybe have only one generic java interface to MIP solvers. In the past, I used jCMPL for that. I guess there are other java alternatives. (BTW. I have no idea if we can legally put gurobi jar to a mvn repo)

> The changes in DRT itself are minimal: (1) Changing DefaultDrtOptimizer to its interface DrtOptimizer at one point, and (2) adjusting the relocation strategies such that they are able to work with multiple "PassengerRequestScheduledEvent" as this happens frequently with AM.

Makes sense.

michalmac (approved):
Looks really nice! 🚀

I have roughly reviewed it, as I am not able to go line by line. So I have more high-level comments (some have been already added to this PR, some we have exchanged by email).

I think the biggest question is how to position the new functionality. To me it is more than just an extension. So I would put it next to the "insertion" algorithm (but since it contains some edrt/shifts stuff, these adaptations should remain in "extensions") or in a separate contrib (but within the matsim repo; I guess this option could be easier to start with).

sebhoerl:
Ok, thanks for the feedback. After our exchange by mail, I was actually leaning towards the solution to have the AM functionality in a separate *repository* next to matsim-libs. There, I would propose to have the following package structure:

```
alonso_mora /
   core
   glpk
   gurobi
   cplex
```

The `core` package will work out of the box with the best-response solver. The others can be used if the system supports it. For Gurobi and CPLEX the challenge is that they rely on external JARs that can not be included via Maven, so in any case we cannot provide them in a combined package. The question is whether this structure should be an external repository, or a contrib. In both cases, it can evolve into an "amodeus2" contrib/repo where we can recover the other re-optimization-based algorithms that were included in Amodeus, but in a cleaner and more integrated way.

michalmac:
> The question is whether this structure should be an external repository, or a contrib. 

I think both options are good. As we now have PR-based releases, using the external repos has fewer downsides than before.

michalmac:
Awesome! 🚀

PR #1822, opened 2022-01-06
Make VehicleType for DVRP modal

nkuehnel:
Allows each dvrp mode to have a different default vehicle type


nkuehnel:
@michalmac this basically follows the proposal from the javadoc in DvrpModule, that it should be modal in the future. Please check if you have some time -  I added the vehicle type binding to fleet module - not sure if it is the right place :)

michalmac:
Thanks @nkuehnel. Please have a look at this PR #1824.

nkuehnel:
thx, will close this one then.

PR #1824, opened 2022-01-07
Dvrp modal vehicle type

michalmac:
Based on #1822 (thanks @nkuehnel). Takes a few more steps for a complete transition, otherwise some examples would still try to override the dvrp-global vehicle type, but that binding would be ignored now.

BTW. We want to support more than one vehicle type per mode in the future (e.g. AVs and non-AVs in one DRT fleet). This is something that should work fine with the selective insertion search (DRT), but we need to introduce a few not so big changes to make it possible.


nkuehnel:
Thanks @michalmac 
The mixed fleet was actually the reason for this change. RIght now we have a simpler approach of having two drt modes with separate vehicle types

Issue #1823, opened 2022-01-06
Put in roadType attributes for emission calculation when creating the network?


kt86:
Hi at all,

we (at VSP) will go more and more forward to make the emission calculation a standard analysis, which should be used as often as possible. For this, one needs the roadType set as an attribute for each link.

Currently, we are doing this ex-post after the simulation run, when starting the emission analysis. This has at least three disadvantages:

1. One has to change the `output_network.xml` and it is not documented in any file, which assignment of `roadType `was made for the analysis. --> Without the analysis script, the analysis is not reproducible.
2. The run script is a constant copy&paste work, which will go into more and more repositories... With each reproduction (and small adaption) we increase the probability, to do there any chances (un)intentionally, which then will spread further.
3. We are guessing the correct type by the number of lanes and the speed limit. In (semi) old networks, there is also no information from OSM available, which may help for a better assignment. (I know, that this is not done consistently in OSM as well: What is a motorway?)

--> Even if the OSM data is available now in more and more cases (see 2.) , I am wondering if one should do the tagging with the` hbefa roadType` in the links attribute when building up the scenario (creating the network for the scenario). Then it is available as input for the emissions analysis and is already there in all runs of the scenario, which helps to do consistent analyses. (A)

Another idea was to put some preprocessing in the emissions contrib which will do that assignment of roadTypes to the untyped network. (B) 
In my opinion, (B) has - besides the advantage that is done every time exactly in the same way - the disadvantage, that it is done independently of the local characteristics and knowledge (which should be there by the people who are building up a new scenario). Such a method can still also implemented for all people who want to have some "default" alternative(?) and to use it for older networks/scenarios.

@Janekdererste As said today, let's have some discussion about it.
@all: Please give us your thoughts :)

Janekdererste:
I am happy to do it. My suggestion would be the following:

- Put the OSM-Road-Type as well as the allowed Speed from OSM into the network as standard link attribute
- Guess the HBEFA Raod type from those attributes as pre-processing in the emissions contrib.
- Provide an option to have HBEFA road types proviede by the emission contrib user
- Fail if neither OSM nor HBEFA road types are provided

This way, assumptions about emission calculations won't spill into our osm package.

JWJoubert:
I vote for any approach where the software makes as little as possible _guessing_. If you want to use the `emissions` contrib, then you ought to be aware of the scenario building process (network creation and the required link attributes). 

[...]

PR #1833, opened 2022-01-10
Allowed speed tag

Janekdererste:
this handles the osm-parser part of #1823 

To figure out HBEFA road types in the emission contrib we want the original max speed tag from osm in the network.


contribs/osm/src/main/java/org/matsim/contrib/osm/networkReader/SupersonicOsmNetworkReader.java
```diff
@@ -299,6 +300,14 @@ private double getFreespeed(Map<String, String> tags, double linkLength, LinkPro
         }
     }
 
+    private double getMaxSpeed(Map<String, String> tags, double linkLength, LinkProperties properties) {
+        if (tags.containsKey(OsmTags.MAXSPEED)) {
+            return parseSpeedTag(tags.get(OsmTags.MAXSPEED), properties);
+        } else {
+            return LinkProperties.calculateSpeedIfNoSpeedTag(linkLength, properties);
```
Janekdererste, line 307: here I use the same logic as in getFreespeed. Alternatively, if ther is no speed tag available, one could also ommit to set the attribute, since it is always going to be the same as the freespeed.

rakow (approved):

PR #1835, opened 2022-01-11
Fix check in SharingRoutingModule

steffenaxer:
This PR ensures that routes that starts and ends at the same InteractionPoint are returned as null. It make no sense to use a sharing mode, insteade agent should walk.

### DIFF
--- a/contribs/sharing/src/main/java/org/matsim/contrib/sharing/service/SharingUtils.java
+++ b/contribs/sharing/src/main/java/org/matsim/contrib/sharing/service/SharingUtils.java
@@ -23,7 +23,7 @@ public class SharingUtils {
 	static public final String STATION_ID_ATTRIBUTE = "sharing:stationId";
 	static public final String SERVICE_ID_ATTRIBUTE = "sharing:service";
 
-	static public final String MODE_PREFIX = "sharing:";
+	static public final String MODE_PREFIX = "sharing_";
 
 	static public final double INTERACTION_DURATION = 60.0;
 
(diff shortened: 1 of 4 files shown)


contribs/sharing/src/main/java/org/matsim/contrib/sharing/service/SharingUtils.java
```diff
@@ -23,7 +23,7 @@
 	static public final String STATION_ID_ATTRIBUTE = "sharing:stationId";
 	static public final String SERVICE_ID_ATTRIBUTE = "sharing:service";
 
-	static public final String MODE_PREFIX = "sharing:";
+	static public final String MODE_PREFIX = "sharing_";
```
sebhoerl, line 26: Do we actually need this change here?

sebhoerl (changes requested):
Thanks, good catch! :) I have one little comment

contribs/sharing/src/main/java/org/matsim/contrib/sharing/service/SharingUtils.java
marecabo: `:` is not a valid char for filenames on Windows. If you run MATSim there, this mode prefix can appear in the filename of the leg mode plots and prevent them from being written.

sebhoerl: Ok, I see. We chose ":" because it clearly distinguishes from cases where we have, for instance, "car_passenger" as a mode. In a way, with ":" we tried to be a bit less "magic constant" style. 

But I guess then, for now, this is fine. We could even make this explicit and define a "mode" attribute per service in the service configuration to be completely transparent.

sebhoerl: Or rather, "sharingMode", because we already have a "mode" attribute... which we could call "routingMode" to be clear. Let's see later :)

sebhoerl (approved):

PR #1840, opened 2022-01-14
Add Qsim SharingVehicles that are used by customers and add reservation

steffenaxer:
This PR adds the following features:

- Add QSim vehicle ressources that are used by customers. Agents are not using their own fallback vehicle, they are now using a shared ressource.
- Usage of a QSim ressource requires a functional ressource reservation mechanism to avoid (vehicle is already in use error). This has been implemented.

Adding a QSim vehicle that enters and leaves links and that is a unique sharing vehicle allows us to create vehicle statistics of shared vehicles. Basic shared vehicle stats will be added in a later commit.


contribs/sharing/src/main/java/org/matsim/contrib/sharing/logic/SharingLogic.java
```diff
@@ -143,8 +154,17 @@ public boolean tryPickupVehicle(double now, MobsimAgent agent) {
 		Activity pickupActivity = (Activity) plan.getPlanElements().get(pickupActivityIndex);
 		Verify.verify(pickupActivity.getType().equals(SharingUtils.PICKUP_ACTIVITY));
 
-		// Find closest vehicle and hope it is at the current station / link
-		Optional<VehicleInteractionPoint> selectedVehicleInteraction = service.findClosestVehicle(agent);
+		SharingVehicle reservedVehicle = service.hasReservationElseNull(agent);
```
sebhoerl, line 157: I would prefer Optional<SharingVehicle> getSharedVehicle(agent);

contribs/sharing/src/main/java/org/matsim/contrib/sharing/logic/SharingLogic.java
```diff
@@ -122,6 +129,10 @@ public boolean tryBookVehicle(double now, MobsimAgent agent) {
 			// Insert new plan elements
 			plan.getPlanElements().addAll(bookingActivityIndex + 1, updatedElements);
 
+			Leg leg = ((Leg) mainElements.get(0));
```
sebhoerl, line 132: This is brittle, there can be more stuff than a leg. Maybe there is some ideas in the TripRouter on how to cleanly set the vehicle for a multi-stage trip. I guess "car with access and egress walk" is not the best example here, because it will probably not happen (except we can define a station as some "virtual" thing and the vehicles are parked in the links around that station). But since we want to be generic there could be all kind of things like "a car trip with a refueling activity in the middle" or whatever :)

contribs/sharing/src/main/java/org/matsim/contrib/sharing/logic/SharingLogic.java
```diff
@@ -197,21 +219,44 @@ public boolean tryPickupVehicle(double now, MobsimAgent agent) {
 						now, agent);
 				updatedElements.addAll(mainElements);
 
+				Leg leg = ((Leg) mainElements.get(0));
```
sebhoerl, line 222: See above

contribs/sharing/src/main/java/org/matsim/contrib/sharing/logic/SharingLogic.java
```diff
@@ -256,6 +301,9 @@ public void tryDropoffVehicle(double now, MobsimAgent agent) {
 			updatedElements.addAll(mainElements);
 			now = timeInterpretation.decideOnElementsEndTime(mainElements, now).seconds();
 
+			Leg leg = ((Leg) mainElements.get(0));
```
sebhoerl, line 304: See above

contribs/sharing/src/main/java/org/matsim/contrib/sharing/service/SharingService.java
```diff
@@ -8,17 +8,23 @@
...
+	IdMap<SharingVehicle, SharingVehicle> getVehicles();
 
-	public IdMap<SharingStation, SharingStation> getStations();
+	IdMap<SharingStation, SharingStation> getStations();
 
-	public void pickupVehicle(SharingVehicle vehicle, MobsimAgent agent);
+	void pickupVehicle(SharingVehicle vehicle, MobsimAgent agent);
 
-	public void dropoffVehicle(SharingVehicle vehicle, MobsimAgent agent);
+	void reserveVehicle(MobsimAgent agent,SharingVehicle vehicle);
```
sebhoerl, line 19: For ease of use, maybe adjust the order of the arguments to the other methods (first vehicle, then agent)

contribs/sharing/src/main/java/org/matsim/contrib/sharing/run/SharingQSimServiceModule.java
```diff
@@ -58,6 +59,17 @@ protected void configureQSim() {
 					specification.getVehicles(), network, serviceConfig.getMaximumAccessEgressDistance());
 		})).in(Singleton.class);
 
+
+		String agentSource = SharingUtils.getServiceMode(serviceConfig)+AGENT_SOURCE_SUFFIX;
+		addQSimComponentBinding(agentSource).toProvider(modalProvider(getter -> {
```
sebhoerl, line 64: Here you can use addModalComponent to avoid the "named component" and having AGENT_SOURCE_SUFFIX.

contribs/sharing/src/main/java/org/matsim/contrib/sharing/service/SharingUtils.java
```diff
@@ -64,6 +65,10 @@ public static QSimComponentsConfigurator configureQSim(SharingConfigGroup sharin
 		return components -> {
 			for (SharingServiceConfigGroup serviceConfig : sharingConfig.getServices()) {
 				components.addComponent(SharingModes.mode(getServiceMode(serviceConfig)));
+
```
sebhoerl, line 68: You should be able to remove this if you use addModalComponent above

sebhoerl (changes requested):
Really nice! I added some comments to make the code a bit cleaner and avoid the additional "named component".

contribs/sharing/src/main/java/org/matsim/contrib/sharing/logic/SharingLogic.java
steffenaxer: changed to getReservedVehicle

contribs/sharing/src/main/java/org/matsim/contrib/sharing/logic/SharingLogic.java
steffenaxer: Implemented setVehicle(Plan plan, List<? extends PlanElement> planElements, SharingVehicle vehicle) with IllegalStateException if more than 1 vehicleId was present

contribs/sharing/src/main/java/org/matsim/contrib/sharing/run/SharingQSimServiceModule.java
steffenaxer: Changed to addModalComponent

steffenaxer (commented):
All things have been adapted

sebhoerl (approved):

sebhoerl:
Looks good, if you could still update `getReservedVehicle` to return `Optional<SharedVehicle>` instead of `null` or not it would be perfect :)

steffenaxer:
Yes, will fix this too.

PR #1846, opened 2022-01-16
Freight: Store MATSim vehicle type during jsprit run

### LINKED ISSUE #1845: Freight: Information from (MATSim) vehicle types get lost due to the twice conversions
When running a freight run, some information from the input data are not available in the output (allVehicles - container), e.g. the `engineInformation`.

Since happens, because we are converting the (MATSim) vehicleType to a (jsprit) vehicle type for the jsprit run. After the jsprit run, we are recreating a  **new** (MATsim) vehicle type. But this only contains data, that were stored  in the jsprit vehType, because not all of our MATSim data has corresponding fields in the jsprit type.

Solution: There exists a container in the jsprit vehicleType, where one can store objects, which will not interact with jsprit --> use this for storing the MATSim vehicleType during the jsprit run and reuse it after the run.

kt86:
and (re)use it afterwards.

This makes it possible that the (MATSim) vehicleType will be available after the jsprit run.
Before, a new (MATSim) vehicleType was created based on the data of the (jsprit) vehicle type. That let to losses of information, that were not available in the jsprit vehicleType, e.g. engineInformation.

Con: If there are made any chances in the (jsprit) vehicleType, they get now lost. But I do not think that that happens anywhere in our workflow, because we are creating a MATSim vehType, which will be transformed to the jsprit one... After that step, there were usually no changes done to the jsprit type.

((Maybe we can add some checks, that the value are the same...))

Closes #1845

### DIFF
--- a/contribs/freight/src/main/java/org/matsim/contrib/freight/jsprit/MatsimJspritFactory.java
+++ b/contribs/freight/src/main/java/org/matsim/contrib/freight/jsprit/MatsimJspritFactory.java
@@ -261,9 +261,14 @@ public class MatsimJspritFactory {
 				Id.create(vehicleId, org.matsim.vehicles.Vehicle.class),
 				Id.create(vehicle.getStartLocation().getId(), Link.class));
 
-//		VehicleType carrierVehicleType = createCarrierVehicleType(vehicle.getType());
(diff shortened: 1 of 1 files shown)


kt86:
Closing this PR because the changes are included in #1849

PR #1859, opened 2022-01-25
Make hbefa road types explicit

### LINKED ISSUE #1823: Put in roadType attributes for emission calculation when creating the network?

Janekdererste:
This PR fixes #1823


Janekdererste:
@kt86 I started working on this. I think we need to think about how to write a meaningfull test for the ```EmissionModule```. Please have a look at the changes.

Janekdererste:
Had a review on zoom. Ready to merge.

PR #1867, opened 2022-02-02
Fix tempStorageCapacity to take into account capacity factor

nejigane:
The lower bound for storage capacity (`tempStorageCapacity`) should take into account flow capacity factor.


michalmac (approved):
Looks good to me. Thanks for contributing to matsim!

michalmac:
merged via https://github.com/matsim-org/matsim-libs/pull/1871

PR #1870, opened 2022-02-04
Speedup drt plot generation

michalmac:
For bigger scenarios, creation of scatter plots could take quite some time. In particular, the recently added "Maximum wait time" plot was taking a lot of time as often all samples (drt legs) have the same max wait time (X value).


michalmac:
Forgot to add: setting `autoSort` to false resolves the issue

PR #1882, opened 2022-02-16
Deploy matsim-release.zip to the maven repository

ksauzz:
Hi,
This PR deploys matsim-x.x-release.zip to the maven repository as releases and PR-labeled releases.

### background

I have used matsim-13.0-release.zip in our project, and would like to use recent changes/fixes on master branch. However, 14.x series of matsim-release.zip has not been published yet. Although I can build it, it would be happy if a PR release includes matsim-release.zip. 

Thanks

### DIFF
--- a/.github/workflows/deploy-on-pr-merge.yaml
+++ b/.github/workflows/deploy-on-pr-merge.yaml
@@ -49,7 +49,7 @@ jobs:
 
       - name: Publish jars to matsim maven repo
         # fail at end to deploy as many jars as possible
-        run: mvn deploy --batch-mode --fail-at-end -DskipTests -Dmaven.resources.skip=true -Dmaven.install.skip=true
+        run: mvn deploy -P release --batch-mode --fail-at-end -DskipTests -Dmaven.resources.skip=true -Dmaven.install.skip=true
         env:
           MAVEN_USERNAME: ${{ secrets.REPOMATSIM_USERNAME }}
           MAVEN_PASSWORD: ${{ secrets.REPOMATSIM_TOKEN }}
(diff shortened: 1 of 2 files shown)


michalmac:
Based on discussions from the time we were releasing 13.0 (@kainagel @mrieser, please correct me if I am wrong):
- we want to focus on releasing pure jars rather than zipped collections of jars+examples (prior to 13.0 we had lots of such zips, whereas for 13.0 we only released `matsim-13.0-release.zip` for the "core" matsim)
- we add the zips as assets to each release on GitHub (https://github.com/matsim-org/matsim-libs/releases) and not as artifacts in the maven repo
- currently, we do not add PR-labelled releases to releases on GitHub (maybe we should?)

@ksauzz What prevents you from using jars directly in your project?

ksauzz:
Thank you for the comment about the team decision for release assets. I understood the policy not to push zip files into the maven repository.

> What prevents you from using jars directly in your project?

In my usecase, we use matsim as an independent tool, not a dependency in a custom java project/repo.
So, in our pipeline process for a simulation, we just download matsim-13.0-release.zip, uncompress it, and run `java -jar matsim-13.jar` with 3rd party dependencies in libs/ dir included in the zip release. To use recent matsim-14-PRxxx.jar, we have to download matsim-14-PRxxx.jar and its all dependencies. It works to me if a fat jar or matsim-14-PRxxx-release.zip is published to somewhere.
BTW, I noticed we can use `mvn dependency:copy-dependencies` for this purpose just now. It's likely to work to me.

According to the release policy, this PR isn't suitable at least. Feel free to close this PR. Thanks.

Issue #1826, opened 2022-01-07
Histogram plots with actual simulation end time


sebhoerl:
Currently, I'm running 7-day taxi service simulations, which made me notice that all histogram plots go only until 30:00:00 by default and  all following trips are aggregated in the last bin (see image). Not a serious problem, but still, it may make sense to adapt the end time of the plot to the actual end time of the simulation.
![0 legHistogram_all](https://user-images.githubusercontent.com/647500/148584209-5f91ac5c-2b6e-4956-8c14-0a80ba73d65f.png)

kainagel:
Geez, all these implicit “magic numbers” in the code.  Please operate them out if you have time and patience to do so.

PR #1883, opened 2022-02-16
Adaptve histograms (for #1826)

nkuehnel:
A simple fix for #1826 which ensures that leg and shift histograms expand until the defined QSim end time. If no end time is provided, the previous default of 30hrs is kept.


nkuehnel:
This does not yet automatically expand to the actual QSim end time if no end time was defined in the config

PR #1886, opened 2022-02-17
Proposal for max absolute detour time for drt

nkuehnel:
Hey, this is a proposal for a max absolute detour time for drt requests such that max travel time becomes min(unsharedRideTime + maxAbsoluteDetour, maxTravelTimeAlpha * unsharedRideTime + maxTravelTimeBeta)

By default, maxAbsoluteDetour is set to positive infinity and therefore ignored.

### DIFF
--- a/contribs/drt/src/main/java/org/matsim/contrib/drt/routing/DrtRouteCreator.java
+++ b/contribs/drt/src/main/java/org/matsim/contrib/drt/routing/DrtRouteCreator.java
@@ -61,7 +61,7 @@ public class DrtRouteCreator implements DefaultMainLegRouter.RouteCreator {
 	 * @return maximum travel time
 	 */
 	static double getMaxTravelTime(DrtConfigGroup drtCfg, double unsharedRideTime) {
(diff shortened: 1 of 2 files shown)


michalmac (approved):
Thanks! Nice addition to the max travel time constraints.

luchengqi7:
Hello Nico! We are now trying to add the detour constraints. We just noticed this maxAbsoluteDetour constraint (added in this PR). This seems to replicate the case where alpha = 1 and beta = maxAbsoluteDetour. 

We are not sure what is the initial idea of this constraint? And maybe we can replace it with the new constraints we are now implementing.

nkuehnel:
@luchengqi7 the idea was that it should be possible to define a maximum absolute extra travel time. You are right it is essentially the same when alpha =1 and beta = maxAbsoluteDetour. But in most cases alpha and beta will be different and then for longer trips can theoretically experience quite a long detour because of the relative alpha component. The new constraint we're adding will again be a relative constraint, right?

michalmac:
> But in most cases alpha and beta will be different and then for longer trips can theoretically experience quite a long detour because of the relative alpha component.

Thanks. It make sense. I think we need to add some more explanation for the existing constraints (and maybe even improve the naming) while working on adding new ones.

PR #1888, opened 2022-02-21
Fix for allowing evs to leave charging queue

nkuehnel:
Hi,

in some of our scenarios vehicles may want to leave the queue at a charger without having started the charging activity. This can happen, for example, when it was initially planned to charge the vehicle during a driver shift break and the queueing time exceeds the break duration.

The ChargingWithQueueingLogic already allows to remove vehicles that are not yet plugged:
https://github.com/matsim-org/matsim-libs/blob/ff60f269b30dc6e57d170d259599880f25bc2685/contribs/ev/src/main/java/org/matsim/contrib/ev/charging/ChargingWithQueueingLogic.java#L94-L106

However, our simulations crashed because the ChargingEventSequenceCollector does not yet seem to be designed for this case. This is what happens:
1. a vehicle is queued at a charger -> the ChargingEventSequenceCollector takes note of the QueuedAtChargerEvent with an ongoing charging sequence
2. the vehicle is removed from the ChargingWithQueueingLogic because the driver has to resume service before being able to charge (or any other reason) -> there is no notification for the sequence collector and the vehicle will be kept with an ongoing sequence
3. at a later time step, the vehicle may wants to charge again at the same charger and is queued -> the sequence collector tries to handle the new QueuedAtChargerEvent and crashes, since the vehicle is still in an ongoing sequence:
https://github.com/matsim-org/matsim-libs/blob/ff60f269b30dc6e57d170d259599880f25bc2685/contribs/ev/src/main/java/org/matsim/contrib/ev/charging/ChargingEventSequenceCollector.java#L74-L78

This PR introduces a possible fix by also introducing a QuitQueueAtChargerEvent (or a better name) next to the QueueAtChargerEvent. This event is thrown when a vehicle is removed from the charger queue without having started charging. The ChargingEventSequenceCollector can then react by removing the vehicle from the ongoing sequences.

It is tested with a large scenario with a lot of charging activity. 

There is a caveat: if the vehicle is removed from the sequence collector, it will not be recorded as a "completed sequence". Not sure if that should be the intended behvavior..

(Sorry for all the review requests, @michalmac ) :)

### DIFF
--- a/contribs/ev/src/main/java/org/matsim/contrib/ev/charging/ChargingEventSequenceCollector.java
+++ b/contribs/ev/src/main/java/org/matsim/contrib/ev/charging/ChargingEventSequenceCollector.java
@@ -38,7 +38,7 @@ import com.google.common.base.Preconditions;
  * @author Michal Maciejewski (michalm)
  */
 public class ChargingEventSequenceCollector
-		implements QueuedAtChargerEventHandler, ChargingStartEventHandler, ChargingEndEventHandler {
+		implements QueuedAtChargerEventHandler, QuitQueueAtChargerEventHandler, ChargingStartEventHandler, ChargingEndEventHandler {
 
 	public static class ChargingSequence {
 		@Nullable //null if no queueing occurred
@@ -75,6 +75,17 @@ public class ChargingEventSequenceCollector
 	public void handleEvent(QueuedAtChargerEvent event) {
 		Preconditions.checkState(ongoingSequences.put(event.getVehicleId(), new ChargingSequence(event)) == null,
 				"Vehicle (%s) already at a charger", event.getVehicleId());
+	}
+
+	@Override
+	public void handleEvent(QuitQueueAtChargerEvent event) {
+		Preconditions.checkState(ongoingSequences.containsKey(event.getVehicleId()),
+				"Vehicle (%s) was not at charger (%s)", event.getVehicleId(), event.getChargerId());
+		Preconditions.checkState(ongoingSequences.get(event.getVehicleId()).queuedAtCharger != null,
+				"Vehicle (%s) was not queued at charger (%s)", event.getVehicleId(), event.getChargerId());
+		Preconditions.checkState(ongoingSequences.get(event.getVehicleId()).queuedAtCharger != null,
+				"Vehicle (%s) already is already plugged", event.getVehicleId(), event.getChargerId());
+		ongoingSequences.remove(event.getVehicleId());
 	}
 
 	@Override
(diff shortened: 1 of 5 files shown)


michalmac:
Thanks @nkuehnel . That's a really valid use case that we should support. I will look into it to give some more specific feedback.

> (Sorry for all the review requests, @michalmac ) :)

No worries. Thanks for keeping me updated and I am always happy to review PRs.

contribs/ev/src/main/java/org/matsim/contrib/ev/charging/ChargingEventSequenceCollector.java
```diff
@@ -77,6 +77,17 @@ public void handleEvent(QueuedAtChargerEvent event) {
...
 
+	@Override
+	public void handleEvent(QuitQueueAtChargerEvent event) {
+		Preconditions.checkState(ongoingSequences.containsKey(event.getVehicleId()),
+				"Vehicle (%s) was not at charger (%s)", event.getVehicleId(), event.getChargerId());
+		Preconditions.checkState(ongoingSequences.get(event.getVehicleId()).queuedAtCharger != null,
+				"Vehicle (%s) was not queued at charger (%s)", event.getVehicleId(), event.getChargerId());
+		Preconditions.checkState(ongoingSequences.get(event.getVehicleId()).chargingStartEvent == null,
+				"Vehicle (%s) already is already plugged", event.getVehicleId(), event.getChargerId());
+		ongoingSequences.remove(event.getVehicleId());
```
michalmac, line 88: Maybe rewrite the handler method in the following way so we call the map only once. I also added the addition of the sequence to the completed sequences (as you pointed out - otherwise we would lose this still valid sequence).
```java
		var sequence = ongoingSequences.remove(event.getVehicleId());
		Preconditions.checkState(sequence != null, "Vehicle (%s) was not at charger (%s)", event.getVehicleId(),
				event.getChargerId());
		Preconditions.checkState(sequence.queuedAtCharger != null, "Vehicle (%s) was not queued at charger (%s)",
				event.getVehicleId(), event.getChargerId());
		Preconditions.checkState(sequence.chargingStartEvent == null, "Vehicle (%s) is already plugged",
				event.getVehicleId(), event.getChargerId());
		completedSequences.add(sequence);
```

michalmac (approved):
Looks good to me!

I added a comment regarding the handler method - IMO it should be revised.

Maybe squash into a single commit.

contribs/ev/src/main/java/org/matsim/contrib/ev/charging/ChargingEventSequenceCollector.java
nkuehnel: Good point, thanks!

PR #1889, opened 2022-02-22
Rename sharing to shared_mobility contrib

balacmi:
This pull request updates the name of the contrib to shared_mobility to better explain its use


sebhoerl:
Why do we want to do this, has anybody requested it? I think there are already quite a bit of people using the code under the name `sharing`. This might lead to more confusion than benefit, maybe? ;-)

balacmi:
Nobody requested it, but I always felt that the name does not explain what is contained in the contrib. 

Why would it be a problem to change the name? Except for the snapshots there is no release that contains it yet.

sebhoerl:
For me it is fine, but if you think it is better to rename, I also don't mind. 
@marecabo @steffenaxer Will this pose any problems? Is anybody else using it who might have a problem with it?

steffenaxer:
Renaming should not be a major problem. I'm fine with this, but I think the difference between sharing and shared_mobility is not super obvious. Thus, there will be always some kind of explanation that describes what’s inside the box. However, shared_mobility is more precise than sharing. Go for it!

sebhoerl (approved):
Ok, so as said, it is fine for me as well. I'll put "approve".

michalmac:
Just wondering: the package name is changed from `sharing` to `shared_mobility`, but at the same time the contrib name remains unchanged (i.e. `sharing`). Is that intended? Should the contrib be also renamed accordingly?

sebhoerl:
Ah, didn't think of that. Yes, @balacmi this should also be changed in the top-level pom and in the local pom.

michalmac:
@balacmi please also rename modules in github action workflows

PR #1911, opened 2022-03-25
Adding a custom function in the demand estimation

luchengqi7:
I want to have a simple rebalancing strategy:
- Keeping 1 vehicle at each villiage (region) where there is at least one request during the time bin.
- More vehicles maybe send to the village if there is a lot of demands during the time bin (e.g., send 1 extra vehicle for every 5 requests)
- Send the vehicle some time in advance (e.g., 5 mintues) if the estimated demand increases (e.g., from 0 to 1, from 1 to 3)

I think we cannot acheive this with the current rebalanicng strategies. This PR is a rather quick fix and is not very elegant. So I just make it a draft. What do you think @michalmac ? Perhaps there is a better way to achieve this purpose?


contribs/drt/src/main/java/org/matsim/contrib/drt/optimizer/rebalancing/targetcalculator/DemandEstimatorWithCustomMappingFunction.java
```diff
@@ -0,0 +1,45 @@
...
+	}
+
+	/**
+	 * Custom transform function (simple example)
+	 */
+	private ToDoubleFunction<DrtZone> transform(double time) {
+		ToDoubleFunction<DrtZone> originalFunction = demandEstimator.getExpectedDemand(time, demandEstimationPeriod);
+
+		double lookAheadTime = 300; //TODO make this a variable (or read from config file: = rebalancing period)
+		ToDoubleFunction<DrtZone> originalFunctionWithLookAhead = demandEstimator.getExpectedDemand(time + lookAheadTime, demandEstimationPeriod);
```
michalmac, line 32: The current implementation of `ZonalDemandEstimator` discretizes predicted demand time-wise and always takes a value from one time bin (no interpolation is done). Therefore, if `lookAheadTime` is similar to or smaller than the time bin, we may not observe a significant changes in the behaviour.

What if we:
- add interpolation to the zonal demand estimation?
- (or alternatively) change the "max of two" function into the "average of two"? (line 35)

luchengqi7: The look ahead time can be helpful for the requests submitted at the start of the time bin (e.g. time bin size = 1800, then the trips submitted right after xx:00 and xx:30 may need to wait longer as there is not yet any rebalnacing vehicles approaching). If we have, say 5 minutes, looking ahead time, then those requests can wait shorter. This can be quite helpful especially for the rural area scenario, where the vehicles density are relatively thin. 

Yeah, I think interpolation is a good idea.

Alternatively, we can also make the demand estimator only return 0 or 1. This is also the original idea from Kai: keeping 1 vehilce in each village (with potential request). And as long as the rebalancing period is short engouh and the zonal system is not too large, then it should provide a good performance. And this can also make the "max" operation more suitable.

michalmac (commented):
Besides a theoretical comment: https://github.com/matsim-org/matsim-libs/pull/1911/files#r838298414, I have also a software-related comment:

Adding a new component to rebalancing means a lot of **additional work** (e.g. new enums, config groups/parameter sets) apart from the **necessary work** (i.e. coding that component and defining/overriding corresponding bindings).

Therefore, for short/mid-term approaches that we are playing around with, but we are not sure if we want to keep them long term, the best IMO is to only do the necessary part.

And it's not only about the amount of extra work, it's also about maintenance and backward compatibility: once we add something to the config group (or add values to enum types) it's much harder to remove it later.

luchengqi7:
Now this strategy is moved to the custom project: 
https://github.com/luchengqi7/project-space

Maybe in the future we will consider to formally set it up here.

PR #1913, opened 2022-03-26
Enable reading and writing of 3D coordinates in facilities files

marecabo:
This change adds support for reading and writing the z-component from/to `Coord`s in facilities files. 

Currently, nodes in networks and activities in plans support reading and writing of 3D-coordinates, but facilities do not. During runtime, it is possible to create an `ActivityFacility` with a 3D-coordinate, but once written to and read from a file, this component of the coordinate is currently lost.

This change adds
- `facilities_v2.dtd`
- `FacilitiesReaderMatsimV2`
- `FacilitiesWriterV2`

based on the corresponding *V1 files with a bit code added to support the third coordinate component.


Janekdererste:
I think this makes sense! Thanks for doing this.

Issue #1916, opened 2022-03-28
Writing & Reading persons `vehicles` attributes from XML


Nitnelav:
I'm using the emissions contrib and to do so I generated a fleet of vehicles for my synethic population.

Then, I'm using `QSimConfigGroup.VehiclesSource.fromVehiclesData` and I set the `vehiclesFile` parameter.
It's quite unclear (and undocumented?), but then vehicles must be assigned to persons for every mode.
That means writing something like that : 

```java
      if (config.qsim().getVehiclesSource() == QSimConfigGroup.VehiclesSource.fromVehiclesData) {
          for (Person person : scenario.getPopulation().getPersons().values()) {
              Id<Vehicle> vehicleId = Id.createVehicleId(person.getId());
              Map<String, Id<Vehicle>> modeVehicle = new HashMap<>();
              modeVehicle.put("car", vehicleId);
              VehicleUtils.insertVehicleIdsIntoAttributes(person, modeVehicle);
          }
      }
```

So far so good.

 - Problem : 
But now I want to assign the vehicles to persons directly as I generate my vehicle fleet, in python (using https://github.com/eqasim-org/ile-de-france) outside of matsim java code. 
That means adding the `vehicles` attribute to persons directly in the XML file. But the  `vehicles` attribute is actually a `Map<String, Id<Vehicle>>` and there is no [AttributeConverter](https://github.com/matsim-org/matsim-libs/blob/master/matsim/src/main/java/org/matsim/utils/objectattributes/AttributeConverter.java) for this.

 - Solution A :
There is a `Map<String, String>` [Converter](https://github.com/matsim-org/matsim-libs/blob/master/matsim/src/main/java/org/matsim/utils/objectattributes/attributeconverters/StringStringMapConverter.java) that should be straightforward to use for vehicles.
That would give us : 
```xml
	<person id="1030955">
		<attributes>
			<attribute name="age" class="java.lang.Integer">61</attribute>
			<attribute name="sex" class="java.lang.String">m</attribute>
			<attribute name="vehicles" class="java.util.Collections$UnmodifiableMap">{"car":"1030955"}</attribute>
		</attributes>
		<plan selected="yes">
			<activity type="home" link="180085" facility="home_458069" x="352759.54" y="6688605.21" >
			</activity>
		</plan>
	</person>
``` 

 - Solution B : 
The "vehicles that belong to person X" seems to be a core element of an agent definition, it would make sense to add it in its own WML tags like : 
```xml
[...]

PR #1922, opened 2022-03-30
writing and reading persons vehicles attributes from XML

Nitnelav:
an attempt to bring a solution to https://github.com/matsim-org/matsim-libs/issues/1916


sebhoerl:
:+1: Any objections to that?

The only thing I see is that with population files we then often have incompatibilities in running projects. Can we have a `@Deprecated` fallback that detects whether the attribute is a `Map<String, String>` and then converts it to `PersonVehicles` before processing it?

Nitnelav:
I'm 99.99% sure the attribute `vehicles`, previous to this patch, could not be anything else than a `Map<String, Id<Vehicle>>` , and so was never written to (or red from) any population XML file.
And as I kept the `VehicleUtils.insertVehicleIdsIntoAttributes()` usage the same, I actually don't think there is anything to depreciate. Am I missing something ?

Janekdererste:
>  was never written to (or red from) any population XML file

My guess would be that you are the first who deemded to use this feature. 

Recently we were leaning towards removing automagic. Maybe it is sufficient to crash when the cast to ```PeronsVehicles``` failes.

sebhoerl:
Ah yes, so the xml's in the unit tests were just to compare the output format, right? So I guess this is fine then

Nitnelav:
Yes indeed, the vehicles are always exported in XML now as they are assigned by default, that's why I had to update some tests that rely on comparing the desired pop.xml file with a reference one.

Nitnelav:
We've been updating this PR branch for a while now, nothing new since the last 12 days, can you merge it ?

Nitnelav:
No worries, I can check that on my todo list now, cheers.

PR #1926, opened 2022-03-31
update comments in PrepareForSimImpl

simei94:


matsim/src/main/java/org/matsim/core/controler/PrepareForSimImpl.java
```diff
@@ -114,20 +114,11 @@ public void run() {
 		 * own single-mode network. However, this assumes that the main mode is car - which PersonPrepareForSim also does. Should
```
Janekdererste, line 114: The comment by dobler is outdated because of your change as well ;-)

PR #1928, opened 2022-04-04
Introduce state profiles for shared mobility

steffenaxer:
Similar to DRT, this PR adds time profiles for vehicle states. The states of shared vehicles are idle, reserved and booked. 
grafik


sebhoerl (changes requested):
Really nice! :) I would just replace all the `Optional.ofNullable(null)` with `Optional.empty()`

steffenaxer:
Fixed and thanks for review!

sebhoerl (approved):

PR #1934, opened 2022-04-07
Post-process option MATSim Application

rakow:
This PR adds the possibility to overwrite a `preparePostProcessing` method in a `MATSimApplication`.
This method may return a list of `MATSimAppCommand` instances that will be run automatically after the simulation has finished.


Janekdererste:
This actually makes much more sense than what I wanted to do.

Janekdererste (approved):

Issue #1933, opened 2022-04-06
Travel time observation: discretization issue


luchengqi7:
Hi @mrieser, I have a small question regarding the travel time observation script. For links that are not used by any vehicle (of the mode), the travel time observation is returning the raw values (lengh/free speed) of the free speed travel time. I think it may be a good idea if we update this so that it match the Qsim discrete time step (see QsimFreeSpeedTravelTime). 

https://github.com/matsim-org/matsim-libs/blob/e28f17a6df119222dd4dc0af81dae5f84008a090/matsim/src/main/java/org/matsim/core/trafficmonitoring/TravelTimeDataArray.java#L115

luchengqi7:
@michalmac I think this is the reason for the fluctuation in the routing of DRT between different iterations. As the unused links are getting shorter travel time (i.e., without any rouding up). The travel time are then changing constantly because of the travel time estimation alpha, even though we set the capcity to a very large value.

sebhoerl:
@luchengqi7 I think we already had this discussion a while ago. This also becomes relevant if we, for instance, want to see if a predicted travel time (in the discrete mode choice model) reflects well the experienced travel time afterwards. I think Kai's position was to leave it as is (because it will introduce a lot of linked effects, as you seem to see in your PR) and only adjust it for specific scenarios when we need this to be consistent specifically. Then you can simply use the QSimFreeSpeedTravelTime as is used already in the DVRP/DRT benchmarks, I think. Not sure if anything has changed regarding this topic :-)

sebhoerl:
What is true is that we might fix the fact that the "baseline" travel time is hardcoded in  `TravelTimeDataArray.java`. It should probably use the currently bound `TravelTime` for the specific modes?

michalmac:
As @sebhoerl already mentioned, this is the way we ran dvrp benchmarks in the past: https://github.com/matsim-org/matsim-libs/blob/63284a47c5e99d86d21094ee97b65bba3c6ea722/contribs/dvrp/src/main/java/org/matsim/contrib/dvrp/benchmark/DvrpBenchmarks.java#L41

This ensures you will always get the same estimated TT values in every iteration.

BTW. You can also read them from an output of an existing simulation instead of going for `QSimFreeSpeedTravelTime`.

michalmac:
Regarding this line:
[...]

PR #1935, opened 2022-04-07
Update Array based travel time data

luchengqi7:
As mentioned in the issue #1933, the travel time observer does not return the correct value for the travel time for the unused links. This difference is due to the discrete QSim time step (see QSimFreeSpeedTravelTime vs FreeSpeedTravelTime). This fix incorporate the QSim config in the Travel Time data construction, such that the travel time for the unused link can be calculated more accurately.

In the future, we should also consider fixing the map based travel time data.


luchengqi7:
Since a lot of other code is depending on this, it is probably not that easy to just fix it directly.

PR #1938, opened 2022-04-08
Less restrictive option for subtour mode-choice

rakow:
Adds a behavior option to subtour mode-choice called `betweenAllLessConstraints`.

With this option, non-closed subtour are also considered as mass-conserving if they span the whole day.
Agents will also always able to freely changes to one mode for the whole day. 
There is also a new analysis class printing some statistic about the subtours to identify if agents don't have a choice set.

### DIFF
--- a/matsim/src/main/java/org/matsim/core/config/groups/SubtourModeChoiceConfigGroup.java
+++ b/matsim/src/main/java/org/matsim/core/config/groups/SubtourModeChoiceConfigGroup.java
@@ -36,6 +36,8 @@ public final class SubtourModeChoiceConfigGroup extends ReflectiveConfigGroup {
 	public final static String MODES = "modes";
 	public final static String CHAINBASEDMODES = "chainBasedModes";
 	public final static String CARAVAIL = "considerCarAvailability";
+	public final static String SINGLE_PROBA = "probaForRandomSingleTripMode";
+
 	private static final String BEHAVIOR = "behavior";
 	
 	private String[] chainBasedModes = new String[] { TransportMode.car, TransportMode.bike };
@@ -100,6 +102,7 @@ public final class SubtourModeChoiceConfigGroup extends ReflectiveConfigGroup {
 		comments.put(MODES, "Defines all the modes available, including chain-based modes, seperated by commas" );
 		comments.put(CHAINBASEDMODES, "Defines the chain-based modes, seperated by commas" );
 		comments.put(CARAVAIL, "Defines whether car availability must be considered or not. A agent has no car only if it has no license, or never access to a car" );
+		comments.put(SINGLE_PROBA, "Defines the probability of changing a single trip for a unchained mode instead of subtour.");
 		{
 			StringBuilder msg = new StringBuilder("Only for backwards compatibility.  Defines if only trips from modes list should change mode, or all trips.  Options: ");
 			for ( SubtourModeChoice.Behavior behavior : SubtourModeChoice.Behavior.values() ) {
@@ -146,10 +149,13 @@ public final class SubtourModeChoiceConfigGroup extends ReflectiveConfigGroup {
 	public final SubtourModeChoice.Behavior getBehavior() {
 		return this.behavior ;
 	}
-	
+
+	@StringGetter(SINGLE_PROBA)
 	public double getProbaForRandomSingleTripMode() {
 		return this.probaForRandomSingleTripMode;
 	}
+
+	@StringSetter(SINGLE_PROBA)
 	public void setProbaForRandomSingleTripMode(double probaForRandomSingleTripMode) {
 		this.probaForRandomSingleTripMode = probaForRandomSingleTripMode;
 	}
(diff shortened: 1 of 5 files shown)


matsim/src/main/java/org/matsim/core/population/algorithms/ChooseRandomLegModeForSubtour.java
```diff
@@ -119,13 +129,49 @@ public ChooseRandomLegModeForSubtour(
...
+	/**
+	 * Return whether a subtour is mass conserving according to configuration.
+	 */
+	public boolean isMassConserving(final Subtour subtour) {
+
+		// non-closed subtours without any parent do not necessarily violate mass conservation
+		if (behavior == SubtourModeChoice.Behavior.betweenAllLessConstraints &&
+				!subtour.isClosed() && subtour.getParent() == null) {
+			return true;
+		}
```
marecabo, line 165: This assumes that all non-closed root subtours are mass conserving, which I think, is not the case:

Here, the car stays parked at work but is later picked up at other, which makes the non-closed root subtour not mass-conserving. Is that deliberate? 
```
            S
           | |
           | |
        walk walk
           | |
           | |
H ---car--- W ---pt--- O ---car--- L
```


In this example, the mass-conserving property depends on the assumption, where vehicles of chain-based modes are initially parked. I suspect, that MATSim in general places chain-based vehicles at the first activity by default, so this should also not be mass-conserving?
```
           S
          | |
          | |
       walk walk
          | |
          | |
H ---pt--- W ---pt--- O ---car--- L
```

rakow: That is true that your first example is not mass conserving. The intention was rather to allow subtour mode choice to change it anyway. Nevertheless, I have changed this method back to avoid any confusion in this regard.

The second example would depend on the assumptions, but the goal is to be less restrictive. If this is used together with the option to run change single trip mode with a certain probably the mass conservation will be violated anyway, which is fine for the intended use-case.

matsim/src/main/java/org/matsim/core/config/consistency/VspConfigConsistencyCheckerImpl.java
```diff
@@ -332,6 +332,22 @@ public void checkConsistency(Config config) {
...
+			if ( DefaultStrategy.SubtourModeChoice.equals( it.getName() ) ) {
+				usingSMC = true ;
+				break ;
+			}
+		}
+		if (usingSMC) {
+			if (config.subtourModeChoice().getProbaForRandomSingleTripMode() <= 0) {
+				problem = true;
+				System.out.flush();
+				log.log(lvl, "SubTourModeChoice 'probaForRandomSingleTripMode' is 0. This probability should be at least above 0.5");
```
tschlenther, line 347: I am uncertain whether we really want to suggest a value higher than 50%

matsim/src/main/java/org/matsim/core/config/groups/SubtourModeChoiceConfigGroup.java
```diff
@@ -36,6 +36,8 @@ public final class SubtourModeChoiceConfigGroup extends ReflectiveConfigGroup {
 	public final static String MODES = "modes";
 	public final static String CHAINBASEDMODES = "chainBasedModes";
 	public final static String CARAVAIL = "considerCarAvailability";
+	public final static String SINGLE_PROBA = "probaForRandomSingleTripMode";
+
```
tschlenther, line 40: further down, the default value is set to 0. In a comment within the same line, it is said that this is done in order to maintain backwards compatibility but should normally be changed. I think this is the point in time where we should consider to conduct the change. Might break a few tests, though.
On the other hand, you incorporated at least a warning the the VspConfigConsistencyChecker

matsim/src/main/java/org/matsim/core/population/PersonUtils.java
```diff
@@ -102,13 +102,22 @@ public static String getCarAvail(Person person) {
         return (String) person.getAttributes().getAttribute(CAR_AVAIL);
     }
 
+    /**
+     * Car is not available if Person.getLicense gives "no" or if
```
tschlenther, line 106: the function actually uses an `AND` operator

contribs/application/src/main/java/org/matsim/application/prepare/population/FixSubtourModes.java
```diff
@@ -0,0 +1,128 @@
...
+
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.SplittableRandom;
+import java.util.stream.Collectors;
+
+@CommandLine.Command(name = "fix-subtour-modes", description = "Fix modes for subtours that contain chain and non-chain based modes.", showDefaultValues = true)
```
tschlenther, line 22: maybe mention that the subtour is set to a random mode found within it's current set of modes?

matsim/src/main/java/org/matsim/core/population/PersonUtils.java
rakow: The comment expresses the negation of the code, which is imo easier to understand as text.

Issue #1942, opened 2022-04-11
Release 14.0


Janekdererste:
We need to tag a new release. For the core repository this means:

## 0. Pre-release 14.0-RC1
- [x] Create a pre-release `14.0-RC1` for branch `master`.
   This step is run to make sure the release workflow is up to date: https://github.com/matsim-org/matsim-libs/blob/master/.github/workflows/deploy-on-release-created.yaml. It will be later used for releasing `14.0` (step 3)

## 1. Prepare the master branch for `15.0-SNAPSHOT`
- [x] Check out the latest master branch and change version numbers in all poms to `15.0-SNAPSHOT`
    ` mvn versions:set -DnewVersion=15.0-SNAPSHOT`
- [x] Push it, make a PR, and merge it.
   From this moment on, the master branch is a development branch for the future version `15.0`.

## 2. Prepare the `14.x` branch
- [x] Create `14.x` branch that points to the last commit with the version `14.0-SNAPSHOT` on `master`.
- [x] Update github workflows by replacing `master` with `14.x` in the following lines:
   - https://github.com/matsim-org/matsim-libs/blob/master/.github/workflows/verify-push.yaml#L6
   - https://github.com/matsim-org/matsim-libs/blob/master/.github/workflows/deploy-on-pr-merge.yaml#L6
   - https://github.com/matsim-org/matsim-libs/blob/master/.github/workflows/deploy-on-pr-merge.yaml#L9
- [x] Change version numbers in all poms to `14.1-SNAPSHOT`
    ` mvn versions:set -DnewVersion=14.1-SNAPSHOT`
- [x] Push `14.x`
- [x] Check if the branch protection rules apply for `14.x`.
   From this moment on, updates and fixes to `14.x` will go through the same PR review and release process like on `master`.

## 3. Release `14.0`
- [x] Create a tag `14.0` that points exactly to the last commit with the version `14.0-SNAPSHOT` on `14.x`.
- [x] Push the tag: `git push origin 14.0`
- [x] Create a release for the `14.0` tag. This will trigger the release workflow that deploys the `14.0` jars to the matsim maven repo and also builds and attaches `matsim-14.0-release.zip` under the `14.0` release assets.

@michalmac please add more things if I forgot anything

michalmac:
> ## 0. Release 14.0-RC1
> * [ ]  Create a release tag `14.0-RC1` for branch `master`.
[...]

PR #1944, opened 2022-04-11
change matsim version to 15.0-SNAPSHOT

Janekdererste:
In preparation of release 14.0 all version numbers in all pom files were bumped to 15.0-SNAPSHOT

This is part of #1942


contribs/vsp/pom.xml
```diff
@@ -15,54 +15,54 @@
 		<dependency>
 			<groupId>org.matsim.contrib</groupId>
 			<artifactId>multimodal</artifactId>
-			<version>14.0-SNAPSHOT</version>
+            <version>15.0-SNAPSHOT</version>
```
michalmac, line 18: This file has some formatting issues as tabs were replaced with spaces.

PR #1948, opened 2022-04-14
Modify preplannedDrtControlerCreator

luchengqi7:
Remove the rebalancing parameter set when preplanned drt controler is created. This is needed because there is a check later. It will fail if the rebalancing parameter is still there (even if we have bound it to the NoRebalancingStrategy)

Alternatively, we can modify the pre-condition check. We will also accept "NoRebalancingStrategy" in the line 93 at PreplannedDrtModeModule. But I dont know how to do that.
https://github.com/matsim-org/matsim-libs/blob/6659f0bb01b46eba88b2d0d7c24d4d055ba53c4b/contribs/drt/src/main/java/org/matsim/contrib/drt/extension/preplanned/run/PreplannedDrtModeModule.java#L93

### DIFF
--- a/contribs/drt/src/main/java/org/matsim/contrib/drt/extension/preplanned/run/PreplannedDrtControlerCreator.java
+++ b/contribs/drt/src/main/java/org/matsim/contrib/drt/extension/preplanned/run/PreplannedDrtControlerCreator.java
@@ -79,6 +79,7 @@ public final class PreplannedDrtControlerCreator {
 				log.warn("The rebalancing parameter set is defined for drt mode: "
 						+ drtConfigGroup.getMode()
 						+ ". It will be ignored. No rebalancing will happen.");
+				drtConfigGroup.removeParameterSet(drtConfigGroup.getRebalancingParams().get());
 				controler.addOverridingQSimModule(new AbstractDvrpModeQSimModule(drtConfigGroup.getMode()) {
 					@Override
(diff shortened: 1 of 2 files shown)


michalmac:
I do not think we should manipulate the config at this point. We should simply fail.

Alternatively, we could override the binding for the rebalancing strategy, but it again a bit too much of automagic (too me). There's already a lot of things going on around dependency injection.

My suggestion: if you do not want to manually modify the input config file, maybe move this new line: 
```java
drtConfigGroup.removeParameterSet(drtConfigGroup.getRebalancingParams().get());
```
outside `PreplannedDrtControlerCreator.createControler()`. Modifying the config just after loading it from the file seems a good location for this.

luchengqi7:
Okay, I see. I will move it.

michalmac:
> Okay, I see. I will move it.

Sorry. I was not specific enough. I would like to keep the "official" example the way it is.

I meant modifying the config in our private scripts that we use for running simulations (in case we want to use exactly the same config for the standard DRT and for the pre-planned one).

michalmac:
> Okay, I see. I will move it.

Sorry. I was not specific enough. I would like to keep the "official" example the way it is.

I meant modifying the config in our private scripts that we use for running simulations (in case we want to use exactly the same config for the standard DRT and for the pre-planned one).

PR #1963, opened 2022-05-04
Introduce simple DRT request acceptance

michalmac:
This PR tries to build the basis for #1954 and similar. The idea is that once the best insertion is selected for a given `DrtRequest`, this proposed insertion needs to be accepted/confirmed/finalised using a mode-specific `DrtOfferAcceptor`. This opens up many opportunities for shaping the final `AcceptedDrtRequest` (e.g. limiting the buffer of the pickup/dropoff times as in #1954). It is also possible to not accept the offer.

There are no special assumptions around `AcceptedDrtRequest`, such as which side accepts the offer (passenger/service provider or both) and how the acceptance process looks like. The purpose is to keep it as simple as possible for the time being (until a more comprehensive prebooking approach is introduced).


nkuehnel:
Nice, thank you very much! This approach is really helpful

PR #1965, opened 2022-05-06
DRT - promised pickup time window offer acceptor

nkuehnel:
update of https://github.com/matsim-org/matsim-libs/pull/1954
in response to
https://github.com/matsim-org/matsim-libs/pull/1963

Left out the config parameter for now.
@michalmac I wasn't sure about the package location, it's not necessarily part of an extension.

### DIFF
--- a/contribs/drt/src/main/java/org/matsim/contrib/drt/passenger/PromisedPickupTimeWindowDrtOfferAcceptor.java
+++ b/contribs/drt/src/main/java/org/matsim/contrib/drt/passenger/PromisedPickupTimeWindowDrtOfferAcceptor.java
@@ -1,0 +1,21 @@
+package org.matsim.contrib.drt.passenger;
+
+import java.util.Optional;
+
+/**
+ * @author nkuehnel / MOIA
+ */
+public final class PromisedPickupTimeWindowDrtOfferAcceptor implements DrtOfferAcceptor {
+	private final double promisedPickupTimeWindow;
+
+	public PromisedPickupTimeWindowDrtOfferAcceptor(double promisedPickTimeWindow) {
+		this.promisedPickupTimeWindow = promisedPickTimeWindow;
+	}
+
+	@Override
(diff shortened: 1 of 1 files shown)


michalmac:
Nice. LGTM

>Left out the config parameter for now.

Makes sense. If we see more use cases, we can add it to the standard config. Maybe we should even group all the time constraint params into one param set. For the time being, you could try out the new possibility of subclassing the drt config.

> @michalmac I wasn't sure about the package location, it's not necessarily part of an extension.

👍

contribs/drt/src/main/java/org/matsim/contrib/drt/passenger/PromisedPickupTimeWindowDrtOfferAcceptor.java
```diff
@@ -0,0 +1,21 @@
...
+
+	public PromisedPickupTimeWindowDrtOfferAcceptor(double promisedPickTimeWindow) {
+		this.promisedPickupTimeWindow = promisedPickTimeWindow;
+	}
+
+	@Override
+	public Optional<AcceptedDrtRequest> acceptDrtOffer(DrtRequest request, double departureTime, double arrivalTime) {
+		double updatedPickupTimeWindow = Math.min(departureTime
+				+ promisedPickupTimeWindow, request.getLatestStartTime());
+		return Optional.of(AcceptedDrtRequest.newBuilder().latestStartTime(updatedPickupTimeWindow).build());
```
michalmac, line 19: For the accepted request to become fully initialised, you need to add:
```java
				.request(request)
				.earliestStartTime(request.getEarliestStartTime())
				.latestArrivalTime(request.getLatestArrivalTime())
```

nkuehnel: Ah yes, forgot about that, thanks!

nkuehnel:
> Maybe we should even group all the time constraint params into one param set

👍

PR #1967, opened 2022-05-09
ResolveGridCoordinates: add warning if network contains non-car links.

tschlenther:
ResolveGridCoordinates sets the linkId for activities, similarly to `PersonPrepareForSim`.
This PR proposes to dump a warn message, when a linkId for an activity shall be set to a non-car link.
We could also think of throwing an exception instead.

In the past, `PersonPrepareForSim` considered only links with car allowed. This changed with #1917  but might actually be reversed. The reason was that, with car being the typical network mode (in old scenarios), activities need to be accessible, also when access and egress legs are not simulated (i.e. for backwards compatibility). Most likely, the linkId in the activity can be set to any link that does not allow for car but still can be reached by car, when accessand egress routing is enabled. However, we need to take care of backwards compatibility. Moreover, we ran into problems with the sharing contrib, when not setting the linkId to a car link, even with access/egress routing enabled.

### DIFF
--- a/contribs/application/src/main/java/org/matsim/application/prepare/population/ResolveGridCoordinates.java
+++ b/contribs/application/src/main/java/org/matsim/application/prepare/population/ResolveGridCoordinates.java
@@ -4,6 +4,7 @@ import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.locationtech.jts.geom.Geometry;
 import org.matsim.api.core.v01.Coord;
+import org.matsim.api.core.v01.TransportMode;
 import org.matsim.api.core.v01.network.Link;
 import org.matsim.api.core.v01.network.Network;
 import org.matsim.api.core.v01.population.*;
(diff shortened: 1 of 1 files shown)


contribs/application/src/main/java/org/matsim/application/prepare/population/ResolveGridCoordinates.java
```diff
@@ -106,6 +111,12 @@ public Integer call() throws Exception {
 
 						if (network != null) {
 							Link link = NetworkUtils.getNearestLink(network, act.getCoord());
+							if(link.getAllowedModes().contains(TransportMode.car)){
```
rakow, line 114: I think you wanted to check if a link does con contain car ?

rakow (changes requested):

contribs/application/src/main/java/org/matsim/application/prepare/population/ResolveGridCoordinates.java
tschlenther: correct. thanks for catching the missing `!`. However, didn't we agree on enforcing to use a car network more strongly? Should I change the warn message to an exception? IllegalArgumentException?

PR #1968, opened 2022-05-09
Add WaitForStopTask to DrtEventsReaders's list

luchengqi7:
The newly added WaitForStopTask is not yet in the list of the recognized DRT activity in the DrtEventsReaders. We need to add it in order to run DRT events reader properly.

### DIFF
--- a/contribs/drt/src/main/java/org/matsim/contrib/drt/util/DrtEventsReaders.java
+++ b/contribs/drt/src/main/java/org/matsim/contrib/drt/util/DrtEventsReaders.java
@@ -26,6 +26,7 @@ import java.util.List;
 import java.util.Map;
 
 import org.matsim.contrib.drt.extension.edrt.schedule.EDrtChargingTask;
+import org.matsim.contrib.drt.extension.preplanned.optimizer.WaitForStopTask;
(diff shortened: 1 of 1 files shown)


michalmac:
This task type is (currently) available only in one of extensions. That's way it should not leak into the standard drt.

@luchengqi7 @nkuehnel I have been thinking about splitting the drt contrib into drt and drt-extensions. This way we can easily notice cases where the standard drt depends on drt-extensions. Avoiding such reverse dependencies may result in a bit of extra work short term, but I think it will also lead to design decisions which are better long term. If no one objects, I will do the split in the coming days.

nkuehnel:
> I have been thinking about splitting the drt contrib into drt and drt-extensions
no objection from our side :)

michalmac:
#1972 made the events processing more open to non-standard drt task types. When reading events from file, you need to provide a list of non-standard types:
```java
DrtEventsReaders.createEventsReader(eventManager, MySpecialTaskType.class, MyVerySpecialTaskType.class);
```
@luchengqi7 will that work for you?

luchengqi7:
Yes, it works for me. Thanks a lot!

michalmac:
> Yes, it works for me. Thanks a lot!

I will then close this PR.

PR #1978, opened 2022-05-12
fix problem with duplicate vehicle Ids

rewertvsp:
fix problem with duplicate vehicle Ids while the creation of vehicles for the carriers

### DIFF
--- a/contribs/application/src/main/java/org/matsim/freightDemandGeneration/CarrierReaderFromCSV.java
+++ b/contribs/application/src/main/java/org/matsim/freightDemandGeneration/CarrierReaderFromCSV.java
@@ -477,6 +477,7 @@ public final class CarrierReaderFromCSV {
 				}
 			}
 			for (String singleDepot : singleNewCarrier.getVehicleDepots()) {
+				int countDepots = 2;
 				for (String thisVehicleType : singleNewCarrier.getVehicleTypes()) {
 					VehicleType thisType = carrierVehicleTypes.getVehicleTypes()
 							.get(Id.create(thisVehicleType, VehicleType.class));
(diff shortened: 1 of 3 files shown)


contribs/application/src/main/java/org/matsim/freightDemandGeneration/CarrierReaderFromCSV.java
```diff
@@ -463,20 +457,21 @@ static void createNewCarrierAndAddVehilceTypes(Scenario scenario,
 				carriers.addCarrier(thisCarrier);
 			}
 			if (singleNewCarrier.getVehicleDepots() == null)
-				singleNewCarrier.setVehicleDepots(new String[] {});
-			while (singleNewCarrier.getVehicleDepots().length < singleNewCarrier.getNumberOfDepotsPerType()) {
-				Random rand = new Random();
+				singleNewCarrier.setVehicleDepots(new ArrayList<String>());
+			Random rand = new Random(4711);
```
michalmac, line 461: Maybe initialise `rand` with a seed like `singleNewCarrier.name.hashCode()`. Otherwise you will always get the same link sequence for all carriers, which may not be desired if you want to have them more spread. `singleNewCarrier.name.hashCode()` makes it also reproducible

contribs/application/src/main/java/org/matsim/freightDemandGeneration/CarrierReaderFromCSV.java
```diff
@@ -485,10 +480,18 @@ static void createNewCarrierAndAddVehilceTypes(Scenario scenario,
...
 								thisType.getId().toString() + "_" + thisCarrier.getId().toString() + "_" + singleDepot
 										+ "_start" + singleNewCarrier.getVehicleStartTime() + "_" + (i + 1),
-								Vehicle.class), Id.createLinkId(singleDepot), thisType)
+								Vehicle.class);
+						while (carrierCapabilities.getCarrierVehicles().containsKey(vehilcelId)) {
+							vehilcelId = Id.create(thisType.getId().toString() + "_" + thisCarrier.getId().toString()
+									+ "_" + singleDepot + "_V" + countDepots + "_start"
+									+ singleNewCarrier.getVehicleStartTime() + "_" + (i + 1), Vehicle.class);
+							countDepots++;
+						}
```
michalmac, line 492: Since now the vehicle depots are unique links, you can revert this change (starting from line 474).

michalmac (approved):
Thanks for fixing it. Looks good to me. I left 2 comments that you may consider.

contribs/application/src/main/java/org/matsim/freightDemandGeneration/CarrierReaderFromCSV.java
rewertvsp: Thanks for this idea. I will change it

contribs/application/src/main/java/org/matsim/freightDemandGeneration/CarrierReaderFromCSV.java
rewertvsp: Now I added the possibility to add a depot on an existing depot location, if no other possible link is found. Thats why I do not revert this change

PR #1979, opened 2022-05-13
Small bug fixing in the output message

luchengqi7:
I guess the second output message should be "timeFromScheduler"


michalmac (approved):
LGTM. Thanks!

PR #1983, opened 2022-05-15
Taxi as drt extension

michalmac:
This PR brings the taxi contrib closer to the DRT contrib.

The most important changes:
- Taxi uses the drt guice modules to set up all core bindings (routing, request creation, etc.) except for the optimiser specific bindings.
- The input taxi config is converted into the drt config, which is then passed to the drt guice modules.
- TaxiRequest is replaced by DrtRequest
- DrtRoute is used instead of generic route
- Taxi trip routing (the replanning phase) now uses the drt routing module (so we can have access egress walks, stops, service areas etc.)

What has not changed (but could/should be addressed in the future):
- At the optimiser level, the drt and taxi are still two completely separate things. They use different task types (**taxi**: pickup, dropoff, empty drive, and occupied drive vs **drt**: stop and drive), dynamic actions, rules for adding/modifying/deleting tasks inside schedules etc.
- Because the task are different and schedules are handled differently, the analysis modules need to remain separate for each contrib.


contribs/taxi/src/main/java/org/matsim/contrib/taxi/run/MultiModeTaxiModule.java
```diff
@@ -36,8 +39,9 @@ public class MultiModeTaxiModule extends AbstractModule {
 	@Override
 	public void install() {
 		for (TaxiConfigGroup taxiCfg : multiModeTaxiCfg.getModalElements()) {
-			install(new TaxiModeModule(taxiCfg));
-			installQSimModule(new TaxiModeQSimModule(taxiCfg));
+			var drtCfg = TaxiAsDrtConfigGroup.convertTaxiToDrtCfg(taxiCfg);
+			install(new DrtModeModule(drtCfg));
+			installQSimModule(new DrtModeQSimModule(drtCfg, new TaxiModeOptimizerQSimModule(taxiCfg)));
```
michalmac, line 44: This is the core change: we switch to `DrtModeModule` and `DrtModeQSimModule`. To make it work, we need to convert the taxi config to a drt config that is passed to those modules.

PR #1990, opened 2022-05-17
put timer into qnetsimengine and print the results to csv

Janekdererste:
I try to measure the time the qsim needs to shuffle agents around. This measures the wallclocktime per timestep in different locations.
- The overall wall clock time in ```QNetsimEngineWithThreadpool```
- The wall clock time divided by the number of links in `QNetsimEngineWithThreadpool`
- The wall clock time divided by the number of nodes in `QNetsimEngineWithThreadpool`

The three items above are the times on the main thread. I.e. the overall time needed including the overhead for threading

Also measured the wall clock time for each worker for working on the links and on the nodes. This is the code in `AbstractQNetsimEngineRunner`. These times are also divided by number of nodes and number of links. 

One should be able to compare the runtime per node/link between the overall code fom `QNetsimEngineWithThreadpool` and the worker Threads from `AbstractQNetsimEngineRunner` like below (numbers are nano seconds)

grafik
grafik


matsim/src/main/java/org/matsim/core/mobsim/qsim/qnetsimengine/QNetsimEngineRunnerForThreadpool.java
```diff
@@ -46,9 +58,15 @@ public Boolean call() {
 		}
 
 		if (this.movingNodes) {
+			var startTime = System.nanoTime();
```
Janekdererste, line 61: Hier beginnt das Time measurement der Worker

matsim/src/main/java/org/matsim/core/mobsim/qsim/qnetsimengine/QNetsimEngineWithThreadpool.java
```diff
@@ -89,23 +145,45 @@ protected void run(double time) {
...
 			engine.setTime(time);
 		}
 
 		try {
+			var overallStartTime = System.nanoTime();
 			for (AbstractQNetsimEngineRunner engine : this.getQnetsimEngineRunner()) {
 				((QNetsimEngineRunnerForThreadpool) engine).setMovingNodes(true);
 			}
+
+			var nodesStartTime = System.nanoTime();
```
Janekdererste, line 159: Hier beginnt das Node measurement für das oberste Diagram.

PR #1986, opened 2022-05-16
extend DrtConfigGroup with DrtWithShiftsConfigGroup

mdziakowski:
extended the DrtConfigGroup as suggested in #1964 and #1937 with DrtWithShiftsConfigGroup

PR #1991, opened 2022-05-17
Draft: allow proper reading of subclassed reflective config groups

nkuehnel:
in https://github.com/matsim-org/matsim-libs/pull/1964 the DrtConfigGroup has been made extendable. 
This has been implemented for drt shifts in https://github.com/matsim-org/matsim-libs/pull/1986

However, it seems like it currently does not work for reading/parsing an existing config xml.
For example, the following
```
URL configUrl = IOUtils.extendUrl(ExamplesUtils.getTestScenarioURL("mielec"), "mielec_drt_config.xml");
Config config = ConfigUtils.loadConfig(configUrl, new MultiModeDrtConfigGroup(DrtWithShiftsConfigGroup::new), new DvrpConfigGroup(),
      new MultiModeAlonsoMoraConfigGroup(), new OTFVisConfigGroup());
```
results in 

> java.lang.IllegalArgumentException: Module drt of type org.matsim.contrib.drt.extension.shifts.config.DrtWithShiftsConfigGroup doesn't accept unknown parameters. Parameter stopDuration is not part of the valid parameters: []

stopDuration is a parameter of the basic drt config group. The reason it fails seems to be that the ReflectiveConfigGroup does not recognize the getter/setter methods of the parent class
https://github.com/matsim-org/matsim-libs/blob/28ec691fce67f029318dd8f0302cb80632d5070f/matsim/src/main/java/org/matsim/core/config/ReflectiveConfigGroup.java#L175-L193
because `c.getDeclaredMethods(); ` does not return methods from the parent class. As such, the DrtWithShiftsConfigGroup does not know the parameters of the DrtConfigGroup anymore.

This PR adds a test as well as a quick solution to this by also checking whether the respective class has a super class which is also a ReflectiveConfigGroup and checks for its declared methods as well. With that, the test included in this PR passes but I'm not sure if this is considered a good solution :)

### DIFF
--- a/contribs/drt-extensions/src/test/java/org/matsim/contrib/drt/extension/shifts/ReadConfigTest.java
+++ b/contribs/drt-extensions/src/test/java/org/matsim/contrib/drt/extension/shifts/ReadConfigTest.java
@@ -1,5 +1,6 @@
 package org.matsim.contrib.drt.extension.shifts;
 
+import org.junit.Assert;
 import org.junit.Rule;
 import org.junit.Test;
 import org.matsim.contrib.drt.extension.shifts.config.DrtShiftParams;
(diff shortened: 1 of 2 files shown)


nkuehnel:
This will probably fail again as soon as we have more layers in the class hierarchy, in which case this approach would need to be recursive (going from one super class to the next)

michalmac:
Good catch! Your solution is good, but as you pointed out, it will not work with more inheritance levels. #1994 should address that.

nkuehnel:
Thanks a lot @michalmac !

Issue #1987, opened 2022-05-16
Carrier capabilities not complete in XML


JWJoubert:
Hi @kt86 and @rgraebe 

Instead of using example files, we are trying to create the required input files for the `freight` contrib. Specifically when a `Carrier` has a an infinite fleet size. According to me, if the `CarrierCapabilities` is set to `CarrierCapabilities.FleetSize.INFINITE`, then one only needs to specify the `VehicleType`s and not the actual vehicles. Yet, when writing the carrier to file, I get something like
```
<capabilities fleetSize="INFINITE">
	<vehicles>
	</vehicles>
</capabilities>
```

with no vehicle types written. For me this will lead to trouble down the road in terms of reproducibility. Can/should we fix this? Do you agree that this is problematic? Am I missing another use case? I am happy to go ahead and work on it and tag you in a pull request.

kt86:
Dear Johan, 

thanks for pointing this out!

We are also using the INFINITE fleet size for our simulation, so it is relevant for us as well! (Even if I have not seen this before, but I run the last things from already existing .xml files...)

Just for clarification: What we are doing, is that we need to define one vehicle per depot (and vehicle type) as representative for this vehicle type. -> It is still a definition of (at least) one vehicle, which then stands for many vehicles of the same type at the same location.

Can you share the code snippet, you are using? 

I totally agree, that this should be written out! -> We need to fix it, if it is missing. So please go ahead and add me as reviewer. I will be off for some days during the next two weeks, but I will try to come back to you as soon as I can.

Another quick thought: I know that @kainagel changed something, that is related with the loading of the vehicleTypes to the carrier a few months ago (part of #1849). Maybe that is somehow relevant here ???

JWJoubert:
> Another quick thought: I know that @kainagel changed something, that is related with the loading of the vehicleTypes to the carrier a few months ago (part of https://github.com/matsim-org/matsim-libs/pull/1849). Maybe that is somehow relevant here ???

Indeed, I believe it might be relevant, @kt86. There are tests commented out, but without clear design decisions. The XML written is not the same as the XML read (there are warnings of ignored fields, like `route`) - there is no document type specified. I do not want to undo changes that had explicit design decisions behind them, but currently it is really hard to navigate the code. There are warnings (in the logger) of deprecated code, yet the code is not tagged as `@Deprecated`.

... and, specifically, the `VehicleType`. This appears in the log

[...]

PR #1992, opened 2022-05-17
VehicleType in CarrierCapabilities

JWJoubert:
This pull request relates to issue #1987. It is not a “fix” because it (still) does not support the spirit of `FleetSize.INFINITE`. This pull request just makes more methods deprecated (for now) so that the code can be cleared as to what happens.

### DIFF
--- a/contribs/freight/src/main/java/org/matsim/contrib/freight/carrier/CarrierCapabilities.java
+++ b/contribs/freight/src/main/java/org/matsim/contrib/freight/carrier/CarrierCapabilities.java
@@ -61,7 +61,13 @@ public class CarrierCapabilities {
 			this.fleetSize = fleetSize;
 			return this;
 		}
-		
+
+		/**
+		 * @deprecated Since the vehicle type is in the {@link CarrierVehicleTypes}
+		 * container, it should not be duplicated here. It is also not written
+		 * to file when writing {@link CarrierPlanXmlWriterV2}.
+		 */
+		@Deprecated
 		public Builder addType( VehicleType type ){
 			if(!typeIds.contains(type.getId())){
 				vehicleTypes.add(type);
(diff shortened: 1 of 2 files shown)


JWJoubert:
I am closing issue #1987 for now. It seems that the `FleetSize.INFINITE` is not (currently) honoured in the spirit with which [it was originally intended](https://github.com/matsim-org/matsim-libs/blob/28ec691fce67f029318dd8f0302cb80632d5070f/contribs/freight/src/main/java/org/matsim/contrib/freight/carrier/CarrierCapabilities.java#L33). My interpretation is that `CarrierCapabilities` should _either_ be created with `CarrierVehicle`s or `VehicleTypes`s, but not both. Yes, from #1987 I understand is that one _sample vehicle_ has to be created per type, but this makes the `jsprit` solution fairly hard to interpret in the end (all `Tour`s have the same vehicle `Id`).

kt86:
Thanks a lot for figuring out, that we missed some deprecations. :( And adding them :)
I will look at it later.

The intention was to fade out that double-definition (since it was never clear, which of the two "vehicleType" is used from which part of the following code...). From our point of view, we can also remove that internal and now deprecated `vehicleType` field, since this information is (now) included in the (MATSim) vehicle and should be used from there. The other thing is more or less a leftover from the old way, where freight has had its one vehicle definitions.

Indeed, that in the tour plan the vehicles (=tour) does not have a unique id (their vehicle type) is something that I am aware of and I also wanted to change it. If it is now also an issue for you, I suggest, that we open an issue here on github and see, who can continue working on that. I already started with it on the `kmtCarrierVehIds` branch some time ago. Since this was for me more a small enhancement, but it came up, that doing it is/was more complicated as expected in the first moment (*) I did not continued working on it... 

(*) relativly easy and clear: DriverAgent and (MATSim) vehicles in the simulation are getting their Ids based on some hard-coded logic based on that field... -> adapt it.
But if i remember correctly, there was the issue, that jsprit (?) or some other code tried to match the Id of the `tour` with an idea of the `vehicles` (or their types??). This is very easy solvable for FINITE fleets, but for infinite FLEETS it was not in the current logic... As an result of that, I started to pass the (MATSim) vehicleType through the jsprit solver (I need to look it up, if I merged it or if it is still in this branch), so it will be also available afterwards ,... -> Need more work and some more intesive thought ... :(

kt86 (approved):

PR #1993, opened 2022-05-18
Include schedule updater in PreplannedDrtOptimizer

luchengqi7:
Add ScheduleTimingUpdater to the PreplannedDrtOptimizer, so that it can update the schedule based on the actual arrival time of the vehicle. This is necessary for the studies with fluctuating network travel time (e.g., through network loading, or network change events).

Currently, if a vehicle arrives earlier than the originaly scheduled time and starts the next task, then the end time of the current task (drive task, with end time based on the schedule) will be different from the start time of the next task (which starts from "now").


michalmac (approved):
Looks good to me 👍

PR #2004, opened 2022-05-25
remove KML support from matsim

michalmac:
The available java libraries that support KML are rather old and force us to stick to JAXB 2, while the world has moved on a bit in the meantime and we should bump to 3 and soon even to 4.

@mrieser @sebhoerl @balacmi As far as I know no one at VSP uses this functionality. I cannot say if this is the case outside VSP. I hope removing KML is fine with you. I will wait a bit with merging this PR.

BTW. There is a hand-crafted KML support inside carsharing. I cannot see it being used (referenced) anywhere, however, I haven't removed it, because it does not depend on the external KML library, so it's not a blocker for bumping JAXB.


mrieser:
Fine with me. KML was introduced like 14 years ago when Google Earth (the desktop app) was a thing. In "recent" versions (like 8-10 years ago) it started producing warnings when loading the files anyway, and later become partially non-working.

JWJoubert:
Not used on our side either. +1 on _GeoPackage_ .

michalmac:
Thanks for all your replies. I am going to merge this PR now and then reopen #2012.

PR #2023, opened 2022-05-27
Bug fixing for drop off duration

luchengqi7:
Extend the drop-off task by the additional stop duration 
- A non-zero value when variable stop duration is used
- The additional stop duration is simply 0 when fixed stop duration is used


nkuehnel (approved):

michalmac:
Nice catch!

Issue #2003, opened 2022-05-24
ShpOptions.getGeometry() only reads first feature of shp


simei94:
When using the getGeometry() method of the contribs/application/src/main/java/org/matsim/application/options/ShpOptions.java class, only the first (or one) feature is read. If I want all the features to be read, I have to iterate over the features myself, although the method should do this already..

Janekdererste:
Looking at the code of `ShpOptions` it looks like it is doing exaclty what you want. 

```
/**
	 * Return the union of all geometries in the shape file.
	 */
	public Geometry getGeometry() {

		Collection<SimpleFeature> features = readFeatures();
		if (features.size() < 1) {
			throw new IllegalStateException("There is no feature in the shape file. Aborting...");
		}
		Geometry geometry = (Geometry) features.iterator().next().getDefaultGeometry();
		if (features.size() > 1) {
			for (SimpleFeature simpleFeature : features) {
				Geometry subArea = (Geometry) simpleFeature.getDefaultGeometry();
				geometry.union(subArea);
			}
		}
		return geometry;
	}
```
This gets the union of all the features geometries right?

The [test class](https://github.com/matsim-org/matsim-libs/blob/master/contribs/application/src/test/java/org/matsim/application/options/ShpOptionsTest.java) for this class doesn't have a test case for this methdo. You could start by adding a test case and see whether it works or fails.

simei94:
Exactly, but somehow only the first feature is used and not all of them as a union. This occured multiple times already and not only in my own scripts..

Janekdererste:
Well, then it is probably a good idea to verify correctnes of the library with a unit test before investigating your project set ups ;-)

Janekdererste:
Closing this, because the autoclose in the pr didn't .

PR #2041, opened 2022-06-08
fix shpOptions.getGeometry()

simei94:
If a shpFile had multiple features, the getGeometry method of class ShpOptions did not include all features into the geometry, but only the first one because when iterating over the shpFile's features the features were collected but the geometry was NOT OVERWRITTEN, which resulted in the geometry just containing the shpFile's first feature.


contribs/application/src/test/java/org/matsim/application/options/ShpOptionsTest.java
```diff
@@ -61,4 +63,35 @@ public void all() {
 				.hasSize(Set.copyOf(ft).size());
 
 	}
+
+	@Test
+	public void testGetGeometry() {
```
Janekdererste, line 68: I found it quite difficult to understand the intend of this test. Naming the variables geometry1,2,3 makes it very hard to understand what the purpose of those variables are. I think a better naming for `geometry1` would be `joinedResult` and for `geometry2` `expectedResult` for example. Of course you can choose any other name, which describes what these things are for. 

I think the comparison of `geometry1` and `geometry3` doesn't make sense. You want to test whether this method correctly joins all the geometries. This would be much clearer without `geometry3`. I can see why you have it in there but the test would also fail if only the geometry of the first feature was returned, as it was the case before your fix. 

Of course one test is better than no test! Nevertheless, I can see four execution paths in the `getGeometry` method. If you want to do it right you could add tests for:

- No features - Exception is expected
- One feature - This could be done the way you handled `geometry3`
- Multiple features - Some feature without geometry 
- Multiple features - Everything is as expected -> The happy path (your test)

contribs/application/src/test/java/org/matsim/application/options/ShpOptionsTest.java
```diff
@@ -61,4 +63,35 @@ public void all() {
...
+
+		ShpOptions shp = new ShpOptions(input, null, null);
+		Geometry geometry1 = shp.getGeometry() ;
+		Geometry geometry2 = null ;
+		Geometry geometry3 = null ;
+
+		List<SimpleFeature> features = shp.readFeatures();
+
+		geometry3 = (Geometry) features.get(0).getDefaultGeometry();
+
```
Janekdererste, line 85: The way you have structured this for loop was also hard to parse for me. I would rather initialize a geometry before the loop and then have no if/else statement in the loop
```
		var geometry = new GeometryFactory().createEmpty(2);
		
		for (var feature : features) {
			var geometryToJoin = (Geometry) feature.getDefaultGeometry();
			geometry = geometry.union(geometryToJoin);
		}
```

Janekdererste (approved):
This works as is. If you're motivated you can do some of the changes I suggested.

Janekdererste:
Fixes #2003

contribs/application/src/test/java/org/matsim/application/options/ShpOptionsTest.java
```diff
@@ -61,4 +63,35 @@ public void all() {
...
+		for(SimpleFeature feature : features) {
+			if(geometry2 == null) {
+				geometry2 = (Geometry) feature.getDefaultGeometry();
+			} else {
+				geometry2 = geometry2.union((Geometry) feature.getDefaultGeometry());
+			}
+		}
+
+		Assert.assertFalse(geometry1.equals(geometry3));
```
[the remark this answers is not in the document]
simei94: I figured that it's better to use the geometry's own .equals method as they are explicitely comparing geometries and not just objects. Also, if I use Assert.assertEquals this test will fail. I think this has to do with the way the method .getGeometry and .readFeatures read in the features. Both now produce the same .shp, but if you look closely on their structure, the readFeatures method produces some brackets, which .getGeometry does not produce. Because of this the result of the functions is not EXACTLY the same, which does not matter for the resulting shape, so I think it's fine.

contribs/application/src/test/java/org/matsim/application/options/ShpOptionsTest.java
simei94: I did some renaming and deleted gemeotry3. For now I won't add further tests, as I have lots of other stuff to do.

contribs/application/src/test/java/org/matsim/application/options/ShpOptionsTest.java
simei94: re-structured the for loop, thanks for the advice

contribs/application/src/test/java/org/matsim/application/options/ShpOptionsTest.java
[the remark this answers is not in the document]
Janekdererste: Ah, I see. That makes sense.

Janekdererste:
This is much easier to read for me now. Very nice!

Issue #1514, opened 2021-05-20
Inconsistent handling of HBEFA 3.1/3.3 and 4.1


JWJoubert:
Some of the `emissions` contrib code was adapted to cater for the newer version of HBEFA (4.1). Unfortunately, code [like  this](https://github.com/matsim-org/matsim-libs/blob/03ef648d01461ff9785466cb9f8a67b5760c9667/contribs/emissions/src/main/java/org/matsim/contrib/emissions/ColdEmissionAnalysisModule.java#L290) in the `ColdEmissionAnalysisModule` and [this](https://github.com/matsim-org/matsim-libs/blob/03ef648d01461ff9785466cb9f8a67b5760c9667/contribs/emissions/src/main/java/org/matsim/contrib/emissions/WarmEmissionAnalysisModule.java#L398) in the `WarmEmissionAnalysisModule` assumes that the user still provides HBEFA 3.1 or 3.3 input files.

Now, the code in the `switch` statement in [EmissionsUtils](https://github.com/matsim-org/matsim-libs/blob/03ef648d01461ff9785466cb9f8a67b5760c9667/contribs/emissions/src/main/java/org/matsim/contrib/emissions/EmissionUtils.java#L404) has no `default` case. This means that if your input data was in fact HBEFA 4.1, then the lookup key will now change because [line 395](https://github.com/matsim-org/matsim-libs/blob/03ef648d01461ff9785466cb9f8a67b5760c9667/contribs/emissions/src/main/java/org/matsim/contrib/emissions/EmissionUtils.java#L395) has set it (by default) to `average`.

I did create a branch in which I added the following:
```
...
            default:
                String emConcept = vehicleInformationTuple.getSecond().getHbefaEmConcept();
                if (emConcept != null) {
                    attribs2.setHbefaEmConcept(emConcept);
                }
                break;
```
and while this works for our code, it breaks some tests. I do think, however, that it is the correct approach.

[...]

PR #2048, opened 2022-06-14
Error message

JWJoubert:
A hard-to-debug NullPointerExcerption is handled with a better error message if the HbefaSizeClass is not set,. This relates to issue #1514.

### DIFF
--- a/contribs/emissions/src/main/java/org/matsim/contrib/emissions/ColdEmissionAnalysisModule.java
+++ b/contribs/emissions/src/main/java/org/matsim/contrib/emissions/ColdEmissionAnalysisModule.java
@@ -319,7 +319,19 @@ final class ColdEmissionAnalysisModule {
 					logger.info(Gbl.FUTURE_SUPPRESSED);
 					detailedReadingInfoCnt++;
 				}
-				if (this.detailedHbefaColdTable.get(efkey) != null) {
+				/* The `this.detailedHbefaColdTable.get(efkey)` call may result in a NullPointerException
+				 * if the vehicle size class was not set. The following check tries to at least throw a
+				 * useful error message. A more complete solution would be to fix the key. (JWJ, June 2022)*/
+				HbefaColdEmissionFactor coldEmissionFactor = null;
+				try{
+					coldEmissionFactor = this.detailedHbefaColdTable.get(efkey);
+				} catch(NullPointerException e){
+					e.printStackTrace();
+					logger.error("Cannot find an emissions factor. One possible cause might be " +
+							"that the HbefaSizeClass is not set, in which case it defaults to null " +
+							"instead of 'not specified' (for HBEFA 4,1).");
+				}
+				if (coldEmissionFactor != null) {
 					HbefaColdEmissionFactor ef = this.detailedHbefaColdTable.get(efkey);
 					logger.debug("Lookup result for " + efkey + " is " + ef.toString());
 					return ef;


kt86 (approved):
Thanks for adding this, @JWJoubert !

contribs/emissions/src/main/java/org/matsim/contrib/emissions/ColdEmissionAnalysisModule.java
```diff
@@ -319,7 +319,19 @@ private HbefaColdEmissionFactor getEmissionsFactor(Tuple<HbefaVehicleCategory, H
 					logger.info(Gbl.FUTURE_SUPPRESSED);
 					detailedReadingInfoCnt++;
 				}
-				if (this.detailedHbefaColdTable.get(efkey) != null) {
+				/* The `this.detailedHbefaColdTable.get(efkey)` call may result in a NullPointerException
```
Janekdererste, line 322: You can get a `NullPointerException` if `this.detailedHbefaColdTable` is `null`. If this is the case the loading of the tables in the 'EmissionModule' https://github.com/matsim-org/matsim-libs/blob/45776ed01dfec9765541b5c89cf133dcbd5fc9f4/contribs/emissions/src/main/java/org/matsim/contrib/emissions/EmissionModule.java#L122 doesn't work properly, or you are trying something which is not covered by your `DetailedVsAverageLookupBehavior`.

The second option where this Exception could come from is during the lookup in the hashmap. The `HbefaVehicleAttributes` override `hashCode` https://github.com/matsim-org/matsim-libs/blob/45776ed01dfec9765541b5c89cf133dcbd5fc9f4/contribs/emissions/src/main/java/org/matsim/contrib/emissions/HbefaVehicleAttributes.java#L95 which is called during the `Map::get` call. Instead of wrapping this in a `try-catch` block, the `HbefaVehicleAttributes.sizeClass` should have a reasonable default. The current default is `average`.https://github.com/matsim-org/matsim-libs/blob/45776ed01dfec9765541b5c89cf133dcbd5fc9f4/contribs/emissions/src/main/java/org/matsim/contrib/emissions/HbefaVehicleAttributes.java#L30 

This means some other code intentionally sets this to `null`. If that's what you want, I'd rather suggest that you fix the `hashCode` method of `HbefaVehicleAttributes`. 

My personal preference is to set this value to something valid instead, because this avoids having null checks everywhere in the code. Maybe the current default would already be sufficent?

Janekdererste (commented):
I guess, you can leave it like this, since it works and it blends quite well into the otherwise messy code. 

As stated in the comment at the code, I think you should solve this differently though, since the problem is somewhere else.

contribs/emissions/src/main/java/org/matsim/contrib/emissions/ColdEmissionAnalysisModule.java
JWJoubert: Noted, thanks @Janekdererste. The problem does seem to be on the `Map::get` call. So the best place to fix it would be in the `hashCode` method when I can make some time. For some reason the `average` seems to be overridden... and that's what I need to figure out. I would prefer that it is rather `not specified` by default. But that is, again, an HBEFA 4.x format that is not backward compatible.

JWJoubert: > because this avoids having null checks everywhere in the code.

That makes sense, yes.

PR #2049, opened 2022-06-14
implement MulitModalLinkChooser

simei94:
Extract functionality of FacilitesUtils into MultiModalLinkChooserDefaultImpl. This enables individual implementations of the functionality, which may be necessary in several scenarios


matsim/src/main/java/org/matsim/core/router/MultimodalLinkChooserDefaultImpl.java
```diff
@@ -0,0 +1,61 @@
...
+            // this is the case where the postal address link is NOT in the subnetwork, i.e. does NOT serve the desired mode,
+            // OR the facility does not have a street address link in the first place.
+
+            if( facility.getCoord()==null ) {
+                throw new RuntimeException("link for facility cannot be determined when neither facility link id nor facility coordinate given") ;
+            }
+
+            accessActLink = NetworkUtils.getNearestLink(network, facility.getCoord()) ;
+            if ( accessActLink == null ) {
+                int ii = 0 ;
```
Janekdererste, line 44: What is this for loop doing? It looks like it is taking 10 links from the network and printing an error message for them?

matsim/src/main/java/org/matsim/core/router/MultimodalLinkChooserDefaultImpl.java
```diff
@@ -0,0 +1,61 @@
...
+import org.apache.log4j.Logger;
+import org.matsim.api.core.v01.Id;
+import org.matsim.api.core.v01.network.Link;
+import org.matsim.api.core.v01.network.Network;
+import org.matsim.core.gbl.Gbl;
+import org.matsim.core.network.NetworkUtils;
+import org.matsim.facilities.FacilitiesUtils;
+import org.matsim.facilities.Facility;
+
+class MultimodalLinkChooserDefaultImpl implements MultimodalLinkChooser {
```
Janekdererste, line 12: So, this is some complicated logic. I don't see any unit test for this. Are you testing this implicitly somewhere else?

simei94: Basically it is tested in the Leipzig [scenario](https://github.com/matsim-scenarios/matsim-leipzig/commit/4cb6fbcdb0fca84185747b1f1220bb68f224b435).

matsim/src/main/java/org/matsim/core/router/MultimodalLinkChooserDefaultImpl.java
vsp-gleich: This is code copied over here from FacilitiesUtils (where we set the original code to deprecated). I have to admit that we just copied and now I'm wondering too whether this is just some debug output committed by accident into FacilitiesUtils.

matsim/src/main/java/org/matsim/core/router/MultimodalLinkChooserDefaultImpl.java
vsp-gleich: I think we need some test in matsim-libs, too

matsim/src/main/java/org/matsim/core/router/MultimodalLinkChooserDefaultImpl.java
vsp-gleich: I talked to @kainagel about these lines of code. They have a debugging purpose. I added a comment in #2051 to explain.

matsim/src/main/java/org/matsim/core/router/MultimodalLinkChooserDefaultImpl.java
Janekdererste: > Basically it is tested in the Leipzig [scenario](https://github.com/matsim-scenarios/matsim-leipzig/commit/4cb6fbcdb0fca84185747b1f1220bb68f224b435).

That's not sufficient I think.

matsim/src/main/java/org/matsim/core/router/MultimodalLinkChooserDefaultImpl.java
Janekdererste: Ah, this makes it clearer. Copy paste development is the best kind of development 🙃

matsim/src/main/java/org/matsim/core/router/MultimodalLinkChooserDefaultImpl.java
simei94: I will write a test for this class, which covers all the possible outcomes of the logic.

Janekdererste: Thanks for doing this. Sorry that you have to do it now. This should have been there in the old place already.

PR #2059, opened 2022-06-19
build: introduce weekly code coverage for matsim core

michalmac:
The report is created by jacoco and then uploaded to CodeCov.

For testing purposes, the workflow is triggered on pushing to master. Later on we will switch to a weekly schedule.

### DIFF
--- a/.github/workflows/code-coverage.yaml
+++ b/.github/workflows/code-coverage.yaml
@@ -1,0 +1,39 @@
+name: weekly-code-coverage
+
+on:
+  push: # TEMP for testing purposes; ultimately it will be weekly (on: schedule)
+    branches:
+      - master
+#  schedule:
+#    - crone: '30 3 * * 0' # Sun 3:30 UTC
+
+jobs:
+  code-coverage:
+    name: deploy snapshot/PR-labelled version
+    # for PR-labelled deployment -- only if closed by merging
+    if: github.event_name == 'push' || github.event.pull_request.merged == true
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout git repo
+        uses: actions/checkout@v3
+
+      - name: Setup Java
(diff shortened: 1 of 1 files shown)


michalmac:
Here you can find the coverage reports: https://codecov.io/gh/matsim-org/matsim-libs. You can explore them using the sunburst chat or just by navigating the directory tree.

kt86:
Very nice! Thanks a lot, @michalmac! 

Do you plan to also introduce it for the contribs? (When we see, that it functions well for matism-core)
I think that may also help us to see some lacks in testing and may help top improve it...
Or are there some restrictions or reasons for not doing it?

PS: I also like the batch for it (and the other workflows) on the Readme page.

michalmac:
Thanks @kt86 

Good question. Adding it for all modules would require some additional work, but nothing blocks it. I was not sure if we really need/use code coverage for contribs. If yes, I can add it later. Right now I want to move all critical build pipelines from Jenkins to GH Actions so we can migrate to Java 17.

kt86:
Sounds good. 

If we really "need" it?  I don't know!
But if we have it, we can have a look at it and _maybe_ improve the coverage incrementally... (and maybe @kainagel forces us, to do so - at least for the more frequently used contribs). 
Without that information, we will  (imo) _never_ do so.

michalmac:
How we test matsim would be an interesting topic for a session during a developer meeting. Specifically, high coverage is good, but maybe should not be one of the main goals for writing tests. For instance, there are many tests that just run some matsim contribs to make sure they complete without an error, but without actually checking the output. Such tests may pump up the coverage, but at the same time add a little value to the software quality.

Janekdererste:
I don't think there is much added value for the contribs. Usually they are maintained by one or few people. If you want to check on your coverage, you can also do that in your IDE already.

PR #2060, opened 2022-06-19
GitHub actions code coverage 2

michalmac:
Add badges to README.md for selected workflows and the test coverage.


michalmac:
Here are the added badges:
image

PR #2077, opened 2022-07-03
Carrier strategy manager

kainagel:
Streamlining the carrier strategy manager so that it can serve as a template for LSP.

The idea is that `CarrierStrategyManager` is central infrastructure.  However, since, as of now, we do not configure it via the config, the only possible way to configure it is now
```
bind( CarrierStrategyManager.class ).toProvider( ...);
```
and then that provider creates an instance of the carrier strategy manager, configures it, and then returns it.  

This is fairly similar to how it was before, except that the separate interface of Carrier(Plan)StrategyManagerFactory is gone, and replaced by the above usage of the guice provider.

So the idea from a user perspective is:
* if you just want to run a zeroth iteration, then `CarrierModule` should be sufficient.  But all scores are set to minus infinity.
* if you want to have decent scores, you need to override the binding for `CarrierScoringFunctionFactory`.
* if you also want to run iterations, you need to override the binding for `CarrierStrategyManager'


kt86 (approved):
Thanks a lot. Looks good for me :)
LGTM and move forward.

PR #2081, opened 2022-07-04
make emissions analysis classes public

Nitnelav:
In the https://github.com/eqasim-org/eqasim-java project need to manually copy classes from the emissions.analyse namespace because they were not made public.
I'm using only the 2 classes modified here but there might be other classes that would benefit from being public.


kt86:
I think that this should be fine.
What do you think @Janekdererste ?

Janekdererste (commented):
With the classes being public it is possible to import them into other packages. It is not possible to do anything with them, because all the methods are package private as well. At least the getters in both classes need to be public, so that it is possible to access the data those classes have collected during event handling. 

The `EmissionsOnLinkHandler` would need a public constructor as well if it is supposed to be used like the [equasim project](https://github.com/eqasim-org/eqasim-java/blob/232ae121ecd1d661b550062111047e1f9d28af52/ile_de_france/src/main/java/org/eqasim/ile_de_france/emissions/RunExportEmissionsNetwork.java#L48) does

Nitnelav:
Oh right, I didn't check that... I'll update my PR and properly check if I can use what I need. Cheers

Nitnelav:
All right it works fine now. Thanks for the review !

PR #2087, opened 2022-07-08
Allow overriding q sim module in abstract module

kainagel:
I have the regular problem that I would like to re-configure the qsim at the level of an `AbstractModule` (and NOT `AbstractQSimModule`).  Re-configuring means overriding already existing material.  Typical use case is the `QNetworkFactory`, e.g. for the `Signals` contrib (see `Signals.configure( controler )`).  We all would prefer `controler.addOverridingModule( new SignalsModule() )` (had a quick VSP survey about this), but this is currently not possible.

Maybe most clear in the sequence of tests (except for the first) in the newly created `qnetsimengine/QSimComponentsTest`.

The suggested implementation in `AbstractModule` would add a method `installOverridingQSimModule` next to `installQSimModule`.  This would make my life _much_ easier.

I agree that one could argue that this goes a bit against Guice, which wants _set_ bindings, not _override_ them.  However, we have violated this for MATSim from the start, since the pre-Guide MATSim approach was to have default factories for everything, and then be able to override these factories if desired.

A resulting disadvantage of the approach is that it depends on the sequence: The override in the AbstractModule has to come _after_ other code that may set/override the same thing.  But we have that problem everywhere.

I am not against moving matsim slowly away from the overriding to the setting approach.  But we would need a roadmap for that which, AFAIK, we currently do not have.

(@sebhoerl : Maybe there is an alternative to register the QNetworkFactory as a QSimModule, and then I could unregister that?  If that is much preferred, I would be ok with it, if you show me how to do that.)


sebhoerl (approved):

sebhoerl:
HI @kainagel, this looks good to me. I also had some situations lately where it would have been cleaner (more user-friendly) to have this functionality.

PR #2089, opened 2022-07-10
build(deps): move log4j deps from matsim to matsim-all pom.xml

michalmac:
The same logging should be used across all modules (contribs). To make sure the same version is used everywhere, we need to have it as a direct dependency (instead of transitive).


Janekdererste:
Very good idea. Thanks for taking care.

PR #2100, opened 2022-07-19
include link2pollutants in emissions contrib

rgraebe:
We frequently require summed emissions per link-id to perform emissions analysis (generating `*.output_events.xml.gz`). In particular, the data structure `link2pollutants` calculates this for us. However, the code to produce `link2pollutants` is duplicated each time we do emission analysis for a different project.

In this commit the `link2pollutants` Map is generated by `EmissionsOnLinkHandler` and can be called by:
```
EmissionsOnLinkEventHandler handler = new EmissionsOnLinkEventHandler(10); 
handler.handleEvent(event);
handler.getLink2pollutants();
```

The required parameter `timeBinSizeInSeconds` (=10 in example) is for generating a similar data structure, but groups link emissions by time bins.

The class `EmissionsOnLinkHandlerTest` has been updated to include cases for testing the new output structure `link2pollutants`.

Thought: I have never seen the use of or worked with the current output of `EmissionsOnLinkHandler`, [getTimeBins()](https://github.com/matsim-org/matsim-libs/blob/eeb6ed5e3d2c3ca9ae5eb794607805e473bb90e9/contribs/emissions/src/main/java/org/matsim/contrib/emissions/analysis/EmissionsOnLinkEventHandler.java#L53). If it is not relevant anymore, it might be worth removing the (current) required parameter `timeBinSizeInSeconds` to make `EmissionsOnLinkEventHandler` a bit more intuitive to use?

### DIFF
--- a/contribs/emissions/src/main/java/org/matsim/contrib/emissions/analysis/EmissionsOnLinkEventHandler.java
+++ b/contribs/emissions/src/main/java/org/matsim/contrib/emissions/analysis/EmissionsOnLinkEventHandler.java
@@ -85,17 +85,12 @@ public class EmissionsOnLinkEventHandler implements WarmEmissionEventHandler, Co
 
         TimeBinMap.TimeBin<Map<Id<Link>, EmissionsByPollutant>> currentBin = timeBins.getTimeBin(time);
 
-        if (!currentBin.hasValue()){
-            currentBin.setValue( new HashMap<>() );
-        }
+        if (!currentBin.hasValue()){ currentBin.setValue( new HashMap<>() ); }
         if (!currentBin.getValue().containsKey(linkId)){
             currentBin.getValue().put( linkId, new EmissionsByPollutant( new HashMap<>( emissions ) ) );
-        } else{
-            currentBin.getValue().get( linkId ).addEmissions( emissions );
-        }
+        } else { currentBin.getValue().get( linkId ).addEmissions( emissions ); }
 
-        if (link2pollutants.get(linkId) == null) {
-            link2pollutants.put(linkId, emissions);
+        if (link2pollutants.get(linkId) == null) { link2pollutants.put(linkId, emissions);
         } else {
             for (Pollutant pollutant : emissions.keySet()) {
                 link2pollutants.get(linkId).merge(pollutant, emissions.get(pollutant), Double::sum);
--- a/contribs/emissions/src/main/java/org/matsim/contrib/emissions/analysis/EmissionsOnLinkEventHandler.java
+++ b/contribs/emissions/src/main/java/org/matsim/contrib/emissions/analysis/EmissionsOnLinkEventHandler.java
@@ -44,7 +44,6 @@ public class EmissionsOnLinkEventHandler implements WarmEmissionEventHandler, Co
 
 
     public EmissionsOnLinkEventHandler(double timeBinSizeInSeconds) {
-
         this.timeBins = new TimeBinMap<>(timeBinSizeInSeconds);
     }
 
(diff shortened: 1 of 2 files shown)


kt86:
Hi @rgraebe ,
thanks for doing it :)

I asked @Janekdererste to have a look at this PR, since a) I am on vacation and b) He is the author of the other (timeBin-based methods). As far as I remember, this is used for another project with a time dependent analysis of emissions. So it has it right to be there ;)

-> I will remove myself from the reviewers list and added @Janekdererste instead. Thanks to Janek for doing it!

Best, Kai MT

rgraebe:
Thank you @kt86 for updating the reviewers and your inputs!

contribs/emissions/src/main/java/org/matsim/contrib/emissions/analysis/EmissionsOnLinkEventHandler.java
```diff
@@ -86,5 +93,13 @@ private void handleEmissionEvent(double time, Id<Link> linkId, Map<Pollutant, Do
         } else{
             currentBin.getValue().get( linkId ).addEmissions( emissions );
         }
+
```
Janekdererste, line 96: This could be simplified like: 

```
var linkEmissions = link2pollutants.computeIfAbsent(linkId, id -> new HashMap<>());
for (var entry : map.entrySet()) {
    linkEmissions.merge(entry.getKey(), entry.getValue(), Double::sum);
}

**Review by Janekdererste — APPROVED** [review]
This is all fine and will work sufficiently well. 

The approach wastes a lot of space though. The TimeBinMap already stores a `Map<Id<Link>, Map<Polluntant, Double>>` just divided into time bins. Instead of storing all the values twice you could also think about calculating the values by link in the getter method. Something like:

```
    public Map<Id<Link>, Map<Pollutant, Double>> getLink2pollutants() { 
        return timeBins.getTimeBins().stream()
            .map(TimeBinMap.TimeBin::getValue)
            .flatMap(linkPollutions -> linkPollutions.entrySet().stream())
            .map(linkEmission -> Tuple.of(linkEmission.getKey(), linkEmission.getValue().getEmissions()))
            .collect(Collectors.toMap(Tuple::getFirst, Tuple::getSecond, (a, b) -> {
                b.forEach((key, value) -> a.merge(key, value, Double::sum));
                return a;
            }));
    }
```

rgraebe:
> ```
>     public Map<Id<Link>, Map<Pollutant, Double>> getLink2pollutants() { 
>         return timeBins.getTimeBins().stream()
>             .map(TimeBinMap.TimeBin::getValue)
>             .flatMap(linkPollutions -> linkPollutions.entrySet().stream())
>             .map(linkEmission -> Tuple.of(linkEmission.getKey(), linkEmission.getValue().getEmissions()))
>             .collect(Collectors.toMap(Tuple::getFirst, Tuple::getSecond, (a, b) -> {
>                 b.forEach((key, value) -> a.merge(key, value, Double::sum));
>                 return a;
>             }));
>     }
> ```

Thank you for adding this, @Janekdererste. I tried this, but ran into the strangest issue that crashes a test where multiple events are considered. When calling `getTimeBins()` this new code changes the values in `timeBins`: the first bin's entry (one link-id in the test) now has a combined emission value totaling ALL emissions across _multiple_ bins (this is the same value that is now contained in `link2pollutants`). Even using a temporary variable for this stream() function did not solve it. I know too little about every aspect of this code to understand how this can be happening?

Janekdererste:
Ah, I didn't test this. I guess the `b.forEach((key, value) -> a.merge(key, value, Double::sum));` puts all the values into map which is ultimately held by the `TimeBinMap`

Depending on whether you want to go with this approach you would have to copy the data at this point.

PR #2106, opened 2022-07-21
Enable tasks other than Stay task at the end of DVRP schedule

luchengqi7:
With this update, the schedule of a DVRP can end with different types of task. 

This is necessary when we want to have partial DVRP schedule during the day. By not computing the full schedule whenever the schedule gets update, we can save some computational workload. This can also be helpful for customize DRT optimizer.

### DIFF
--- a/contribs/drt/src/main/java/org/matsim/contrib/drt/optimizer/VehicleDataEntryFactoryImpl.java
+++ b/contribs/drt/src/main/java/org/matsim/contrib/drt/optimizer/VehicleDataEntryFactoryImpl.java
@@ -141,7 +141,7 @@ public class VehicleDataEntryFactoryImpl implements VehicleEntry.EntryFactory {
 	}
 
 	static double calcVehicleSlackTime(DvrpVehicle vehicle, double now) {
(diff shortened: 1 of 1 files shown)


michalmac (approved):
Looks good to me.

There is at least one more case where we assume the last task to be an instance of `StayTask`: https://github.com/matsim-org/matsim-libs/blob/05fcff323bb0db8790764ba190b05b6609198930/contribs/drt/src/main/java/org/matsim/contrib/drt/optimizer/rebalancing/RebalancingUtils.java#L64

However, I would not change it for now unless it is needed.

luchengqi7:
> Looks good to me.
> 
> There is at least one more case where we assume the last task to be an instance of `StayTask`:
> 
> https://github.com/matsim-org/matsim-libs/blob/05fcff323bb0db8790764ba190b05b6609198930/contribs/drt/src/main/java/org/matsim/contrib/drt/optimizer/rebalancing/RebalancingUtils.java#L64
> 
> However, I would not change it for now unless it is needed.

Thanks a lot for the information. Yeah, currently, I will probably not be using it. We can discuss that tomorrow in the meeting.

PR #2113, opened 2022-07-27
Accept custom task based on DefaultStayTask

luchengqi7:
Now, we also want to include the "WaitForStopTask" in this loop. I think it makes sense to use the superclass DefaultStayTask when doing the casting, such that we can also accept custom tasks that are based on task type STAY.


michalmac (approved):
Looks good to me!

Actually, if we want to be open for any other task, we should switch to interfaces (i.e. `DriveTask` and `StayTask`). `DrtStopTask` is already an interface - no need to change it.

michalmac:
We should also keep in mind that having it too generic is not always good. In some cases, it may be easier/better/safer to provide a different implementation of `VehicleEntry.EntryFactory`.

PR #2120, opened 2022-07-27
make it possible to transfer final SoC of an EV as initial SoC to the…

tschlenther:
… next iteration

This is useful in some scenarios, e.g. when inevstigating charging behavior in urban areas (over multiple days).
By default, this feature is not enabled.

I know that normally, the intention behind `ElectricVehicleSpecification` is to be immutable. This PR kind of moves around that. However, i feel like especially the `initalSoC` variable should be allowed to change over iterations.

So far, no unit tests are included in this PR. BTW: The `ev` contrib seems not to be heavily covered by tests in general.
I have run the RunEVExample for multiple iterations and compared the SoC line plots; it seems to work :)

Until now, this kind of functionality was included in the `vsp` contrib in the `playground.vsp.ev` package but in a more ugly way. There, the initial SoC is still read from / written to the vehicle type. If @michalmac approves to make this functionality more central, i.e. include it in the ev contrib, I will clean up the `playground.vsp.ev` package as well.

### DIFF
--- a/contribs/ev/src/main/java/org/matsim/contrib/ev/EvConfigGroup.java
+++ b/contribs/ev/src/main/java/org/matsim/contrib/ev/EvConfigGroup.java
@@ -58,6 +58,10 @@ public final class EvConfigGroup extends ReflectiveConfigGroup {
 	public static final String TIME_PROFILES = "timeProfiles";
 	static final String TIME_PROFILES_EXP = "If true, SOC time profile plots will be created";
 
+	private static final String TRANSFER_FINAL_SOC_TO_NEXT_ITERATION_EXP = "determines whether the resulting SoC at the end of the iteration X is set to be the initial SoC in iteration X+1 for each EV." +
+			" If set to true, bear in mind that EV might start with 0% battery charge.";
+	private static final String TRANSFER_FINAL_SOC_TO_NEXT_ITERATION = "transferFinalSoCToNextIteration";
+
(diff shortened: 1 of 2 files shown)


michalmac (approved):
Looks good to me!

I suggest some changes around setting up the QSim component binding. I pushed them to a separate branch (it's not tested, so please check if it works). See comparison: https://github.com/matsim-org/matsim-libs/compare/transferrableSoCForEV...transferrableSoCForEV-refactored

michalmac:
> I know that normally, the intention behind ElectricVehicleSpecification is to be immutable. This PR kind of moves around that. However, i feel like especially the initalSoC variable should be allowed to change over iterations.

It's OK to replace one vehicle specification with another between iterations. By immutability, we reduce the risk that the changes to the specification will impact the ongoing iteration, which in most cases is rather unintended.

If we want to modify the vehicle during the iteration, then we should modify the actual vehicle (it contains the vehicle current state), but not its specification (=fixed input).

michalmac:
> If @michalmac approves to make this functionality more central, i.e. include it in the ev contrib

Good idea to move it to ev!

tschlenther:
> I suggest some changes around setting up the QSim component binding. I pushed them to a separate branch (it's not tested, so please check if it works). 

I checked, it works. Thank you. Will go ahead and merge this PR now

PR #2121, opened 2022-07-27
vrp/ev: change scenario to non-static field in FinalSoc2VehicleTypeTest

michalmac:
IDs are automatically reset between unit tests (see: AutoResetIdCaches), so we do not transfer state between tests.
However, if we have some static fields that refer to IDs (e.g. scenario), we may run into some inconsistencies (two "identical" IDs, old and new, each having a different index).

The rule to avoid such issues in the future: keep tests independent (so the execution order does not matter) and therefore do not use static variables (except for immutable objects).

(This is my best guess what could be the reason. Let's see if this helps.)

### DIFF
--- a/contribs/vsp/src/test/java/playground/vsp/ev/FinalSoc2VehicleTypeTest.java
+++ b/contribs/vsp/src/test/java/playground/vsp/ev/FinalSoc2VehicleTypeTest.java
@@ -20,7 +20,12 @@
 
 package playground.vsp.ev;
 
-import org.junit.*;
+import java.util.HashMap;
+import java.util.Map;
+
+import org.junit.Assert;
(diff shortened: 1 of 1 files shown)


michalmac:
So it worked.
FYI Two related PRs around resetting the ID cache and keeping the tests isolated in general: https://github.com/matsim-org/matsim-libs/pull/1502 and https://github.com/matsim-org/matsim-libs/pull/1507

tschlenther:
thank you so much for taking care of this! And please excuse the mess I made in the first place. Will keep that in mind for the future

michalmac:
No worries. These kind of things just happen from time to time. It's good that we run unit tests on Windows (Appveyor), which increases chances of catching some issues.

PR #2126, opened 2022-08-01
Update GTFS Library to 14.0

rakow:
* Update the downstream gtfs converter to 14.0 and extended the gtfs cli tool
* Added option to use the new prefix setter to ensure generated ids are unique
* Options to use multiple shp files when creating schedule


contribs/application/src/main/java/org/matsim/application/prepare/population/CleanPopulation.java
```diff
@@ -65,7 +65,7 @@ public Integer call() throws Exception {
 
 		// Using the analysis main mode identifier instead of the routing mode based one on purpose
 		// to be able to process older population files without any routing modes!
-		TripsToLegsAlgorithm trips2Legs = new TripsToLegsAlgorithm(new DefaultAnalysisMainModeIdentifier());
+		TripsToLegsAlgorithm trips2Legs = new TripsToLegsAlgorithm(new RoutingModeMainModeIdentifier());
```
vsp-gleich, line 68: Hi, there is a comment above the line where DefaultAnalysisMainModeIdentifier() was replaced by RoutingModeMainModeIdentifier(). That comment still holds true for old (pre 2020) population files. Is this class used for old population files? Or is this for another use case?

vsp-gleich: @rakow

PR #2139, opened 2022-08-15
Adjust warning suppression in distance calculation with mixed coordinates

nkuehnel:
We recently ran into the problem that we had to halt our simulation because of overwhelming warning logs coming from distance calculations with mixed coords (i.e., with/out z coordinates) in large simulations with millions of agents.

The reason was that the warning counter had an integer overflow, effectively turning negative and spamming the mixed coord warning repeatedly. This PR tries to prevent this by having a boolean memorizing the state of warning logs.


michalmac:
what about switching to `AtomicLong`?

nkuehnel:
Ah yes, will do, thanks!

PR #2147, opened 2022-08-21
Also add AttributeConverters to HouseholdsReader in ScenarioLoaderImpl

marecabo:
Currently, the method `ScenarioLoaderImpl.loadHouseholds()` is not using the `AttributeConverter`s although other MATSim containers like population and network use them. This change also adds the supplied `AttributeConverter`s to the HouseholdsReader object before parsing the households file.


Janekdererste:
I guess this is adjacent to #2148

PR #2148, opened 2022-08-21
Allow loading of Scenario with AttributeConverters

marecabo:
The `loadScenario` methods in `ScenarioUtils` load a scenario with all its containers (population, network, ...) from a config. However, it is not possible to supply [`AttributeConverter`](https://github.com/matsim-org/matsim-libs/blob/master/matsim/src/main/java/org/matsim/utils/objectattributes/AttributeConverter.java)s to be used during that process.

This change overloads the `loadScenario` methods with a map of `AttributeConverter` as additional argument, such that they will be used by the (private) [`ScenarioLoaderImpl`](https://github.com/matsim-org/matsim-libs/blob/master/matsim/src/main/java/org/matsim/core/scenario/ScenarioLoaderImpl.java) when loading the content of the containers.

### DIFF
--- a/matsim/src/main/java/org/matsim/core/scenario/ScenarioUtils.java
+++ b/matsim/src/main/java/org/matsim/core/scenario/ScenarioUtils.java
@@ -19,7 +19,10 @@
  *                                                                         *
  * *********************************************************************** */
 
- package org.matsim.core.scenario;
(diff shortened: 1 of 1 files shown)


Janekdererste:
This makes sense. Thanks for contributing!

PR #2156, opened 2022-08-26
add a config switch for the event type when to create scoring functions

ikaddoura:
Adding a config switch to specifiy the event type when to create scoring functions for the population. So far, this was done in the iteration starts event which will now become the default. A new option is to create the scoring functions a bit later in the before mobsim event. This allows to change scoring relevant parameters (e.g. the person attributes) during the replanning operation.

In a previous version, I tried to change it to the before mobsim event without using a config switch. matsim itself was okay with that, but some contribs were not and tests failed.


jfbischoff:
Looks good to me :)

PR #2158, opened 2022-08-28
freight: switch to slf4j-reload4j and bump to 2.0.0

michalmac:
log4j 1.x has been declared EOL and the slf4j authors strongly encourage to use slf4j-reload4j binding (instead of slf4j-log4j12).
Reload4j is a drop-in replacement for log4j version 1.2.7.
For more info see: https://www.slf4j.org/manual.html#swapping

### DIFF
--- a/contribs/freight/pom.xml
+++ b/contribs/freight/pom.xml
@@ -76,7 +76,7 @@
 		</dependency>
 		<dependency>
 			<groupId>org.slf4j</groupId>
-			<artifactId>slf4j-log4j12</artifactId>
+			<artifactId>slf4j-reload4j</artifactId>
 		</dependency>
 
 		<dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -231,8 +231,8 @@
 
             <dependency>
                 <groupId>org.slf4j</groupId>
-                <artifactId>slf4j-log4j12</artifactId>
-                <version>1.7.36</version>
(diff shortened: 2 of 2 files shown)


kt86:
Thanks a lot, @michalmac !

BTW: Why do we have this dependency also in the freight contrib? I think, that other contribs do not have it as (duplicated?) dependency.
Should it not be possible to get it from the central (matsim) pom?

michalmac:
- We have this dependency only in freight. The `slf4j-reload4j` entry in the top (parent) pom.xml is a part of `dependencyManagement` section where we only define the version, but the actual dependency is declared in the freight pom.xml (section `dependencies`)
- The `slf4j-reload4j` dependency is declared only in freight, because jsprit uses slf4j and we want jsprit to log with Log4j (but it's not used directly by freight).

BTW. I just noticed that
- in matsim we already use Log4j 2.x (so we should use the `log4j-slf4j-impl` binding (i.e. implementation) rather than `slf4j-reload4j`)
- in `freight/pom.xml` we do not need to declare slf4j-api (it's only used in jsprit). we should only declare the implementation /binding (currently `slf4j-reload4j`, but should be changed to `log4j-slf4j-impl`)
I will address both issues in another PR.

kt86:
Thanks a lot for the explanation! Once again something learned for me :)
And thanks a lot for taking care of it!

PR #2162, opened 2022-09-02
freight: switch from slf4j-reload4j (log4j 1.2) to log4j-slf4j-impl (log4j 2)

michalmac:
At the same time, we have to downgrade slf4j-api to 1.7.36 (as 1.8+ is not yet supported by log4j-slf4j-impl)


michalmac:
This PR addresses issues listed here: https://github.com/matsim-org/matsim-libs/pull/2158#issuecomment-1230914639

PR #2163, opened 2022-09-02
log4j: remove log4j-1.2-api bridge between log4j 1.2 and 2.x

michalmac:
In the past, we migrated from Log4j 1.2 to 2.x by means of the Log4j 1.x bridge (log4j-1.2-api; see Option 1 in https://logging.apache.org/log4j/2.x/manual/migration.html).

This PRs removes the need for using that bridge by adapting the logging code so it directly calls the Log4j 2 API (Option 2 in the guide linked above). The changes are mostly minimal. The only difference is that now we cannot set the logging level on the logger object, but I think this is a good change as the use of `Logger.setLevel()` goes against the general idea of configuring the logging outside the matsim-lib code (i.e. configuring it by the user).


michalmac:
Re the use of `Logger.setLevel()` -- it was mostly used in tests, util scripts or some "ex-playground" code (in the vrp contrib).

michalmac:
I just added another commit that removes old log4j 1.2 configuration files:

After we migrated to log4j 2.x (2.5 years ago), these configuration files became ignored.
The easiest way to have them back is to copy and adapt log4j2.xml from the matsim module.

I am deleting them as most of them are very old with commit notes suggesting they were meant
as workarounds for some Jenkins limitations.

PR #2174, opened 2022-09-12
Add convenience method to time bin map

Janekdererste:
Add getValue with default supplier. Something like computIfAbsent in Map

### DIFF
--- a/contribs/analysis/src/main/java/org/matsim/contrib/analysis/time/TimeBinMap.java
+++ b/contribs/analysis/src/main/java/org/matsim/contrib/analysis/time/TimeBinMap.java
@@ -3,6 +3,7 @@ package org.matsim.contrib.analysis.time;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.function.Supplier;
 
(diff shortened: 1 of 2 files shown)


contribs/analysis/src/main/java/org/matsim/contrib/analysis/time/TimeBinMap.java
```diff
@@ -14,111 +15,121 @@
...
+		}
+
+		public void setValue(T value) {
+			this.value = value;
+		}
+
+		public T getValue(Supplier<T> computeIfAbsent) {
+
+			if (!hasValue()) {
+				setValue(computeIfAbsent.get());
```
michalmac, line 126: The name of the method indicates no side effect behaviour (like `Map.getOrDefault()`), whereas there is potentially one. I suggest changing the name to `computeIfAbsent()` (following the Java Collections nomenclature).

Janekdererste: Thanks for the suggestion. I will do so!

PR #2179, opened 2022-09-14
Add getvalue with default

Janekdererste:
Change method name as suggested here https://github.com/matsim-org/matsim-libs/pull/2174#discussion_r968368132


Janekdererste:
Somehow this was never merged.

PR #2186, opened 2022-09-16
MATSimApplication with other loadScenario signatures

steffenaxer:
This PR provides only a minor change but allows to exchange the way how a scenario is loaded for the MATSimApplication. There is now a exchangeable createScenario(Config config) that is pre-configured with the orignal default behavior. However, this likewise protected method could be overridden and thus allows to load a scenario with attribute converters, which might be mandatory for some users (see #2148 from @marecabo).


rakow:
Looks good! Sorry, did not get any notification from github, probably need to change some settings...

PR #2192, opened 2022-09-19
Ev contrib maintenance

kainagel:
Changes so far:
* added checks for events equality and plans equality at integration tests
* moved *IT to *Test so that we are sure that they always run (they are fairly fast)
* moved material from user code to `EvModule`.


kainagel:
This one ran into problems, see the above problems to move master into this one here.  (Seems like there were changes in master at the same places.).  I am closing this one here, and will instead continue on evContribMaintenance2

Issue #1335, opened 2021-01-07
Generate WarmEmissionEvents for last link of leg


Janekdererste:
**Current Behaviour**

Currently, no emissions are generated for the last link of a leg. The emissions for this link are generated on the next leg when the agent leaves the activity in between those two legs. 

**Desired Behaviour**

The emissions for the last leg should be generated upon VehicleLeavesTrafficEvent. When the next leg is started no emissions should be generated for that link.

**Comments**
@kt86 I discussed this with @kainagel . And we both thought this should be changed because in some cases our analysis is time dependent where it actually is important at what time the emissions are generated.

JWJoubert:
> in some cases our analysis is time-dependent where it actually is important at what time the emissions are generated.

For what it is worth, @Janekdererste, I support this.

PR #2200, opened 2022-09-23
FIXED: Issue 1335 Generate warm emission events for last link of leg

rgraebe:
This PR implements the _Desired Behaviour_  from [this issue](https://github.com/matsim-org/matsim-libs/issues/1335):

> **Current Behaviour**
> 
> Currently, no emissions are generated for the last link of a leg. The emissions for this link are generated on the next leg when the agent leaves the activity in between those two legs.
> 
> **Desired Behaviour**
> 
> The emissions for the last leg should be generated upon VehicleLeavesTrafficEvent. When the next leg is started no emissions should be generated for that link.

To this end, a few changes were made to the [WarmEmissionHandler](https://github.com/matsim-org/matsim-libs/blob/28e04f977ad2873f73ad55a4aad3749440225d21/contribs/emissions/src/main/java/org/matsim/contrib/emissions/WarmEmissionHandler.java):

1. the `VehicleLeavesTrafficEvent` now (also) triggers an emissions calculation event, as a result,
2. (at the `linkLeaveEvent`) we no longer require the logic where travel time should account for possible time spent outside of the vehicle to perform an activity. Therefore, this logic is removed.
3. The only logic now performed upon `linkLeaveEvent` is that of calculating emissions when the agent traversed the _entire_ link (with no activity).

HBEFA files in [emissions-sampleScenario](https://github.com/matsim-org/matsim-libs/tree/master/examples/scenarios/emissions-sampleScenario) (used in the tests for the emissions contrib) have been changed from txt to csv format.

Some refactoring reduces duplicate code between the [WarmEmissionHandler](https://github.com/matsim-org/matsim-libs/blob/28e04f977ad2873f73ad55a4aad3749440225d21/contribs/emissions/src/main/java/org/matsim/contrib/emissions/WarmEmissionHandler.java) and the [ColdEmissionHandler](https://github.com/matsim-org/matsim-libs/blob/28e04f977ad2873f73ad55a4aad3749440225d21/contribs/emissions/src/main/java/org/matsim/contrib/emissions/ColdEmissionHandler.java).


contribs/emissions/src/main/java/org/matsim/contrib/emissions/ColdEmissionHandler.java
```diff
@@ -86,19 +86,14 @@ public void reset(int iteration) {
 
     @Override
     public void handleEvent(LinkLeaveEvent event) {
+
+        // extract event details
 		Id<Vehicle> vehicleId = event.getVehicleId();
```
Janekdererste, line 91: formatting

contribs/emissions/src/main/java/org/matsim/contrib/emissions/ColdEmissionHandler.java
```diff
@@ -86,19 +86,14 @@ public void reset(int iteration) {
...
-            if (zeroLinkLengthWarnCnt == 0 ){
-                logger.warn("Length of the link "+ linkId + " is zero. No emissions will be estimated for this link. Make sure, this is intentional.");
-                logger.warn(Gbl.ONLYONCE);
-                zeroLinkLengthWarnCnt++;
-            }
-            return;
-        }
+        warnIfZeroLinkLength(linkId, linkLength);
 
         Double previousDistance = this.vehicleId2accumulatedDistance.get(vehicleId);
```
Janekdererste, line 98: this should be lower case `double`

contribs/emissions/src/main/java/org/matsim/contrib/emissions/ColdEmissionHandler.java
```diff
@@ -129,18 +124,19 @@ public void handleEvent(LinkLeaveEvent event) {
             }
         }
     }
-    static void handleNullVehicle( Id<Vehicle> vehicleId, EmissionsConfigGroup emissionsConfigGroup ){
+
+    static void handleNullVehicleECG(Id<Vehicle> vehicleId, EmissionsConfigGroup emissionsConfigGroup ){
         if( emissionsConfigGroup.getNonScenarioVehicles().equals( EmissionsConfigGroup.NonScenarioVehicles.abort ) ){
```
Janekdererste, line 129: Maybe this would bekome easier to read as a switch statement. But maybe it doesnt matter

contribs/emissions/src/main/java/org/matsim/contrib/emissions/ColdEmissionHandler.java
```diff
@@ -201,6 +197,16 @@ public void handleEvent(VehicleEntersTrafficEvent event) {
         }
     }
 
+    private void warnIfZeroLinkLength(Id<Link> linkId, double linkLength) {
```
Janekdererste, line 200: good refactoring

contribs/emissions/src/main/java/org/matsim/contrib/emissions/WarmEmissionAnalysisModule.java
```diff
@@ -328,7 +329,7 @@ Map<Pollutant, Double> calculateWarmEmissions(double travelTime_sec, String road
 				ef_gpkm = (fractionFreeFlow * efFreeFlow_gpkm) + (fractionStopGo * efStopGo_gpkm);
 
 			} else if (ecg.getEmissionsComputationMethod() == AverageSpeed) {
-				ef_gpkm = getEf(vehicleInformationTuple, efkey).getFactor();
+				ef_gpkm = getEf(vehicleInformationTuple, efkey).getFactor(); // this produces errors for vehicle leaves traffic events... ~rjg
```
Janekdererste, line 332: What kind of errors does it produce?

contribs/emissions/src/main/java/org/matsim/contrib/emissions/WarmEmissionHandler.java
```diff
@@ -65,14 +65,14 @@ class WarmEmissionHandler implements LinkEnterEventHandler, LinkLeaveEventHandle
 	private int nonCarWarn = 0;
 	private int noVehWarnCnt = 0;
 
-	private final Map<Id<Vehicle>, Tuple<Id<Link>, Double>> linkenter = new HashMap<>();
-	private final Map<Id<Vehicle>, Tuple<Id<Link>, Double>> vehicleLeavesTraffic = new HashMap<>();
-	private final Map<Id<Vehicle>, Tuple<Id<Link>, Double>> vehicleEntersTraffic = new HashMap<>();
+	private final Map<Id<Vehicle>, Tuple<Id<Link>, Double>> linkEnterMap = new HashMap<>();
+//	private final Map<Id<Vehicle>, Tuple<Id<Link>, Double>> vehicleLeavesTraffic = new HashMap<>(); // not needed anymore ~rjg 07.22
```
Janekdererste, line 69: Please remove if not needed :-)

contribs/emissions/src/main/java/org/matsim/contrib/emissions/WarmEmissionHandler.java
```diff
@@ -104,14 +103,27 @@ public void handleEvent(VehicleLeavesTrafficEvent event) {
...
+			this.linkEnterMap.remove( vehicleId );
+			// this is "so that no second emission event is computed for travel from parking to link leave" (kn)
+
+		} else {
+			handleNullVehicle(vehicleId);
+		}
+}
 
 	@Override
 	public void handleEvent(VehicleEntersTrafficEvent event) {
```
Janekdererste, line 129: Interesting, whether this is still true 🤔. I guess you know better than me.

contribs/emissions/src/main/java/org/matsim/contrib/emissions/WarmEmissionHandler.java
```diff
@@ -123,106 +135,110 @@ public void handleEvent(VehicleEntersTrafficEvent event) {
...
-			if (zeroLinkLengthWarnCnt == 0) {
-				logger.warn("Length of the link " + linkId + " is zero. No emissions will be estimated for this link. Make sure, this is intentional.");
-				logger.warn(Gbl.ONLYONCE);
-				zeroLinkLengthWarnCnt++;
-			}
-			return;
-		}
+		warnIfZeroLinkLength(linkId, linkLength);
 
 		// excluding links with zero lengths from leaveCnt. Amit July'17
```
Janekdererste, line 158: Is this comment still correct? zero length links seem to be counted as well.

contribs/emissions/src/main/java/org/matsim/contrib/emissions/WarmEmissionHandler.java
```diff
@@ -123,106 +135,110 @@ public void handleEvent(VehicleEntersTrafficEvent event) {
...
 		linkLeaveCnt++;
 
-		if(!this.linkenter.containsKey(vehicleId)){
+		if (!this.linkEnterMap.containsKey(vehicleId)) { // vehicle has NOT entered this link
 			int maxLinkLeaveFirstActWarnCnt = 3;
-			if(linkLeaveFirstActWarnCnt < maxLinkLeaveFirstActWarnCnt){
+			if (linkLeaveFirstActWarnCnt < maxLinkLeaveFirstActWarnCnt) {
 				logger.info("Vehicle " + vehicleId + " is ending its first activity of the day and leaving link " + linkId + " without having entered.");
```
Janekdererste, line 164: I find this info absolutely useless. As it says. It does something because of the way matsim is designed. This is nothing which has to appear in every log.

contribs/emissions/src/main/java/org/matsim/contrib/emissions/WarmEmissionHandler.java
```diff
@@ -123,106 +135,110 @@ public void handleEvent(VehicleEntersTrafficEvent event) {
...
 				logger.info("This is because of the MATSim logic that there is no link enter event for the link of the first activity");
 				logger.info("Thus, no emissions are calculated for this link leave event.");
 				if (linkLeaveFirstActWarnCnt == maxLinkLeaveFirstActWarnCnt) logger.warn(Gbl.FUTURE_SUPPRESSED);
 			}
 			linkLeaveFirstActWarnCnt++;
-		} else if (!this.linkenter.get(vehicleId).getFirst().equals(linkId)){
+		} else if (!this.linkEnterMap.get(vehicleId).getFirst().equals(linkId)) { // vehicle HAS entered a link but not THIS one
```
Janekdererste, line 170: I find this log message also not helpful

contribs/emissions/src/main/java/org/matsim/contrib/emissions/WarmEmissionHandler.java
```diff
@@ -123,106 +135,110 @@ public void handleEvent(VehicleEntersTrafficEvent event) {
...
-				travelTime = leaveTime - enterTime - departureTime + arrivalTime;
-
-				// now that we subtracted the time spent on the link outside of the vehicle
-				// we delete these events (so that when the link gets crossed once more
-				// the departure/arrival time are not wrongly taken into account)
-				// Thanks @markusstraub for providing this
-				this.vehicleLeavesTraffic.remove(vehicleId);
-				this.vehicleEntersTraffic.remove(vehicleId);
-			}
+			// the vehicle traversed the entire link, DO calculate emissions IF this was not after an activity
```
Janekdererste, line 180: The following logic is also too complicated. It uses `this.linkEnterMap.get(vehicleId).getSecond()` and then tests `this.linkEnterMap.containsKey(vehicleId)` on the next line. 

You could remove the two cases for logging basic Matsim stuff above. Then down here, something like the following should happen (not tested)

```
if (this.linkEnterMap.contains(vehicleId)) {
    var tuple = this.linkEnterMap.get(vehicleId);
    if (tuple.getFirst().equals(linkId)) {
        Vehicle vehicle = VehicleUtils.findVehicle(vehicleId, scenario);
	if (vehicle != null) {
            emissionsCalculation(vehicleId, vehicle, link, leaveTime, travelTime);
	    this.linkEnterMap.remove(vehicleId);
        } else {
            handleNullVehicle(vehicleId);
        }
    }
}

````

Janekdererste (approved):
Very nice. Thank you for taking this on. I have added suggestions for even more refactorings, in case you are in the mood.

kt86 (approved):
Thanks @rgraebe!

One question: Why do we have now (additonally) sample files (and configs -> testv2_Vehv1 / 2) in target/classes/test .... ?
Can't we use the existing scenario/emissions-sampleScenario directory?

--> Looks like it is somehow a copy of the original data, that was (unintendedly) commited(?)

Please clean that up before merging.

Nevertheless, I will approve this PR, so you are not stuck and are not forced to wait for my "go" after doing it. :)

rgraebe:
> One question: Why do we have now (additonally) sample files (and configs -> testv2_Vehv1 / 2) in target/classes/test .... ? Can't we use the existing scenario/emissions-sampleScenario directory?
> 
> --> Looks like it is somehow a copy of the original data, that was (unintendedly) commited(?)

Hi Kai, so this was also strange for me until I saw why it looks like duplicates/added files. The `target/classes/test` files aren't usually tracked by git, but I had to add them to change them: I changed the hbefa files from txt to csv (but under `target/classes/test` it looks like I "added" them). The config files needed to update to include the (renamed) hbefa files. That is why they also appear "added".

contribs/emissions/src/main/java/org/matsim/contrib/emissions/ColdEmissionHandler.java
rgraebe: I'd rather not change this, because if changed to `double`, it cannot be compared to `null` anymore and it seems like the `!= 0` test is a good "safety check". It is also used in the `coldEmissions` map within the `if` statement... so it doesn't seem like it was a typo by the original author?

contribs/emissions/src/main/java/org/matsim/contrib/emissions/WarmEmissionHandler.java
rgraebe: Nope, I had to add an if clause to fix this... thanks.

PR #2201, opened 2022-09-23
Introduce Parameter and Comment annotations to ReflectiveConfigGroup

michalmac:
Creating and maintaining config groups may be time consuming and it is always easy to make some errors. Since we are thinking about introducing something like "drt config 2.0", I decided to spent some time on simplifying the way we work with config groups etc.

This PR introduces new annotations to the reflective config group: `Parameter` (marking the parameters) and `Comment` (attaching the comment to the parameter). Thanks to them, we can reduce the current code like this:
```java
	public static final String MAX_WAIT_TIME = "maxWaitTime";
	static final String MAX_WAIT_TIME_EXP = "Max wait time for the bus to come (optimisation constraint).";

	// other constants ...

	@PositiveOrZero
	private double maxWaitTime = Double.NaN;// seconds

	// other fields ...

	@Override
	public Map<String, String> getComments() {
		Map<String, String> map = super.getComments();
		map.put(MAX_WAIT_TIME, MAX_WAIT_TIME_EXP);
		//other params here
		return map;
	}

	/**
	 * @return -- {@value #MAX_WAIT_TIME_EXP}
	 */
	@StringGetter(MAX_WAIT_TIME)
	public double getMaxWaitTime() {
		return maxWaitTime;
	}

	/**
	 * @param maxWaitTime -- {@value #MAX_WAIT_TIME_EXP}
	 */
	@StringSetter(MAX_WAIT_TIME)
	public DrtConfigGroup setMaxWaitTime(double maxWaitTime) {
		this.maxWaitTime = maxWaitTime;
		return this;
	}

	// other getters and setters ...
```
to just a few lines:
```java
	@Parameter
	@Comment("Max wait time for the bus to come (optimisation constraint).")
	@PositiveOrZero
	public double maxWaitTime = Double.NaN;// seconds
```

What is important, all information is grouped in one place (parameter, comment, constraint, default value). And that info is visible in javadoc and in an IDE (e.g. whenever hovering a mouse pointer over that field, at least in IntelliJ):
image


michalmac:
A few more notes:
- Sometimes the parameter has a different name than the corresponding field, then we need to provide the parameter name (if we do  not want to align the field name) like here:
```java
	@Parameter("minChargingTime")
	public int minimumChargeTime = 1200;
```
- We can mix the old and new approache, i.e. use `@StringGetter`/`@StringSetter` for some params, while using `@Parameter` for other ones
- We can use `@Comment` only for parameters annotated with `@Parameter`
- @Parameter does not yet support comma separated lists/sets of strings (can be easily added in another PR)
- Not having getters and setters means that we need to expose the corresponding fields. However, I do not see it as an issue, because in (almost?) all cases I saw the existing getters and setters give full access to the field (see: https://dzone.com/articles/getter-setter-use-or-not-use-0, in particular section "Getters and Setters Are Highly Overused"  and references)

michalmac:
In the first commit in this PR I also removed support for `@DoNotConvertNull` from `ReflectiveConfigGroup`. It is not used anywhere. IMO it only added to the complexity. Besides that, `checkConvertNullAnnotations()` had a bug (it does not check getter vs setter, but getter vs getter instead).

As for the added complexity, it is both about `ReflectiveConfigGroup` (more complex code) and about actual XML config files (while looking at the config file, we should not need to navigate to the corresponding code to check if for this specific parameter, `value="null"` means `null` or `"null"`).

However, if we want to keep this feature, I can revert this commit.

michalmac:
I migrated ~25 config groups/parameters (in the dvrp, drt, taxi and ev contribs) to this new approach, each migration in a separate commit.

neuma:
> What is important, all information is grouped in one place (parameter, comment, constraint, default value). And that info is visible in javadoc and in an IDE (e.g. whenever hovering a mouse pointer over that field, at least in IntelliJ): image

Works in Eclipse too. 👍

PR #2207, opened 2022-09-29
Option to delete ITERS directory at the end

rakow:
In many cases, the ITERS directory is not needed, especially when a run finished successfully. This PR changes the default behavior to delete the ITERS directory at the end.  Can be reverted to the old behavior via config:
```
config.controler().setHandleItersAtEnd(ControlerConfigGroup.HandleIterations.keep);
```


neuma:
> In many cases, the ITERS directory is not needed, especially when a run finished successfully. This PR changes the default behavior to delete the ITERS directory at the end. Can be reverted to the old behavior via config:
> 
> ```
> config.controler().setHandleItersAtEnd(ControlerConfigGroup.HandleIterations.keep);
> ```

This must be the opposite! Please make this an opt-in feature. Updating MATSim should not start deleting any files without consent of the user.

steffenaxer:
I would totally agree @neuma! Per default nothing should be deleted. Especially because many contrib specific detailed information is only stored within the ITERs.

rakow:
We discussed this internally and agreed that the default behavior should be what most people want and what saves the most storage space. We have quite some issues at the moment because of the amount and size of output files.  
Do you see a strong need for these files in the ITERS directory? Maybe other people who need these files regularly also want to comment on that.

But in my opinion, not changing it simply because it is a different behavior is not a convincing argument.

steffenaxer:
Need to discuss it with my working group. But from my point of view copying content from the last ITER folder might be a potential solution. 
However, I know from some use cases in which the ITERS information is somehow a general source of information, that shows how the model is evolving. 
I totally understand the aspect of data storage savings, but one could argue, if there is this problem, one could actively enable ITER deletion, but not per default?
Just to mention another aspect, most of the problem is generated by large xml.gz files, why no just setting the write interval to a higher value, to handle this problem?

neuma:
> I totally understand the aspect of data storage savings, but one could argue, if there is this problem, one could actively enable ITER deletion, but not per default?

I second that. This PR introduces a solution to a problem some users might have and I certainly welcome the feature as such. 

> But in my opinion, not changing it simply because it is a different behavior is not a convincing argument.

In my opinion, MATSim should never delete or overwrite files without the user actively asking for it. I would have the same issue with changing the default of `OverwriteFileSetting` to something like `overwriteExistingFiles` because it is more convenient. It's a matter of [Fail-safe](https://en.wikipedia.org/wiki/Fail-safe) design.

rakow:
> Just to mention another aspect, most of the problem is generated by large xml.gz files, why no just setting the write interval to a higher value, to handle this problem?

Many small files are as big of a problem, especially on network files systems. 

> In my opinion, MATSim should never delete or overwrite files without the user actively asking for it. I would have the same issue with changing the default of OverwriteFileSetting to something like overwriteExistingFiles because it is more convenient. It's a matter of [Fail-safe](https://en.wikipedia.org/wiki/Fail-safe) design.

Ok, I see the issue that one could have put files there that would then be deleted. I'm fine with keeping the default behavior and there are some good arguments for it. In any case, I will leave the PR open for a while in case anybody has an additional opinions.

PR #2220, opened 2022-10-05
Refactor electric fleet/vehicle specification

michalmac:
Also clean up the ev package inside the vsp contrib (besides the other refactorings in the ev contrib).


tschlenther:
Thank you really much! Especially for also cleaning up in the vsp-contrib package!!

nkuehnel:
@michalmac will DVRP fleet also be merged in to vehicles?

michalmac:
> @michalmac will DVRP fleet also be merged in to vehicles?

Long-term, yes. However, first I want to see how it works with EVs. I still find it useful to have both `ElectricVehicleSpecification` (esp. due to unit conversions) and `ElectricFleetSpecification` (so we have a known and injectable set of EVs).

nkuehnel:
Alright, thanks!

PR #2223, opened 2022-10-06
fix minShareIdleVehicles by excluding times without vehicles in service

vsp-gleich:
and add minCountIdleVehicles


contribs/drt/src/main/java/org/matsim/contrib/drt/analysis/DrtAnalysisControlerListener.java
```diff
@@ -214,6 +205,38 @@ public void notifyIterationEnds(IterationEndsEvent event) {
...
+			}
+		}
+		return minCountIdleVehiclesOverDay < Double.MAX_VALUE ?
+				format.format(minCountIdleVehiclesOverDay) : notAvailableString;
+	}
+
+	private String getMinShareIdleVehiclesOverDayFormatted() {
+		double minShareIdleVehiclesOverDay = Double.MAX_VALUE;
+		double[] stayTaskProfile = vehicleOccupancyProfileCalculator.getNonPassengerServingTaskProfiles().get(new DrtTaskType(DrtTaskBaseType.STAY));
+		double[] numberOfVehiclesInServiceProfile = vehicleOccupancyProfileCalculator.getNumberOfVehiclesInServiceProfile();
```
michalmac, line 227: `numberOfVehiclesInServiceProfile` gets computed twice, i.e. in `getMinCountIdleVehiclesOverDayFormatted()` and `getMinShareIdleVehiclesOverDayFormatted()`. Maybe we want to compute it once and then pass to both methods. Quite likely, it's not a lot of computation, so we can keep it the way it is.

contribs/drt/src/main/java/org/matsim/contrib/drt/analysis/DrtAnalysisControlerListener.java
```diff
@@ -214,6 +205,38 @@ public void notifyIterationEnds(IterationEndsEvent event) {
 				filename(event, "drt_boardings", ".csv"), filename(event, "drt_alightments", ".csv"), network);
 	}
 
+	private String getMinCountIdleVehiclesOverDayFormatted() {
+		double minCountIdleVehiclesOverDay = Double.MAX_VALUE;
+		double[] stayTaskProfile = vehicleOccupancyProfileCalculator.getNonPassengerServingTaskProfiles().get(new DrtTaskType(DrtTaskBaseType.STAY));
```
michalmac, line 210: you can also use `DrtStayTask.TYPE` instead of `new DrtTaskType(DrtTaskBaseType.STAY)`

vsp-gleich: Thanks!

contribs/drt/src/main/java/org/matsim/contrib/drt/analysis/DrtAnalysisControlerListener.java
```diff
@@ -214,6 +205,38 @@ public void notifyIterationEnds(IterationEndsEvent event) {
...
+					// only consider time intervals in which vehicles were in operation. Otherwise, any time period without a vehicle in operation will make for count 0.
+					minCountIdleVehiclesOverDay = Math.min(minCountIdleVehiclesOverDay, stayTaskProfile[i]);
+				}
+			}
+		}
+		return minCountIdleVehiclesOverDay < Double.MAX_VALUE ?
+				format.format(minCountIdleVehiclesOverDay) : notAvailableString;
+	}
+
+	private String getMinShareIdleVehiclesOverDayFormatted() {
```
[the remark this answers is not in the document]
vsp-gleich: I first wrote it as 1 method and was not sure what is best. Merged it back to one method. Thanks for reviewing!

contribs/drt/src/main/java/org/matsim/contrib/drt/analysis/DrtAnalysisControlerListener.java
vsp-gleich: Should be fixed by merging the two methods using numberOfVehiclesInServiceProfile to one method only.

PR #2229, opened 2022-10-11
Ev contrib maintenance2

kainagel:
evContribMaintenance ran into problems.  Backtracking a bit and then branching off to this one.


kainagel:
This branch ran into a dead end because of too many conflicts with other things.  After backtracking and merging, the work was resumed and merged with https://github.com/matsim-org/matsim-libs/pull/2333 .

Issue #2225, opened 2022-10-11
Empty sets not loaded correctly from ConfigGroup


marecabo:
Thanks for the parameter annotation also for sets/lists in ConfigGroups by @michalmac in #2202!

I noticed, when saving and loading config groups with empty sets, that they are not loaded the same again.

Say we have a config group with

```java
    @Parameter
    @NotNull
    public Set<String> licenseModes = Set.of(TransportMode.car);

    @Parameter
    @NotNull
    public Set<String> drtModes = Collections.emptySet();
```

and save and load that config group, the resulting values will be:

```java
System.out.println(configGroup.licenseModes); // {"car"}
System.out.println(configGroup.drtModes); // {""} <-- set with empty string as element, not an empty set
```

Should it be like it is currently or should an empty set also be loaded as an empty set? Intuitively, I would prefer the latter.

michalmac:
I agree, we should return `{}` rather than `{""}`. I will fix it.

marecabo:
Thanks! Fortunately, there is only one test which fails when changing to the desired behavior: https://github.com/matsim-org/matsim-libs/actions/runs/3226559716/jobs/5280224784#step:6:997

marecabo:
Thanks for the quick fix. This makes creating config groups a lot more convenient!

PR #2230, opened 2022-10-12
core: disallow blank strings as elements of collections in config parameters

michalmac:
In response to https://github.com/matsim-org/matsim-libs/issues/2225

(De)serialisation rules:
- "null" (string) <-> `null` (pointer)
- blank string <-> empty collection
- blank element in a comma separated string - thrown exception during deserialisation
- blank element in a collection - thrown exception during serialisation


michalmac:
This PR does not impact the old setters/getters used for converting between Strings and Collections.

PR #2239, opened 2022-10-17
fix bug for check of located in relevantArea

rewertvsp:
The relevantArea check up was not correct when the acitivity csr and the shape crs are different. This should be fixed now.

### DIFF
--- a/contribs/application/src/main/java/org/matsim/application/prepare/freight/tripExtraction/ExtractRelevantFreightTrips.java
+++ b/contribs/application/src/main/java/org/matsim/application/prepare/freight/tripExtraction/ExtractRelevantFreightTrips.java
@@ -144,8 +144,8 @@ public class ExtractRelevantFreightTrips implements MATSimAppCommand {
             Coord endCoord = endActivity.getCoord();
             double departureTime = startActivity.getEndTime().orElse(0);
 
-            boolean originIsInside = relevantArea.contains(MGC.coord2Point(startCoord));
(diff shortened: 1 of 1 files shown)


contribs/application/src/main/java/org/matsim/application/prepare/freight/tripExtraction/ExtractRelevantFreightTrips.java
```diff
@@ -144,8 +144,8 @@ public Integer call() throws Exception {
             Coord endCoord = endActivity.getCoord();
             double departureTime = startActivity.getEndTime().orElse(0);
 
-            boolean originIsInside = relevantArea.contains(MGC.coord2Point(startCoord));
-            boolean destinationIsInside = relevantArea.contains(MGC.coord2Point(endCoord));
+			boolean originIsInside = relevantArea.contains(MGC.coord2Point(sct.transform(startCoord)));
```
rakow, line 147: The indentation is now messed up probably mixed spaces and tabs,

luchengqi7 (approved):

rakow (changes requested):

contribs/application/src/main/java/org/matsim/application/prepare/freight/tripExtraction/ExtractRelevantFreightTrips.java
rewertvsp: I changed the indentations

rakow (approved):

PR #2244, opened 2022-10-20
emissions: add case when vehicle does not enter link but leaves traffic

rgraebe:
For some reason, in the [Berlin scenario](berlin-v5.4-0.1pct.output_events.xml.gz), there are (in the output events) agents that enter traffic after an activity and immediately leave traffic again. This `vehicleLeavesTrafficEvent` triggers an emissions calculation. The calculation requires an entry time to compute travel time. This entry time comes from the `linkEnterMap`. However, for these rare cases in the Berlin scenario (9 total!) there is no `linkEnterEvent` (and thus no link enter time) meaning that the `linkEnterMap` returns "Null" when queried for the specific vehicle, so the two emission tests in [matsim-berlin](https://github.com/matsim-scenarios/matsim-berlin/tree/a32dad8bf2b1231ae12982187f6dfdb833f6900e/src/test/java/org/matsim/analysis/emissions) fail (@vsp-gleich picked this up).

This PR adds a few lines of code (including a test -- TODO) to prevent these failures: if no `linkEnterEvent` exists for the agent triggering the `VehicleLeavesTrafficEvent`, the `vehicleEntersTrafficEvent`'s time is used. If a `vehicleEntersTrafficEvent` **also** does not exist (very unlikely), the code aborts with a descriptive message.


kt86 (commented):
Hi @rgraebe,
thanks for fixing this :) 
Calculation the travelTime from that events sounds even cleaner than what was coming into my mind (directly return a zero emission event, because the vehicle has not moved on the link.)

One Question: have you tested it with that rare matsim-berlin events? 
Reason for my question: I do not have in mind, what it will do later on, when calculating the data for the emission event: Will it then be something like traveling 1 (?)sec (plus our 1 bonus second from the other PR) on the whole link length) ?--> we have a very high speed for the link, and maybe we also have calculated that emissions for that link, even we should not do it(?) ....

Two minor suggestions - after checking the behavior:
- Maybe it would be good, to have a small extra test for it - to prevent removing it without a notice -> leading again to some errors later on.
- Add a short comment, for this extra text new introduced case, so one will know it later, why_ this was added...

rgraebe:
> Will it then be something like traveling 1 (?)sec (plus our 1 bonus second from the other PR) on the whole link length) ?

This made me think... the rare cases we have here are interpreted wrong with our travel time calculation --> you are right that the travel time will be very short (only our extra one second, because the vehicle enter+leaves traffic in the SAME second!)... so it will seem like this agents travels very fast on the entire link, which is NOT the case. 

Therefore, I actually think these rare cases should throw a zero emissions event like you mentioned, because this is closer to reality (and the alternative of having super short travel times causes inaccuracies in the interpretation during mobsim). What do you think?

rgraebe:
> Maybe it would be good, to have a small extra test for it - to prevent removing it without a notice -> leading again to some errors later on.

I agree that we need a test and comments for this @kt86, also because the current tests' `output_events` file does not have these cases.

kt86:
> Therefore, I actually think these rare cases should throw a zero emissions event like you mentioned, because this is closer to reality (and the alternative of having super short travel times causes inaccuracies in the interpretation during mobsim). What do you think @kt86?

Then let's do it in this way :) 

Maybe add log.warn(..) message when throwing it in the first step, so one sees that there is a special handling. That could be useful to either have a closer look at the simulation run (why do we have this behavior (*)) OR in cases, where - maybe at some point in the future and for whatever reason - the vehicles enters/leaves positions are not the same anymore.

(*) We may check, what freight driver agents with several activities at the same link are doing. I can imagine that they will also behave like this - and that would be ok, I think. If so, then the "0" emission throwing would be indeed a correct answer ;)

rgraebe:
@kt86 , I added a test for this handling of the `vehicleLeavesTrafficEvent`. The test attempts to write out emissions (like the currently [failing test(s)](https://github.com/matsim-scenarios/matsim-berlin/tree/master/src/test/java/org/matsim/analysis/emissions) in `matsim-berlin`) for a sample events file taken from [`matsim-berlin`](https://svn.vsp.tu-berlin.de/repos/public-svn/matsim/scenarios/countries/de/berlin/berlin-v5.4-0.1pct/output-berlin-v5.4-0.1pct/berlin-v5.4-0.1pct.output_events.xml.gz). If the emission events can be computed, the test passes after checking if the emission events output file is similar to the expected file (`smallBerlinSample.output_events.xml.gz`).

I believe this is what we wanted to add before concluding this PR?

contribs/emissions/src/main/java/org/matsim/contrib/emissions/WarmEmissionHandler.java
```diff
@@ -144,7 +161,7 @@ public void handleEvent(LinkLeaveEvent event) {
 		double linkLength = link.getLength();
 		double leaveTime = event.getTime();
 
-		if ( !warnIfZeroLinkLength( linkId, linkLength ) ) { linkLeaveCnt++; }
+		if ( warnIfZeroLinkLength( linkId, linkLength ) ) { linkLeaveCnt++; }
 		// excluding links with zero lengths from leaveCnt. Amit July'17
```
kt86, line 165: Do we need this if statement here anymore?
I think, that we NOW just can do the `linkLeaveCnt++`, or?

contribs/emissions/src/main/java/org/matsim/contrib/emissions/WarmEmissionHandler.java
```diff
@@ -174,12 +191,12 @@ private void emissionsCalculation(Id<Vehicle> vehicleId, Vehicle vehicle, Link l
...
 			if (zeroLinkLengthWarnCnt == 0) {
-				logger.warn("Length of the link " + linkId + " is zero. No emissions will be estimated for this link. Make sure, this is intentional.");
+				logger.warn("Length of the link " + linkId + " is zero. No emissions will be estimated for this link. Make sure that this is intentional.");
 				logger.warn(Gbl.ONLYONCE);
 				zeroLinkLengthWarnCnt++;
 			}
-			return true;
-		} else { return false; }
+			return false;
+		} else { return true; }
```
kt86, line 199: Why have you changed the return values?
I would expect the previous ones: If linkLength == 0 -> true, if not -> false.

contribs/emissions/src/test/java/org/matsim/contrib/emissions/events/VehicleLeavesTrafficEventsTest.java
```diff
@@ -0,0 +1,95 @@
...
+ * This class tests the (proper) handling of these cases as a result of
+ * the VehicleLeavesTrafficEvent method in the WarmEmissionHandler.
+ *
+ * @author Ruan J. Gräbe
+ */
+public class VehicleLeavesTrafficEventsTest {
+    private static final String hbefaColdFile = IOUtils.extendUrl( ExamplesUtils.getTestScenarioURL( "emissions-sampleScenario" ), "sample_41_EFA_ColdStart_vehcat_2020average.csv" ).toString();
+    private static final String hbefaWarmFile = IOUtils.extendUrl( ExamplesUtils.getTestScenarioURL( "emissions-sampleScenario" ), "sample_41_EFA_HOT_vehcat_2020average.csv" ).toString();
+    private static final String eventsFile =  "test/input/org/matsim/contrib/emissions/events/VehicleLeavesTrafficEventTest/smallBerlinSample.output_events.xml.gz";
+    private static final String vehiclesFile = "test/input/org/matsim/contrib/emissions/events/VehicleLeavesTrafficEventTest/smallBerlinSample_emissionVehicles.xml";
```
kt86, line 36: Please use MatsimTestUtils.getInputDirectory() OR MatsimTestUtils.getClassInputDirectory() + < Filename > instead of defining the path manually "test/input.../"

contribs/emissions/src/test/java/org/matsim/contrib/emissions/events/VehicleLeavesTrafficEventsTest.java
```diff
@@ -0,0 +1,95 @@
...
+    private static final String networkFile = "https://svn.vsp.tu-berlin.de/repos/public-svn/matsim/scenarios/countries/de/berlin/berlin-v5.5-10pct/input/berlin-v5.5-network.xml.gz";
+    private static final String outputDirectory = "test/output/org/matsim/contrib/emissions/events/VehicleLeavesTrafficEventTest/";
```
kt86, line 38: Analogous with the outputDir: MatsimTestUtils.getOutputDirectory() + < Filename >

contribs/emissions/src/test/java/org/matsim/contrib/emissions/events/VehicleLeavesTrafficEventsTest.java
```diff
@@ -0,0 +1,95 @@
...
+    private static final String inputDirectory = "test/input/org/matsim/contrib/emissions/events/VehicleLeavesTrafficEventTest/";
```
kt86, line 39: You do not need this seperatly, see MatsimTestUtils

contribs/emissions/src/test/java/org/matsim/contrib/emissions/events/VehicleLeavesTrafficEventsTest.java
```diff
@@ -0,0 +1,95 @@
...
+            eventWriterXML.closeFile();
+        } catch ( Exception e ) {
+            throw new RuntimeException( "Failing because there is no proper handling for the case where" +
+                    " a vehicle leaves traffic WITHOUT entering the link (no link enter time). A zero" +
+                    " emissions event should occur for these instances because there is no significant travel" +
+                    " on the link" );
+        }
+        // If the try-block executes, this ensures emission events occur as expected:
+        String expected = inputDirectory + emissionEventsFileName;
+        String actual = outputDirectory + emissionEventsFileName;
```
kt86, line 90: You can inline these both variables, they are short enough :)

kt86 (commented):
Why did you added the `network.xml.gz` as own file? Looks like you are using it from public-svn in the test, or?

I think both ways are ok: i) Using it from public-svn -> you do not need it here or ii) take this file here, but then please adapt the input of your test (and maybe reduce the network to the necessary data)

PR #2258, opened 2022-10-30
Fix wrong iteration handling for MATSimApplication

steffenaxer:
This PR fixes a wrong handling of the iterations argument of a MATSimApplication. Without this fix and a given iterations argument, there would be always one iteration too less.


rakow:
This was by design, because the number of iterations is the number of the last Iterations + 1. But probably it is more useful to use the last iteration instead, if that is what it is expected.

steffenaxer:
Yes in principle you are right, but everybody (at least me and three others) assumed independently that this number is last iteration as configured in controller config group, because this it was the user would expect.

rakow (approved):

PR #2264, opened 2022-11-04
add regression to electric taxi tests

kainagel:
Added testing for plans and events file equality for the etaxi tests.  Reason is that I changed the EvModule, and want to make sure that despite these changes behavior is still same.

At least two of the tests are non-deterministic.  I set the number of threads to 1; did not help.


michalmac:
There are a couple of ignored integration tests in the taxi contrib. This was due to segmentation faults on Jenkins (and only there). Now we can un-ignore all of them, since we migrated from Jenkins.

michalmac:
@kainagel Right now I am looking into the issue with non-deterministic runs.

kainagel:
@michalmac : Not sure if you saw my email.  There is a race condition between "charging_start" and "discharging" at the end of a link ... in some cases, that last discharge is included, in some not.  In some sense, this is even already noted in the code: 
```
//XXX The current implementation is thread-safe because no other EventHandler modifies battery charge
// (for instance, AUX discharging and battery charging modifies charge outside event handling
// (as MobsimAfterSimStepListeners)
//TODO In the long term, it will be safer to move the discharging procedure to a MobsimAfterSimStepListener
private EvDrive dischargeVehicle(Id<Vehicle> vehicleId, Id<Link> linkId, double eventTime) {
```
I added a discharging events, and with that I get
```
<event time="28356.0" type="charging_start" charger="charger_2" vehicle="taxi_1_4" chargerType="default" charge="2.00055E7"  />
<event time="28356.0" type="discharging" vehicle="taxi_1_4" deltaCharge="-85400.0" reason="drive" chargeAfter="1.99201E7"  />
```
in one case, and
```
<event time="28356.0" type="discharging" vehicle="taxi_1_4" deltaCharge="-85400.0" reason="drive" chargeAfter="1.99201E7"  />
<event time="28356.0" type="charging_start" charger="charger_2" vehicle="taxi_1_4" chargerType="default" charge="1.99201E7"  />
```
in the other.  One could argue that the writing of events is somewhat random, but one clearly sees that in the first case the discharge of that link taken from the battery is not included when charging starts.

kainagel:
What if we add an interaction time of 1 sec in `EvNetworkRoutingModule`?

michalmac:
@kainagel Thanks for looking into the problem. It seems you have spotted the right place where the things go wrong. This is how I see it:

1. General mobsim issue:

This EV functionality was written when DynActivityEngine was used in place of the default activity engine (and DynActivityEngine delegated to the default activity engine). This way both activity engines were called before the netsim engine was called.

Currently, DynActivityEngine and the default one are separate and DynActivityEngine is executed after the NetsimEngine, while the default one is executed before. This is a bit confusing that some activities are executed before and some after the netsim.

General question: should we depend on a specific order of engines or not? I guess this must have been discussed before.

2. EVRP-specific problem:

Driving-related energy consumption is computed from netsim events (that means the computation is in another thread and therefore not "fully" synced with the mobsim). All other changes to the battery charge (AUX consumption and charging) are computed inside mobsim, mostly from MobsimAfterSimStepListeners, however the initialisation of charging is executed as ChargingActivity simulated from DynActivityEngine (and this is where we have the race condition between processing the netsim events and ChargingActivity.initialize()).

Solution: I will try to move the charging initialisation from DynActivityEngine to MobsimAfterSimStepListeners.

BTW. This issue is not related to the generic EV functionality, it's only related to simulating EVRPs (i.e. e-drt, e-taxi)

michalmac:
@kainagel I updated this PR. Now the e-taxi simulations are reproducible.

michalmac:
@kainagel I rebased this PR on the latest master and enabled auto-merge. It's a valuable PR and it's good to have it merged given that we are going to make changes to the ev contrib in the near future.

PR #2269, opened 2022-11-08
Feature/coding conventions in hermes

jfbischoff:


vsp-gleich:
The @Deprecated annotation was introduced by mrieser in 2013 together with the comment "Should be replaced by some more generic VehicleArrivalEvent which supports both links and facilities." and maybe for cars that field delay really does not make sense, but it is really useful for pt. This is maybe part of why that single event type has not arrived in the 9 years after the comment.

vsp-gleich (approved):

Issue #1333, opened 2021-01-07
Should we put a editorconfig or similar into the project


Janekdererste:
I use autoformatting and clean up of imports and such things in IntelliJ, because I don't want to bother with formatting my code. I also am too lazy to change any of the default settings of my idea, because I don't really care how the code is formatted. Now, this leads to situations where my idea removes spaces in a case like the following

```
someMethod( someParameter );
```
because it thinks the spaces around the parameter don't belong there. If another person then edits this file the spaces are put in again because the other idea thinks spaces are important. This leads to a lot of edits and makes finding the actual changes in commits dificult. 

Should we place a code style file in the repository to make importing common formatting rules easy? Options are editorconfig, IntelliJ-style xml, Eclipse style xml, I think. 

In the JavaScript world opinionated formatters like [Prettier](https://prettier.io/) are often used. This prevents discussions about the right style. Maybe there is an option like this for Java as well?

I guess @kainagel @mrieser @michalmac @rakow @kt86 @vsp-gleich @jfbischoff might have an opinion on this.

michalmac:
We started a similar discussion some time ago: https://matsim.atlassian.net/browse/MATSIM-918.

In the meantime I added .editorconfig to a few dvrp-related modules. Seems to work, but it does not bring us close to having a unified formatting, which would greatly improve coding experiences.

rakow:
I always tend to add a  .editorconfig to my repos. The example of @Janekdererste is not covered by the standard, though. 

You can set them for Intelij anyway (https://blog.jetbrains.com/idea/2019/06/managing-code-style-on-a-directory-level-with-editorconfig/), but I don't know about other editors.

michalmac:
In the jira thread, one proposed option was to use google-java-format. It's really well maintained and widely supported (IDEs, editors...). But its formatting is so much different to ours, so that would be a huge change.

So maybe in our case having a MATSim-customised Eclipse formatting config is a better option? There is some support for mvn and IntelliJ:
- https://github.com/revelc/formatter-maven-plugin - validate code formatting or re-formatting code from mvn
- https://github.com/krasa/EclipseCodeFormatter - use the Eclipse formatter in IntelliJ

kainagel:
If those people who participate at this discussion here can agree on a solution, I will be ok with it.

My wishlist would be as follows:

(1) indentation by tabs (since then I can choose myself if I want wide or narrow indents)

[...]

PR #2277, opened 2022-11-09
Global .editorconfig

rakow:
This PR puts a top level .editorconfig into the project in order to make the indentation uniform.

Some vsp members were really frustrated with the different code styles, and especially with differences in tabs and spaces.
There also was already some discussion in #1333.

Many people still prefer quite different code styles. Internally, we agreed to make at least the indentation consistent, but don't force any of the other options.

Any suggestions of additions to the config are welcome. I also have removed the configs that were present in subprojects.


.editorconfig
```diff
@@ -1,20 +1,21 @@
...
+
 [*.java]
 indent_style = tab
-tab_width = 4
-trim_trailing_whitespace = true
 insert_final_newline = true
-ij_continuation_indent_size = 8
-max_line_length = 120
+trim_trailing_whitespace = true
+max_line_length = 150
```
michalmac, line 13: 120 is already a lot (when diffing in the split mode). Google Java Style Guide says even 100: https://google.github.io/styleguide/javaguide.html#s4.4-column-limit

.editorconfig
```diff
@@ -1,20 +1,21 @@
...
+
 
 [*.{xml, xsd, dtd}]
 max_line_length = off
 indent_style = tab
 trim_trailing_whitespace = true
 insert_final_newline = true
-indent_size = 4
+indent_size = 4
```
michalmac, line 21: Add a new line here? (just to be consistent with our config `insert_final_newline = true` :-) )

.editorconfig
```diff
@@ -1,20 +1,21 @@
...
 
+
 [*.java]
 indent_style = tab
-tab_width = 4
-trim_trailing_whitespace = true
 insert_final_newline = true
-ij_continuation_indent_size = 8
-max_line_length = 120
+trim_trailing_whitespace = true
```
michalmac, line 12: I see you removed: `tab_width = 4` and `ij_continuation_indent_size = 8`. I think I had to add both to make the continuation indentation work in IntelliJ (it got translated to 1 tab for normal indentation and 2 tabs for continuation indentation).

michalmac (commented):
Nice to make it standard across the whole repo.

My suggestions:
 - shorter lines (e.g. 120)
 - please test the settings in both IntelliJ and Eclipse before merging to master.

.editorconfig
rakow: I and KN are in favor of longer lines. Of course, some people prefer a more vertical code style., but one could always go lower than the limit.

.editorconfig
rakow: I think defining the tab size defeats the purpose of tabs, because one could make the indentation larger or narrower without changing the style. I haven't noticed any problems in IntelliJ, and we used a similar config for a long time.
Maybe it also has to do with the “smart tabs” features (https://www.jetbrains.com/help/idea/code-style-sql-tabs-and-indents.html) but I'm not sure.

.editorconfig
Janekdererste: Screens are wide nowadays.

michalmac: My screen is not small either. However, I often work with the editor window split vertically (so I see two files side by side) and I view git diffs in the split mode. In this case, anything beyond 120 lines is not visible.

However, I do not want to be the blocker here. 150 is OK. Anything between 120 and 150 is even better ;-)

.editorconfig
michalmac: I tried the new settings out and the continuation indentation works fine. For some reason, a few years ago, I had to add these two settings as a workaround. There is even issues reported for that:
- https://youtrack.jetbrains.com/issue/IDEA-266666/editorconfig-messing-up-javas-continuation-indent-of-long-lines (Java, older one)
- https://youtrack.jetbrains.com/issue/IDEA-254520/Continuation-indent-should-not-be-overriden-by-editorconfig (Kotlin, recent one).

michalmac: > I think defining the tab size defeats the purpose of tabs, because one could make the indentation larger or narrower without changing the style. 

@rakow Actually, not specifying `tab_width` and at the same time specifying `max_line_length` is also inconsistent. For instance, if we have 3 tabs of indentation at the beginning of a line, should we break the line after 126, 138 or 144 characters (tab sizes: 8, 4 and 2). I think we should declare the tab size to avoid different behaviours across different IDEs.

PR #2282, opened 2022-11-13
Initial version of DrtCompanion extension

steffenaxer:
**What it does:**
This PR provides the functionality to sample additional rides to an existing booked drt trip. Such a behavior is somehow required, if one would like to replicate more realistic occupancy rates in drt services. In other words. Without such an approach you will very likely underestimate real world vehicle occupancy and fleet sizes. The provided default sampling distribution is just a functional placeholder. Please fill it with your empirically observed values.

**How it works:**
Before starting the mobsim, additional, so called companions, are generated (sampled) by observing existing drt trips in the selected plan. The sampling distribution gets provided by DrtCompanionParams. At the end of the mobsim all companion agents are removed. They will never reach the replanning phase of MATSim.

**Remarks:**
Because the current drt contrib does not allow to book multiple seats per request, this approach is in my point of view the only straight-forward way to reach the desired objective. However @michalmac, @sebhoerl  maybe this PR is the initial spark to implement a more sophisticated approach. In between, have fun with this approach.

### DIFF
--- a/contribs/drt-extensions/src/main/java/org/matsim/contrib/drt/extension/DrtWithExtensionsConfigGroup.java
+++ b/contribs/drt-extensions/src/main/java/org/matsim/contrib/drt/extension/DrtWithExtensionsConfigGroup.java
@@ -1,0 +1,60 @@
+/* *********************************************************************** *
+ * project: org.matsim.*
+ *                                                                         *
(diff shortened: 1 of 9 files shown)


nkuehnel (approved):
LGTM, thanks! :)

PR #2283, opened 2022-11-14
drt: make DetourPathDataCache reusable by making it public final

nkuehnel:
Proposal to make DetourPathDataCache reusable by making it public final

We are currently in the midst of implementing an alternative insertion heuristic and would like to re-use this class


michalmac (approved):
LGTM

PR #2307, opened 2022-12-05
Emissions maintenance

kainagel:
mostly removing material that has been deprecated for many years

this may hick up, since it also changes material in matsim-examples: some config switches now do not exist any more and thus had to be removed from config files there.


contribs/emissions/src/main/java/org/matsim/contrib/emissions/ColdEmissionHandler.java
```diff
@@ -119,6 +119,7 @@ public void handleEvent(LinkLeaveEvent event) {
                     this.vehicleId2accumulatedDistance.put( vehicleId, distance );
                 }
                 // yyyy I have absolutely no clue what the distance stuff is doing here.  kai, jan'20
+                // I now think that this has to do with the fact that the cold emissions are smeared out over the initial distance.  Don't know the details, though.  kai, dec'22
```
Janekdererste, line 122: That is correct. We have two cold emission events. One right at the start of a trip and one after 1km.

PR #2311, opened 2022-12-08
Various optimizations, mostly reducing memory in large populations

mrieser:
- Activities and Legs both always had an Attributes object, even if no attributes are stored. Changed code such that the Attributes-object is only created when an actually attribute needs to be stored.
- Leg.routingMode was stored as an attribute only, although it is now used by default since a few years. Make `routingMode` a member of LegImpl to store it directly instead of storing it in the Attributes, resulting in a now often empty Attributes, which can be optimized away.
- Don't use OptionalTime as members in various places (ActivityImpl, LegImpl, various Routes) as the memory-overhead is just gigantic. The default java `Optional<T>` should also not be used as member, but only as parameter or return type in methods.
- Use a special implementation of Activity for interaction activities (sometimes called "stage activities" in the code), as such interaction activities typically do not have a start or end time and have the duration always set to 0.
- Adapt the existing CompressedNetworkRoute to provide better compression in multimodal networks. Renamed it to `HeavyCompressedNetworkRoute`.
- Introduce a new `MediumCompressedNetworkRoute` which provides substantial memory savings (but not as much as the HeavyCompressedNetworkRoute`), but with a lower performance overhead than the HeavyCompressedNetworkRoute.
- Integrate the members of `DefaultTransitPassengerRoute.RouteDescription` directly into `DefaultTransitPassengerRoute`. Use the inner class `RouteDescription` only for (de)serializing the route as short-lived objects.
- Reuse `ObjectMapper` for (de)serializing data in `DefaultTransitPassengerRoute` and `DRTRoute`. ObjectMapper is thread-safe, and creating a new object for every requests add quite a lot of performance overhead.
- When loading attributes from a file (e.g. person attributes, link attributes), each attribute name was its own String instance, even if all the names were the same, resulting in sometimes millions of identical Strings in the Java heap.

All these optimizations result in Populations using about 30% less memory with the default LinkNetworkRoutes, and simulations running 1-2% faster. When using HeavyCompressedMemoryRoutes, memory savings can reach up to 50% for large populations in our benchmarked scenario. Actual numbers will depend on network, modal split and other scenario characteristics.


jfbischoff:
Nice early santa present 👍

michalmac:
Thanks. Nice improvements! 🚀 

It's a good change that we stop storing `OptionalTime` in plans where we create millions of such.

General comment from my side: `OptionalTime` is a bit different to `Optional<T>`. Returning a nullable `T` is less risky than returning `UNDEFINED_TIME` (which is a valid `double` value). The latter may easily get unnoticed and lead to bugs that are hard to detect and fix. So we always need to make sure to wrap the `double` time with `OptionalTime` so that `UNDEFINED_TIME` will not leak.

mrieser:
@michalmac: We kept the API identical, so there is still `OptionalTime` used as parameter or as return value, we just don't use it to store the value but have the "interpretation" of `UNDEFINED_TIME` internally. It's not perfect either, as now each class like `ActivityImpl`, `LegImpl` etc. that uses `OptionalTime` in the API is now independently responsible for correctly storing and returning undefined times, and we potentially create more temporary and short-lived objects when the API is called.

PR #2316, opened 2022-12-11
ev: report MissingEnergyEvents in simulation instead of throwing exception

michalmac:
We still throw an exception in the planning (UrbanEVTripsPlanner).


michalmac:
A missing energy event looks like this:
```xml
	<event time="33562.0" type="missing_energy" vehicle="9" link="314328603_0" energy="136333.32392805014"  />
```

PR #2320, opened 2022-12-12
More accurate (slow) speeds with Hermes

KasiaKoz:
Right now Hermes speeds are rounded to integer and stored as an 8-bit unsigned integer. The maximum allowed speed is then 255m/s, which is quite fast. We would like to capture differences in slower speeds when modelling active modes like walking and cycling, especially if we give different agents different speeds---right now all walking speeds would be rounded from 1m/s for example.

This PR changes the way speeds are encoded:
- speeds between 0.0-9.9 have 1 decimal point accuracy: stored between integers 0-99
- speeds 10-165 (the new max speed, still pretty fast) is encoded between 100-255

❗ I had to change a couple of tests:
- `AgentTest.prepareLinkEntry` is fair enough since the speed encoding changed
- `TravelTimeTest.testEquilOneAgent` (and similarly in `TravelTimeTest.testEquilTwoAgents` - which was commented out.. I uncommented it, since it works - let me know if there was a reason to comment it out!). The travel time on one of the links (the only one with 5,000 length, all others are longer and unaffected by these changes) increased by a second. Before, in the travel time computation there was integer division returning `178`. Now, the velocity used is a double, which means the division returns a floating point answer of `178.57...` which is then rounded up when cast to integer.


jfbischoff:
Nice additions to Hermes, thanks @KasiaKoz !
--
If we ever need speeds >165m/s one could think to reduce the accuracy for high speeds. After all, it doesn't really matter if an aeroplane or bullet train travels with 200m/s or 201... but then again these modes would probably be simulated using deterministic PT and thus the whole thing doesn't really matter.

PR #2329, opened 2022-12-15
bicycle contrib now using additive PersonScoreEvents ...

kainagel:
and thus no longer overwriting scoring function.  Still overwriting the link speed calculation.


Janekdererste (approved):
I checked all failing tests events files were the same except for additional scoring events. Plans files were equal as well, scores only differed in last bit of a fraction. This is probably due to rounding errors during the summation of scoring events.

PR #2332, opened 2022-12-16
DRT: proposal for slack time calculator interface

nkuehnel:
In some projects we have a slightly different interpretation of slack times. If we had this interface, it would allow us to be more flexible and reduce code duplication.

### DIFF
--- a/contribs/drt-extensions/src/main/java/org/matsim/contrib/drt/extension/edrt/optimizer/EDrtVehicleDataEntryFactory.java
+++ b/contribs/drt-extensions/src/main/java/org/matsim/contrib/drt/extension/edrt/optimizer/EDrtVehicleDataEntryFactory.java
@@ -20,6 +20,7 @@ package org.matsim.contrib.drt.extension.edrt.optimizer;
 
 import java.util.List;
 
+import org.matsim.contrib.drt.optimizer.SlackTimeCalculator;
 import org.matsim.contrib.drt.optimizer.VehicleEntry;
 import org.matsim.contrib.drt.optimizer.VehicleDataEntryFactoryImpl;
 import org.matsim.contrib.drt.run.DrtConfigGroup;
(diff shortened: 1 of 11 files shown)


michalmac (commented):
@nkuehnel The PR is correct, but before approving it I wanted to ask for some more explanation. Maybe some of your answers could be added as comments to the new interface.

Until now, the slack calculation was used as an internal performance enhancement. With this PR, it becomes something visible outside and customisable, which means the slack may be interpreted/modelled in different ways. I am wondering what is your slack interpretation and if it is consistent with the insertion algorithm.

michalmac:
I do not know exactly the use case, but I think it is safer to create `Waypoint.Stop`s with customised latest departure times rather than manipulating cached slack times. Otherwise, it is hard to guarantee that everything will work correctly when the genuine input values (departure times) and the derived slack times do not match.

There is already a constructor that takes in the latest departure time:
```
public Stop(DrtStopTask task, double latestArrivalTime, double latestDepartureTime, int outgoingOccupancy)
```
Maybe this is enough. If not, please suggest required changes.

nkuehnel:
Yes that was my other suggestion regarding the Waypoint.Stop class. This would require to make the class public:
https://github.com/matsim-org/matsim-libs/blob/26e14fa14b79b1c418ebf7f5c7e27e8ea6b0c238/contribs/drt/src/main/java/org/matsim/contrib/drt/optimizer/Waypoint.java#L140

This has the downside that one has to duplicate the VehicleDataEntryFactoryImpl class to actually call the constructor differently but I guess that's okay..

michalmac:
> Yes that was my other suggestion regarding the Waypoint.Stop class

Yes, that wasn't clear from my reply, but I referred there to your other approach  :-)

> This would require to make the class public

It's already public (it is defined inside an interface)

> one has to duplicate the VehicleDataEntryFactoryImpl class 

:-( However, a bit of duplication is often better than adding complexity around making every piece reusable. But let's see how this goes. If you find some parts could be reused, please suggest the changes.

nkuehnel:
> It's already public (it is defined inside an interface)

Right!

I will then close the PR for now and see how that goes, thanks for your feedback :) 
We are currently looking at adding transfers to DRT - the committed transfer waypoints should then make sure that the slack time does not allow shifting them around too much.

We will probably have something to open source for that soon, happy to discuss the implementation further at that time

PR #2333, opened 2022-12-16
ev contrib maintenance

kainagel:
There some similar branches, which did not complete because I did not have time, and then master had to many changes to still merge.  The goal is:

* change ev contrib integration tests so that they also do regression (based on plans and events)
* pull code that is currently in the user code into `EvModule`.

I more or less got this done.  See `RunEvExample`.  It can now be run by _only_ using `EvModule`, i.e. it is more difficult to "silently" forget other pieces.  What I did _not_ include into `EvModule` was the `EvNetworkRoutingProvider`.  Reason is that this is just one way to come up with charging activities, i.e. when the route is longer than the initial charge can last.  There are other ways to generate charging activities.  ( Can't say if this is a good design decision. )


michalmac (approved):
Nice refactorings. Also great to have regression unit tests.

I added one comment regarding improving some potentially misleading javadocs.

kainagel:
After commit 9a7b8e29f09eb5933cad82255c5f900dca18a1a0: 

when I run, on branch evMaintenance3, RunETaxiScenarioIT#testRuleBased alone, then it passes.

When I run the whole class, then it fails with 

2022-12-21T10:23:07,882  WARN EventsFileComparator$EventsComparator:167 The event:
2022-12-21T10:23:07,883  WARN EventsFileComparator$EventsComparator:168  | fromLink=291 | latestDropoffTime=38188.0 | latestPickupTime=37424.0 | mode=taxi | person=0013664 | request=taxi_361 | time=33824.0 | toLink=534 | type=DrtRequest submitted | unsharedRideDistance=4183.0 | unsharedRideTime=382.0 | 
2022-12-21T10:23:07,883  WARN EventsFileComparator$EventsComparator:169 is missing in events file:test/output/org/matsim/contrib/etaxi/run/RunETaxiScenarioIT/testRuleBased//output_events.xml.gz
2022-12-21T10:23:07,883  WARN EventsFileComparator$EventsComparator:187 The event:
2022-12-21T10:23:07,883  WARN EventsFileComparator$EventsComparator:188  | fromLink=291 | latestDropoffTime=38188.0 | latestPickupTime=37424.0 | mode=taxi | person=0013664 | request=taxi_361 | time=33824.0 | toLink=534 | type=DrtRequest submitted | unsharedRideDistance=4254.0 | unsharedRideTime=382.0 | 
2022-12-21T10:23:07,883  WARN EventsFileComparator$EventsComparator:189 is missing in events file:test/input/org/matsim/contrib/etaxi/run/RunETaxiScenarioIT/testRuleBased//output_events.xml.gz

Any idea what that could be?  Evidently, the difference is in the unsharedRideDistance.  Where I cannot imagine why the typical things, race conditions or public static non-final variables, might play a role.

There was no regression test before I started this, so maybe this was never caught.

What should I do?  

From a purely technical perspective I could make the events file comparison ignore the unshared ride distance.  Intuitively, this does not make sense.

michalmac:
This is related to the caching of Ids which are shared by multiple tests:
- when running separate tests - everything is OK
- when running a sequence of tests from maven - everything is OK as I added `AutoResetIdCaches` to mvn surefire and failsafe configs (https://github.com/matsim-org/matsim-libs/blob/master/pom.xml#L330 and https://github.com/matsim-org/matsim-libs/blob/master/pom.xml#L342)
- when running a sequence of tests from IDE - we either need a plugin for IDE (I used [JUnit 4 Surefire Listener](https://plugins.jetbrains.com/plugin/15718-junit-4-surefire-listener), unfortunately now incompatible with the latest IntelliJ) or we need to manually reset Ids in all tests (which is very tedious)

PR #2335, opened 2022-12-20
EventsFileComparator: remove deprecated method 

### LINKED ISSUE #2327: MatsimTestUtils.compareEventsFiles(...) calls deprecated comparing method
https://github.com/matsim-org/matsim-libs/blob/bc5b9f001d020d4b82804adbdd82db7ea8fdf759/matsim/src/test/java/org/matsim/testcases/MatsimTestUtils.java#L360-L362

calls a deprecated comparison method: 

https://github.com/matsim-org/matsim-libs/blob/bc5b9f001d020d4b82804adbdd82db7ea8fdf759/matsim/src/main/java/org/matsim/utils/eventsfilecomparison/EventsFileComparator.java#L73-L97

The correct one is - avoiding the transformation of the `Result` to an `int:`
https://github.com/matsim-org/matsim-libs/blob/bc5b9f001d020d4b82804adbdd82db7ea8fdf759/matsim/src/main/java/org/matsim/utils/eventsfilecomparison/EventsFileComparator.java#L128-L130

Should we just change it, accepting, that this will maybe break code / let tests behave differently, because the return value of this method changes...?

Or should we add (another) method to `MATSimTestUtils`?

kt86:
fixes #2327 

Removing the (long-time) deprecated method `compareAndReturnInt(f1, f2)`, which returns the result as an int.
Using `compare(f1, f2)` instead, which will directly return the Result as an enum value.


michalmac (approved):
Nice!

PR #2336, opened 2022-12-20
MATSimTestUtils: some renamings to align task and name of methods

### LINKED ISSUE #2328: MATSimTestUtils.compareEventsFiles(...) does not lead to failing test ...
... if event files differ.

It only prints the differences as a warning to the console/logfile. 
Additionally, it returns an int of the result (see also #2327), which at least I am **not** using by default :
https://github.com/matsim-org/matsim-libs/blob/bc5b9f001d020d4b82804adbdd82db7ea8fdf759/matsim/src/test/java/org/matsim/testcases/MatsimTestUtils.java#L360-L362

_I am not sure if that is the expected behavior?_ 

For me, it is NOT. I would assume, that it will lead to failing tests (by throwing an exception) - as e.g. ` compareFilesBasedOnCRC(...)` or `compareFilesLineByLine(...)` in the same class are doing.

Possible solution approaches:
1. Adding another method, which will fail if event files differ.
2. Adapt the behavior to the other compare classes -> Assertion Error / RunTimeException. **Keep the return value** as it is. 
3. Adapt the behavior to the other compare classes -> Assertion Error / RunTimeException. **Change the return value** to void. 
4. Adding another method, with an extended signature to include a behavioral decision: failIfDiffer?: true/false
5.  ...

Any opinion on that? @kainagel @michalmac

kt86:
resolves #2328 

`compareXYZ` should return a result of the comparing step.
`assertEqualXYZ` should return lead to an Assertion error, if the result of the comparing step is _not_ equal. 

This was a bit confusing before.

@michalmac : I decided to not throw an exception and instead use `Assert.assertEqual(...)`. For me, this feels to fit better for the testing scope. If you would have an Exception instead, please feel free to change this.


michalmac (approved):
Nice refactoring!

PR #2338, opened 2022-12-23
Add parking costs

jfbischoff:
This PR adds the functionality to easily charge parking costs depending on link attributes. Both network and (pseudo)-teleported modes can be accessed.
Parking Charges are a major factor in mode choice - we have been using this code internally a long time and think it might be useful for others too. 
Effects of parking charges will depend on the handling of money events in the MATSim Scoring process.

### DIFF
--- a/contribs/parking/README.md
+++ b/contribs/parking/README.md
@@ -1,5 +1,15 @@
-
 # Parking
 
-The MATSim simulation by default does not consider parking infrastructure or supply constraints.
-For technical reasons, the modeling efforts around parking in MATSim are divided in two parts parking choice and parking search.
+The MATSim simulation by default does not consider parking infrastructure or supply constraints. There are different
+approaches as to how parking can be handled in MATSim, depending on the use case.
+
+- Parking Choice, based on work of Rashid Waraich. Parts of the contrib are used in the carsharing contrib. Its main
+  goal, as far as we understand it, is the simulation of parking choice (e.g. between garages). There is no modelling of
+  circulating traffic searching for parking or walking of agents to/from parking.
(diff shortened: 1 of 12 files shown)


tschlenther:
Thank you :) Looks like a proper and tested variant of a [package in the vsp contrib](https://github.com/matsim-org/matsim-libs/tree/master/contribs/vsp/src/main/java/playground/vsp/simpleParkingCostHandler) I have recently played with. The latter also allows to model residential parking fees or daytickets etc. We might want to move some of that over to the parking contrib.
BTW: We were thinking about potentially (re-) moving the parking choice package. Are you aware of any current applications or have any other objections?

jfbischoff:
> The latter also allows to model residential parking fees or daytickets etc.
Nice additions. Daypassses / daily cappings could be handy indeed. 


>. Are you aware of any current applications or have any other objections?

No, I have no clue what it is good for. Maybe @balacmi knows more about it. I think some parts have been used in Carsharing (at least the documentation states that), but that could relate to some utilities only.

manserpa:
That's awesome, thanks @jfbischoff - really helpful for us (@gac55 )

manserpa:
BTW: As mentioned, capped fares (besides parking also public transport, road pricing, ...) are a very big topic. An example for an appropriate implementation in MATSim would be cool :)

tschlenther:
For road pricing, see the corresponding 'roadpricing' contrib ;) Fees in public transport are currently modeled in different ways, as far as I know. However, the tendency is to use a custom handler that throws `PersonMoneyEvent`s. One has just to be careful in terms of taking this into account while routing, then....

PR #2345, opened 2023-01-02
bind LinkSpeedCalculator by mode

kainagel:
Steps towards https://github.com/matsim-org/matsim-libs/issues/2337 .  Text from there:

I am just looking at the bicycle contrib. It overwrites the QNetworkFactory so that it can set the LinkSpeedCalculator.

This means that it cannot be combined with anything else that overwrites the QNetworkFactory. (At least not without manual programming.)

My idea to make this more orthogonal would be:

allow things like LinkSpeedCalculator to be injected into the Default(!?)QNetworkFactory

make this injection by mode so that different contribs can set these things independently by mode

Wanted to ask if anybody can see a reason why this would be a stupid idea.

---

A disadvantage would be that we would get something like this in the inner QSim loop:
```
... linkSpeedCalculators.get(vehicle.getMode()).getMaximumVelocity(...)
```
That is, a get that would be an associative array. Can't say how much speed this would cost.

### DIFF
--- a/contribs/bicycle/src/main/java/org/matsim/contrib/bicycle/BicycleModule.java
+++ b/contribs/bicycle/src/main/java/org/matsim/contrib/bicycle/BicycleModule.java
@@ -20,13 +20,17 @@ package org.matsim.contrib.bicycle;
 
 import com.google.inject.Inject;
 import com.google.inject.Singleton;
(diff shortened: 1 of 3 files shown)


kainagel:
see #2617 .

PR #2348, opened 2023-01-04
Deprecate MATSimTestCase

kt86:
The `MatsimTestCase` is the "old" infrastructure for providing some standardized helper methods for JUnit-testing.
We do have the `MatsimTestUtils` now for a while (since using JUnit 4). The `MatsimTestUtils` are also providing more methods. 

Thus, it is time to phase out the old infrastructure - and avoid that people will use start using it again.

This PR includes:
- Deprecate `MatsimTestCase`
- Replace `MatsimTestCase.EPSILON` by `MatsimTestUtils.EPSILON` (inside matsim-libs)
- Update the tests in the freight contrib


kt86:
> Nice. Is it possible to remove `MatsimTestCase` or is it still used somewhere in matsim-lib?

It is flying around more or less everywhere (approx. 150 inheritors of that class) -.-

And replacing it needs some manual work and is not only direct refactoring ... (except the EPSILION, which was easy to replace.) 

-> I started only with the contrib, I feel responsible for (freight). Hope that some other developer will do it for their stuff ;) Or it is a possible task for a code-sprint.

But I think it is also a good first step, if we no one starts using the "old" infrastructure anymore.

PR #2404, opened 2023-02-08
Freight: Introduce an EventsReader for freight events

kt86:
For reading in the freight specific events.

It bases on the [DvrpEventsReaders](https://github.com/matsim-org/matsim-libs/blob/master/contribs/dvrp/src/main/java/org/matsim/contrib/dvrp/util/DvrpEventsReaders.java).

Please not: It is not complete yet --> More Events-converter will come soon.


kt86:
Please note: The `Freight<...>Event`s were renamed to `Carrier<...>Event` in the meantime.

PR #2417, opened 2023-02-16
abort handler

kainagel:


contribs/drt/src/test/java/org/matsim/contrib/drt/run/examples/DrtAbortTest.java
```diff
@@ -0,0 +1,425 @@
...
+		private InternalInterface internalInterface;
+		private final List<MobsimAgent> agents = new ArrayList<>();
+
+		List<String> drtModes = new ArrayList<>();
+		Map<String,Double> alphas = new HashMap<>();
+		Map<String,Double> betas = new HashMap<>();
+
+		@Inject MyAbortHandler(Network network, Map<String, TravelTime> travelTimeMap, Config config ){
+			travelTime = travelTimeMap.get( TransportMode.car );
+			router = new SpeedyALTFactory().createPathCalculator( network, new OnlyTimeDependentTravelDisutility( travelTime ), travelTime );
```
michalmac, line 230: This router is created with for the default network and car travel times. This may not work for all drt modes.

PR #2442, opened 2023-03-09
vsp: fix UrbanVehicleChargingHandler

michalmac:
Add vehicles queued at the charger to the ongoing charging procedures.

Currently, we can observe the following issue:
- A vehicle gets queued at a charger (plugin interaction activity)
- A plugout interaction activity is executed while the vehicle is still in the queue
- Because the vehicle was not included in the ongoing charging procedures, it does not get removed from the charger logic
- The charging continues, often despite the vehicle moving somewhere in the network (!)
- At some point the vehicle gets fully charged and so the charging end event is triggered
- In response to this event, UrbanVehicleChargingHandler tries to remove the vehicle from the ongoing charging procedures, which causes an NPE (as the vehicle has not been registered there)

### DIFF
--- a/contribs/vsp/src/main/java/playground/vsp/ev/UrbanVehicleChargingHandler.java
+++ b/contribs/vsp/src/main/java/playground/vsp/ev/UrbanVehicleChargingHandler.java
@@ -52,7 +52,7 @@ import java.util.*;
  */
 class UrbanVehicleChargingHandler
 		implements ActivityStartEventHandler, ActivityEndEventHandler, PersonLeavesVehicleEventHandler,
-		ChargingEndEventHandler, ChargingStartEventHandler, MobsimScopeEventHandler {
+		ChargingEndEventHandler, ChargingStartEventHandler, QueuedAtChargerEventHandler, MobsimScopeEventHandler {
 
 	private final Map<Id<Person>, Id<Vehicle>> lastVehicleUsed = new HashMap<>();
 	private final Map<Id<Vehicle>, Id<Charger>> vehiclesAtChargers = new HashMap<>();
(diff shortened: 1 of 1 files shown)


michalmac:
This fixes issue matsim-org/matsim-code-examples#890. In this simulation the bug caused the vehicle was being charged at two chargers at the same time:
```xml
        <!-- the vehicle gets queued (but no registration happens) -->
        <event time="47429.0" type="queued_at_charger" charger="23" vehicle="262_car"  />
        <!-- the vehicle leaves the charger, but the charging procedure is not stopped (so the vehicle remains queued) -->
        <event time="69120.0" type="vehicle enters traffic" person="262" link="93" vehicle="262_car" networkMode="car" relativePosition="1.0"  />
        <event time="78020.0" type="vehicle leaves traffic" person="262" link="84" vehicle="262_car" networkMode="car" relativePosition="1.0"  />
        <!-- the vehicle arrives at another charger, here charging starts immediately (without queueing). -->
        <!-- a 4-second gap between arriving and the start of charging, because charging is simulated every `n` seconds -->
        <event time="78024.0" type="charging_start" charger="18" vehicle="262_car" chargerType="default" charge="2406505.180896909"  />
        <!-- the vehicle finally starts charging at the charger from which it should have been removed; now charging happens at two chargers in parallel -->
        <event time="82064.0" type="charging_start" charger="23" vehicle="262_car" chargerType="default" charge="4.684650518089691E7"  />
        <!-- now the vehicle is fully charged and gets disconnected from the second charger ... -->
        <event time="82879.0" type="charging_end" charger="18" vehicle="262_car" charge="7.2E7"  />
        <!-- ... and from the first charger, which causes the reported NPE -->
        <event time="82879.0" type="charging_end" charger="23" vehicle="262_car" charge="7.2E7"  />
```

michalmac:
ping @tschlenther

tschlenther:
I am very grateful! Thanks for taking care! ping @rgraebe

PR #2455, opened 2023-03-17
Add late diversion constraint for drt detours

nkuehnel:
proposal:
usually, customers do not like being diverted shortly before their arrival, even though their time window would permit.
This PR introduces a hard constraint to the default DRT algorithm to allow setting a time from which it is no longer allowed to divert for new requests if a boarded passenger is very close to dropoff. By default it is set to 0 (as before).

Feedback / reviews welcome
@michalmac


contribs/drt/src/main/java/org/matsim/contrib/drt/run/DrtConfigGroup.java
```diff
@@ -184,6 +184,15 @@ public enum OperationalScheme {
 	@PositiveOrZero
 	public double advanceRequestPlanningHorizon = 0; // beta-feature; planning horizon for advance (prebooked) requests
 
+	@Parameter
+	@Comment(
+			"Time before reaching a planned dropoff from which it is not allowed to insert new detours for new requests. I.e.," +
+					" if set to 180, then a vehicle will not divert to pickup a new passenger once a boarded passenger is only " +
```
michalmac, line 190: Suggestion: "divert to pickup a new passenger" -> "divert to pick up or drop off a new passenger"

contribs/drt/src/main/java/org/matsim/contrib/drt/optimizer/insertion/DefaultInsertionCostCalculator.java
```diff
@@ -55,6 +60,23 @@ public double calculate(DrtRequest drtRequest, Insertion insertion, DetourTimeIn
 			return INFEASIBLE_SOLUTION_COST;
 		}
 
+		// divert right now
+		if(insertion.pickup.index == 0) {
+			Waypoint nextWaypoint = insertion.pickup.nextWaypoint;
+			// there is an existing stop following the new insertion
+			if(nextWaypoint instanceof Waypoint.Stop && nextWaypoint != insertion.pickup.newWaypoint) {
+				// passengers are being dropped off == may be close to arrival
+				if(!((Waypoint.Stop) nextWaypoint).task.getDropoffRequests().isEmpty()) {
```
michalmac, line 69: What if the second next waypoint is dropping off and is within the defined time limit?

contribs/drt/src/main/java/org/matsim/contrib/drt/run/DrtConfigGroup.java
```diff
@@ -184,6 +184,15 @@ public enum OperationalScheme {
...
+					"3 minutes away from her destination, even though her time window would allow it." +
+					" Delayed detour just before arrival are usually perceived very negatively.")
+	@PositiveOrZero // used only for stopbased DRT scheme
+	public double fixedApproachTime = 0;// [m];
```
michalmac, line 194: `fixedApproachTime` is too general. It will be very hard to memorise what it means. I think we need a name that would be more expressive.

contribs/drt/src/main/java/org/matsim/contrib/drt/optimizer/insertion/DefaultInsertionCostCalculator.java
```diff
@@ -55,6 +60,23 @@ public double calculate(DrtRequest drtRequest, Insertion insertion, DetourTimeIn
...
+					double nextArrival = nextWaypoint.getArrivalTime();
+					double departureTime = insertion.vehicleEntry.start.getDepartureTime();
+					//arrival is very soon
+					if (nextArrival - departureTime < drtConfigGroup.fixedApproachTime) {
+						return INFEASIBLE_SOLUTION_COST;
+					}
+				}
+			}
+		}
+
```
michalmac, line 79: Have you maybe thought about wrapping `DefaultInsertionCostCalculator` with `FixedArrivalTimeCostCalculator`. This way we can use guice to add this additional condition while keeping the config untouched?

contribs/drt/src/main/java/org/matsim/contrib/drt/run/DrtConfigGroup.java
nkuehnel: How would you divert for a drop off of a passenger that hasn't boarded yet? Or do you mean the case where both pickup and dropoff of the new passenger are inserted before the next stop?

contribs/drt/src/main/java/org/matsim/contrib/drt/optimizer/insertion/DefaultInsertionCostCalculator.java
nkuehnel: That's a good point. This means we need to iterate over all scheduled stops!

contribs/drt/src/main/java/org/matsim/contrib/drt/run/DrtConfigGroup.java
michalmac: Yes, that's possible, esp. if someone sets this param to larger values.

contribs/drt/src/main/java/org/matsim/contrib/drt/optimizer/insertion/DefaultInsertionCostCalculator.java
michalmac: ... all scheduled stops within the `fixedApproachTime` horizon

contribs/drt/src/main/java/org/matsim/contrib/drt/optimizer/insertion/DefaultInsertionCostCalculator.java
nkuehnel: Yes I thought about this but I liked the fact of having it configurable through the config. If you think it is too specific we can also go with the wrapper option - but that means one would need to always set it programmatically right?

michalmac: I think this is a nice feature that can be potentially used in many scenarios, so having it in the config sounds good to me.

I am not sure if all algorithms will be able to support this feature, but that is a different problem that we need to address in "drt config v2.0"...

contribs/drt/src/main/java/org/matsim/contrib/drt/run/DrtConfigGroup.java
nkuehnel: Done.

contribs/drt/src/main/java/org/matsim/contrib/drt/run/DrtConfigGroup.java
```diff
@@ -184,6 +184,15 @@ public enum OperationalScheme {
 	@PositiveOrZero
 	public double advanceRequestPlanningHorizon = 0; // beta-feature; planning horizon for advance (prebooked) requests
 
+	@Parameter
+	@Comment(
+			"Time before reaching a planned dropoff from which it is not allowed to insert new detours for new requests. I.e.," +
+					" if set to 180, then a vehicle will not divert to pickup a new passenger once a boarded passenger is only " +
+					"3 minutes away from her destination, even though her time window would allow it." +
+					" Delayed detour just before arrival are usually perceived very negatively.")
```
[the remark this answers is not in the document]
nkuehnel: Thanks, done!

contribs/drt/src/main/java/org/matsim/contrib/drt/optimizer/insertion/DefaultInsertionCostCalculator.java
nkuehnel: Hi @michalmac 
I updated the code to check for all scheduled stops and check whether the actual detour caused is >0 (in some configurations it could also be that the new insertion is after the next stop and only adds an additional passenger to an existing stop without increasing stop duration, in which case there wouldn't be any additional detour)

contribs/drt/src/main/java/org/matsim/contrib/drt/run/DrtConfigGroup.java
nkuehnel: How about "allowDetourBeforeArrivalThreshold"?

contribs/drt/src/main/java/org/matsim/contrib/drt/optimizer/insertion/DefaultInsertionCostCalculator.java
nkuehnel: 👍

michalmac:
@nkuehnel Let me know when it's ready for re-review. Thanks!

nkuehnel:
@michalmac it is, I guess :)

contribs/drt/src/main/java/org/matsim/contrib/drt/optimizer/insertion/DefaultInsertionCostCalculator.java
```diff
@@ -55,6 +60,32 @@ public double calculate(DrtRequest drtRequest, Insertion insertion, DetourTimeIn
 			return INFEASIBLE_SOLUTION_COST;
 		}
 
+		// all stops after the new (potential) pickup but before the new dropoff
+		// are delayed by pickupDetourTimeLoss
+		double detour = detourTimeInfo.pickupDetourInfo.pickupTimeLoss;
```
michalmac, line 65: `detour` -> `timeLoss`?

contribs/drt/src/main/java/org/matsim/contrib/drt/optimizer/insertion/DefaultInsertionCostCalculator.java
```diff
@@ -55,6 +60,32 @@ public double calculate(DrtRequest drtRequest, Insertion insertion, DetourTimeIn
...
+		if(vEntry.stops != null) {
+			for (int s = insertion.pickup.index; s < vEntry.stops.size(); s++) {
+				Waypoint.Stop stop = vEntry.stops.get(s);
+				// passengers are being dropped off == may be close to arrival
```
michalmac, line 69: This should limit the loop to the stops within the the time threshold:
```java
if (nextArrival - departureTime >= drtConfigGroup.allowDetourBeforeArrivalThreshold) {
    break;
}
```

contribs/drt/src/main/java/org/matsim/contrib/drt/optimizer/insertion/DefaultInsertionCostCalculator.java
```diff
@@ -55,6 +60,32 @@ public double calculate(DrtRequest drtRequest, Insertion insertion, DetourTimeIn
...
+				if (!stop.task.getDropoffRequests().isEmpty()) {
+					double nextArrival = stop.getArrivalTime();
+					double departureTime = insertion.vehicleEntry.start.getDepartureTime();
+					//arrival is very soon
+					if (nextArrival - departureTime < drtConfigGroup.allowDetourBeforeArrivalThreshold &&
```
michalmac, line 74: The condition `nextArrival - departureTime < drtConfigGroup.allowDetourBeforeArrivalThreshold` can be removed if the other `if` is added (see comment above).

contribs/drt/src/main/java/org/matsim/contrib/drt/optimizer/insertion/DefaultInsertionCostCalculator.java
```diff
@@ -55,6 +60,32 @@ public double calculate(DrtRequest drtRequest, Insertion insertion, DetourTimeIn
...
+							detour > 0) {
+						return INFEASIBLE_SOLUTION_COST;
+					} else {
+						// all following stops are above the threshold
```
michalmac, line 78: Not sure if this is correct if `detour == 0` and we are still within the time threshold.

If this is indeed incorrect, consider adding such a case to your tests (BTW. Nice that you have already added some tests!).

contribs/drt/src/main/java/org/matsim/contrib/drt/optimizer/insertion/DefaultInsertionCostCalculator.java
```diff
@@ -55,6 +60,32 @@ public double calculate(DrtRequest drtRequest, Insertion insertion, DetourTimeIn
...
+						break;
+					}
+				}
+				if (s == insertion.dropoff.index) {
+					// all stops after the new (potential) dropoff are delayed by totalTimeLoss
+					detour = detourTimeInfo.getTotalTimeLoss();
```
michalmac, line 84: Consider adding:
```java
if (detour == 0) {
    break;
}
```

contribs/drt/src/main/java/org/matsim/contrib/drt/run/DrtConfigGroup.java
```diff
@@ -184,6 +184,15 @@ public enum OperationalScheme {
...
 	public double advanceRequestPlanningHorizon = 0; // beta-feature; planning horizon for advance (prebooked) requests
 
+	@Parameter
+	@Comment(
+			"Time before reaching a planned dropoff from which it is not allowed to insert new detours for new requests. I.e.," +
+					" if set to 180, then a vehicle will not divert to pickup or dropoff a new passenger once a boarded passenger is only " +
+					"3 minutes away from her destination, even though her time window would allow it." +
+					" Delayed detours just before arrival are usually perceived very negatively.")
+	@PositiveOrZero
+	public double allowDetourBeforeArrivalThreshold = 0; // [s];
```
michalmac, line 194: This name is much better than the previous one!

michalmac (commented):
I added a few comments. Please consider them.

nkuehnel:
Picked this old PR up again and updated it to latest matsim/ addressed old review comments

contribs/drt/src/main/java/org/matsim/contrib/drt/optimizer/insertion/DefaultInsertionCostCalculator.java
nkuehnel: done

contribs/drt/src/main/java/org/matsim/contrib/drt/optimizer/insertion/DefaultInsertionCostCalculator.java
nkuehnel: updated

PR #2461, opened 2023-03-22
CarrierPlan will contain no services tag if services are empty

nixlaos:
writeServices method discontinued if carrier contains no services


kt86:
also fix, that more than one tour can be available (change in xsd.)

PR #2462, opened 2023-03-22
replace http with https for dtd file paths

vsp-gleich:
(http://www.matsim.org/files/dtd -> https://www.matsim.org/files/dtd)


tschlenther:
@vsp-gleich do we still need this PR ord can you close it?

kt86:
I think that this makes totally sence. I also stumbled over this issue a few weeks ago. During that I tried to change it at least for some freight related stuff, but cancelled it, because currently we cannot automatically update the DTD/XSD files on matsim.org/files/dtd .. 
(and in these files all the things must be updated as well).

So from my point of view: I make sense to update, but let's wait until the automatic update of the files works again.
(And maybe restart with a new PR instead of trying to fix this one)

vsp-gleich:
We can close this PR, it did not work out a year ago and now there are too many conflicts.

Issue #2463, opened 2023-03-23
ChartUtils not using the tmp folder inside the outputDirectory


tkchouaki:
Hi, 
The ChartUtil class in org.matsim.core.utils.charts relies on the JFreeChart library which in turn relies on javax.imageio. The latter uses the FileCacheImageOutputStream class from javax.imageio.stream. This class creates a temporary file as a cache for the image files (e.g, leghistogram) and it is currently not stored in the tmp folder inside the outputDirectory but rather in the system-dependent tmp folder.

I encountered this issue after simulations stopped due to a "No space left on device" error while the drive in which my output directory is located is far from capacity, this was because the drive containing the tmp folder was full. 

The solution would be to call ImageIO.setCacheDirectory with the right directory path. 
If this is an issue, I can quickly make a PR.

tkchouaki:
I suggested a fix for this issue in #2468. However I'm not sure it would fix #2052

Janekdererste:
Thanks, I willl merge the PR :-)

PR #2468, opened 2023-03-24
Fixing #2463

tkchouaki:
Sets the Cache File directory used by ChartUtils to the tmp folder in the output directory


michalmac (approved):
Thanks! LGTM.

PR #2488, opened 2023-04-04
ev: individual time profiles of power supplied from charging stations

mattiasingelstrom:
This added stat will be valuable to analyse the power demand of each charging station as well as its utility rate. The output file has the following appearance with the values for each charging station being in kW: 
```
time charger1 charger2 charger3     
09:05	0.0	0.0	0.0
09:10	0.0	0.0	0.0
09:15	0.0	0.0	65.0
09:20	0.0	0.0	130.0
09:25	0.0	0.0	195.00000000000148
09:30	0.0	0.0	195.0
09:35	0.0	0.0	195.0
09:40	0.0	0.0	260.0
09:45	0.0	0.0	260.0
09:50	0.0	0.0	260.0
09:55	0.0	0.0	325.0
10:00	0.0	0.0	325.0
```

### DIFF
--- a/contribs/freight/src/main/java/org/matsim/contrib/freight/jsprit/MatsimJspritFactory.java
+++ b/contribs/freight/src/main/java/org/matsim/contrib/freight/jsprit/MatsimJspritFactory.java
@@ -72,51 +72,27 @@ public final class MatsimJspritFactory {
 	@SuppressWarnings("unused")
 	private static final  Logger log = LogManager.getLogger(MatsimJspritFactory.class);
 
-	// How to deal with a multi-depot VRP? Which depotLink should be used? kmt
-	// jul/18
-
-	/**
-	 * Creates (MATSim) CarrierShipment from a (jsprit) service
-	 *
-	 * @param service   to be transformed to Shipment
-	 * @param depotLink as from-link for Shipment
-	 * @return CarrierShipment
-	 * @see CarrierShipment, Service
-	 */
-	static CarrierShipment createCarrierShipmentFromService(Service service, String depotLink) {
-		CarrierShipment carrierShipment = CarrierShipment.Builder
-				.newInstance(Id.create(service.getId(), CarrierShipment.class), Id.create(depotLink, Link.class),
-						Id.create(service.getLocation().getId(), Link.class), service.getSize().get(0))
-				.setDeliveryServiceTime(service.getServiceDuration())
-				.setDeliveryTimeWindow(
-						TimeWindow.newInstance(service.getTimeWindow().getStart(), service.getTimeWindow().getEnd()))
-				.build();
-		for (String skill : service.getRequiredSkills().values()) {
-			FreightUtils.addSkill(carrierShipment, skill);
-		}
-		return carrierShipment;
-	}
 
 	/**
 	 * Creates (MATSim) {@link CarrierShipment} from a (jsprit) {@link Shipment}
 	 *
-	 * @param shipment to be transformed to MATSim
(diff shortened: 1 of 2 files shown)


contribs/ev/src/main/java/org/matsim/contrib/ev/stats/EvMobsimListener.java
```diff
@@ -85,4 +100,32 @@ public void notifyMobsimBeforeCleanup(MobsimBeforeCleanupEvent event) {
...
+					} else {
+						chargerPowerTimeProfileCollectorProvider.setChargerEnergy(charger, 0.0);
+					}
+				} else {
+					chargerPowerTimeProfileCollectorProvider.setChargerEnergy(charger, 0.0);
+				}
+			});
+
+		}
+	}
```
michalmac, line 130: Please move this method to `ChargerPowerTimeProfileCollectorProvider`, so all logic is in one place.

michalmac:
Right now we have 3 commits and 3 changed files. Each commit changes one file. It does not need to be like this. If all file changes are related (e.g. a feature change, or a cleanup/refactoring), they can go in one commit.

contribs/ev/src/main/java/org/matsim/contrib/ev/stats/EvMobsimListener.java
mattiasingelstrom: I would love to, but I cannot move the `MobsimAfterSimStepListener `to `ChargerPowerTimeProfileCollectorProvider` in a way that actually works. By just moving it, `notifyMobsimAfterSimStep` is never called. I guess there is something wrong with the binding of the class. Do you have any idea of what could be the cause?

michalmac: A few things will need to be rearranged here to make it work. I will try to look into it next week and suggest some fixes.

michalmac:
@mattiasingelstrom Last week I merged a few PRs that will simplify collecting stats in EV (#2529, #2530, #2532). I will work on this PR and suggest changes probably this Friday. I hope this is not blocking you too much.

michalmac:
I did a few more PRs to reduce the number of ways to collect time profiles. Even I had trouble to suggest you what is the bast way to proceed. Now there are two approaches:
- listening to events and collecting data (custom class), and then using `ProfileWriter` to create output files
- reading internal state, collecting data and creating output files using `TimeProfileCollector`

Given that you want to listen to events (which is the preferred way for creating output), I suggest you to use the first approach. You may have a look at `VehicleTaskProfileCalculator` and `VehicleTaskProfileView` to see how this is achieved.

This is how things are coupled together in `DrtModeAnalysisModule`:
```java
bindModal(VehicleTaskProfileCalculator.class).toProvider(modalProvider(
		getter -> new VehicleTaskProfileCalculator(getMode(), getter.getModal(FleetSpecification.class), 300,
				getter.get(QSimConfigGroup.class)))).asEagerSingleton();
addEventHandlerBinding().to(modalKey(VehicleTaskProfileCalculator.class));

addControlerListenerBinding().toProvider(modalProvider(getter -> {
	MatsimServices matsimServices = getter.get(MatsimServices.class);
	String mode = drtCfg.getMode();
	var profileView = new VehicleTaskProfileView(getter.getModal(VehicleTaskProfileCalculator.class), taskTypeComparator, taskTypePaints);
	return new ProfileWriter(matsimServices, mode, profileView, "task_time_profiles");
}));
```

PR #2490, opened 2023-04-04
add a break for a switch in emissions contrib

kainagel:


kainagel:
Unfortunately, I do not have time to add a testcase for the `ignore` execution path.

PR #2492, opened 2023-04-05
Change version to 16

Janekdererste:
Prepare the master branch for version 16.0 after the 15.0 release

Issue #2491, opened 2023-04-05
Release 15.0


Janekdererste:
We need to tag a new release. For the core repository this means:

## 0. Pre-release 15.0-RC1
- [x] Create a pre-release `150-RC1` for branch `master`.
   This step is run to make sure the release workflow is up to date: https://github.com/matsim-org/matsim-libs/blob/master/.github/workflows/deploy-on-release-created.yaml. It will be later used for releasing `15.0` (step 3)

## 1. Prepare the `15.x` branch
- [x] Create `15.x` branch that points to the last commit with the version `15.0-SNAPSHOT` on `master`.
- [x] Update github workflows by replacing `master` with `15.x` in the following lines:
   - https://github.com/matsim-org/matsim-libs/blob/master/.github/workflows/verify-push.yaml#L6
   - https://github.com/matsim-org/matsim-libs/blob/master/.github/workflows/deploy-on-pr-merge.yaml#L6
   - https://github.com/matsim-org/matsim-libs/blob/master/.github/workflows/deploy-on-pr-merge.yaml#L9
- [x] Change version numbers in all poms to `15.1-SNAPSHOT`
    ` mvn versions:set -DnewVersion=15.1-SNAPSHOT`
- [x] Push `15.x`
- [x] Check if the branch protection rules apply for `15.x`.
   From this moment on, updates and fixes to `15.x` will go through the same PR review and release process like on `master`.
- [x] Create PR from master to 15.x labelled "LAST CALL FOR 15.0!!!!!" This way people might be aware that the release is coming

## 2. Prepare the master branch for `16.0-SNAPSHOT`
- [x] Check out the latest master branch and change version numbers in all poms to `16.0-SNAPSHOT`
    ` mvn versions:set -DnewVersion=16.0-SNAPSHOT`
- [x] Push it, make a PR, and merge it.
   From this moment on, the master branch is a development branch for the future version `16.0`.

## 3. Release `15.0`
- [x] Create a tag `15.0` that points exactly to the last commit with the version `15.0-SNAPSHOT` on `15.x`.
- [x] Push the tag: `git push origin 15.0`
- [x] Create a release for the `15.0` tag. This will trigger the release workflow that deploys the `15.0` jars to the matsim maven repo and also builds and attaches `matsim-15.0-release.zip` under the `15.0` release assets.

michalmac:
[...]

PR #2493, opened 2023-04-05
build: stop deploying SNAPSHOT versions on merging to master

michalmac:
After discussions with some of you, it seems that stopping deploying SNAPSHOT versions is a good step towards increasing the maintainability and reproducibly.

Now the idea is to merge this PR shortly before we bump to `16.0-SNAPSHOT` on master, so `16.0-SNAPSHOT` will never get deployed.

Ping @Janekdererste (related to #2491 and #2492 )

### DIFF
--- a/.github/workflows/deploy-on-pr-merge.yaml
+++ b/.github/workflows/deploy-on-pr-merge.yaml
@@ -1,9 +1,6 @@
 name: deploy-on-pr-merge
 
 on:
-  push: # snapshot deployment
-    branches:
-      - master
   pull_request_target: # pr-labelled deployment
     branches:
       - master
@@ -12,7 +9,7 @@ on:
 
 jobs:
   deploy-snapshot:
-    name: deploy snapshot/PR-labelled version
+    name: deploy PR-labelled version
     # for PR-labelled deployment -- only if closed by merging
     if: github.event_name == 'push' || github.event.pull_request.merged == true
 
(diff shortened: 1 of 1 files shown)


Janekdererste:
I don't understand why removing the push trigger causes snapshots not to be deployed anymore. Could you explain this @michalmac ?

michalmac:
> I don't understand why removing the push trigger causes snapshots not to be deployed anymore. Could you explain this @michalmac ?

Good question, I guess the setup is a bit too complicated because of the historical reasons.

Initially we deployed a snapshot on merging to master. Then deploying a PR-labelled version was added. However, to deploy  a PR-labelled version we need to know the PR number, which is not available when the workflow is triggered on `push`. Therefore, triggering on `pull_request` was added.

So this way we ended up with two triggers: `push` for snapshots and `pull_request` for PR-labelled versions. This PR removes the first trigger and consequently deployment of snapshots.

michalmac:
BTW. In the meantime, we had to change `pull_request` to `pull_request_target`, so PRs coming from forks could access secrets (necessary to deploy to mvn repo).

Janekdererste:
Thanks @michalmac 

Should I just merge the PR then?

michalmac:
> Thanks @michalmac
> 
> Should I just merge the PR then?

This is the tricky part now. We merge it now and then it goes to 15.x (via #2495), but since we have already deployed 15.1-SNAPSHOT, maybe we want to continue doing that for 15.x? If that's the case, then let's merge this PR after #2495 gets merged. Anyway, that's only a suggestion. I will leave it up to you to decide :-)

michalmac:
> let's merge this PR after #2495 gets merged

... but of course before #2492 gets merged

Janekdererste:
> but since we have already deployed 15.1-SNAPSHOT

Oh, I hadn't thought about that 🙈. Then let's merge it after https://github.com/matsim-org/matsim-libs/pull/2495 and remove the snapshot deployments for 16.x

PR #2497, opened 2023-04-05
New contrib: Informed mode choice

rakow:
This contrib provides mode choice algorithms that can be used to replace the standard random based algorithms in MATSim.

The main idea is that utilities of available modes are estimated in the planning phase and higher ranked alternatives selected with higher probability. Unpromising plan candidates can also be removed altogether, with a flexible pruning functionality. Unlike the default mode choice implementation, it is also possibly to freely configure constraints on mode combinations.

For more information, please refer to the README and paper, which was presented at ABMTrans 2023.

> C. Rakow and K. Nagel (2023) [An annealing based approach to informed mode choice in agent-based transport simulations](https://doi.org/10.1016/j.procs.2023.03.086), *Procedia Computer Science*, **220**, 667-673.

Some additional changes beyond the contrib:
- Added new example scenario "kelheim"
 - Very small sample of a realistic scenario to test a full integration with drt, pt, etc.


michalmac (commented):
It is hard to review this PR because of its huge size (123 commits, 145 files changed).
I suggest splitting it into smaller parts and proceed in this (or similar) order:
1. Introduce changes to the "main" matsim via separate PRs - there are a few changes that should be clearly communicated (e.g. modal injections, changes in scoring, reflective config group).
2. Introduce `informed-mode-choice` as one PR (could include changes to other contribs if they are small an obvious); however, I do not feel competent to review this PR.
3. Changes to the DRT should go first to `drt-extensions` (this could also be a separate PR), so I can review it and we can discuss more in detail some changes (e.g. I would like to discuss the modal binding of `DrtSpeedUpParams`).

PR #2510, opened 2023-04-12
Add new standard output csvs

rakow:
This PR adds two new csv files to standard output.

`links.csv`, which is useful for any link-based analysis and can be used instead of the network.xml

`activities.csv`, to analyze performed activities, similar to trip.csv


jfbischoff:
Nice :)

PR #2514, opened 2023-04-14
Use providers for drt shifts

nkuehnel:


michalmac:
@nkuehnel I am wondering why you need `Provider<DrtShiftsSpecification>` (instead of `DrtShiftsSpecification`)? It looks a bit too complex, but maybe I am missing something.

michalmac:
> @michalmac it's for usage together with the simulated annealing extension, where the shifts may change during iterations. Since these listeners are only bound once and live during multiple QSims, they need access to the respective current version of the specification

I could not find a place where you modify the specifications (I guess it may be somewhere outside matsim-libs).

However, you can have a look at this code, where we modify the specifications between subsequent QSim runs: https://github.com/matsim-org/matsim-libs/blob/20a9fa3169fbfe556768e7d296113a4cafa1b5c0/contribs/dvrp/src/main/java/org/matsim/contrib/dvrp/fleet/VehicleStartLinkToLastLinkUpdater.java#L50
Here `fleetSpecification` is modified, but since it is always the same object (only the contained data changes), we can refer to it directly (no provider needed).

nkuehnel:
@michalmac yes I was aware of that possibility - however, during simulated annealing I need to keep track of old solutions to be able to "reset" in certain cases or output the "best" solution at the end, even if the current accepted solution is different

nkuehnel:
One could of course keep track of old solutions and replace all of the content of the original container object, if you think that is preferable...

michalmac:
Thanks for explaining. Since I do not see that code, it is hard to judge which approach is better. I think the pattern that we modify the container (e.g. fleet, population, chargers) is more popular and maybe even simpler.

If you prefer the current approach, please do not change it. I just wanted to understand why you needed `Provider<DrtShiftsSpecification>`. Now I know it. Thanks!

PR #2523, opened 2023-04-19
Remove non-standard random utils

michalmac:
Remove `RandomUtils` (uses `commons-math3` for generating random sequences) that duplicates the default `MatsimRandom` functionality. Use `MatsimRandom` instead.


michalmac:
@nkuehnel @steffenaxer The removal of `RandomUtils` required minor changes in simulated annealing and drt-extensions (companions). Now you do not need to use two different classes to generate random numbers.

nkuehnel:
nice, thanks for letting us know :)

PR #2539, opened 2023-04-25
minor cleanup: Use Assertion from junit

kt86:
This is only a minor change. It just came up because for me IntelliJ had compile errors here (I do not know why)

@michalmac: I believe, it is the correct way to go into this direction?

### DIFF
--- a/contribs/drt-extensions/src/test/java/org/matsim/contrib/drt/extension/companions/RunDrtWithCompanionExampleIT.java
+++ b/contribs/drt-extensions/src/test/java/org/matsim/contrib/drt/extension/companions/RunDrtWithCompanionExampleIT.java
@@ -20,16 +20,7 @@
 package org.matsim.contrib.drt.extension.companions;
 
 
-import java.io.IOException;
-import java.net.URL;
-import java.nio.file.Files;
-import java.nio.file.Paths;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.stream.Collectors;
-
-import org.assertj.core.api.Assertions;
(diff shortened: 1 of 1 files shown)


kt86:
Alright, that is most probably the reason for it - thanks! Have not done it, because for me, changing it to the (for me known standard) looks like the "cleaner" approach at that moment and it was a very easy "fix".

_Is there a specific reason to use AssertJ (in the future) instead of junit?_ 
If so, I will try to use it when writing (new) tests... otherwise I will stay with junit to avoid the growing zoo of dependencies for similar tasks. (And we should communicate it (better*), so other people will also switch to it.)

@michalmac Please feel free to either close this PR (we will stay with/switch to AssertJ) or merge it.

*maybe I only missed it :(

michalmac:
We use AssertJ in dvrp, ev and all related contribs (and maybe a few unrelated). It is very handy in many situations, especially when working with collections or exceptions. I think it is OK to have it also in drt-extensions.

michalmac:
> _Is there a specific reason to use AssertJ (in the future) instead of junit?_
> If so, I will try to use it when writing (new) tests... otherwise I will stay with junit to avoid the growing zoo of dependencies for similar tasks. (And we should communicate it (better*), so other people will also switch to it.)

No, I do not think we have a policy wrt which library we use. I would say it's good to be consistent in using one or the other (per contrib or package).

kt86:
Great, thanks a lot! I never realized, that it is already used in parallel in so many places. -.- 
So, I am very sorry for the confusion and the extra work for you, to update me!

PR #2555, opened 2023-04-30
dvrp: simplify TimeDiscretizer

michalmac:


michalmac:
Yes, that was intended (see the unit test), but maybe not communicated well. I wanted to simplify the class, because the functionality is very simple, whereas this class had grown to something big given its purpose (because all the configuration options). At the same time, it wasn't used in many places.

Now I think we have three options:
- rounding down after calling `getIdx()`
- adding a second method that will replicate the old `OPEN_ENDED` behaviour (probably then we need to have more descriptive names)
- a boolean parameter to `getIdx()`

I would like to avoid adding any switches inside the discretizer, so that we know the actual behaviour just by looking at the call site (no need to backtrack to the place the discretizer is created).

PR #2570, opened 2023-05-08
Freight: Store the score from jsprit in an Attribute of the CarrierPlan

kt86:
The score of the plan is now free for the "MATSim" score, which is subject to change over the iterations and is the basis for (optional) replanning strategies ⇾ a step more towards standard MATSim behavior.

For not losing the information of the score from the jsprit run, this is now stored in an attribute and accessible via additional getter/setter methods. The (MATSim) score remains empty (== null?) as long MATSim is not run and thus, no scoring from MATSim happened.

Using the `Attributes `logic, the Reader/writer should work well with it, but I will check it later. 

Things to do before merging:

- [x] Check the CarrierPlanWriter: Dealing with "null" score and the attribute.
- [x] Check the CarrierPlanReader: Dealing with "null" score and the attribute.
- [x] Adapt the Reader/WriterTests, so this new attribute is also covered.
- [x] Adapt the Reader/WriterTests, so the "null" score is also covered.


kt86:
"null" score means in practice that it is not set (e.g. by FreightUtils.runJsprit())  --> It is not written out anyway: 
grafik

PR #2579, opened 2023-05-12
core: remove flaky MemoryObserverTest

michalmac:
It has recently failed a couple of times in GitHub Actions (linux):
```
Error: MemoryObserverTest.testStartStop:59 [Number of log messages]
Expecting actual: 6 to be between: [2, 4]
```
In the past it also failed on AppVeyor (windows) and had to be adapted.

It seems that any non-negative number is theoretically possible (in all assertions).


michalmac:
Recent failures:
- https://github.com/matsim-org/matsim-libs/actions/runs/4949337000/jobs/8851347314#step:6:977
- https://github.com/matsim-org/matsim-libs/actions/runs/4954277462/jobs/8862632486#step:4:983

PR #2582, opened 2023-05-13
build: stop building on AppVeyor

michalmac:
After merging #2581, now we have both windows and linux covered by full integration tests that are run each night.

This PR turns down the windows builds on AppVeyor (base unit tests run after each merge to master, no integration tests).


michalmac:
I disconnected AppVeyor from matsim-libs, but without removing the project from AppVeyor. This means that the build history is still available: https://ci.appveyor.com/project/michalmac/matsim-libs/history.

PR #2583, opened 2023-05-15
changed getter access to map

alex94263:
Changed getter Methods for Config, such that the actual representation of the map is returned and not the private Field, i.E. if you remove a CoreModule it is actually removed.


michalmac:
@alex94263 Shouldn't we also remove these fields?

https://github.com/matsim-org/matsim-libs/blob/master/matsim/src/main/java/org/matsim/core/config/Config.java#L89-L115

PR #2601, opened 2023-05-28
Bugfix: Ensure switchOffFISSLastIteration is used

steffenaxer:
- Ensure that config switch for FISS is used. 
- Add a test, that counts number of linkLeaveEvents with disabled FISS.


contribs/drt-extensions/src/test/java/org/matsim/contrib/drt/extension/fiss/RunFissDrtScenarioIT.java
```diff
@@ -146,18 +149,43 @@ public void test() {
...
+				@Override
+				public void install() {
+					addEventHandlerBinding().toInstance(linkCounter);
+				}
+			});
+
 		}
 
 		run.run();
+		assert(linkCounter.getLinkLeaveCount()==23961);
```
mrieser, line 175: `assert` is a java-keyword, which musst explicitly be enabled to have any effect (`--enable-assertions`, `-ea`). Maven enables this when running tests, so it should be fine. But I would still propose using JUnit's assert-methods instead (`Assert.assertEquals(23961, linkCounter.getLinkLeaveCount())`) as this will provide the better error message when it fails.

steffenaxer: Thanks, done!

PR #2612, opened 2023-06-02
ev: individual time profiles of power supplied from charging stations

mattiasingelstrom:
This added stat will be valuable to analyse the power demand of charging stations as well as its utility rate. Only chargers with charging activity are presented in the output file. The output file has the following appearance with the power being in kW:

```
time	        charger3
09:11:50	0.0
09:11:55	0.0
09:12:00	0.0
09:12:05	64.92843121500694
09:12:10	64.92843121500694
...	        ...
09:19:10	64.92843121500694
09:19:15	64.92843121500694
09:19:20	64.92843121500694
09:19:25	129.85686243001388
09:19:30	129.85686243001388
09:19:35	129.85686243001388
```

The average power during a charging activity is calculated and added to a charging station's total power transfer for the activity's duration. Draft PR #2488 aimed to calculate the actual power curve from the increase in energy of an electric vehicle during its charging activity, which would be beneficial for complex charging curves. This proved difficult since the calculator does not have access to the EV fleet objects and could only rely on events, and not the MobsimAfterSimStepListener.


tschlenther (approved):

tschlenther:
Hello @mattiasingelstrom there seem to be problems with your handler in tests in taxi and drt-extensions etc.
Can you have a look?

PR #2640, opened 2023-06-16
Add DrtCompanionUtils.isDrtCompanion() method

marecabo:
This adds the small convenience method `DrtCompanionUtils.isDrtCompanion(Id<Person> personId)`.

### DIFF
--- a/contribs/drt-extensions/src/main/java/org/matsim/contrib/drt/extension/companions/DrtCompanionUtils.java
+++ b/contribs/drt-extensions/src/main/java/org/matsim/contrib/drt/extension/companions/DrtCompanionUtils.java
@@ -21,6 +21,7 @@ package org.matsim.contrib.drt.extension.companions;
 
 import java.util.List;
 
+import org.matsim.api.core.v01.Id;
 import org.matsim.api.core.v01.population.Person;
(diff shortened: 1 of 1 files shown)


kainagel:
I would recommend to return an enum, not a boolean.  Reason: Quite often, at some later stage, one desires additional differentiation, such as "childCompanion".  Then one adds another boolean, and one ends up with hierarchies of booleans, out which many combinations have no meaning.  An enum avoids that problem.

I would also strongly advise against using IDs implicitly for meaning.  Persons can have attributes, and those are explicitly meant for such functionality.  Is maybe 10% more work, but much more longterm stable.

steffenaxer:
I will change this. Thanks for this tipp!

kainagel:
thanks ...

marecabo:
Hi @kainagel,
thanks for your comments, I agree that identification via prefixes is not robust and introduced a `companionType` attribute in #2641.
As I do not know, what kind of companion types users come up with, I chose to use a String attribute and not an enum. This seems to be the more in line how I see attributes used in MATSim.

kainagel:
Indeed.  For attributes, "String" often makes sense.  And the `isCompanion(...)` method then essentially is a helper method for `( getCompanion(...)==null )`.

PR #2644, opened 2023-06-19
feat: change access/egress default

kainagel:
(Actually, this is a "change of a default setting".  Is a source change, is a test change, it is NOT a bug fix, but it also does not add functionality.)

If I see this correctly, using the access egress router _also_ means that a vehicle type is required for routing. This is not structurally necessary; it just was a design decision to require the vehicle for routing, but this was never enforced for the "old" router.  This makes debugging the test failures a bit confusing since quite often we have a "missing vehicle" problem.


paulheinr:
#3655 changes the default. Hopefully, it can get merged soon.

kainagel:
One would have thought that this fails because many mobsim regression tests generate different results ... but at this stage, it often (mostly?) fails because some guice bindings are now no longer fulfilled and/or some config options have become inconsistent.

paulheinr:
Done in #3655, this is why i close this PR.

kainagel:
Thanks.  Also for noting that this was done somewhere else.

PR #2665, opened 2023-06-29
MultiThreaded Population Writer

steffenaxer:
This PR provides a multithreaded population writer. This PR has been motivated by the long waiting time for population dumping, when working with large population data sets. I was wondering, why we are not creating the person strings, that contains the xml on different threads. So I tried to design an implementation, that fits with the current PersonWriter but allows parallel string creation and writing. In detail we have:
- 2 threads to convert persons to strings (more threads doesn't make things faster)
- 1 thread for writing
- OutputQueue (for writing) is limited to reduce the memory footprint

The dump duration of a population could be reduced by approx. 50-60 %


steffenaxer:
I don't know why RunRoadPricingExampleIT fails. When I run all tests this contrib it fail, if I run this test in a single call, it passes?!

mrieser:
nice work. For large populations, this might reduce the time to write populations by several minutes!

mrieser (approved):
the many commits should be squashed when merging.
Thanks for the effort!

PR #2669, opened 2023-07-02
Add VehicleUtils.hasVehicleId() method

marecabo:
This change adds the method `VehicleUtils.hasVehicleId(Person person, String mode)` that checks whether a person has a designated vehicle for a specific mode assigned - without throwing an exception.

Motivation for this method is, that
* the existing methods `getVehicleIds` and `getVehicleId` always throw an exception, if the `"vehicles"` attribute is not filled with a `PersonVehicles` object, and
* the attribute name `VehicleUtils.VEHICLE_ATTRIBUTE_KEY = "vehicles"` is private, so for checking it manually, a hard-coded copy of the `"vehicles"` string is required, which is ugly.

### DIFF
--- a/matsim/src/main/java/org/matsim/vehicles/VehicleUtils.java
+++ b/matsim/src/main/java/org/matsim/vehicles/VehicleUtils.java
@@ -115,6 +115,21 @@ public final class VehicleUtils {
 	}
 
 	/**
+	 * Checks whether a person has a vehicle id for mode - without throwing an
+	 * exception if not.
+	 * 
+	 * @param person the person one wants to check for a vehicle id
+	 * @param mode   the mode for the vehicle id to check
+	 * @return whether person has a vehicle id for that mode
+	 * 
+	 * @see {@link VehicleUtils#getVehicleId(Person, String)}
+	 */
+	public static boolean hasVehicleId(Person person, String mode) {
+		PersonVehicles personVehicles = (PersonVehicles) person.getAttributes().getAttribute(VehicleUtils.VEHICLE_ATTRIBUTE_KEY);
+		return personVehicles != null && personVehicles.getVehicle(mode) != null;
+	}
+
+	/**
 	 * Retrieves vehicleIds of all vehicles that are assigned to the person.
 	 *
 	 * @param person the person one wants to retrieve vehicles for


kainagel:
ok with me.  There are, evidently, different standards of what could happen if something is not there:
1. return "null"
2. return an object of type `Optional`
3. throw an exception
We have proponents of all three solutions.  Not sure if we could have a developer meeting and agree on one standard.  As long as we don't have a standard here, I find your solution ok.  Certainly better than making the static string public ... that is deliberately private.  An alternative would be to make `getVehicleId(s)` return null rather than throwing an exception.  Would be ok with me ... but maybe not others.

(java collections differentiates between `peek` and `poll`.  Maybe we should consider a similar thing so that users could choose which one they want in a given situation.)

marecabo:
@kainagel Thanks for your comment. I agree, that making such private member public should not be an option as everyone will write own helpers that might do things slightly different.

Then, to throw my two cents into the discussion: ;)
- One consideration would be, that this helper methods are as efficient as possible regarding computational and memory allocation demands. This is especially relevant for large populations when these methods might be called, say, ten million times. 
- If returning an exception, it should be a custom exception class, and not some generic RunTimeException, where it is required to compare parts of the exception message (with a hard-coded substring of it) to identify what happend - which introduces another antipattern.

kainagel:
> @kainagel Thanks for your comment. I agree, that making such private member public should not be an option as everyone will write own helpers that might do things slightly different.
> 
> Then, to throw my two cents into the discussion: ;)
> 
> * One consideration would be, that this helper methods are as efficient as possible regarding computational and memory allocation demands. This is especially relevant for large populations when these methods might be called, say, ten million times.
> * If returning an exception, it should be a custom exception class, and not some generic RunTimeException, where it is required to compare parts of the exception message (with a hard-coded substring of it) to identify what happend - which introduces another antipattern.

I have started a new issue #2689, and put a start for a discussion into a google docs https://docs.google.com/document/d/1CSvPcMKrPy-0j_RBvMEyCyqqDq_JH1snFkGUlTrLIKM/edit?usp=sharing .  I find google docs better for this type of discussing since there is a chance ( :-) ) to have a consolidated document at the end.

PR #2677, opened 2023-07-05
Allow RoutingMode as MainModeIdentifier in trips csv 

rakow:
There was a clause that disabled RoutingModeMainModeIdentifier in the trip csv, because the routing mode was not available at that time. Since the routing mode is now also part of the experienced leg, it can be used without any errors.

### DIFF
--- a/matsim/src/main/java/org/matsim/analysis/TripsAndLegsCSVWriter.java
+++ b/matsim/src/main/java/org/matsim/analysis/TripsAndLegsCSVWriter.java
@@ -117,15 +117,6 @@ public class TripsAndLegsCSVWriter {
         Tuple<Iterable<?>, Iterable<?>> record = new Tuple<>(tripRecords, legRecords);
         List<TripStructureUtils.Trip> trips = TripStructureUtils.getTrips(experiencedPlan);
 
-        /*
-         * The (unlucky) default RoutingModeMainModeIdentifier needs routing modes set in the legs. Unfortunately the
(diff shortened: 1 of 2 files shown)


vsp-gleich:
Hi @rakow, there is a significant difference between routing mode and main mode. If we put routing mode into a field called main mode there are cases in which this skews up analysis. E.g. routing mode was "pt", no suitable pt route was found and therefore the actual trip was of mode "walk". Then for mode choice calibration we need this to be recognized as "walk" and not "pt", because it is de facto "walk" (real world reference data to calibrate with gives only the really used mode "walk", not the attempted mode "pt"). 
So it is better to leave the field main mode empty than to insert misleading information.
If you need routing mode it should be a separate field.

vsp-gleich:
If you need the field main mode to never be empty, then we should rather enforce having a main mode identifier or set TransportPlanningMainModeIdentifier as default. However, this will cause all simulations to crash if new modes are used which are not contained in the default TransportPlanningMainModeIdentifier and if no suitable AnalysisMainModeIdentifier is bound (overwriting the default implementation). This could be another barrier for beginners, but would certainly enforce better results.

rakow:
Ok, I see.  I don't need the main_mode main mode to never be empty. Probably the RoutingModeMainModeIdentifier should be removed as default, or do you know if this will cause problems elsewhere?

vsp-gleich:
Unfortunately, e.g. modestats does not work if we have no main mode identifier bound. I got some questions on that topic since I last changed it, so maybe it's best to remove RoutingModeMainModeIdentifier as default.

PR #2680, opened 2023-07-06
add new default analysis main mode identifier

vsp-gleich:
add new default analysis main mode identifier to stop misleading usage of routing mode main mode identifier which is not suitable for analysis purposes. Runs with modes not included in DefaultAnalysisMainModeIdentifier will need to bind their own AnalysisMainModeIdentifier.


vsp-gleich:
Fixes misleading usage of routing mode, e.g. in #2677

PR #2686, opened 2023-07-13
Use Woodstox XML Parser for DTD validated XMLs

steffenaxer:
This PR replaces the regular XML Parser with a much faster version from woodstox (https://github.com/FasterXML/woodstox).
In #2685 I already posted a picture of the performance improvements. The performance gain by just switching the parser is about 30 % speed up. For XSD validated XMLs regular XML Parser is used as a fallback.


rakow:
Very cool, I can't wait to try it!  
I have one question, though: Did you perform your benchmarks with gzipped populations ? 
LZ4 should be around 10x faster than gzip, but it barely made a difference because the XML processing was the bottleneck. With your recent improvements, it might be possible now to gain more speed (or utilize more threads) with a faster compression algorithm.

steffenaxer:
Fun fact, all the activitites about faster XML processing have been motivated by the fact that zstd-Compression http://facebook.github.io/zstd/ is approx. 10x faster in decompression thatn gz. So I wanted to get this 10x faster. While testing I found out, that XML reading and object generation was the real bottleneck. So I started to make readers and writers of the population to make use of multiple threads. So to come back to your question, I tested only with gz compression.

PR #2709, opened 2023-07-26
fix: warn for parallell links in OneToManyPathCalculator

sebhoerl:
Currently, there is a comment in `OneToManyPathCalculator` that already explains that having parallel links (same start and end node) may produce unexpected results when reconstructing the link sequence (see `constructLinkSequence`). 

Yesterday, I was debugging timing issues in DRT where I had a scenario in which I never managed to get the timing 100% congruent between insertion/scheduling and the simulation. After digging for several hours, I arrived at the comment in `OneToManyPathCalculator` so I thought it might be useful to have a check and a warning about the issue :)

### DIFF
--- a/contribs/dvrp/src/main/java/org/matsim/contrib/dvrp/path/OneToManyPathCalculator.java
+++ b/contribs/dvrp/src/main/java/org/matsim/contrib/dvrp/path/OneToManyPathCalculator.java
@@ -24,22 +24,26 @@ import static java.util.stream.Collectors.toList;
 import static org.matsim.contrib.dvrp.path.LeastCostPathTreeStopCriteria.allEndNodesReached;
 import static org.matsim.contrib.dvrp.path.LeastCostPathTreeStopCriteria.withMaxTravelTime;
 import static org.matsim.contrib.dvrp.path.VrpPaths.FIRST_LINK_TT;
-import static org.matsim.core.router.util.LeastCostPathCalculator.Path;
 
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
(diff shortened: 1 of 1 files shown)


sebhoerl:
Actually, it might make sense to move this further upstream, what do you think @michalmac? The closest `public` access point to this seems to be `OneToManyPathSearch.createSearch`. To avoid multiple verifications in DRT, it would probably make sense to have the check in `MultiInsertionDetourPathCalculator`, but then individual users of the class won't see the warning.

sebhoerl:
It might make sense to add `new NetworkSegmentDoubleLinks().run(subnetwork);` to `DvrpModeRoutingNetworkModule` if we have `useModeFilteredSubnetwork=true`

sebhoerl:
Plus, I just noticed it is a tricky thing to solve automatically: We cannot just add `NetworkSegmentDoubleLinks.run` in `DvrpGlobalRoutingNetworkProvider` for instance, because then the network gets modified and we will have links in the routing network that don't exist in the QSim network, so agents will be stuck. In a standard scenario, this really needs to be cleaned downstream before loading the network.

michalmac (approved):
Great to have this check!

michalmac:
> Actually, it might make sense to move this further upstream, what do you think @michalmac? The closest `public` access point to this seems to be `OneToManyPathSearch.createSearch`. However, to avoid multiple verifications in DRT, it would probably make sense to have the check in `MultiInsertionDetourPathCalculator`, but then individual users of the class won't see the warning.

IMO it's a good place, because the check is close to the actual functionality that may be affected by this issue.

PR #2710, opened 2023-07-26
fix: repair treatment of extended stop durations in drt task scheduling

sebhoerl:
Currently, there is an issue when using varying stop durations (`StopDurationEstimator` + `IncrementalStopDurationEstimator`) in the scheduling of the DRT tasks. Everything works well in the case when new tasks are created, but for stops to which pickups or dropoffs are added, only their own duration is updated and the scheduling finishes. This is fine if the `stopDuration` stays constant (e.g. 60s), but if the stop duration increases, this has an impact on the following tasks. So whenever the stop duration is updated during scheduling, we should use `ScheduleTimingUpdater` to repair the following start and end times. Otherwise, we have an inconsistent schedule. It may be repaired eventually somewhere, but if we have subsequent insertions in one iteration, this may lead to confusion. 

- Added an integration test with varying stop duration
- Added a verification in `DefaultRequestInsertionScheduler` that detects inconsistent schedules and fails without further intervention
- Added fixes in `DefaultRequestInsertionScheduler` that repair the schedule timing


michalmac (approved):
Nice catch. Great to have `verifyTaskContinuity()` and to have an IT with incremental stop duration.

PR #2716, opened 2023-08-01
Repeated selective insertion search

steffenaxer:
This PR provides a novel insertion search implementation that is a blend between extensive and selective search. In a nutshell:

-	Introduction of a time dependent travel time cache (see AdaptiveTravelTimeMatrix) to get most accurate detour time estimates. First, I played around and developed a straightforward time dependent TravelTimeMatrix, but the initialization and updating is painful slow when having 30-time bins. Second, I developed an AdaptiveTravelTimeMatrix that gets initialized from the existing FreeSpeedMatrix but it is updated lazily within the insertion search whenever a regular calculatePaths method is called. As we have multiple millions routing calls over the iterations one can produce passively a travel time matrix that is quite accurate without addition runtime burden. 
-	In addition, I implemented a repeated selective search approach, in which insertions are firstly estimated (speedfactor is 1.0) via the AdaptiveTravelTimeMatrix. Second, insertions are sorted according to their costs. Third, the n-cheapest insertion (config switch, retryInsertion) candidates are than routed and validated according to the travel time constrains. If a valid candidate is found, the validation stops. In many cases only 1-3 trials of different insertions are needed. I found out that 5 candidates are a good tradeoff between performance and rejection rate. 

With this novel approach, the rejection rate is up to 2 % worse than the extensive insertion but likewise 4x faster. Approx. 73 % of the requests can be inserted with the first trial. Which reveals the main reason, why selective search insertion is much faster than extensive search. However, to reach a nearly similar rejection rate, additional trials are needed.
<p align="center">
grafik
</p>

### DIFF
--- a/contribs/drt/src/main/java/org/matsim/contrib/drt/optimizer/DrtModeOptimizerQSimModule.java
+++ b/contribs/drt/src/main/java/org/matsim/contrib/drt/optimizer/DrtModeOptimizerQSimModule.java
@@ -31,6 +31,8 @@ import org.matsim.contrib.drt.optimizer.insertion.InsertionCostCalculator;
 import org.matsim.contrib.drt.optimizer.insertion.UnplannedRequestInserter;
 import org.matsim.contrib.drt.optimizer.insertion.extensive.ExtensiveInsertionSearchParams;
 import org.matsim.contrib.drt.optimizer.insertion.extensive.ExtensiveInsertionSearchQSimModule;
+import org.matsim.contrib.drt.optimizer.insertion.repeatedselective.RepeatedSelectiveInsertionSearchParams;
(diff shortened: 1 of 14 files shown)


michalmac (approved):
Nice! This is an interesting approach to overcome the current shortcomings/bugs around the selective search, i.e.:
- selective search may lead to rejections even if rejections are disabled (bug)
- travel time matrices may give significantly incorrect estimates if a zone (grid cell) contains links which are "locally disconnected" (e.g. two sides of a river or a motorway in a rural area)

FYI. We want to address these two issues and hopefully, once they get resolved, we can have only one insertion search as it was the case a few years ago (for the ease of development and maintenance).

steffenaxer:
Thank you @michalmac for the positive feedback!

PR #2728, opened 2023-08-08
Freight: using centralised constants for parser and writer

### LINKED ISSUE #2194: Freight: Use centralized fields for Reader/Writer
Instead of defining the general strings in each parser
https://github.com/matsim-org/matsim-libs/blob/1cc92b34650451300bd3dba63a2ea92dfb11ed49/contribs/freight/src/main/java/org/matsim/contrib/freight/carrier/CarrierPlanXmlParserV2.java#L53-L69

and having it hard-coded in the corresponding Writer, e.g. https://github.com/matsim-org/matsim-libs/blob/1cc92b34650451300bd3dba63a2ea92dfb11ed49/contribs/freight/src/main/java/org/matsim/contrib/freight/carrier/CarrierPlanXmlWriterV2.java#L162-L179

, we could put them to a more centralized location, e.g. https://github.com/matsim-org/matsim-libs/blob/master/contribs/freight/src/main/java/org/matsim/contrib/freight/carrier/FreightConstants.java, where already some of them are located.

(( And I am sure, that there are more location inside the FreightContrib, where these Strings are defined again -.- ))

nixlaos:
resolves #2194


kt86 (approved):
Thanks a lot!

PR #2730, opened 2023-08-09
feat: avoid short wait times in drt

sebhoerl:
This is a follow-up PR for #2723. See that PR for more detailed information and see comparison between the two PRs to see the individual changes here.

**Problem**

- Currently, wait times of certain requests can be smaller than the configured `stopDuration`. This happens when their pickup is merged into an ongoing `DrtStopTask` / `DrtStopActivity`.
- This is especially critial when working with variable stop durations, for instance for wheelchair requests. Merging such a request into an already ongoing one will not extend the duration properly.
- *Additional problem*: When fixing the bullets above it might happen that an already existing request will be picked up later than allowed due to shifting the end time of the task beyond its latest pickup constraint. Such an insertion should not be valid anymore.

**Changes**

- Introduce *start slack* which indicates by how much time the (ongoing) start task in an insertion schedule can be extended without violating request departure time constraints. This requires passing down *now* to the insertion. This change represents the majority of changed files.
- Integrate an additional check for the *start slack* in `InsertionGenerator`.
- Make `DrtStopActivity` depend dynamically on the end time of the generating `DrtStopTask` to react on changing task timing.

**Further discussion**

- The current configuration of the `StopTimeCalculator` reproduces the current behavior: Stop tasks will never chance duration, also when new requests are added to a task. All integration tests, hence, pass, without any changes. The test `RunDrtExampleIT::testRunDrtStopbasedExampleWithFlexibleStopDuration` shows the new set-up with reactive stop tasks that end once the full `stopDuration` for each pickup has elapsed. I would propose to change the default behavior to this set-up while keeping the relevant classes available to reproduce existing behavior:
  - *Pro*: Avoids zero-duration or very short vehicle interactions and wait times. Removes incoherent behavior when using variable stop durations by default.
  - *Contra*: Changes current default behavior. Integration tests need to be updated (changes are in the magnitude of ~3s additional wait and travel time, rejection rate stays the same)
- With these changes (but already since introducing variable stop durations), the configuration option `stopDuration` is not exact anymore. I propose to rename it to `minimumStopDuration` as a stop will always have this duration (from the vehicle perspective) but may be longer, dependent on the requests that are served.


michalmac:
@sebhoerl I will review this PR after #2730 is merged and this one gets rebased on master.

sebhoerl:
@michalmac Here is the consolidated version after merging https://github.com/matsim-org/matsim-libs/pull/2730 :)

contribs/drt/src/main/java/org/matsim/contrib/drt/optimizer/VehicleEntry.java
```diff
@@ -64,6 +64,10 @@ public boolean isAfterLastStop(int index) {
 	}
 
 	public double getSlackTime(int index) {
-		return slackTimes[index];
+		return slackTimes[index + 1];
+	}
+	
+	public double getStartSlack() {
```
michalmac, line 70: `getStartSlackTime()`?

contribs/drt/src/main/java/org/matsim/contrib/drt/optimizer/Waypoint.java
```diff
@@ -52,12 +52,14 @@ class Start implements Waypoint {
 		public final Link link;
 		public final double time;
 		public final int occupancy;
+		public final double now;
```
michalmac, line 55: Maybe `createTime` is less confusing than `now`? Maybe we could have `createTime` even at the `VehicleEntry` level?

Why not pass the current time to `InsertionDetourTimeCalculator`?

michalmac (approved):
Nice PR. My main questions were around introducing `now` to `Start` - I just wanted to know your thoughts. Otherwise, it's ready to be merged.

contribs/drt/src/main/java/org/matsim/contrib/drt/optimizer/VehicleEntry.java
sebhoerl: Ok, will update

contribs/drt/src/main/java/org/matsim/contrib/drt/optimizer/Waypoint.java
sebhoerl: Yes, `createTime` sounds good to me, I didn't put `now` to `VehicleEntry` because there it was even less clear what it means. But putting `createTime` into `VehicleEntry` is perfect, then it means "when the entry was created".

My initial attempt was to pass the time over to `InsertionDetourTimeCalculator`, but this created a whole chain of passing it through the whole infrastructure, over the Extensive/Selective classes etc., which exploded the number of changed files + the unit tests needed to be adapted. So adding it to VehicleEntry/Start looks clearner to me.

sebhoerl:
Thanks for the review, what is your view on the two points in "further discussion"?
- Changing default behaviour 
- Renaming `stopDuration` > `minimumStopDuration`

michalmac:
> Thanks for the review, what is your view on the two points in "further discussion"?
> 
> * Changing default behaviour
> * Renaming `stopDuration` > `minimumStopDuration`

I do not mind changing the default behaviour. However, I am not sure if we should rename the parameter name (due to backward compatibility). Maybe just changing the comment would be enough?

Let's do it in a separate PR, then maybe others will share their opinions.

PR #2735, opened 2023-08-16
core: accept TeleportedModeParams as arg in PlansCalcRouteConfigGroup.addTeleportedModeParams()

michalmac:
After recent refactorings (#2646), ModeRoutingParams is now a deprecated subclass of TeleportedModeParams.


michalmac:
This PR is also needed to properly fix https://github.com/matsim-org/matsim-code-examples/pull/1008.

PR #2750, opened 2023-08-30
Add priority to ControlerListener

rakow:
This PR adds a default method to the controler interface, which allows defining a priority. Listeners with higher priority are now executed first. This gives more control over the execution order of listeners, which is nondeterministic at the moment. 

This PR also ensures that the SimWrapper listener is executed later than others (makes #2749 obsolete)


tschlenther:
great :)
I think it might be a good idea to state each listeners priority in the log info message !? 
`e.g. log.info("calling notifyScoring on " + aListener.getClass().getName() + " with priority " + aListener.priority() )`
It could clear everything up for users and help debugging...

PR #2773, opened 2023-09-19
Add railsim contrib

munterfi:
A large-scale hybrid micro- and mesoscopic simulation approach for railway operation.

The *railsim* contrib provides a custom QSim-Engine to simulate trains:

- Trains are spatially expanded along several links. Additional events indicate when the end of the train leaves a link.
- Trains accelerate and decelerate based on predefined vehicle attributes (along a single link or along several links).
- The infrastructure ahead of each train is blocked (reserved train path) depending on the braking distance which is
  computed based on the vehicle-specific deceleration and the current speed.
- Capacity effects are modeled at the level of resources. A resource consists of one link or several links.
- Trains may deviate from the network route given in the schedule, e.g. to avoid a blocked track (dispatching,
  disposition).

This contrib was initiated and initially funded by the [Swiss Federal Railways](https://www.sbb.ch) (SBB).


steffenaxer:
Thank you for this great and very interesting contribution. This adds a completely new functionality to matsim. Pretty cool!

contribs/pom.xml
```diff
@@ -90,6 +90,7 @@
 		<module>parking</module>
 		<module>protobuf</module>
 		<module>pseudosimulation</module>
+		<module>railsim</module>
```
michalmac, line 93: Please add this new contrib to the list of contribs in https://github.com/matsim-org/matsim-libs/blob/master/.github/workflows/verify-push.yaml

ikaddoura:
thanks everyone!

munterfi:
Great! Big thanks to @rakow for implementing a major part of the current version and resolving the remaining issues!

PR #2808, opened 2023-10-05
Allow idle vehicle to stay in field, till timeout is exceeded

steffenaxer:
This PR introduces a feature to configure a certain timeout interval (returnToDepotTimeout) that needs to be exceeded to send back a vehicle into the nearest depot. The current implementation had a quite strict rule, that enforced idle vehicles to return to depot directly after getting idle. Especially when working with eDrt, this strict rule caused a lot of relocations (into the depot and maybe back again) that could be mitigated by allowing a vehicle to stay at its last position at least for a certain timeout interval. 

To keep nearly the very same behavior:
-	vehicles get checked every 60 seconds. 
-	default returnToDepotTimeout is set to 60 second.


nkuehnel:
Nice!

contribs/drt/src/main/java/org/matsim/contrib/drt/scheduler/EmptyVehicleRelocator.java
```diff
@@ -41,6 +41,7 @@
  */
 public class EmptyVehicleRelocator {
 	public static final DrtTaskType RELOCATE_VEHICLE_TASK_TYPE = new DrtTaskType("RELOCATE", DRIVE);
+	public static final DrtTaskType RELOCATE_VEHICLE_TO_DEPOT_TASK_TYPE = new DrtTaskType("RELOCATE_DEPOT", DRIVE);
```
michalmac, line 44: Maybe "RELOCATE_DEPOT" -> "RELOCATE_TO_DEPOT"?

contribs/drt/src/main/java/org/matsim/contrib/drt/optimizer/DefaultDrtOptimizer.java
```diff
@@ -95,6 +97,7 @@ public void notifyMobsimBeforeSimStep(@SuppressWarnings("rawtypes") MobsimBefore
 			requestInserter.scheduleUnplannedRequests(unplannedRequests.getSchedulableRequests());
 		}
 
+		returnToDepot(drtCfg.returnToDepotEvaluationInterval, drtCfg.returnToDepotTimeout);
```
michalmac, line 100: Maybe change `returnToDepot()` to `relocateVehiclesToDepot()?  Or `planReturnsToDepot()` so it's clearer what the method does?

michalmac (approved):
Nice feature!

contribs/drt/src/main/java/org/matsim/contrib/drt/scheduler/EmptyVehicleRelocator.java
steffenaxer: Done

contribs/drt/src/main/java/org/matsim/contrib/drt/optimizer/DefaultDrtOptimizer.java
steffenaxer: Done

steffenaxer:
Thanks for the review!

PR #2823, opened 2023-10-09
fix: missing mode params

tkchouaki:
The org.matsim.benchmark.Benchmark class runs the config file located in examples.scenarios.benchmark which uses the population file of the berlin scenario. However the config file for benchmark is missing some mode parameters that are necessary. They are added in this PR

### DIFF
--- a/examples/scenarios/benchmark/config.xml
+++ b/examples/scenarios/benchmark/config.xml
@@ -27,9 +27,9 @@
 	<module name="qsim">
 		<param name="startTime" value="00:00:00" />
 		<param name="endTime" value="00:00:00" />
-	
+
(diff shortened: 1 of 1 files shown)


tschlenther:
I think that if when you have resolved the conflicts we can merge this PR here ...  :)

mrieser:
I think the Benchmark-class in the repository is from the old benchmark. The new benchmark essentially uses the Run-class from the OpenBerlin-scenario, which creates these missing parameters, see https://github.com/matsim-scenarios/matsim-berlin/blob/31af2a18d0c7a1eaf9e037a7ad7512c020999573/src/main/java/org/matsim/run/RunOpenBerlinScenario.java#L40

Issue #2768, opened 2023-09-15
Plan Inheritance


neuma:
## Idea
At its core, MATSim is a co-evolutionary algorithm that repeatedly performs the steps mutation, scoring, and selection. However, what happens during the so-called relaxation process is mostly invisible to the user.

Some questions this proposal tries to answer
- Which strategies prevail in the end?
- Which strategies fail immediately?
- Which strategies are the most robust and produce the most descendants?
- What does the lineage of an agent's choice-set look like?
- How many iterations do we actually need until the choice-set of most agents only produce inferior short-lived descendants?

At the end of a MATSim run, the following record is available for analysis for each plan ever created.
- agentId: Id of the agent the plan belongs to
- planId: Globally unique plan id
- ancestorId: NONE for plans read initially from file or the planId of the plan from which this plan had been mutated
- mutatedBy: INITIAL for plans read initially from file or the name of the PlanStrategy that altered and thus created this plan
- iterationCreated: Iteration in which the plan had been added to the choice-set of the agent
- iterationRemoved: Iteration in which the plan had been removed from the choice-set of the agent
- iterationsSelected: List of iterations in which this plan had been the selected plan.

Further data like executed scores could be added in a later implementation.

## Current situation
In MATSim, each agent has a choice-set of n plans. The choice-set is altered by PlanStrategies in two steps
1. The PlanSelector of a PlanStrategy chooses one plan of the choice-set and marks it as the new selected plan.
2. [Optional] The PlanStrategy may apply PlanStrategyModules to a deep-copy of the then selected plan. Each PlanStrategyModule alters/mutates the plan, e.g. the transport mode of a leg. Eventually, the mutated plan becomes the new selected plan for the current iteration.

This loop is performed at the start of each iteration during the Replanning phase. Each mutation of a plan is a candidate for further mutations in upcoming iterations.

The selected plan of the final iteration is considered the outcome of the MATSim relaxation process.

## Implementation Proposal
Once the GenericPlanStrategyImpl mutates a plan during the optional step 2, the then selected plan adds/overrides two new attributes
- planMutator: The name of the PlanStrategy - most likely the result of GenericPlanStrategyImpl.toString()
- iterationCreated: The current iteration as derived from the GenericPlanStrategyImpl.replanningContext

The actual book-keeping is done by a second component - the PlanInheritanceModule. Implemented as BeforeMobsimListener and ShutdownListener it performs the following actions

[...]

PR #2831, opened 2023-10-10
Adding plan inheritance

alex94263:
Pull Request for Proposal/Issue as written down in https://github.com/matsim-org/matsim-libs/issues/2768.  
Essentially it allows to analyze the creation, lifespan and origin of plans in a matsim-run.  
Additional Config Group has been added in order to en-/disable planInheritance specific attributes and files.


matsim/src/main/java/org/matsim/api/core/v01/population/Plan.java
```diff
@@ -49,6 +49,18 @@ public interface Plan extends MatsimPopulationObject, Customizable, BasicPlan, A
 	public abstract String getType();
 
 	public abstract void setType(final String type);
+	
+	public abstract void setPlanId(final String planId);
+	
+	public abstract String getPlanId();
```
marecabo, line 55: Thanks for this great work! Maybe it would make sense to let `Plan` implement the [`Identifiable`](https://github.com/matsim-org/matsim-libs/blob/25fefcd4b04ba6909d2db583c4cbe267aa5fbc2f/matsim/src/main/java/org/matsim/api/core/v01/Identifiable.java#L23) interface and use the regular MATSim Id class for ids. This would allow using the specialized container classes `IdMap`, which are very efficient.

alex94263: This should be implemented now.

marecabo: Looks great!

PR #2834, opened 2023-10-11
Draft: max detour constraint drt

nkuehnel:


contribs/drt/src/main/java/org/matsim/contrib/drt/optimizer/insertion/MaxDetourInsertionCostCalculator.java
```diff
@@ -0,0 +1,75 @@
...
+
+	private boolean violatesDetour(InsertionGenerator.Insertion insertion, VehicleEntry vEntry, DrtRequest drtRequest) {
+		// Check if the max travel time constraint for the newly inserted request is violated
+		double rideDuration = insertion.dropoff.newWaypoint.getArrivalTime() - insertion.pickup.newWaypoint.getDepartureTime();
+		if(drtRequest.getMaxRideDuration() < rideDuration) {
+			return true;
+		}
+
+		final int pickupIdx = insertion.pickup.index;
+		Map<Id<Request>, Double> pickUps = new HashMap<>(ongoingRequests);
```
michalmac, line 43: What is the point of copying this map?

contribs/drt/src/main/java/org/matsim/contrib/drt/optimizer/insertion/MaxDetourInsertionCostCalculator.java
```diff
@@ -0,0 +1,75 @@
...
+import org.matsim.contrib.dvrp.passenger.PassengerDroppedOffEventHandler;
+import org.matsim.contrib.dvrp.passenger.PassengerPickedUpEvent;
+import org.matsim.contrib.dvrp.passenger.PassengerPickedUpEventHandler;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class MaxDetourInsertionCostCalculator implements InsertionCostCalculator, PassengerPickedUpEventHandler, PassengerDroppedOffEventHandler {
+
+	private final Map<Id<Request>, Double> ongoingRequests = new HashMap<>();
```
michalmac, line 19: This map can be read and written at the same time (events handling is executed in parallel to optimisation)

contribs/drt/src/main/java/org/matsim/contrib/drt/optimizer/insertion/MaxDetourInsertionCostCalculator.java
michalmac: I think I understand the idea. However, I think copying the map and then adding even more stuff to it is a bit too heavy.

luchengqi7:
Hi Nico, it seems that our hack does not work with the current insertion logic. The arrival time of the WayPoint is not implemented. 
https://github.com/matsim-org/matsim-libs/blob/5695f6309935b71d5ccb59fcbf8b13a23dbc0811/contribs/drt/src/main/java/org/matsim/contrib/drt/optimizer/Waypoint.java#L260

I have discussed with Michal (@michalmac), and now we think it may be better to have some more fundamental implementation, rather than changing the current basic structure of the insertion. 

We can have a discussion after you return from your vacation.

luchengqi7:
Hi Nico,

How are you? Hope everything is running smoothly :) 

May I know if there is any update from your side on this topic? Shall we go ahead to change the structure of the insertion logic, or perhaps we can think of an alternative quick fix? 

Best regards,
Chengqi Lu

sebhoerl:
Just as comment, it might also be worth it thinking about adding this as a constraint here :)
https://github.com/matsim-org/matsim-libs/pull/2947

michalmac:
> > Just as comment, it might also be worth it thinking about adding this as a constraint here :) #2947
> 
> That's a great point! Thanks a lot, Sebastian!

My suggestion would be to use the max detour time to pre-compute slack times before generating insertions. Otherwise, if we only post-process insertions, we may experience an "unexplainable" increase in computation time when changing from max travel time to max detour time.

nkuehnel:
Hey @luchengqi7 ! sorry for my late reply, it was a bit busy.

Maybe we can have a chat next week? I'll write a mail

michalmac:
I just pushed a commit that reverts the changes in `AbstractRoute` (this was not intended) and instead introduces them in `DrtRoute`.

@nkuehnel @luchengqi7 I wanted to bring this up for your next week sync: `DrtConfigGroup.maxAbsoluteDetour` is used for the total travel time (wait+ride). I guess we need something else for the max detour wrt ride time? Right now this logic is not correct (see the first argument in both `min()` methods):
```java
	static double getMaxTravelTime(DrtConfigGroup drtCfg, double unsharedRideTime) {
		return Math.min(unsharedRideTime + drtCfg.maxAbsoluteDetour, drtCfg.maxTravelTimeAlpha * unsharedRideTime + drtCfg.maxTravelTimeBeta);
	}

	static double getMaxRideTime(DrtConfigGroup drtCfg, double unsharedRideTime) {
		return Math.min(unsharedRideTime + drtCfg.maxAbsoluteDetour, drtCfg.maxDetourAlpha * unsharedRideTime + drtCfg.maxDetourBeta);
	}
```

luchengqi7:
Hi @nkuehnel, today I have discussed with Michal, and he is happy about this new idea. Since there are some other changes and conflicts in this branch, shall I open a new PR in the current matsim-lib (upstream)?

luchengqi7 (approved):

luchengqi7:
The proposed changes in this PR have been merged to the master branch by PR #2997.

PR #2840, opened 2023-10-11
Counts v2 format

### LINKED ISSUE #2780: More flexible counts format
The current counts format is very limited, as it can only store volumes for one mode at a fixed hour interval.

@hzoerner was working on a new format that allows for different mode-specific observations with flexible interval. `counts-v2` contains the current implementation, there are a few tasks left:

- Consolidate the xml schema and create a xsd file
- Backward compatibility and possible replacement of the old Counts class. Currently, everything is separated.

rakow:
This PR updates the counts format to v2 introducing more flexibility in the type of data that can be stored. In summary, the counts format can now store:

- Any observations additional to counts (volumes)
- Observations differentiated by mode
- Data at any desired (but fixed) time interval

Example file:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<counts xmlns="http://www.matsim.org/files/dtd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.matsim.org/files/dtd http://www.matsim.org/files/dtd/counts_v2.xsd" source="Bundesanstalt für Straßenwesen" name="BASt Counts" description="Aggregated daily traffic volumes for car and freight traffic." year="2021">

	<location refType="link" refId="2423" id="2423" name="Marienfelde_S">

		<measurements type="volumes" networkMode="car" interval="3600">
				<value t="0" val="56.28772173437474" />
				<value t="3600" val="34.67126398386528" />
				<value t="7200" val="37.18496486930159" />
<!-- Other values are left out... -->
		</measurements>
		<measurements type="volumes" networkMode="truck" interval="3600">
				<value t="0" val="11.100505172776847" />
				<value t="3600" val="9.51377028264906" />
				<value t="7200" val="12.541180784315472" />
<!-- Other values are left out... -->
				<value t="82800" val="14.506341987151561" />
		</measurements>
	</location>
```

The new API involves using `createAndAddMeasureLocation`, `MeasurementLocation` and `Measurable` (see javadoc on usage) instead of existing `Count` and `Volume` classes.

All changes are supposed to be backwards compatible, since the old `Count` and `Volume` classes are still present and functional.
They will internally use volume measurements of a one-hour interval for the mode car.

Closes #2780


kainagel:
Question: If it says "interval = 3600" and "t=3600", do you mean the time period from 0 to 3599, or from 3600 to 7199?  I think that it should be the latter.  (Otherwise, the second that is mentioned in the "t" entry would not even be part of the interval.) If this is already the case, it would help to have the examples start at "t=0".  If this is not the case, then I really really think we should change it.

We had lots of problems with the fact that the old format started at 1 and not at 0.  @mrieser , I think that you said the same two days ago.

Issue #2820, opened 2023-10-09
Freight: Improve NetworkbasedTransportationCost calculation / storage


kt86:
... in the `freight` contrib.

I know that @michalmac has invested a lot of effort to use faster methods in the context of DRT and for investigation related to the Vulkaneifel scenario.

@michalmac I would like to talk with you to see, what of these things can be taken to improve the speed of the freight contrib.

michalmac:
> @michalmac I would like to talk with you to see, what of these things can be taken to improve the speed of the freight contrib.

Definitely, let's talk.

kt86:
Update from yesterdays code-sprint discussion:

I have discussed this with @michalmac. This issue will be done in two steps:
1) Update the current calculation in a way that the routing is done for the time at the beginning of the time bin (and not for  the first time that requests the calculation within a time bin). This avoids different results if routing requests within the same time bin are changed. This change was done in a quickfix in #2850 together with @michalmac .
2) @michalmac will replace the current calculation and storage with the approach he developed in another part. This will include some adaptions to the needs here. This will be done in the coming week(s). 

Once again, thank you very much, @michalmac for your help and work!

PR #2850, opened 2023-10-12
Friehgt: minor change in NetbasedTransportCostCalculation≈

kt86:
The router now calculates the values for the time at the beginning of a timeslice.


kt86:
first step for #2820

Issue #2829, opened 2023-10-10
Consider turn restrictions during routing


marecabo:

paulheinr:
If done, #3461 should also be fixed.

PR #2855, opened 2023-10-12
[DO NOT MERGE] Add DisallowedNextLinks

marecabo:
**Depends on https://github.com/matsim-org/matsim-libs/pull/2849 to be merged first!**

This adds a class storing sequences of disallowed next links, which is used to model turn restrictions. See https://github.com/matsim-org/matsim-libs/issues/2829 for the description of the format.


marecabo:
I unfortunately had to adjust the `discrete-mode-choice` test in https://github.com/matsim-org/matsim-libs/pull/2855/commits/17962686a9d3afe58f195fe9660547826b2afe64. This test even fails on another [pr](https://github.com/matsim-org/matsim-libs/actions/runs/6516817266/job/17700714521?pr=2828) which should also be unrelated to `discrete-mode-choice`.

matsim/src/main/java/org/matsim/core/network/DisallowedNextLinks.java
```diff
@@ -0,0 +1,105 @@
...
+	 */
+	public boolean addDisallowedLinkSequence(String mode, List<Id<Link>> linkSequence) {
+		List<List<Id<Link>>> linkSequences = this.linkIdSequencesMap.computeIfAbsent(mode, m -> new ArrayList<>());
+
+		boolean result = false;
+		// prevent adding empty/duplicate link id sequences, or duplicate link ids
+		if (!linkSequence.isEmpty() && new HashSet<>(linkSequence).size() == linkSequence.size()
+				&& !linkSequences.contains(linkSequence)) {
+			result = linkSequences.add(ImmutableList.copyOf(linkSequence));
+			if (result) { // sorting is required for a.equals(b) <=> a.hashCode() == b.hashCode()
```
michalmac, line 47: Because we store link sequences as `ArrayList`, the return of `add()` is always `true`, so no need to check `if (result) {`

michalmac, line 47: I am not sure I understand this comment. Did you mean: sorting is required because we use a `List` instead of a `Set` for storing all link sequences and we want to easily compare them?

matsim/src/main/java/org/matsim/core/network/DisallowedNextLinks.java
```diff
@@ -0,0 +1,105 @@
...
+		this.linkIdSequencesMap.clear();
+	}
+
+	public boolean isEmpty() {
+		return this.linkIdSequencesMap.isEmpty();
+	}
+
+	@Override
+	public boolean equals(Object obj) {
+		if (obj instanceof DisallowedNextLinks dnl) {
```
michalmac, line 79: Early return if not `instanceof`?

matsim/src/main/java/org/matsim/core/network/DisallowedNextLinksUtils.java
```diff
@@ -0,0 +1,110 @@
...
+				errors.addAll(isNextLinkSequenceOf(links, link, linkSequence));
+			}
+		}
+		return errors != null ? errors : Collections.emptyList();
+	}
+
+	private static List<String> isNextLinkSequenceOf(Map<Id<Link>, ? extends Link> links,
+			Link link, List<Id<Link>> nextLinkIds) {
+
+		List<String> messages = null;
```
michalmac, line 77: For simplicity, I would pre-allocate an array list here: `var messages = new ArrayList();`. It's a relatively cheap operation.

Same for the `getErrors()` method.

contribs/discrete_mode_choice/src/test/java/org/matsim/contrib/discrete_mode_choice/examples/TestSiouxFalls.java
```diff
@@ -56,9 +56,9 @@ public void install() {
 
 		controller.run();
 
-		assertEquals(44196, (int) listener.counts.get("pt"));
+		assertEquals(44195, (int) listener.counts.get("pt"));
 		assertEquals(132316, (int) listener.counts.get("car"));
-		assertEquals(82140, (int) listener.counts.get("walk"));
+		assertEquals(82139, (int) listener.counts.get("walk"));
```
michalmac, line 61: If you rebase on the current master branch, it should be resolved (maybe not in the nicest way though).

michalmac (commented):
Nice! Looking forward to have turn restrictions in matsim!

I guess we will not merge it until the routing part is ready?

matsim/src/main/java/org/matsim/core/network/DisallowedNextLinks.java
marecabo: Thanks, I missed that!

matsim/src/main/java/org/matsim/core/network/DisallowedNextLinks.java
```diff
@@ -0,0 +1,105 @@
...
+		boolean result = false;
+		// prevent adding empty/duplicate link id sequences, or duplicate link ids
+		if (!linkSequence.isEmpty() && new HashSet<>(linkSequence).size() == linkSequence.size()
+				&& !linkSequences.contains(linkSequence)) {
+			result = linkSequences.add(ImmutableList.copyOf(linkSequence));
+			if (result) { // sorting is required for a.equals(b) <=> a.hashCode() == b.hashCode()
+				Collections.sort(linkSequences, (l, r) -> l.toString().compareTo(r.toString()));
+			}
+		}
+		return result;
```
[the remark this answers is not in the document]
marecabo: Thanks, that looks simpler!

matsim/src/main/java/org/matsim/core/network/DisallowedNextLinks.java
marecabo: I don't see how this would be simpler, maybe you can help me out?
I only could prepend these lines, but it would be too redundant, I think:
```java
if (!(obj instanceof DisallowedNextLinks)) {
	return false;
}
```

matsim/src/main/java/org/matsim/core/network/DisallowedNextLinksUtils.java
marecabo: Thanks, makes it a lot nicer!

contribs/discrete_mode_choice/src/test/java/org/matsim/contrib/discrete_mode_choice/examples/TestSiouxFalls.java
marecabo: I did a merge from master an manually took the lines from the master.

marecabo:
Thanks for the comments, @michalmac! I fixed all but one and did a merge. 

If this is fine, I would recommend to merge this when done (and not wait with merging this until the routing part is also done), so we can use this PR-release in pt2matsim to work on the OSM conversion in parallel to the routing.

matsim/src/main/java/org/matsim/core/network/DisallowedNextLinks.java
michalmac: Thanks for confirming my thinking.

I think this part of the comment is confusing `a.equals(b) <=> a.hashCode() == b.hashCode()`. Instead you could add part of your response as a comment.

matsim/src/main/java/org/matsim/core/network/DisallowedNextLinks.java
michalmac: Yes, this is what I was thinking about. This would reduce the level of nesting by 1 in the following `if` statements.

But it's just a hint, you can ignore it :-)

michalmac:
> If this is fine, I would recommend to merge this when done (and not wait with merging this until the routing part is also done), so we can use this PR-release in pt2matsim to work on the OSM conversion in parallel to the routing.

I would also recommend merging it. But perhaps we need to add clear error/warning message that this feature is not ready yet (in case someone specifies such restrictions in their network and expects them to be obeyed).

matsim/src/main/java/org/matsim/core/network/DisallowedNextLinks.java
marecabo: Sure, just added it.

matsim/src/main/java/org/matsim/core/network/DisallowedNextLinks.java
marecabo: You are right, it was possible to remove that one nesting, thanks!

marecabo:
Good point with the warning. I added a one-time warning in the constructor - and comments to remove it, once the routing part is finished.

PR #2863, opened 2023-10-13
Improve chance of compiling hybridsim and protobuf contrib

marecabo:
This change adds a plugin to the protobuf and hybridsim contrib that made them compile on our machines right after checking out the repo. 
During the Code Sprint we talked about that these contribs beared some challenges in getting matsim compile from source successfully.


michalmac:
We already have `os-maven-plugin` declared as extension. According to this [readme](https://github.com/trustin/os-maven-plugin#issues-with-eclipse-m2e-or-other-ides), we should remove that declaration if we go for the "plugin" option.

BTW. Which IDE do you use?

marecabo:
I'm mostly using [VS Code](https://code.visualstudio.com/) with its [Extension Pack for Java](https://marketplace.visualstudio.com/items?itemName=vscjava.vscode-java-pack).

michalmac (approved):
Nice!

PR #2865, opened 2023-10-13
DRT Leg Estimators

### LINKED ISSUE #2800: Review and consolidate DRT estimators
On the branch [drt-estimator](https://github.com/matsim-org/matsim-libs/tree/drt-estimator) there is some existing work that introduces estimators for drt travel and waiting times. Currently, it is designed to be used for estimating utility during mode choice. 

Apparently, similar functionality is also needed in the drt speed-up module and possibly even more places. The branch needs to be reviewed and discussed if this functionality and API needs to be consolidated.

rakow:
Adds DRT estimators to the  `drt-extensions` contrib. These can be used to estimate the expected distance, travelTime, waitingTime and fare of drt trips during replanning phase.

This module can be used together with the `informed-mode-choice` contrib, but could be useful at other places too, (e.g drt speed-up module)

At the moment, there is only one implementation that uses global estimates (Intended for small fleets and services areas)

Closes #2800


contribs/drt-extensions/src/main/java/org/matsim/contrib/drt/extension/estimator/BasicDrtEstimator.java
```diff
@@ -0,0 +1,174 @@
...
+	private GlobalEstimate currentEst;
+	private RegressionResults fare;
+
+	public BasicDrtEstimator(DrtEventSequenceCollector collector, DrtEstimatorConfigGroup config,
+							 DrtConfigGroup drtConfig) {
+		//zones = injector.getModal(DrtZonalSystem.class);
+		this.collector = collector;
+		this.config = config;
+		this.drtConfig = drtConfig;
+		this.speedUpParams = this.drtConfig.getDrtSpeedUpParams().isPresent() ? this.drtConfig.getDrtSpeedUpParams().get() : null;
```
michalmac, line 48: This could be replaced with a shorter version: `this.drtConfig.getDrtSpeedUpParams().orElse(null)`.

We could also have `Optional<DrtSpeedUpParams>` (despite IntelliJ suggesting that this is a bad practice), then line 55-56 would need to be adjusted accordingly.

michalmac: Or you can always simply call `drtConfig.getDrtSpeedUpParams()` (without storing it) - same way as the drt fare params are obtained in `estimate()`.

contribs/drt-extensions/src/main/java/org/matsim/contrib/drt/extension/estimator/MultiModalDrtLegEstimator.java
```diff
@@ -0,0 +1,74 @@
...
+			DrtEstimator instance = null;
+			try {
+				instance = injector.getInstance(Key.get(DrtEstimator.class, DvrpModes.mode(el.mode)));
+			} catch (Exception e) {
+				log.warn("Could not retrieve drt estimator for {}: {}", el.mode, e.getMessage());
+			}
+
+			estimators.put(el.mode, instance);
+		}
+	}
```
michalmac, line 50: I think you should use a map binder, i.e. `MapBinder<DvrpMode, DrtEstimator>` and bind estimators by mode in each `DrtEstimatorModule`.

rakow: But the estimators use modal bindings and will need other modal components. Will it work when binding them as map only?

michalmac: I think you should be able to do something like this:
1. In each `DrtEstimatorModule.install()` call:
```java
MapBinder.newMapBinder(this.binder(), DvrpMode.class, DrtEstimator.class) .mapBinder.addBinding(DvrpModes.mode(getMode())).to(DrtEstimator.class);
```
2. Inject the map into the `MultiModalDrtLegEstimator` constructor:
```java
@Inject
	public MultiModalDrtLegEstimator(Map<DvrpMode, DrtEstimator> drtEstomators) {...}
```

This way you avoid handing DI (e.g. using injector) inside the "business logic" code.

BTW. I haven't tried this example with your code, because there is no unit test to play with, so it may not be 100% correct.

rakow: Thanks, I got it working.

PR #2875, opened 2023-10-14
minibus: fix full integration on windows by adding another value to allowed results

michalmac:
After switching to the speedy router, `SubsidyTestIT` started to behave non-deterministically. It's a "quick fix" following the similar fix in this commit:
```
non-deterministic test, allow for 2 outcomes... :-/ Marcel Rieser Yesterday 20:58 31fe21b6
```


michalmac:
https://github.com/matsim-org/matsim-libs/actions/workflows/full-integration.yaml shows that the latest two builds failed (ids: 486 and 487)

marecabo:
Thanks, the Maps-in-ConfigGroups pr was also affected from this. 

I think, one discrete-mode-choice test now has a similar issue: https://github.com/matsim-org/matsim-libs/pull/2855#issuecomment-1762783104

PR #2876, opened 2023-10-16
Freight events reader test AND adaptions/fixes on carrierFreightXXXEventHandlers

### LINKED ISSUE #2771: Freight / Carriers: Add tests for EventsReader
This is a follow-up to #2769 (and thus #2404). Add additional tests when reading in events-files
At least two tests are needed - based on the type of the jobs:
- Services
- Shipments

nixlaos:
added tests for and CarrierEventsReaders. This closes #2771.


kt86:
This PR will also include the following fixes:
- renaming of `freight` * `EventHandler` to `carrier` * `EventHandler`
- add two missing EventHandlers: `pickup` and `dropoff` events both have two (and not only one) event: `start` and `end`
- Error in `convert` method in `CarrierShipmentDelivery` * `Event`

I also updated the test input files and added tests, that will count the number of carrier related events.

In a next step, I will also check the events itself – at least one per type.

PR #2877, opened 2023-10-16
Person vehicle types attribute

### LINKED ISSUE #2801: Store vehicle type in PersonVehicles
Currently, it is presumably not possible to have agents with differentiated vehicle types in `PersonVehicles` when using mode vehicles.

This feature would be useful when working with populations that may use a variety of types. E.g freight agents with different types of trucks. The differentiation is especially important when working with count data, since trucks and smaller cars are often counted separately.

rakow:
Introduces a `vehicleTypes` attribute in persons and a new `PersonVehicleTypes` class that allow to store mode specific vehicle types.  Use `VehicleUtils.insertVehicleTypesIntoAttributes` to set this attribute.

This class is similar to the existing `PersonVehicles`, but instead of the id it allows to store a mode specific vehicle type.
This attribute is completely optional, and the current behavior is not changed, unless this attribute is set.  Setting the type explicitly is useful for use-cases where mode specific vehicle types are not sufficient, and the types need to be further differentiated on a per-person basis. 

Closes #2801 .


rakow:
@mrieser: I think minibus and dmc contrib are still failing undeterministicly.

PR #2878, opened 2023-10-16
Clean up TimeAllocationMutators, add a new feature

jfbischoff:
This PR reduces the number of TimeAllocationMutators in MATSim from 3 to 1. 
Also, the functionality to provide different mutationRanges per Subpopulation has been removed; it has not been accessible from config for several years.
Optionally, it is possible to allow timeMutation only around the initial endTime of an Activity.


jfbischoff:
Fixes #2842 and solves #2843

jfbischoff:
Due to the diferent way how randoms are generated, several integration tests return slightly different values. These have been adjusted.

PR #2884, opened 2023-10-18
Bugfix/test sioux falls

jfbischoff:
Test results a different int on windows or linux, this should please both.


michalmac:
FYI This is not a platform-dependent issue. The same results (and consequently a failure) happened during the latest weekly build: https://github.com/matsim-org/matsim-libs/actions/runs/6527887455/

PR #2893, opened 2023-10-23
Listener priorities

Royal2Flush:
We introduced listener priorities but don't make widespread use of them yet as far as I can see. This merge request proposes some priorities of Listeners I found in the code.

Note: I will push a couple of more commits in the next days


matsim/src/main/java/org/matsim/core/controler/corelisteners/PlansDumpingImpl.java
```diff
@@ -66,6 +66,11 @@ final class PlansDumpingImpl implements PlansDumping, BeforeMobsimListener {
 		this.writePlansInterval = config.getWritePlansInterval();
 		this.writeMoreUntilIteration = config.getWritePlansUntilIteration() ;
 	}
+	
+	@Override
+	public double priority() { // this should happen after other BeforeMobsimListeners did their thing
+		return -1000;
```
rakow, line 72: You should make this a public final static constant, so other listeners can make use of it to control if they want to run before or after the plans dumping.

tschlenther:
@Royal2Flush please go ahead with this PR or otherwise we will soon close it, in order to stay on top of the PR list and get it a bit cleaner :) Thank you

PR #2900, opened 2023-10-25
drt (analysis): only include time windows where fleet is in operation in ProfileWriter

tschlenther:
The DRT contrib produces output files (png and txt) that provide information about fleet occupancy and vehicle tasks.
So far, the files included all time bins for the corresponding mobsim time span, even though the drt fleet might only be in operation for a specific time period in the day, e.g. from 6 am to 8pm.
This PR proposes to reduce outputted data to those time bins, where the drt fleet is in operation. More specifically, begin with the first vehicle operation start and end with the last vehicle operation end. If there is something like a lunch break, where no vehicle is in operation, this will be still included in the output.

Wrt to `RunDrtExampleIT.testRunDrtExampleWithNoRejections_ExtensiveSearch`,
this is how the well-known occupancy plot looked before:

this is how it would look after this PR:

The corresponding txt files will also be reduced, accordingly. This is actually intended, as it will help with the corresponding SimWrapper viz.

Let me know if you think this is a bad idea.

(This PR is a somewhat 'cleaner' version of #2894)

### DIFF
--- a/contribs/common/src/main/java/org/matsim/contrib/common/timeprofile/ProfileWriter.java
+++ b/contribs/common/src/main/java/org/matsim/contrib/common/timeprofile/ProfileWriter.java
@@ -19,12 +19,8 @@
  */
 package org.matsim.contrib.common.timeprofile;
 
-import java.awt.Paint;
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Map;
-import java.util.stream.Stream;
-
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Lists;
 import org.jfree.chart.JFreeChart;
 import org.jfree.chart.plot.XYPlot;
 import org.jfree.data.xy.DefaultTableXYDataset;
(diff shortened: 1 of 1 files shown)


michalmac:
I am just wondering about comparability with other modes/charts. Let's say we have drt1 and drt2, each starting at a different time. It will be harder to compare two charts if they have different x axes. Otherwise, I think it's a good change.

neuma:
Keeping the full simulation time visible gives me a quick double-check opportunity to verify that there are indeed no actions beyond the period of operation - e.g. stay. Also, I second @michalmac regarding the comparison of different drt operators and scenarios.

For publication-ready figures, I normally just exclude zero-only lines from the plots as required.

tschlenther:
> Keeping the full simulation time visible gives me a quick double-check opportunity to verify that there are indeed no actions beyond the period of operation - e.g. stay. 

That would still be possible, because we do not directly tie the output to the inputted operation times but to time bins where there was at least 1 vehicle doing 'something', i.e. we just cut off the white space in the first figure.

> Also, I second @michalmac regarding the comparison of different drt operators and scenarios.

Hmm... I see the point..

PR #2909, opened 2023-11-02
Bug fix regarding multiple plug in

auzpatwary37:
I have added necessary comments in he file for a smooth understanding. Basically, when the charging end event happens the vehicles are removed from the vehiclesAtCharger map. But if the charging is not started, they are never put into the map in the first place. So, assuming if the vehicles are not there they are full charge is not right as they might be still in the queue and was never plugged in the first place. As a result, the queue in the charging logic is not cleaned and they still await for plug in. The car can then be put into the logic of another charger and gets plugged in simultaneously in two chargers, hence the bug.

### DIFF
--- a/contribs/vsp/src/main/java/playground/vsp/ev/UrbanVehicleChargingHandler.java
+++ b/contribs/vsp/src/main/java/playground/vsp/ev/UrbanVehicleChargingHandler.java
@@ -115,9 +115,14 @@ class UrbanVehicleChargingHandler
 			if (tuple != null) {
 				Id<Vehicle> evId = Id.create(tuple.getFirst(), Vehicle.class);
 				if(vehiclesAtChargers.remove(evId) != null){ //if null, vehicle is fully charged and de-plugged already (see handleEvent(ChargingEndedEvent) )
+					//this is not necessarily true as vehicles can be also queued and waiting for plug in Nov23 Ashraf
 					Id<Charger> chargerId = tuple.getSecond();
 					Charger c = chargingInfrastructure.getChargers().get(chargerId);
 					c.getLogic().removeVehicle(electricFleet.getElectricVehicles().get(evId), event.getTime());
+				}else {
+					Id<Charger> chargerId = tuple.getSecond();
+					Charger c = chargingInfrastructure.getChargers().get(chargerId);
+					if(c.getLogic().getQueuedVehicles().contains(electricFleet.getElectricVehicles().get(evId)))c.getLogic().removeVehicle(electricFleet.getElectricVehicles().get(evId), event.getTime());// so it is important to remove the vehicle from queue as well.
 				}
 			} else {
 				throw new RuntimeException("there is something wrong with the charging procedure of person=" + event.getPersonId() + " on link= " + event.getLinkId());
@@ -134,6 +139,7 @@ class UrbanVehicleChargingHandler
 	public void handleEvent(ChargingEndEvent event) {
 		vehiclesAtChargers.remove(event.getVehicleId());
 		//Charging has ended before activity ends
+		// again this will only fire if the vehicle is picked from queue and plugged in//ashraf Nov2023
 	}
 
 	@Override
--- a/contribs/vsp/src/main/java/playground/vsp/ev/UrbanVehicleChargingHandler.java
+++ b/contribs/vsp/src/main/java/playground/vsp/ev/UrbanVehicleChargingHandler.java
@@ -114,15 +114,16 @@ class UrbanVehicleChargingHandler
 			Tuple<Id<Vehicle>, Id<Charger>> tuple = chargingProcedures.get(event.getLinkId()).remove(event.getPersonId());
 			if (tuple != null) {
 				Id<Vehicle> evId = Id.create(tuple.getFirst(), Vehicle.class);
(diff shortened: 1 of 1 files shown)


contribs/vsp/src/main/java/playground/vsp/ev/UrbanVehicleChargingHandler.java
```diff
@@ -115,9 +115,14 @@ public void handleEvent(ActivityEndEvent event) {
...
 				Id<Vehicle> evId = Id.create(tuple.getFirst(), Vehicle.class);
 				if(vehiclesAtChargers.remove(evId) != null){ //if null, vehicle is fully charged and de-plugged already (see handleEvent(ChargingEndedEvent) )
+					//this is not necessarily true as vehicles can be also queued and waiting for plug in Nov23 Ashraf
 					Id<Charger> chargerId = tuple.getSecond();
 					Charger c = chargingInfrastructure.getChargers().get(chargerId);
 					c.getLogic().removeVehicle(electricFleet.getElectricVehicles().get(evId), event.getTime());
+				}else {
+					Id<Charger> chargerId = tuple.getSecond();
+					Charger c = chargingInfrastructure.getChargers().get(chargerId);
+					if(c.getLogic().getQueuedVehicles().contains(electricFleet.getElectricVehicles().get(evId)))c.getLogic().removeVehicle(electricFleet.getElectricVehicles().get(evId), event.getTime());// so it is important to remove the vehicle from queue as well.
```
michalmac, line 125: Please reformat and rearrange your code (e.g. by adding local variables) so it is easier to read.

auzpatwary37: I have added a local variable for the EV for easier reading

tschlenther:
this makes me wonder whether we need the `vehiclesAtChargers` map at all...

By definetely removing the vehicle both from the charger logic and it's queue, I think we don't need this helper memory map and can thus also get rid of three methods, i.e. do not need to implement neither of the following itnerfaces

`ChargingEndEventHandler`, `ChargingStartEventHandler`, `QueuedAtChargerEventHandler`

As a result, the method in focus here would only be as follows:

```
@Override
	public void handleEvent(ActivityEndEvent event) {
		if (event.getActType().endsWith( UrbanEVModule.PLUGOUT_INTERACTION )) {
			Tuple<Id<Vehicle>, Id<Charger>> tuple = chargingProcedures.get(event.getLinkId()).remove(event.getPersonId());
			if (tuple != null) {
				Id<Vehicle> evId = Id.create(tuple.getFirst(), Vehicle.class);
				ElectricVehicle ev = electricFleet.getElectricVehicles().get(evId);
				Id<Charger> chargerId = tuple.getSecond();
				Charger c = chargingInfrastructure.getChargers().get(chargerId);
				c.getLogic().removeVehicle(ev, event.getTime());
				// it is important to remove the vehicle from queue as well.
				if(c.getLogic().getQueuedVehicles().contains(ev)){
					c.getLogic().removeVehicle(ev, event.getTime());
				}
			} else {
				throw new RuntimeException("there is something wrong with the charging procedure of person=" + event.getPersonId() + " on link= " + event.getLinkId());
			}
		}
	}
```


@auzpatwary37 can you confirm ?

auzpatwary37:
@tschlenther this would though an error in the charging with queuing logic. This is because if the vehicle is not found anywhere the precondition checker in the remove vehicle function in the ChargingWithQueuingLogic class throws an error that the vehicle is neither plugged in nor in queue in this charger.

tschlenther:
> @tschlenther this would throw an error in the charging with queuing logic. This is because if the vehicle is not found anywhere the precondition checker in the remove vehicle function in the ChargingWithQueuingLogic class throws an error that the vehicle is neither plugged in nor in queue in this charger.

right. thanks.
I think we should merge then (if the tests pass). Just on eminor comment: the code would be easier to read if you would separate the statement from the if clause more clearly, in line 125. Best would be to use braces and a new line, but at least a tab between the bracket and the beginning of the statement would be necessary to help the eye ;)

Thank you for the bug fix!!

contribs/vsp/src/main/java/playground/vsp/ev/UrbanVehicleChargingHandler.java
tschlenther: the code should be shorter and easier to read like this:

```
if(vehiclesAtChargers.remove(evId) != null || c.getLogic().getQueuedVehicles().contains(ev)){ 
	Id<Charger> chargerId = tuple.getSecond();
	Charger c = chargingInfrastructure.getChargers().get(chargerId);
	c.getLogic().removeVehicle(ev, event.getTime());
}
```

tschlenther:
thanks again for providing this bug fix. i will address in another PR with the attempt for clean and short code

tschlenther:
actually, this should already have been fixed in #2442 
@auzpatwary37 yo must have used a (local) version from before #2442 , right? can you confirm?

auzpatwary37:
@tschlenther maybe, seems like the same issue. Adding the vehicles that are queued in the chargers as well will indeed fix the issue without modifying the condition. We are still using an old version and currently trying to shift to the most recent one.

Issue #2925, opened 2023-11-10
edrt: results are not deterministic


sebhoerl:
While writing additional unit tests for #2826 (prebooking in combination with `edrt`), I noticed that the output of `edrt` is not deterministic. In particular, events appear at random times during the iteration:

![Screenshot from 2023-11-10 22-03-54](https://github.com/matsim-org/matsim-libs/assets/647500/d8935786-b5fa-4655-93a3-6b7de0e27181)

Here I was writing out the events of 10 runs of the `RunEDrtScenarioIT`, every time it has a different hash:

```
md5sum *
a9f047a4f63a71698a256c4c85c46bc8  events_0000.xml
525fdffcf9fdd148e65db5c9517c9749  events_0001.xml
d61901df6ca0d13f28f806521b779205  events_0002.xml
8202a42c67aa41f130e9ed5a91b2fedc  events_0003.xml
8de07129df24b47b8628736d463df501  events_0004.xml
fc260ed09667c25520fa6da808d2cbc7  events_0005.xml
2a1c9de1b973b7b8bde61da21300b769  events_0006.xml
79af77419d846a9fd21fba1bdb3015c1  events_0007.xml
927b83a35028ec1d5706a942314214d9  events_0008.xml
7f96411df0357fb6e05e2cb45ba20960  events_0009.xml
2800fcba68ade3881ba6ba89c71a56bc  events_0010.xml
```

Different from #2826 I couldn't yet find a case where it actually changes something in the dispatching decisions.

sebhoerl:
Potential fix is in #2925. When I run the `RunEDrtScenarioIT` in a loop, I get the following now:

```
md5sum *
d8aa9c92ecdb7ba56f14255a46d441eb  events_0000.xml.gz
1e151ba46007a3aa7cb5068ca76c3286  events_0001.xml.gz
1e151ba46007a3aa7cb5068ca76c3286  events_0002.xml.gz
1e151ba46007a3aa7cb5068ca76c3286  events_0003.xml.gz
1e151ba46007a3aa7cb5068ca76c3286  events_0004.xml.gz
1e151ba46007a3aa7cb5068ca76c3286  events_0005.xml.gz
1e151ba46007a3aa7cb5068ca76c3286  events_0006.xml.gz
1e151ba46007a3aa7cb5068ca76c3286  events_0007.xml.gz
1e151ba46007a3aa7cb5068ca76c3286  events_0008.xml.gz
[...]

PR #2926, opened 2023-11-10
fix: order of drt discharge events

sebhoerl:
Relatively simple fix for #2925: All discharging events are collected and only fired at the end of the simstep.

If somebody who is more familar with the `edrt` code could take over, that would be great :)

### DIFF
--- a/contribs/ev/src/main/java/org/matsim/contrib/ev/discharging/DischargingModule.java
+++ b/contribs/ev/src/main/java/org/matsim/contrib/ev/discharging/DischargingModule.java
@@ -41,6 +41,7 @@ public final class DischargingModule extends AbstractModule {
 			protected void configureQSim() {
 				this.bind(DriveDischargingHandler.class).in( Singleton.class );
 				addMobsimScopeEventHandlerBinding().to(DriveDischargingHandler.class);
+				addMobsimListenerBinding().to(DriveDischargingHandler.class);
 				// event handlers are not qsim components
 
 				this.bind(IdleDischargingHandler.class).in( Singleton.class );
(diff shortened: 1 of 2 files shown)


contribs/ev/src/main/java/org/matsim/contrib/ev/discharging/DischargingModule.java
```diff
@@ -41,6 +41,7 @@ public void install() {
 			protected void configureQSim() {
 				this.bind(DriveDischargingHandler.class).in( Singleton.class );
 				addMobsimScopeEventHandlerBinding().to(DriveDischargingHandler.class);
+				addMobsimListenerBinding().to(DriveDischargingHandler.class);
```
michalmac, line 44: This line should be replaced with:
`this.addQSimComponentBinding(EvModule.EV_COMPONENT).to(DriveDischargingHandler.class);`

Right now I doubt there is any event process. Maybe this is why so may tests failed.

contribs/ev/src/main/java/org/matsim/contrib/ev/discharging/DriveDischargingHandler.java
```diff
@@ -104,6 +111,7 @@ public void handleEvent(VehicleLeavesTrafficEvent event) {
 	// (for instance, AUX discharging and battery charging modifies charge outside event handling
 	// (as MobsimAfterSimStepListeners)
 	//TODO In the long term, it will be safer to move the discharging procedure to a MobsimAfterSimStepListener
+	// -> has been implemented now as a hack /sebhoerl
```
michalmac, line 114: While this change will definitely help with the order of events (which is good!), I was thinking about updating the SOC in the main thread (outside the event handling threads) to avoid race conditions (and also to have events deterministically ordered).

I will create a PR for moving the whole logic outside the event handling.

PR #2930, opened 2023-11-13
feat: canceling prebooked drt requests

sebhoerl:
**Context**
This is a follow-up PR to #2826. While #2826 introduces functionality to prebook requests in DRT, it is mainly aimed at fleet simulations where each person represents a single trip. Once we run simulations with a whole population, congestion, etc., things become trickier: What happens if we have a prebooked request, but the agent gets stuck during the day. Then the vehicle would wait indefinitely for the passenger. So as a remedy, we necessarily need some functionality to cancel requests during the day. And since we need the functionality in any case, it makes sense to also offer it as a clear interface, which allows modellers to set up an even more complex logic if they need to.

**Interface**
The main interface is added to the `PrebookingManager` that now allows, besides prebooking an agent's leg, to also cancel the associated request:

```PrebookingManager.cancel(Leg)```

The `PrebookingManager` also automatically cancels requests of agents for which a `PersonStuckEvent` is detected.

**Scheduling**
Technically, *canceling* a request means *unscheduling* the request if it has already been assigned to a vehicle. Currently, there are two implementations that can be chosen from the `PrebookingParams`:

- `SimpleUnscheduler` simply removes `DrtRequest`s from their associated `DrtStopTask`. This means that the vehicle will still visit the stop location, even if the stop becomes empty. However, this is a very robust and simple way to unschedule requests.
- `ComplexUnscheduler` is a more complicated implementation that will perform the necessary rewriting of the schedule if in the process described above, some stops are left without any pickups or dropoffs. There is an extensive unit test covering this class, but I'm still not 100% sure if it will cover all potential edge cases. This is why I made it optional, with the simple version being the default.

**DRT Changes**
Major changes in DRT are that the `DrtStopTask` now has `remove*` methods, otherwise only prebooking code is affected.

**Abandoning**
Referring to the example from above (agent gets stuck), the concept of "abandoning a request" is introduced. Technically, the `PrebookingStopActivity` will always wait for a passenger until the `earliestDepartureTime` of the request has been reached. Beyond that point, it will make use of the `AbandonVoter` interface to check if the request should be left behind (which means *canceling* it). The default implementation, `MaximumDelayAbandonVoter`, will wait for an additional *maximum delay* and then abandon the request. This delay can be configured in the `PrebookingParams` and is set to infinity (= never) by default.

### CODE REVIEW

**Comment by sebhoerl** [comment]
Right now all the commits from #2826 are included here. Once this one is merged, I will do a rebase.

**Comment by sebhoerl** [comment]
Voilà, here is the rebase. The only major change in DRT (modulo prebooking) is that `DrtStopTask` gets two additional methods `removePickupRequest` and `removeDropoffRequest`.

File: contribs/drt/src/test/java/org/matsim/contrib/drt/prebooking/ComplexUnschedulerTest.java [hunk 1]
```diff
@@ -0,0 +1,633 @@
...
+		assertTrue(schedule.getTasks().get(3) instanceof DrtDriveTask);
+		assertTrue(schedule.getTasks().get(4) instanceof DrtStayTask);
+		assertTrue(schedule.getTasks().get(5) instanceof DrtStopTask);
+		assertTrue(schedule.getTasks().get(6) instanceof DrtDriveTask);
+		assertTrue(schedule.getTasks().get(7) instanceof DrtStayTask);
+		assertTrue(schedule.getTasks().get(8) instanceof DrtStopTask);
+		assertTrue(schedule.getTasks().get(9) instanceof DrtDriveTask);
+		assertTrue(schedule.getTasks().get(10) instanceof DrtStayTask);
+		assertTrue(schedule.getTasks().get(11) instanceof DrtStopTask);
+		assertTrue(schedule.getTasks().get(12) instanceof DrtStayTask);
```
michalmac, line 104: Consider asserting begin and end time of each task.

contribs/drt/src/test/java/org/matsim/contrib/drt/prebooking/StuckPrebookingTest.java
```diff
@@ -0,0 +1,227 @@
...
+import org.matsim.core.controler.AbstractModule;
+import org.matsim.core.controler.Controler;
+import org.matsim.core.mobsim.framework.events.MobsimBeforeSimStepEvent;
+import org.matsim.core.mobsim.framework.listeners.MobsimBeforeSimStepListener;
+import org.matsim.testcases.MatsimTestUtils;
+
+/**
+ * @author Sebastian Hörl (sebhoerl) / IRT SystemX
+ */
+public class StuckPrebookingTest {
```
michalmac, line 39: Maybe `PersonStuckPrebookingTest` or `StuckPersonPrebookingTest` would be more descriptive?

michalmac (approved):
Nice to have this functionality! 💯 

I think there are a few issues that require addressing before merging (especially the ones around thread safety).

contribs/drt/src/main/java/org/matsim/contrib/drt/prebooking/PrebookingManager.java
```diff
@@ -180,30 +196,204 @@ public PassengerRequest retrieveRequest(MobsimAgent agent, Leg leg) {
...
+		if (item != null) {
+			item.vehicleId = event.getVehicleId();
+		}
+
+		if (unscheduleUponVehicleAssignment.contains(event.getRequestId())) {
+			// this is the case if a request has been rejected / canceled after submission
+			// but before scheduling
+			unscheduler.unscheduleRequest(event.getTime(), event.getVehicleId(), event.getRequestId());
+			unscheduleUponVehicleAssignment.remove(event.getRequestId());
+		}
```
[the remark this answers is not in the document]
sebhoerl: True, I will investigate this further. There might be an issue then that in the next time step the agent enters before it can be unscheduled. I'll check how to do this cleanly.

michalmac:
> The race condition problem is a tricky one, I thought that at last in `doSimStep` we have synchronized on all events, but this doesn't seem to be the case ...

I think `MobsimAfterSimStepListener` is safe (at least as long as you do not trigger new events there).

sebhoerl:
Should be thread-safe now. I collect all `scheduled`/`rejected`/`stuck` events during the simulation step and process everything in `notifyMobsimAfterSimStep`. Also, I collect all outgoing events (`booked`/`rejected`) and publish them in the next simulation step.

contribs/drt/src/main/java/org/matsim/contrib/drt/prebooking/PrebookingManager.java
```diff
@@ -106,60 +115,94 @@ public Id<Request> getRequestId(Leg leg) {
 		return Id.create(rawRequestId, Request.class);
 	}
 
+	// Event handling: We track events in parallel and process them later in
+	// notifyMobsimAfterSimStep
+
+	private final List<PassengerRequestScheduledEvent> scheduledEvents = Collections
+			.synchronizedList(new LinkedList<>());
```
michalmac, line 122: This could be also `ConcurrentLinkedQueue`.

sebhoerl:
Actually, forgot to switch `processEvent`/`flushEvents` to the next simstep after testing compatibility with the current version. This probably shifts some things by one second. Coming up.

PR #2933, opened 2023-11-14
EV: fix bugs related to charger queue

tschlenther:
This PR fixes 2 bugs:

1) in standard ev contrib: plug vehicles that are already in the charger queue before plugging in newly arriving vehicles
2) in vsp ev contrib (i.e. charging during activities): remove vehicle from charger queue when plugging out. before, the vehicle qas only removed from the charger if it was already plugged, i.e. not in the waiting queue. This address #2909 by also shortening the code.


tschlenther:
bug 2) was actually in the central ev contrib as well and should now  also be adressed, there.

PR #2942, opened 2023-11-17
Remove CSS style from SimWrapper dashboard warning panel

billyc:
The embedded <style> tags used on the `summary.md` file in the "Overview" panel are mostly OK, but one of them is interfering with the upcoming UI updates.

Removing the `.dash-card-frame` CSS selector solves the problem. 

I tested on the existing site and it does not seem to cause problems.


rakow:
It still works fine, but this line removed some margins, and I think it looks better without these in the old layout.
Somehow the new layout still has problems with it, so you need to take a look at it again before we can merge.

PR #2944, opened 2023-11-17
ev: handle waiting time

tschlenther:
This PR includes one new feature including a test (1) and some analysis enhancement (2)

1. In long-haul truck traffic, drivers are obliged to take a 45 minute break regularly. During this break, the must not move the vehicle for 1m. Consqeuently, if they can start their break only *after* moving forward in the charger queue, i.e. after they plugged in their vehicle.
This means that the charging activity duration should be 45 minutes *plus* the waiting time.
With the EV contrib, this could not be modeled so far, because the activity duration for the charging id preplanned and the execution included the time waiting in the charger queue.

This PR proposes `evConfigGroup.enforceChargingInteractionDuration` which prolongs charging interaction activitities for the time waiting in the charger queue (plus 1 to `evConfigGroup.chargeTimeStep` additional seconds). Per default, it is disabled. In consequence, the `maximumDuration` which is set by the `EVNetworkRoutingModule` would not need to account for waiting time, any more (I would adress this in a future PR - I think that the computation of the maximumDuration includes a bug, anyways).

The `RunEvExampleTest` is extended by one corresponding test case. I checked the reasonability of the outputs. Maybe this should be (and will be) tested more explicity, i.e. testing the execution against the preplanning for the charging duration based on events.

2. The handling of queueing, including quitting the queue without any charging (which is marked by a `QuitQueueAtChargerEvent`) is corrected in the `ChargingEventSequenceCollector`. Moreover, the iteration output file `<it>.chargingStats.csv` now contains information on the waiting times. To accomplish that, I replaced the `ChargerPowerCollector` with a new class `ChargingProceduresCSVWriter`. The latter uses the `ChargingEventSequenceCollector` and thus does not need to be an `EventHandler` itself. I also implemented the new analysis class as an `IterationEndsListener` as opposed to a `MobsimBeforeCleanupListener`, which is more align with most other statistics writers in MATSim. Moreover, it needs less infrastructure/objects to get injected.. One consequence of the things mentioned in the 2 two previous sentences is that the `ChargingProceduresCSVWriter` dumps out the linkId instead of X and Y coordinates, i.e. those **two previously contained columns are now not available any more**. As the coordinates were basically representing `link.getCoord()` I thought this should be alright. **Feedback is welcome!**


michalmac:
Thanks for fixing the queueing logic. I haven't noticed I was asked for review. Now I looked at the PR and it looks good to me.

PR #2946, opened 2023-11-18
escalate exceptions during parallel population reading (Fix for #2929)

### LINKED ISSUE #2929: core: ParalellPopulationReader fails without exception
I just came across an example where the ParallelPopulationReader processes a population, but inside there are inconsistencies (here, a referenced facility that does not exist). In the example, `PopulationUtils.decideOnCoordForActivity` causes a `RuntimeException` through `Gbl.assertNotNull`. While the error is printed in the console, there is no mechanism that let's the "top level" reader fail at the end of the parallel processing. Probably some custom exception handler is missing on the executing thread. If one doesn't look at the log carefully, the only thing that one notices is that there are agents missing ;) 

Screenshot from 2023-11-13 13-04-50

mrieser:
before, the reading of the population just continued, even if one of the reader-threads had an exception. Now, the exception from a reader-thread is propagated to the main-thread.

Closes #2929


mrieser:
Sorry, the PR also contains automatic whitespace-removal my IDE does…

steffenaxer (approved):

steffenaxer:
Approved! Thanks for the fix

PR #2947, opened 2023-11-20
feat: flexibly configure the drt insertion process

sebhoerl:
This PR adds functionality to customize the insertion process of DRT in terms of **constraints** and **objectives**. The idea is that the user can add a `DrtInsertionModule` to the run script and activate various constraints and change the goal of the insertion algorithm. The module is available in `drt-extensions/insertion`:

```java
DrtInsertionModule insertionModule = new DrtInsertionModule(drtConfig) //
	.withVehicleRange(vehicle -> 150.0 * 1e3) //
	.withSkills(vehicle -> { //
		return equippedVehicleIds.contains(vehicle.getId()) ? Set.of("wheelchair") : null; //
	}, request -> { //
		return (Boolaan) population.getPersons().get(request.getPassengerId()).getAttributes()
				.getAttribute("needsWheelchair") ? Set.of("wheelchair") : null; //
	}) //
	.withPromisedPickupTimeWindow(120.0) //
	.minimizeVehicleDistance();

controller.addOverridingQSimModule(insertionModule);
```

In this example, we define that vehicles should not move more than 150km (afterwards requests cannot be inserted), we give some vehicles that `wheelchair` skill, and we determine which request needs to be matched with those specific vehicles because they required the `wheelchair` skill. Additionally, when inserting, we require that already schedule requests don't get their initially assigned pickup time shifted by two minutes, and finally we define that the insertion algorithm should select those insertions that generate the least additional distance (instead of minimizing the vehicle activate time as it is done by default in DRT).

Technically, the module installs a custom version of `CostCalculationStrategy` that returns `InsertionCostCalculator.INFEASIBLE_SOLUTION_COST` if a constraint is not fulfilled and otherwise returns a custom value dependent on the configured objective. Note that this means that this module operates *after* all insertions have been generated and right *before* `{Selective,Extensive,...}InsertionProvider` selects one candidate based on the calculated objectives. Some constraints could very well be integrated already at the time of generating all insertions (like the exclusivity constraint). For now, this is a very straight-forward implementation without any changes to DRT itself, but we might think of a further integration in the future.

## Constraints
Currently, the following constraints are implemented.

### Exclusivity
Defines that certain requests are not allowed to share a vehicle with other passengers. 

```java
DrtInsertionModule insertionModule = new DrtInsertionModule(drtConfig)
	.withExclusivity(request -> exclusiveRequestIds.contains(request.getId()));
```

### Single request constraint
Defines that a person that submitted one request, cannot submit a second one. This is a rather technical constraint that is useful when running prebooking simulations. To perform a fair comparison between a prebooked and non-prebooked simulation this is necessary, because otherwise a person may get a prebooked request rejected, then wait until the departure is due and send an immediate request. The total number of requests with and without prebooking would, hence, be different when calculating the rejection rate and other metrics. On the implementation side, the constraint is interesting between it is an event handler. 

```java
DrtInsertionModule insertionModule = new DrtInsertionModule(drtConfig)
	.withSingleRequestPerPerson();
```

### Skills
The constraint guides the matching between vehicle and requests. For each vehicle, we can define a set of skills (defined as strings) and for requests, we define a set of requirements. A request can only be scheduled to a vehicle if it covers all the required skills of a person.

```java
DrtInsertionModule insertionModule = new DrtInsertionModule(drtConfig)
	.withSkills(
		vehicle -> { ... },
		request -> { ... });
```

There is one callback that returns a set of skills for each vehicle and another one that returns the requirements per request. The latter can, for instance, perform a lookup in the person attributes. Returning `null` here corresponds to returning an empty set (i.e., no special skills or requirements).

### Vehicle range
This constraint defines a daily drive range for the vehicles. Note that reconstructing the schedule's change in distance for every insertion can become an expensive task. Therefore, it is possible to define a (pessimistic) approximation using the Euclidean distance multiplied by a factor. Only if this pessimistic violates the constraint, an actual re-routing on the network is performed. In the long-term, it might make sense to add distance information in the `DetourInfo` objects generated by the insertion process.

```java
DrtInsertionModule insertionModule = new DrtInsertionModule(drtConfig)
	.withRangeEstimationFactor(2.0) //
	.withVehicleRange(vehicle -> 150.0 * 1e3);
```

### Custom constraints
It is possible to define custom constraints, either by providing an instance of `DrtInsertionConstraint` or by providing a class name of an implementation of `DrtInsertionConstraint` which will be searched in the DVRP mode's dependency injection space (`modalKey`).

```java
DrtInsertionModule insertionModule = new DrtInsertionModule(drtConfig)
	.withConstraint((request, insertion, detourTimeInfo) -> {
		// return true/false
	);
```

```java
DrtInsertionModule insertionModule = new DrtInsertionModule(drtConfig)
	.withConstraint(MyConstraint.class);
```

## Objectives
The following objectives are currently defined.

### Vehicle active time
This is the standard objective of DRT which minimizes the active time of the vehicles.

```java
DrtInsertionModule insertionModule = new DrtInsertionModule(drtConfig)
	.minimizeVehicleActiveTime();
```

### Passenger delay
This objective minimizes the delays of the passengers at pickup ("wait time") or dropoff ("travel time"). The objective is similar to the *vehicle activate time*, but the delays at each stop after insertion are weighted by the number of requests that are picked up or dropped off at each stop. The total pickup delay and the total dropoff delay for one insertion are multiplied by a weighting factor, respectively. Those are the values that can be provided when activating the objective.

```java
DrtInsertionModule insertionModule = new DrtInsertionModule(drtConfig)
	.minimizePassengerDelay();
```

```java
DrtInsertionModule insertionModule = new DrtInsertionModule(drtConfig)
	.minimizePassengerDelay(1.0, 2.0);
```

### Vehicle distance
This objective aims at minimizing distance when inserting a request. The distances that can be minimized are *occupied distance*, *empty distance*, and *passenger distance* (including occupancy) and they can be weighted individually. The same caveats apply as for the *range constraint*: Instead of performing the actual routing (= default), it may make sense to define a factor that is multiplied with the Euclidean distance.

```java
DrtInsertionModule insertionModule = new DrtInsertionModule(drtConfig)
	.withDistanceObjectiveFactor(1.3) //
	.minimizeDistances(1.0, 1.0, 0.0);
```

```java
DrtInsertionModule insertionModule = new DrtInsertionModule(drtConfig)
	.minimizeEmptyDistance();
```

### Custom objective
It is also possible to define custom objectives, either by providing an instance of `DrtInsertionObjective` or by providing a class name of an implementation of `DrtInsertionObjective` which will be searched in the DVRP mode's dependency injection space (`modalKey`).

```java
DrtInsertionModule insertionModule = new DrtInsertionModule(drtConfig)
	.minimize((request, insertion, detourTimeInfo) -> {
		// return objective value
	);
```

```java
DrtInsertionModule insertionModule = new DrtInsertionModule(drtConfig)
	.minimize(MyObjective.class);
```

## Other functionality
Some other functionality is currently present.

### Promised time windows
Based on some earlier work from MOIA, we can define the amount by which pickups and dropoffs can be shifted forward depending on the what has been promised upon first scheduling of a request.

```java
DrtInsertionModule insertionModule = new DrtInsertionModule(drtConfig)
	.withPromisedPickupTimeWindow(120.0) //
	.withPromisedDropoffTimeWindow(300.0);
```

### Soft insertion
When inserting a new request, we may detect that the latest pickup time or the latest dropoff time cannot be fulfilled. Normally, this is a hard constraint and the insertion is not feasible. However, we can decide to evaluate insertions using a soft constraint, i.e. we add the difference between the requested pickup/dropoff time and the one that can be proposed to the objective, weighted by their respective factors.

```java
DrtInsertionModule insertionModule = new DrtInsertionModule(drtConfig)
	.withSoftInsertionPenalty(1.0, 100.0);
```


tschlenther:
This reads really promising, thank you! Looking forward to (test) use cases...

sebhoerl:
Any objections? :)

contribs/drt-extensions/src/main/java/org/matsim/contrib/drt/extension/insertion/distances/RoutingDistanceCalculator.java
```diff
@@ -0,0 +1,26 @@
...
+	}
+
+	// Currently, simply protected through synchonized. Later on would be better to
+	// set this up in a more intelligent way, I'm not sure, but I guess that the
+	// constraints are accesses in parallel by DRT. /sh nov'23
+	@Override
+	public synchronized double estimateDistance(double departureTime, Link fromLink, Link toLink) {
+		VrpPathWithTravelData path = VrpPaths.calcAndCreatePath(fromLink, toLink, departureTime, router, travelTime);
+		return VrpPaths.calcDistance(path);
+	}
```
michalmac, line 25: This will be a bottleneck then (even if you have a separate calculator per each tread so the calls do not have to be synchronised). I would rather think about:
- exposing more data that is computed while processing each insertion
- using distance matrices instead

contribs/drt-extensions/src/main/java/org/matsim/contrib/drt/extension/insertion/distances/InsertionDistanceCalculator.java
```diff
@@ -0,0 +1,205 @@
...
+			}
+		}
+
+		return new VehicleDistance(occupiedDistance, emptyDistance, passengerDistance);
+	}
+
+	public VehicleDistance calculateInsertionDistance(Insertion insertion, DetourTimeInfo detourTimeInfo,
+			DistanceCalculator distanceEstimator) {
+		// pairs containing driven distance and occupany
+		List<Pair<Double, Integer>> addedDistances = new LinkedList<>();
```
michalmac, line 56: Maybe use a record instead of `Pair<Double, Integer>`?

contribs/drt-extensions/src/main/java/org/matsim/contrib/drt/extension/insertion/distances/RoutingDistanceCalculator.java
sebhoerl: Yes, this is just an implementation example for small use cases. I can add a warning that is printed when this class is used.

michalmac (approved):
I think this is a very valuable extension, but I think we should be careful when using this approach (or suggesting others to use it). My main concern is that drt is relatively slow, and by properly constraining the problem we can (often significantly) reduce the computation time. This however will not be the case when using this approach.

Of course, we should merge this PR, but it should be documented somehow that this approach is not optimal wrt computation speed and there are other ways of implementing such constraints.

For instance:
- skills or single request constraint, can be checked even before we start generating insertions for a given request-vehicle pair
- vehicle range - this kind of constraints are perfect fit for including them in the min cost path tree calculations

Despite my concerns, I look forward to have it merged :-)

contribs/drt-extensions/src/main/java/org/matsim/contrib/drt/extension/insertion/objectives/VehicleDistanceObjective.java
```diff
@@ -0,0 +1,32 @@
...
+
+import org.matsim.contrib.drt.extension.insertion.DrtInsertionObjective;
+import org.matsim.contrib.drt.extension.insertion.distances.DistanceCalculator;
+import org.matsim.contrib.drt.extension.insertion.distances.InsertionDistanceCalculator;
+import org.matsim.contrib.drt.extension.insertion.distances.InsertionDistanceCalculator.VehicleDistance;
+import org.matsim.contrib.drt.optimizer.insertion.InsertionDetourTimeCalculator.DetourTimeInfo;
+import org.matsim.contrib.drt.optimizer.insertion.InsertionGenerator.Insertion;
+import org.matsim.contrib.drt.passenger.DrtRequest;
+
+public class VehicleDistanceObjective implements DrtInsertionObjective {
```
michalmac, line 11: If this is the objective then probably the path search should also use the distance at the objective to keep things consistent.

sebhoerl: Hm, it depends on how you mean it. As long as the insertion algorithm is based on travel time, I think here we should also use travel time in the router when we "reconstruct" the routes that have been used to build the routes, to be coherent. Otherwise, I agree, if we can base the insertion on distances (see my other comment) then it would make sense to do that here as well. But then ideally we just have this information available in the detourInfo so we don't need to route at all at this point :)

sebhoerl:
@michalmac I completely agree, it would be good to move most of this way further upstream. It really should be thought of as a first attempt as a minimally invasive approach to add some constraints. However, it also (except the vehicle constraint) doesn't really impact performance because it just filters out options that would have been generated in any case if no constraints were set. 

Currently, the insertion logic is very dependent on travel times, travel time estimation etc. and there is not much of a notion of distance. So introducing distance (and we were also thinking of tracking energy directly on the insertion level) is quite a bit of effort, but we might tackle it at some point.

Regarding energy, we were even thinking of then generating some insertions to which charging activities are attached, this way we could really force insertions of charging activities in order to keep the itineraries feasible :)

neuma:
I find this an interesting approach. Thanks. 

> @michalmac I completely agree, it would be good to move most of this way further upstream. It really should be thought of as a first attempt as a minimally invasive approach to add some constraints. However, it also (except the vehicle constraint) doesn't really impact performance because it just filters out options that would have been generated in any case if no constraints were set.

Since I am also a bit concerned regarding the performance... Did you perform any tests with/wo vehicle constraints?

sebhoerl:
@neuma Every constraint comes with an impact. For instance, for the exclusivity constraint, we need to verify whether a request is inserted at a position where already someone is on board. The most impactful is probably the distance constraint as, theoretically, we have to reroute every insertion to obtain the distance delta. So this is kind of experimental and these distances should already be estimated on the fly during the insertion algorithm when we estimate the travel times. However, we already provide a remedy by letting users define the Euclidean distance factors that can be used instead (as a replacement of the routing for the distance objective, and as a pessimistic estimate for the constraint). We have tested with a large-scale scenario and having, for instance, a reasonable pessimistic distance estimation already reduces the additional run time a lot.

neuma:
@sebhoerl Thanks for looking into this. I find the insertion process is indeed the greatest bottleneck with respect to the computing performance. It is, however, also the most important part when it comes to the overall performance of the Drt system. The Euclidean distance might be a descent proxy for well connected networks. Considering rivers, railway tracks, and other obstacles - even the extensive use of one-ways in some cities - the Euclidean approach would add a lot of odd vehicle allocations. So, I understand that this part requires some tradeoff - more computational time or less accuracy. 

This is probably going offtopic... If the estimate is based on distance only, I would prefer to preserve the network connectivity and instead move to some more sophisticated static routing approach. Where can I start looking into this?

neuma:
@sebhoerl  Awesome. Thanks.

PR #2950, opened 2023-11-20
ev: make a reasonable assumption about power compatibility during routing

tschlenther:
In standard ev routing, the duration of a charging (interaction) activity is estimated. To estimate the charging time, the power needs to be determined. For that, we we take the minimum of the charger plug power and the vehicles power compatibility. For the determination of the latter, the vehicles power capacity was divided by 3.6, which results in very high power (compatibility) values. Pretty likely, this was meant to be `batteryCapacity / 3.600`, i.e. assume that the vehicle is compatible with a power that charges it within one hour (cf. `FixedSpeedCharging`). So far, the result was that we were taking the charger plug power for the estimation of the charging time. The RunEVExample shows the possible impacts: The vehcile capacitiy is 65 kWh, the charger has a power of 100kW, we used the latter for the duration estimation, but then during the execution we use min(100kW, 65kWh/3.600s)....
I changed the estimation calculation and the test accordingly...


tschlenther:
The bug was mentioned #2944. But here, we do not enhance the logic of the estimation, only the assumption behind it....

PR #2955, opened 2023-11-23
Make mode analysis faster

paulheinr:
With increased number of iterations the duration of `IterationEndListeners` rises linearly. 
image
The reason is that both `PHbyModeCalculator` and `PKMbyModeCalculator` write their **whole** output csv files each iteration. Thus, their writing overhead increases with each iteration. 

I think it's better just to append the output each iteration.

### DIFF
--- a/matsim/src/main/java/org/matsim/analysis/PHbyModeCalculator.java
+++ b/matsim/src/main/java/org/matsim/analysis/PHbyModeCalculator.java
@@ -34,10 +34,7 @@ import org.apache.commons.csv.CSVPrinter;
 import org.apache.logging.log4j.LogManager;
 import org.jfree.chart.axis.CategoryLabelPositions;
 import org.matsim.api.core.v01.IdMap;
-import org.matsim.api.core.v01.population.Activity;
-import org.matsim.api.core.v01.population.Leg;
-import org.matsim.api.core.v01.population.Person;
-import org.matsim.api.core.v01.population.Plan;
+import org.matsim.api.core.v01.population.*;
 import org.matsim.core.config.groups.ControllerConfigGroup;
 import org.matsim.core.config.groups.GlobalConfigGroup;
 import org.matsim.core.controler.OutputDirectoryHierarchy;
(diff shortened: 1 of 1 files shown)


matsim/src/main/java/org/matsim/analysis/PHbyModeCalculator.java
```diff
@@ -72,118 +69,116 @@ void addIteration(int iteration, IdMap<Person, Plan> map) {
...
+													 .toArray();
+			chart.addSeries(mode + WAIT_TIME_SUFFIX, valueWaitTime);
+		}
+		chart.addMatsimLogo();
+		chart.saveAsPng(controllerIO.getOutputFilename(FILENAME+ ".png"), 1024, 768);
+	}
+
+	private TreeSet<String> getAllModes() {
+		return this.phtPerIteration.values()
+								   .stream()
```
tschlenther, line 175: the indentation looks a bit odd, at least here on GitHub

paulheinr: You are right. For some reason IntelliJ used tabs instead of spaces for line indent. Thanks for the hint!

kt86: @paulheinr please have in mind that tabs are preferred, while spaces are not (in MATSim). See https://matsim.org/docs/devguide/conventions 
So if you fix it, please stay with tabs ;)

paulheinr: Okay, that's interesting. Then I will stay with tabs. But they seem a bit outdated to me.

At the code sprint we were talking about introducing the google code formatter. They also use spaces. But this is maybe not the right place for this discussion.

paulheinr:
My first intuition was not right: Writing the csv file doesn't cause the problems. It's the png file. An easy fix would be to turn off the "createGraphs" option. But then you don't get any graph. 

The benefit of the current implementation is that even if the simulation crashes at some point you get graphs until that point. So for testing this is nice, but normally you are interested in the last png. So maybe it's time for another config parameter like `createGraphInterval`, which could be a number (most flexible) or an enum like `eachIteration`, `lastIteration`...

@rakow @mrieser What do you think about it?

mrieser:
While I am — in general — always for better performance, this here looks for me to be "premature optimization". We're talking about potentially saving  half a second per iteration. As long as iterations usually take several minutes, I think this can be ignored. And if not, then people can disable the creation of graphs, as mentioned.

I think the potential overhead of having the events-handlers running at all is likely higher, and not a lot of (power) users use these statistics (they rather build their own statistics based on trips.csv or similar). I think we could have larger gains if we would bundle all those event handlers that produce informational statistics in a separate module that can be enabled and disabled by the config. Then users could disable it completely, likely saving more time (and memory) during iterations.

rakow:
I think Paul used a very simple toy model. We do notice some severe slowdown when doing a lot of iterations:

kexi-seed2-low-mode-innovation stopwatch

mrieser:
> I think Paul used a very simple toy model. We do notice some severe slowdown when doing a lot of iterations:

😲  wow, okay, but that's with like 3000 iterations? And the question is: is it due to the mode-analysis, or due to something else? I guess it's hard to profile this in iteration 2999 only ;-)

The log file should include the information when it calls which iterationEndsHandler (e.g. `calling notifyIterationEnds on org.matsim.core.controler.corelisteners.EventsHandlingImpl`. Maybe this information can be used to pinpoint which IterationEndsHandler is using more and more time?

In the chart above, interestingly, the overhead seems to get smaller again in the very last few iterations, when the mobsim is also faster. Why does the time of the mobsim change so much? Somehow it seems as if the iterationEnds-overhead is relative to the mobsim-runtime? Strange... (or was it a shared machine and there was something else running in iterations 1500–2900?)

mrieser:
Just did some performance test for the chart creation:

```java
import org.matsim.core.utils.charts.StackedBarChart;

import java.util.HashMap;
import java.util.Map;
import java.util.Random;

class Scratch {
	public static void main(String[] args) {
		int[] maxIterations = new int[] { 10, 50, 500, 1000, 3000 };

		for (int maxIteration : maxIterations) {
			System.out.println("# iterations: " + maxIteration);
			HashMap<String, Map<Integer, Double>> modeHistories = new HashMap<>();
			Random r = new Random(20231127L);
			String[] modes = new String[]{"car", "pt", "bike", "ride", "walk"};

			for (String mode : modes) {
				HashMap<Integer, Double> iterationValues = new HashMap<>();
				modeHistories.put(mode, iterationValues);
				for (int iteration = 0; iteration < maxIteration; iteration++) {
					iterationValues.put(iteration, r.nextDouble() * 100);
				}
			}

			long start = System.currentTimeMillis();
			StackedBarChart chart2 = new StackedBarChart("Mode Statistics", "iteration", "share");
			for (Map.Entry<String, Map<Integer, Double>> entry : modeHistories.entrySet()) {
				String mode = entry.getKey();
				Map<Integer, Double> history = entry.getValue();
				double[] historyArray = new double[history.size()];
				int i = 0;
				for (Map.Entry<Integer, Double> entryHistory : history.entrySet()) {
					historyArray[i] = entryHistory.getValue();
					i++;
				}
				chart2.addSeries(mode, historyArray);
			}
			chart2.addMatsimLogo();
			long end1 = System.currentTimeMillis();
			chart2.saveAsPng("test_stackedbar.png", 800, 600);
			long end2 = System.currentTimeMillis();
			System.out.println("preparing chart took " + (end1 - start) + " milliseconds");
			System.out.println("rendering chart as png took " + (end2 - end1) + " milliseconds");
		}
	}
}
```

Result:
```txt
# iterations: 10
preparing chart took 1439 milliseconds
rendering chart as png took 621 milliseconds
# iterations: 50
preparing chart took 10 milliseconds
rendering chart as png took 85 milliseconds
# iterations: 500
preparing chart took 93 milliseconds
rendering chart as png took 241 milliseconds
# iterations: 1000
preparing chart took 267 milliseconds
rendering chart as png took 266 milliseconds
# iterations: 3000
preparing chart took 3019 milliseconds
rendering chart as png took 573 milliseconds
```

So with very large datasets, the chart creation indeed takes several seconds. As we have multiple of such charts (score, modes, modesByPkm, modesByPh, …) it might indeed result in several seconds of overhead per iteration. Didn't expect that (or rather never thought about that).

mrieser:
Solutions?

- Write the chart only every 10th/50th/? iteration?
- Replace the `createGraphs` boolean config-parameter with something like `createGraphsInterval`?
- Automatically switch to an interval if `lastIteration > 300`?
- Try another library and see if it has the same problem?
- ...?

paulheinr:
I used your code to plot a chart with XChart and got these runtimes: 
```
# iterations: 10
preparing chart took 65 milliseconds
rendering chart as png took 1099 milliseconds
# iterations: 50
preparing chart took 1 milliseconds
rendering chart as png took 111 milliseconds
# iterations: 500
preparing chart took 2 milliseconds
rendering chart as png took 323 milliseconds
# iterations: 1000
preparing chart took 3 milliseconds
rendering chart as png took 305 milliseconds
# iterations: 3000
preparing chart took 5 milliseconds
rendering chart as png took 889 milliseconds
```
(I don't know what's happening there with 10 iterations)

It's only faster with 3000 iterations, but still close to 1s. So I think I prefer a numeric parameter `createGraphsInterval`.

mrieser:
My computer needs 3000 milliseconds what your computer does in 5 milliseconds?? I think I need to upgrade :-D

The first run with 10 iterations likely needs to do some static initialization that the other runs then benefit from, so I wouldn't give to much on that outlier.

paulheinr:
No no, I wanted to see, how fast other libraries are and used XChart instead of MATSim's default JFree. With JFree the computation times were similar to yours.

neuma:
> So with very large datasets, the chart creation indeed takes several seconds. As we have multiple of such charts (score, modes, modesByPkm, modesByPh, …) it might indeed result in several seconds of overhead per iteration. Didn't expect that (or rather never thought about that).

Just want to confirm that JFreeChart is an excellent pick for smaller datasets but struggles with larger ones. It's not only bar charts or XY-charts - which might be expected from any plotting library. But it also struggles with simple box and whisker charts whenever it tries to plot the same few outlier indicators on top of each other.

I have no solution though except for switching the graph creation off for larger runs.

mrieser:
XChart seems to be faster (and creates cleaner charts, IMHO).
But even then, it probably takes 1/3 or 1/4 of the time, it would still use up many seconds in late iterations.
Would it make sense to put the chart-generation into a separate thread, so it would not delay other code?

neuma:
A separate thread would mitigate albeit not solve the issue. 👍

There is also the bonus of a more fail-safe design. Currently, JFreeChart has the habit of crashing on newly installed Linux machines because of missing fonts that used to be installed per default. Crashing in a separate thread would allow MATSim itself to continue its run.

tthunig:
The minimal example of @paulheinr above was one of my illustrative examples. There, I need thousands of iterations. Thus, this drawing issue was severe. A run that took one week before only needs 30min when I disable createGraphs. Would be nice, though, if we would have a possibility to at least draw the graphs once in the last iteration.

paulheinr:
I've now changed the parameter to `createGraphsInterval`. The old parameter is still valid as input, but the output_config only contains the new one. To do so, I had to refactor some code. 
Each module, which is writing graphs, has to check itself if should do so in the current iteration. This leads to same copied code, but from architectural view these checks make more sense in the listeners than in the controller.

PR #2958, opened 2023-11-23
fix : handle spaces in comma-separated qsim config values

Nitnelav:
The qsim.mainMode definition in the XML was not trimmed (after being split). So definitions such as the one below returned ["car", " motorcycle"] for exemple.  

```xml
...
	<module name="qsim" >
                ...
		<param name="mainMode" value="car, motorcycle" />
                ...
	</module>
...
```
Same story for sqim.seepMode


Nitnelav:
Sure ! I don't have merge permissions though

PR #2969, opened 2023-11-28
feat: interaction activities in withinday replanning

sebhoerl:
I just noticed something interesting: I'm working with within-day replanning and I'm using `RoutingModule`s to replace certain parts of the agent schedule at the beginning of the simulation. The predefined `RoutingModule`s may create `InteractionActivity`s which I then integrate into the *modifiable plan* of an agent. This is a problem, because `BasicAgentImpl` eventually will call `setStartTime` on the activity one the agent arrives there, but this method is not defined for the specialized `InteractionActivity`. The reason why things work normally is that upon simulation start-up all `InteractionActivity` are automatically copied into a classic `Activity` that is used in the *modifiable plan* of the agent during the QSim.

This means that whenever we use `RoutingModule`s and insert the results in an ongoing agent plan within-day, we need to be able to manually convert the interaction activities to standard activity objects. This PR adds this functionaity:
- `PopulationUtils.convertInteractionToStandardActivity` converts an activity to a standard one if it is an interaction activity
- `WithinDayAgentUtils.convertInteractionActivities` fixes a list of plan elements coming from a routing module if necessary


tschlenther:
Nice!
A remark: I recently came across the EVNetworkRoutingModule, where similar things happen:

```
Activity chargeAct = PopulationUtils.createStageActivityFromCoordLinkIdAndModePrefix(selectedChargerLink.getCoord(),
	selectedChargerLink.getId(), stageActivityModePrefix);
// createStageActivity... creates a InteractionActivity where duration cannot be set.
chargeAct = PopulationUtils.createActivity(chargeAct);
```

That use case is a little special, I guess, because it wants to define an interaction activity that has a non-zero duration and therefore does this hacky copying... Maybe should be adopted using your suggested method ??

kainagel:
why are interaction acts not normal acts? to save memory? if so, is this really necessary?  in my intuition, giving a durstion to an i. a. is rather plausible.  If possible, i would prefer cleaning up the core data structure. Rather than building balconies around the problems. Thanks …

Sent from my iPhone

On 28 Nov 2023, at 12:15, tschlenther ***@***.***> wrote:

﻿

Nice!
A remark: I recently came across the EVNetworkRoutingModule, where similar things happen:

Activity chargeAct = PopulationUtils.createStageActivityFromCoordLinkIdAndModePrefix(selectedChargerLink.getCoord(),
        selectedChargerLink.getId(), stageActivityModePrefix);
// createStageActivity... creates a InteractionActivity where duration cannot be set.
chargeAct = PopulationUtils.createActivity(chargeAct);


That use case is a little special, I guess, because it wants to define an interaction activity that has a non-zero duration and therefore does this hacky copying... Maybe should be adopted using your suggested method ??

sebhoerl:
@kainagel probably @mrieser can explain if there was a big performance impact :)

PR #2972, opened 2023-11-29
Bugfix edrt failIfNotStarted

steffenaxer:
This PR is a fix for failIfNotStarted error in the EDrtVehicleDataEntryFactory. It is currently not checked whether a vehicle has already reached its out of service period within the VehicleEntry generation.


marecabo:
I think, you could reduce the nesting and integrate these new lines in an `else if` of the outer if statement.

contribs/drt-extensions/src/main/java/org/matsim/contrib/drt/extension/edrt/optimizer/EDrtVehicleDataEntryFactory.java
```diff
@@ -74,7 +74,11 @@ public VehicleEntry create(DvrpVehicle vehicle, double currentTime) {
 		if (schedule.getStatus() == ScheduleStatus.PLANNED) {
 			nextTaskIdx = 0;
 			chargeBeforeNextTask = battery.getCharge();
-		} else { // STARTED
+			// STARTED, but the vehicle is out of service time
```
michalmac, line 77: I think this comment should be moved one line down.

contribs/drt-extensions/src/main/java/org/matsim/contrib/drt/extension/edrt/optimizer/EDrtVehicleDataEntryFactory.java
```diff
@@ -74,7 +74,11 @@ public VehicleEntry create(DvrpVehicle vehicle, double currentTime) {
...
+		} else if (currentTime >= vehicle.getServiceEndTime()) {
```
michalmac, line 78: Consider changing this series of `if`s to a switch statement with cases for PLANNED, STARTED and default.

contribs/drt-extensions/src/main/java/org/matsim/contrib/drt/extension/edrt/optimizer/EDrtVehicleDataEntryFactory.java
michalmac: Is it really `STARTED, but the vehicle is out of service time`? Then you would not get `failIfNotStarted`?

contribs/drt-extensions/src/main/java/org/matsim/contrib/drt/extension/edrt/optimizer/EDrtVehicleDataEntryFactory.java
steffenaxer: I'm not so familiar with those `ScheduleStatus`, if a started state ensures also a vehicle is in operational time window? Maybe you might refactor this? Till this, I would go without switches.

PR #2975, opened 2023-12-01
Add static factory method for ChargingInfrastructureSpecificationDefaultImpl

nkuehnel:


nkuehnel:
See
https://github.com/matsim-org/matsim-libs/pull/2888#issuecomment-1835783587

michalmac (approved):
:+1

PR #2984, opened 2023-12-06
DRT: Sharing factor and pooling rate

nkuehnel:


nkuehnel:
This PR adds code to evaluate pooling rates and sharing factors for DRT systems.

Pooling rate is a common metric used in the evaluation of on-demand pooling systems and is a simple true/false on whether there was any temporal overlap between two rides/bookings (even if just for 1s).

As pooling rate itself is not very expressive (since it doesn't say anything about the length of shared trips), MOIA internally uses and defines a "sharing factor" that takes into account how long trips are shared. It is defined as:

$$
\phi_i = \frac{t_i^{actual}}{\sum_{s \in segments_i} \frac{t(s)}{o^{passenger}(s)}}
$$

where $t_i^{actual}$ is the total duration of trip $i$, $t(s)$ is the duration of segment $s$ and $o^{passenger}(s)$ is the occupancy during each segment of the trip. A trip is split into segments between each stop where additional passenger may board or leave. The sharing factor is >=1 (1 meaning that the whole trip was on its own). If, i.e., a trip is completely shared with another booking, the trips sharing factor equals 2.

In detail, it adds 
- a per iteration output with sharing factor and pooling rate per request
- a cross iteration output with aggregated stats (number of pooled rides, average sharing factor, overall pooling rate)

A small test case is included.

contribs/drt/src/main/java/org/matsim/contrib/drt/sharingfactor/SharingFactorModule.java
```diff
@@ -0,0 +1,38 @@
...
+
+    @Override
+    public void install() {
+        bindModal(SharingFactorTracker.class).toProvider(modalProvider(getter ->
+                new SharingFactorTracker(new SharingFactorTracker.GroupPredicate() {
+                    @Override
+                    public boolean isGroupRepresentative(Id<Person> personId) {
+                        return SharingFactorTracker.GroupPredicate.super.isGroupRepresentative(personId);
+                    }
+                }))).asEagerSingleton();
```
michalmac, line 29: This is super complex, especially `SharingFactorTracker.GroupPredicate.super.isGroupRepresentative(personId)`.

I think it is equivalent to this:
```java
bindModal(SharingFactorTracker.class).toProvider(modalProvider(getter ->
                new SharingFactorTracker(personId -> true))).asEagerSingleton();
```

contribs/drt/src/main/java/org/matsim/contrib/drt/sharingfactor/SharingFactorTracker.java
```diff
@@ -0,0 +1,112 @@
...
+			double total = 0;
+			double portion = 0;
+
+			boolean pooled = false;
+
+			Segment last = finishedSegments.get(0);
+			if (last.occupancy > 1) {
+				pooled = true;
+			}
+			for (int i = 1; i < finishedSegments.size(); i++) {
```
michalmac, line 64: I think these 5 lines may get simplified to:
```java
var last = new Segment(0,1);
for (var next : finishedSegments) {
```

contribs/drt/src/test/java/org/matsim/contrib/drt/sharingfactor/SharingFactorTest.java
```diff
@@ -0,0 +1,205 @@
...
+import org.matsim.core.events.ParallelEventsManager;
+import org.matsim.testcases.MatsimTestUtils;
+
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * @author nkuehnel / MOIA
+ */
+public class SharingFactorTest {
```
michalmac, line 20: Nice to have this part tested!

michalmac (approved):
Looks good to me. Nice!

I left some suggestions for simplifying code.

contribs/drt/src/main/java/org/matsim/contrib/drt/sharingfactor/SharingFactorTracker.java
```diff
@@ -0,0 +1,112 @@
...
+
+	private final Map<Id<DvrpVehicle>, List<Id<Request>>> map = new HashMap<>();
+
+	private final Map<Id<Request>, List<Segment>> segments = new HashMap<>();
+
+	private final Map<Id<Request>, Double> sharingFactors = new HashMap<>();
+	private final Map<Id<Request>, Boolean> poolingRate = new HashMap<>();
+
+	public interface GroupPredicate {
+		default boolean isGroupRepresentative(Id<Person> personId) {
```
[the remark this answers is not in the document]
nkuehnel: ok, I'll remove it

PR #2989, opened 2023-12-08
Make TripsAndLegsWriter public again

steffenaxer:
#2955 removed the public constructor of the `TripsAndLegsWriter `. I would kindly ask to make this constructor again public. @jfbischoff developed this very valuable class, which is also a strong utility within post-processing tasks. 

However, post-processing would require the creation of a certain instance of this class. I observed in the recent past some changes that removed the visibility of classes and constructors. Maybe we require a general discussion about how to deal with this kind of an issue. Removing the visibility might be good in terms of code maintenance, however it pushes users/devs into the direction to make local code copies, which feels like a silent fork. In my eyes, MATSim is a framework that should allow users/devs to reuse code instead of making various replicas of the very same functionality.

### DIFF
--- a/matsim/src/main/java/org/matsim/analysis/TripsAndLegsWriter.java
+++ b/matsim/src/main/java/org/matsim/analysis/TripsAndLegsWriter.java
@@ -79,7 +79,7 @@ public class TripsAndLegsWriter {
     private static final Logger log = LogManager.getLogger(TripsAndLegsWriter.class);
 
 	@Inject
-	TripsAndLegsWriter(Scenario scenario, CustomTripsWriterExtension tripsWriterExtension, CustomLegsWriterExtension legWriterExtension,
+	public TripsAndLegsWriter(Scenario scenario, CustomTripsWriterExtension tripsWriterExtension, CustomLegsWriterExtension legWriterExtension,
 					   AnalysisMainModeIdentifier mainModeIdentifier, CustomTimeWriter customTimeWriter) {
 		this.scenario = scenario;
 		this.tripsWriterExtension = tripsWriterExtension;


mfitz:
> ...however it pushes users/devs into the direction to make local code copies, which feels like a silent fork

We've done exactly this "silent/local fork" in the past when something we were using in the road pricing module went from public to private. Removing things from the public API is a breaking change and carries risk for users, so it seems like something that should not be done lightly.

steffenaxer:
Think on running this class without injection, without running a whole simulation. In order to do so, you would need your own controlled that creates the injection. Use case would e.g. to re-create a trips and legs file with additional columns without running a simulation

jfbischoff:
Trips And Legs Writer should be callable for Post Processing purposes - so I'm also in favor :)

kainagel:
Some comments:
1. Some constructors (and other methods) in MATSim are unfortunately public which should not have been public to begin with.  We know that these are breaking API changes, but there is also no point to support an erroneous design decisions until the bitter end.  Could you please have a look at https://github.com/google/guice/wiki/KeepConstructorsHidden .
2. You do not need a full `Control(l)er` to get injection.  Could you please have a look at `ControlerUtils.createAdhocInjector(...)`?  Evidently, one still binds the controler infrastructure, which may still be overkill in many situations, but we could think about a simpler version.
3. If you end up copying code, could you please let us know?  In most cases, this should indeed not be necessary.  We should either be able to provide an alternative.  Or change back to public constructor.

Having said all this, I would say that `TripsAndLegsWriter` is a borderline case.  
1. Evidently, we do not use injection at `Scenario` level so far, and in consequence it does not make sense to enforce injection to classes that are "at the same level" (i.e. can be used with a `Scenario` without having a `Control(l)er`).  
2. On the other hand,`TripsandLegsWriter` already has a fair number of dependencies, implying that there might come more, pointing towards guice injection.

In the balance, I would say that 1 wins over 2 (in this case by quite some margin).

PR #2990, opened 2023-12-09
DRT - consolidate request insertion schedulers for default and shift implementation

nkuehnel:


nkuehnel:
This PR gets rid of the duplication of the DRT request insertion scheduler for the driver shift extension. It's meant as a first step forward consolidating it with the latest changes including prebooking and group bookings. The updated implementation of @sebhoerl for the request insertion scheduler seems to work for the shift case as well.

The only thing I would like some feedback on is how to deal with the initialization of the fleet. So far this happens in the ReuqestInsertionScheduler (maybe it doesn't even have to?). However, in the shift extension, vehicles should start wth a "WaitForShift" task. A proposal that is presented in this PR is that the task factory interface provides a createInitialTask method that can be implemented to serve different needs and returns the task vehicles should be initialized with.

WDYT?

@michalmac @sebhoerl

michalmac (approved):
Nice. LGTM

sebhoerl:
Nice, next stop, merging the PassengerEngines ;)

nkuehnel:
@sebhoerl I'm on it, expect a draft later today ;)

PR #2995, opened 2023-12-13
ev: handle driving energy consumption via MobsimEngine (in the next time step)

michalmac:
The recent fix for non-determinism wrt drive-related energy consumption (both battery update and sending events) https://github.com/matsim-org/matsim-libs/pull/2935 resulted in the following issue reported by @mattiasingelstrom:

> After PR2935, our EventHandler logging and penalising MissingEnergyEvents doesn’t work as before. I get an error saying events in the queue are not ordered chronologically.

The implemented behaviour that stopped working was: in response to receiving a `MissingEnergyEvent`, create and emit a corresponding `PersonMoneyEvent` to penalise for getting undercharged. The second event resulted in a `RuntimeException` saying `"Events in the queue are not ordered chronologically. [...]"`

After looking into `SimStepParallelEventsManagerImpl`, it seems that throwing `MissingEnergyEvent`s from a `MobsimAfterSimStepListener` and then responding to these events with `PersonMoneyEvent`s is not really supported.

Therefore, we should move the handling of driving energy consumption from a `MobsimAfterSimStepListener` to a `MobsimEngine`. However, this means we should also delay discharging by one time step, that is:
- in time step `n`: collect all `LinkLeaveEvent`s and `VehicleLeavesTrafficEvent`s
- in time step `n+1`: update battery charge levels, and emit `DrivingEnergyConsumptionEvent`s and potentially `MissingEnergyEvent`s )

This also means that if step `n` is the last step of a simulation, the corresponding battery events will be emitted during `afterSim()` (with time set to `Infinity`).

===

This whole setup seems a bit complex. The biggest issue is that there is no easy way to feedback events (e.g. `LinkLeaveEvents`) into simulation (e.g. updating battery charge level) in a deterministic way without introducing a one second delay.

But maybe there is a way to do so? Or a simpler workaround?


steffenaxer:
@michalmac with this PR I have one edrt test failing with non-deterministic result. It's pretty hard to find the root cause. But I can say with 100 % certainty that with PR #3003 the same simulation behaves as expected.

In my test I am counting passenger pickup events. Since PR #2995 slightly different results occur in which vehicles show slightly different discharging curves. That leads finally to different vehicle availabilities causing slightly different passenger pickup events. May I kindly ask to check for this behavior? I will try to create a test. However, any assistance in the meantime would be wonderful.

PR #2997, opened 2023-12-15
Add DRT detour constraints (based on PR #2834)

luchengqi7:
This PR is based on the finished version of PR #2834, built based on the latest matsim-lib.

For more details, please refer to the previous PR.


contribs/drt/src/main/java/org/matsim/contrib/drt/passenger/MaxDetourOfferAcceptor.java
```diff
@@ -0,0 +1,23 @@
...
+	private final double promisedPickupTimeWindow;
+
+	public MaxDetourOfferAcceptor(double promisedPickupTimeWindow) {
+		this.promisedPickupTimeWindow = promisedPickupTimeWindow;
+	}
+
+	@Override
+	public Optional<AcceptedDrtRequest> acceptDrtOffer(DrtRequest request, double departureTime, double arrivalTime) {
+		double updatedPickupTimeWindow = Math.min(departureTime
+			+ promisedPickupTimeWindow, request.getLatestStartTime());
```
michalmac, line 15: Consider a different name than `promisedPickupTimeWindow`. We usually think of a time window as a `[a,b)` period, where `a` and `b` are time stamps. Here, `promisedPickupTimeWindow` is a duration and represents a time buffer.

In the drt config group, we use `maxAllowedPickupDelay` instetad of `promisedPickupTimeWindow`. I think we should also use this name in this class.

michalmac: Also `updatedPickupTimeWindow` should be renamed. Maybe `updatedLatestStartTime`?

contribs/drt/src/main/java/org/matsim/contrib/drt/routing/DrtRouteCreator.java
```diff
@@ -65,17 +65,30 @@ static double getMaxTravelTime(DrtConfigGroup drtCfg, double unsharedRideTime) {
...
 
+	/**
+	 * Calculates the maximum ride time defined as: drtCfg.maxDetourAlpha * unsharedRideTime + drtCfg.maxDetourBeta
+	 *
+	 * @param drtCfg
+	 * @param unsharedRideTime ride time of the direct (shortest-time) route
+	 * @return maximum ride time
+	 */
+	static double getMaxRideTime(DrtConfigGroup drtCfg, double unsharedRideTime) {
+		return Math.min(unsharedRideTime + drtCfg.maxAbsoluteDetour, drtCfg.maxDetourAlpha * unsharedRideTime + drtCfg.maxDetourBeta);
```
michalmac, line 76: I am not sure if we should use `unsharedRideTime + drtCfg.maxAbsoluteDetour` in both `getMaxTravelTime()` and `getMaxRideTime()`.

Imposing this constraint on `maxTravelTime` implies it is also imposed on `maxRideTime`, because `maxTravelTime >= maxRideTime`.

contribs/drt/src/main/java/org/matsim/contrib/drt/run/DrtConfigGroup.java
```diff
@@ -113,6 +113,27 @@ public static DrtConfigGroup getSingleModeDrtConfig(Config config) {
...
+	@Parameter
+	@Comment(
+		"Defines the maximum allowed absolute detour based on the unsharedRideTime. A linear combination similar to travel time constrain is used"
+			+ "This is the constant part")
+	@PositiveOrZero
+	public double maxDetourBeta = Double.POSITIVE_INFINITY;// [s]
+
+	@Parameter
+	@Comment(
+		"Defines the maximum delay allowed from the initial scheduled pick up time. Once a estimated pick up time is determined, the DRT optimizer"
```
michalmac, line 132: Maybe rewrite the second sentence in this way?
"Once the initial pickup time is offered, he DRT optimizer must not exceed the promised time by the `maxAllowedPickupDelay` amount of time."

contribs/drt/src/main/java/org/matsim/contrib/drt/run/DrtConfigGroup.java
```diff
@@ -113,6 +113,27 @@ public static DrtConfigGroup getSingleModeDrtConfig(Config config) {
...
+			+ "This is the ratio part")
+	@DecimalMin("1.0")
+	public double maxDetourAlpha = Double.POSITIVE_INFINITY;
+
+	@Parameter
+	@Comment(
+		"Defines the maximum allowed absolute detour based on the unsharedRideTime. A linear combination similar to travel time constrain is used"
+			+ "This is the constant part")
+	@PositiveOrZero
+	public double maxDetourBeta = Double.POSITIVE_INFINITY;// [s]
```
michalmac, line 128: These two comments should mention that the detour is computed from the latest promised pickup time and not from the actual pickup time.

I think we should also mention the following somewhere (maybe in `MaxDetourInsertionCostCalculator`):
By taking into account the latest promised pickup time (instead of the actual pickup time), we simplify the handing of this constraint, because the actual pickup time remains unknown until the moment of picking up the passenger.

BTW. In a simulation setup without congestion, the actual pickup time is always less than or equal to the promised time. With congestion there is no such a guarantee.

michalmac (approved):
Nice work!

I added a few comments. Please consider them so it is easier to reason for the users and developers how this constraint really works.

nkuehnel:
Thanks @luchengqi7 !

@michalmac would it make sense to consolidate the OfferAcceptor and InsertionCostCalculator to implement the new constraint in the default implementations? The parameters are already part of the default drt config and request anyway

michalmac:
> @michalmac would it make sense to consolidate the OfferAcceptor and InsertionCostCalculator to implement the new constraint in the default implementations? The parameters are already part of the default drt config and request anyway

Good point! Actually, that was the main reason for adding these params to the config (I assume).

contribs/drt/src/main/java/org/matsim/contrib/drt/run/DrtConfigGroup.java
luchengqi7: Thanks, @michalmac! I will update the descriptions and comments based on the comments.

luchengqi7:
> > @michalmac would it make sense to consolidate the OfferAcceptor and InsertionCostCalculator to implement the new constraint in the default implementations? The parameters are already part of the default drt config and request anyway
> 
> Good point! Actually, that was the main reason for adding these params to the config (I assume).

Oh yeah, I will update that! 😀

michalmac:
@luchengqi7 I reenabled IT tests in taxi and rerun them on my laptop (intel CPU). The problem with CI failures is that the workflows are run on a linux machine with an intel CPU. When runnin on a mac machine with an apple cpu, you get slight differences in scores (output plans).

I think, in the long term, we need to change the way how we compare output plans to account for differences between these platforms.

nkuehnel:
thanks so much @luchengqi7 !

PR #3002, opened 2023-12-18
Drt: fix occupancy change calculation at waypoints to account for group requests

nkuehnel:


nkuehnel:
So far occupancy change for waypoints always assumed one passenger per request. It should, however, sum up all passengers of all requests entering/leaving.

michalmac:
Good catch! We have some tests where we verify feasibility of insertions (also wrt the vehicle capacity). Maybe we should add/modify some of them to reflect that there may be more passengers?

nkuehnel:
done, added a small test :)

Issue #3010, opened 2023-12-21
PR #2995 might produce non-deterministic results for edrt simulations


steffenaxer:
@michalmac with PR #2995, I have one edrt test failing with non-deterministic result. It's pretty hard to find the root cause. But I can say with 100 % certainty that with PR #3003 the same simulation behaves as expected.

In my test I am counting passenger pickup events. Since PR #2995 slightly different results occur in which vehicles show slightly different discharging curves. That leads finally to different vehicle availabilities causing slightly different passenger pickup events. May I kindly ask to check for this behavior? I will try to create a test. However, any assistance in the meantime would be wonderful. Maybe @sebhoerl has an idea regarding this issue?

michalmac:
@steffenaxer #2995 obviously results in a different behaviour, so that is fine. But could you confirm: is the behaviour always reproducible (i.e. deterministic) or not?

If it is non-deterministic, then it would be helpful if you could provide some samples of events of two or more different runs (for the same matsim version). Thanks!

steffenaxer:
Yes in can confirm that the simulation produces different non deterministic results. I will provide you tomorrow the requested files!

steffenaxer:
Unfortunately debugging is quite painful. I tried to strip down code and variants as best as possible and found out that pickUpEvents behaves non-deterministic if we have two drt modes. These two modes are the only available modes. So agent needs to use drt1 or drt2. Both drt service specs are exactly the same. If I remove the second drt mode everything is again deterministic. However, I have also a test without eDrt and the remaining setup is identical. This non eDrt test produces after 10 iterations deterministic results.
So the problem needs to be somehow related to the following aspects:
1.	There are two drt modes 
2.	eDrt is used
3.	in the first iteration (Iteration 0) results are identical. Simulation deviates in later iterations.

steffenaxer:
Please find attached three simulations, executed with (PR #2995)
[output_events_1.xml.gz](https://github.com/matsim-org/matsim-libs/files/13751966/output_events_1.xml.gz) (pickUpEvents=2747)
[output_events_2.xml.gz](https://github.com/matsim-org/matsim-libs/files/13751963/output_events_2.xml.gz) (pickUpEvents=2754)
[output_events_3.xml.gz](https://github.com/matsim-org/matsim-libs/files/13751965/output_events_3.xml.gz) (pickUpEvents=2755)

steffenaxer:
Please find attached three simulations, executed with (PR #3003)
[...]

PR #3016, opened 2023-12-27
ev: fix missing drive discharging events due to concurrent access to a hash map

michalmac:
@steffenaxer this is a potential fix for #3010.


michalmac:
@steffenaxer Thanks for creating a test instance, that helped with narrowing down the potential cause of this non-determinism. I squashed some of your commits, where you were trying out different setups. The final version of the test is 1:1 with the current state of https://github.com/steffenaxer/matsim-libs/tree/edrtDeterminism.

steffenaxer:
Thanks Michal for this wonderful Christmas present 🎁!

PR #3021, opened 2023-12-28
DRT: add maxShiftStartDelay and log message when a shift is deleted due to…

TomE168:
Before, shifts with start times smaller than timestep were removed directly from unscheduledShifts-queue. With the new parameter, it is possible to define a max Delay, which postpones the removal. Also, a log warning message is triggered, when shifts are removed.


nkuehnel:
LGTM, thanks @TomE168

Issue #2851, opened 2023-10-12
Remove lanes


tthunig:
As discussed in the code sprint, we want to [enable turn restrictions on the link level](https://github.com/matsim-org/matsim-libs/issues/2829). With this we don't need the extra lane format in MATSim anymore: We can **instead have multiple parallel links** if we want to model separate waiting queues for different turn directions or traffic signals.

As far as we see, this only affects 
1. _visualization_ (in current Via links would be placed on top of each other; but current Via was also not able to visualize lanes) and 
2. the _insertion of agents into traffic_ from their activities, i.e. the choice of the appropriate link to enter the traffic (which is currently done at the first, common part of the link in case lanes are used).

Both issues could be managed by providing an alignment attribute on the link level (as before on the lane level) which determines the order of parallel links (see also the detailed proposal below).

Note, that signals in MATSim already work without lanes, as they only control outflow of queues (QueueWithBuffer). OTFVis is also already able to visualize traffic signals on links (and lanes).

Background:
MATSim's QSim is quite complex. For the link/queue representation for example we have AbstractQLink, QLinkImpl, QLinkLanesImpl and QueueWithBuffer which all depend on each other. If we are able to remove lanes, we could further simplify the remaining structure and make it easier to maintain/extend/replace functionality.

![Screenshot 2023-10-12 at 14 02 48](https://github.com/matsim-org/matsim-libs/assets/9266830/9765298f-88d4-4309-932c-606e74fe42a6)
The picture shows different possible link/lane representations. Format 0 corresponds to the standard MATSim case without lanes, format 1 corresponds to the current MATSim setup with lanes, format 2 corresponds to a similar queue behavior as format 1 but without lanes (as soon as turn restrictions are allowed on the link level) and format 3 is a simplified version of format 2 with less accuracy in terms of spillback behavior. 
As both, format 2 and 3, would be possible with the new turn-specific links, we can basically leave it to the user which one is chosen. For conversion, I suggest to go for format 2 (see proposal below).

Proposal:

- [ ]  [Enable **turn restrictions on the link level**](https://github.com/matsim-org/matsim-libs/issues/2829)
- [ ] Add **alignment information on the link level** (to define the order of parallel links; as before on the lane level). 
[...]

PR #3024, opened 2023-12-29
Support turn restrictions in SpeedyALT

mrieser:
This PR adds support for taking turn restrictions into account when using SpeedyALT routing. It is part of #2829.

Turn restrictions in MATSim are modeled with link attributes named `disallowedNextLinks` (see #2855).

When creating the routing graph for `SpeedyALT` (or `SpeedyDijkstra`, respectively), the builder for `SpeedyGraph` can take the disallowed-next-links into account and encode the restrictions into the routing graph. As turn restrictions are mode specific, creating the routing graph requires knowledge of the mode for which turn restrictions should be applied.

Technical documentation of how turn restrictions are incorporated into the routing graph used by the Speedy*-routers is available in the wiki: https://github.com/matsim-org/matsim-code-examples/wiki/turn-restrictions


mrieser:
@michalmac , @tduberne , @marecabo : Hi all, I implemented turn restrictions in SpeedyALT, but to fully integrate it into MATSim requires some additional work, where I could use your input. What is your opinion to the following points:

Turn restrictions (i.e. `DisallowedNextLinks`) are mode specific. Thus `SpeedyGraphBuilder.build()` needs to know for which mode the turn restrictions should be applied. This information is currently often missing.

Background: MATSim-Networks can be multimodal. The `NetworkRoutingProvider` filters the multimodal network for a specific mode. This filtered network is then passed to a `LeastCostPathCalculatorFactory` to create the actual routing algorithm instance. The `LeastCostPathCalculatorFactory` has in the past thus effectively ignored the modes and just used whatever links were available in the network.

While the `NetworkRoutingProvider` knows for which mode the network should be filtered for, the `LeastCostPathCalculatorFactory` currently does not know about the mode with its current API. But to create the routing graph with turn restrictions, we need this information.

The easiest solution would be to add a mode-parameter to `LeastCostPathCalculatorFactory.createPathCalculator()`, but is it the right solution? Would users then probably expect that a multimodal-network is filtered by such a factory when they pass a specific mode to it? (Currently, the network is filtered and cached outside of this method).

Also, DRT et al makes use of `SpeedyGraph` in various places outside of `LeastCostPathCalculatorFactory` and will need to adapt as well. What would the best solution be there? Does it already pre-filter the network for a specific (drt-)mode or might this be an oversight currently (I haven't dug deep enough into the code to answer this myself)?

As we pass the `mode` to `SpeedyGraphBuilder.build()`, would users expect that the graph is also filtered by that mode? Would that be a "safer" solution? Currently, the network is not filtered, as it is already filtered in `NetworkRoutingProvider`.

What would a good API look for to fully integrate the turn restrictions into MATSim?

Should we generally move the mode-filtering from `NetworkRoutingProvider` to the `LeastCostPathCalculatorFactory`? (But as we have multiple factories, e.g. for Dijkstra, AStarLandmarks, SpeedyALT, ... that filtering would have to happen multiple times = potential code duplication)

michalmac:
Great to have this feature! I have not looked into code changes. I will try to quickly answer your questions:

> Also, DRT et al makes use of `SpeedyGraph` in various places outside of `LeastCostPathCalculatorFactory` and will need to adapt as well. What would the best solution be there? Does it already pre-filter the network for a specific (drt-)mode or might this be an oversight currently (I haven't dug deep enough into the code to answer this myself)?

Each DVRP mode (drt, taxi, etc.) can use exactly one of these networks:
1. a global DVRP routing network (could be potentially filtered by modes defined in `DvrpConfigGroup.networkModes`)
2. a mode-specific DVRP network (that network is created by filtering the global DVRP routing network, so if the global network was already pre-filtered, this is a second filtering step)

By default, `DvrpOfflineTravelTimeEstimator` and `DvrpOfflineTravelTimes` use the global network, so there is no redundancy here (estimating TTs for each DVRP mode). However, this can also be changed if we want to address some special cases.

==

When it comes to the proposed turn restriction functionality, this means that:
- we can still use the global DVRP routing network for offline/online TT estimation
- we cannot support turn restrictions for DVRP modes that use the global network (i.e. case 1), but we can enable turn restrictions for DVRP modes that use their mode-specific networks (i.e. case 2), which I think is fine

> What would a good API look like to fully integrate the turn restrictions into MATSim?
>
> Should we generally move the mode-filtering from `NetworkRoutingProvider` to the `LeastCostPathCalculatorFactory`? (But as we have multiple factories, e.g. for Dijkstra, AStarLandmarks, SpeedyALT, ... that filtering would have to happen multiple times = potential code duplication)

I think that's the cleanest/safest path.

Maybe we should consider deleting some of the older routers like `AStarEuclidean` and `AStarLandmarks` (I always feel that having too much old legacy code only slows us down, at least in the long term).

I am not sure what to do with the routers that have been moved to `locationchoice`. Migrating to the speedy ones might be cleanest, but that means some extra work.

marecabo:
This is a great feature, thanks a lot! 

Currently, it is specific to `SpeedyAlt` but I think it could become generic for all routers. We transform the network with this subgraphs, generate routes, and then we transform the routes back to only using links existing in the network. However, it would require introducing some e.g.,  `NetworkTransform` and `RouteDetransform` steps before and after all routers. What are your thoughts on this? ;)

Anyway, I am excited to use this feature in our models! 🚀

tduberne:
Hi, and first of all thank you for the implementation and sorry for the radio silence... Apparently I need to check my github notification settings and/or e-mail filters.

I agree with Michal: moving mode filtering to the least cost path algorithm factory feels like the (second) cleanest approach. I actually remember wanting it to be the case in the past already.

I say "second cleanest", because from a distance, it feels like the turning restrictions should be packageable with the filtered network (or better, one would not need to create a filtered network apart from an internal data structure for the algorithm). I will keep that in the back of my mind when diving in the code, but I am sure that if it was possible with reasonable effort you would already have made it possible.

I will have a look at moving mode filtering into the router factories in the coming days.

kainagel:
Thanks a lot.  Didn't see this before, sorry.

From a higher level, I agree with @tduberne : The `LeastCostPathCalculator` is the "computer science" router, and that does not know anything about modes.

Again from my higher level, I do not understand why for mode-dependent turning restrictions we need the mode at the level of `LeastCostPathCalculator` ... it seems to me that it should possible to generate a modal network with turning restrictions for each mode separately.  There is probably something in the text above that I did not grasp.

Also, there were routers with turning restrictions before (`InvertedLeastPathCalculator` and related material).  What is happening to them?  Are we deprecating them?  Are we using the file formats that were already there, or are we also deprecating them?  If we are deprecating the old material, I am not against it since these features were not used a lot, indicating that they may not have served people's needs.  But I would prefer to truly deprecated them if they are superseded.

@tthunig , do you have an opinion?

tthunig:
The goal was to deprecate and later remove the old material afterward (see [#2851](https://github.com/matsim-org/matsim-libs/issues/2851)).

tduberne:
Proposition, as an alternative to making the LeastCostPathCalculatorFactory filter the network: embrace the fact that the least cost path algorithms use a monomodal network, and extend the DisallowedNextLinks structure to allow monomodal semantics.

Proposition: in a monomodal setting, all "disallowed next link sequences" present in the network have to be obeyed, irrespective of the mode they are defined for.

Consequence: the mode filtering algorithm would also need to consider those, and filter them as well.

@mrieser @marecabo what do you think of this solution?

marecabo:
Thanks for your comments and your proposal in #3221, @tduberne!

> Consequence: the mode filtering algorithm would also need to consider those, and filter them as well.

Yes, I think, that makes sense. The additional methods in #3221 for `DisallowedNextLinks` look good to me! 

For (potentially?) better performance/memory usage, we could consider just storing the `Collection<List<Id<Link>>>` from `DisallowedNextLinks.getMergedDisallowedLinkSequences()` in the mono-modal network in some designated internal attribute (e.g., `disallowedNextLinkSequences`), as we only need this information for building a graph.

tduberne:
> For (potentially?) better performance/memory usage, we could consider just storing the Collection<List<Id<Link>>> from DisallowedNextLinks.getMergedDisallowedLinkSequences() in the mono-modal network in some designated internal attribute (e.g., disallowedNextLinkSequences), as we only need this information for building a graph.

I did consider it, but decided against it, because:
- in the current state of the code, as far as I can tell, getMergedDisallowedLinkSequences will only be called once per simulation per link, to create the immutable routing network, so a cached value would not be used.
- but adding an internal cache attribute to DisallowedNextLink, even if initialized to null, would add the memory usage of one reference (4 or 8 bytes depending on architecture) to each DisallowedNextLink
- - the code is written in a way that one can "use a multimodal network in a monomodal way", because the network filter allows to filter several modes, but in practice there should probably be only one, which leads to limited overhead. One potential optimization would be to handle the monomodal case specially, but I really do not expect it to make any difference in practice

I will try to add a few tests before the end of the week, and then we can have a look at merging.

marecabo:
> * in the current state of the code, as far as I can tell, getMergedDisallowedLinkSequences will only be called once per simulation per link, to create the immutable routing network, so a cached value would not be used.

Thanks for the clarification, then I do not see it as necessary and let's keep it simple.

tduberne:
@mrieser @marecabo from my perspective work is done here:
- the router is able to be restricted to the proper mode, including turn restrictions, based on the proposed "a-modal" interpretation of a network with turn restrictions and the inclusion of turn restrictions in network filtering
- new tests cover network filtering and routing with mode restrictions

Is it OK to merge?

marecabo:
Thanks a lot! I think, we can then remove the warning in the constructor of DisallowedNextLinks, because DNLs are now considered.

PR #3026, opened 2023-12-30
Ensure deterministic depot selection

steffenaxer:
This PR ensures a deterministic depot selection for eDrt in situation in which multiple depots are having the very same distance - unfortunately, yes this might happen 😓 and would cause a non-deterministic depot selection. Took me quite some time to narrow down this problem.

### DIFF
--- a/contribs/drt/src/main/java/org/matsim/contrib/drt/optimizer/depot/Depots.java
+++ b/contribs/drt/src/main/java/org/matsim/contrib/drt/optimizer/depot/Depots.java
@@ -50,9 +50,9 @@ public class Depots {
 		if (!STAY.isBaseTypeOf(currentTask)) {
 			return false;
 		}
-		
+
 		// only if stay task is last task: with prebooking we may also idle during the day, but
-		// currently all the downstream relocation/charging logic assumes that we only stay at 
+		// currently all the downstream relocation/charging logic assumes that we only stay at
 		// the end of the schedule
 		if (currentTask.getTaskIdx() < schedule.getTaskCount() - 1) {
 			return false;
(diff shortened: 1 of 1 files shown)


contribs/drt/src/main/java/org/matsim/contrib/drt/optimizer/depot/Depots.java
```diff
@@ -64,13 +64,15 @@ public static boolean isSwitchingFromStopToStay(DvrpVehicle vehicle) {
...
-						.min(Comparator.comparing(
-								l -> DistanceUtils.calculateSquaredDistance(currentLink.getToNode().getCoord(),
-										l.getFromNode().getCoord())))
-						.get();
+			null /* already at a depot*/ :
+			links.stream().map(l -> new DepotCandidates(l, DistanceUtils.calculateSquaredDistance(currentLink.getToNode().getCoord(),
+					l.getFromNode().getCoord())))
+				.sorted(Comparator.comparing(DepotCandidates::distance)
+					.thenComparing(h -> h.link.getId()))
+				.findFirst().get().link();
```
michalmac, line 74: Use `min(...).get()` instead of `sorted(...).findFirst().get()`. There is no point in sorting the whole stream if we only want to get the minimum element.

contribs/drt/src/main/java/org/matsim/contrib/drt/optimizer/depot/Depots.java
```diff
@@ -64,13 +64,15 @@ public static boolean isSwitchingFromStopToStay(DvrpVehicle vehicle) {
 	}
 
 	public static Link findStraightLineNearestDepot(DvrpVehicle vehicle, Set<Link> links) {
```
michalmac, line 66: FYI Another option would be to pass a list of links to `findStraightLineNearestDepot()` instead of a set and shift the responsibility for reproducibility to the client code. But I think using the link ID for draws is safer.

michalmac (approved):
LGTM. Please remove the sorting of the stream.

contribs/drt/src/main/java/org/matsim/contrib/drt/optimizer/depot/Depots.java
steffenaxer: Done

PR #3027, opened 2023-12-30
Provides a way to provide custom handling for unknown parameters in ReflectiveConfigGroup

### LINKED ISSUE #2960: Migration path between config version
One big puzzle piece I'm missing is a migration path between config group versions. Currently matsim throws an error if there is an unknown parameter. In my eyes it would require a handler that could be overwritten per ConfigGroup that allows to say what should happen with an unknown parameter. If nothing is overwritten  the current error is thrown. What do you think @mrieser, @kainagel?

mrieser:
If a ConfigGroup based on ReflectiveConfigGroup has to add an unmapped parameter, usually an exception is thrown.

In some cases, it could make sense to provide special handling in such cases, e.g. to implement backward compatibility to older parameter names, to support a migration from older to newer parameters, etc.

This PR adds two methods `handleAddUnknownParam(...)` and `handleGetUnknownValue(...)` (matching `addParam(...)` and `getValue(...)`) that are designed to be overwritten. The default implementation reuses the existing logic in ReflectiveConfigGroup. Custom config groups can overwrite the methods and add custom functionality to deal with unknown parameters.

Fixes #2960

### DIFF
--- a/matsim/src/main/java/org/matsim/core/config/ReflectiveConfigGroup.java
+++ b/matsim/src/main/java/org/matsim/core/config/ReflectiveConfigGroup.java
@@ -285,14 +285,29 @@ public abstract class ReflectiveConfigGroup extends ConfigGroup implements Matsi
 			return;
 		}
 
-		Preconditions.checkArgument(storeUnknownParameters, "Module %s of type %s doesn't accept unknown parameters."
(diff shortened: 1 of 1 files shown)


steffenaxer:
Nice job! Thanks 🙏

PR #3032, opened 2024-01-02
Cleanup small scale traffic generation

rakow:
Various improvements for the small scale traffic generation
- Don't read the shape file unnecessarily often
- Use interfaces as types and not implementations
- Improve run script, so that a separate config only for generation should not be necessary


rewertvsp (approved):
Thx Christian for the changes. I also did some changes so that PR can be done.

PR #3036, opened 2024-01-05
change input vehicle file for EventBasedAnalysis

rewertvsp:
Changes the input vehicleFile for the analysis.
- the reason is that the allVehicles file only contains the types of the used vehicles 

Problem: if we read a carrier which uses **not** all available vehicles the types of this vehicles are not in the allVehicles file, and it creates this [`exception`](https://github.com/matsim-org/matsim-libs/blob/95612817ff151219caea5b7e13a5cd721b4f2665/contribs/freight/src/main/java/org/matsim/freight/carriers/CarrierPlanXmlParserV2_1.java#L203-L205)
Solution:
- using the carriersVehicleTypes file instead


kt86:
I would like to stay for the analysis with the "allVehicles" container.

If the CarrierReader is the problem, Then we may need the option to read the carriers file even, if some vehicleTypes are missing. This should not be the default because for the normal simulation and tour planning, we require all vehicle types, but here for the analysis, we do not.

kt86:
Will close this PR here. 
I have created an issue instead, that should be addressed when implementing another approach.

PR #3065, opened 2024-01-22
Only use EvBaseModule in EDrt

nkuehnel:


tschlenther:
definetely a good way to quick-fix your problems.
As I am not heavily familiar with edrt, and especially not with your custom behavioral models:
Will your charging behavior for edrt work without the VehicleChargingHandler?
From the code, it reads like it wasn't meant to be working with DynAgents, in the first place. But, as it seems, it worked for DynAgents - until #2944. As @steffenaxer pointed out, we will probably need to come up with test cases for simualtins with both private and commercial (edrt) charging behavior.
But for a start, it would be good if you can confirm that both run successfully, when separated.

steffenaxer:
Yes the VehicleChargingHandler is in my point of view not required anywhere for eDrts.

tschlenther (approved):

PR #3067, opened 2024-01-22
avoid NullPointerException in situations where we have charging/queui…

### LINKED ISSUE #3025: PR #2944 Might not work in combination with eDrt and eTaxis
PR #2944 introduced some major changes in the `VehicleChargingHandler` which is regularly part of the `EvModule`. The `EvModule` is also regularly part of the `EDrtControlerCreator`.  In context of eDrt I'm observing NPEs at least in this code line https://github.com/matsim-org/matsim-libs/blob/b3af4975e9daffb82917d605828a01b7ba76b708/contribs/ev/src/main/java/org/matsim/contrib/ev/charging/VehicleChargingHandler.java#L153

This might be caused in a situation in which vehicles are send to charging but the driver has not left the vehicle and was not registered in the `lastDriver` hashmap. In my case I have a feature that I call proactive charging for eDrts, that charges vehicles that are idle but their SOC is higher than the minSOC, which needs to be exceeded normally in order to charge the vehicle. (Might be published in the future)

While debugging the code I observed some other mysterious comments. The `EvModule` has a comment that says this module should not be used with DynAgents, however all of us are using it when working with eDrt.
https://github.com/matsim-org/matsim-libs/blob/b3af4975e9daffb82917d605828a01b7ba76b708/contribs/ev/src/main/java/org/matsim/contrib/ev/EvModule.java#L40

**Some thoughts I would like to discuss:**
1. It should be possible to simulate eDrt and ownership EVs together. So, one should check actively the relevance of the comments in the `EvModule`. There will be a great demand in the future to simulate both aspects in combination. 
2. The newly introduced code @tschlenther should also work with situations in which a vehicle could be charged, even if the driver has not left the vehicle. Otherwise, one needs to ensure that the `VehicleChargingHandler` is only used for non DynAgents.

@sebhoerl please have a look on this, as you already implemented a lot ev charging behavior.

tschlenther:
…ng with drivers that do not leave their vehicle for that (DynAgents). those won't be able to prolong their charging activities (obviously)

An attempt to fix #3025.

@steffenaxer can you possibly test this?

### DIFF
--- a/contribs/ev/src/main/java/org/matsim/contrib/ev/charging/VehicleChargingHandler.java
+++ b/contribs/ev/src/main/java/org/matsim/contrib/ev/charging/VehicleChargingHandler.java
@@ -52,7 +52,10 @@ import org.matsim.core.mobsim.qsim.QSim;
 import org.matsim.core.mobsim.qsim.agents.WithinDayAgentUtils;
 import org.matsim.vehicles.Vehicle;
 
-import java.util.*;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 
 /**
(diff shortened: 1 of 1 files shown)


tschlenther:
@nkuehnel gave positive feedback. so i will merge (if checks run through)

PR #3068, opened 2024-01-22
Moves PassengerGroupIdentifier for Drt Companions from Person to Leg

steffenaxer:
This PR moves the PassengerGroupIdentifier information (stored in attributes) from the person level to the leg level. This is necessary, because an agent might travel with drt multiple times a day with different groups of companions. Each group needs to have its own PassengerGroupIdentifier.


nkuehnel (approved):
LGTM :)

PR #3075, opened 2024-01-25
Railsim: Moving blocks & Deadlock avoidance

rakow:
This PR extends the recently introduced railsim contrib (#2773) with two major feature updates:

-  Allow modelling of mesoscopoic links as in ETCS Level 3 "moving block"
    - Links can be explicitly set as "fixed block" (existing logic) or "moving block"
    - On moving blocks, trains will not block the whole segment, but instead reserve the distance required to ensure safe operation

- Introduction of deadlock avoidance algorithms
   - Bottleneck segments (e.g. links with capacity of 1) can easily lead to a deadlock when two trains want to traverse a bottleneck from different directions at the same time
   - To avoid this problem, a new component was introduced which can prevent trains from reserving links that will lead to a deadlock
   - A simple deadlock avoidance is provided, which will prevent many cases. More algorithms can be implemented and tested easily

A more in depth description will soon be published and presented (Most likely at vsp seminar & ABMTRANS2024).


rewertvsp:
Since we have the RailSim contrib since September 2023 and there is still some development, it could be a suitable time that the developer can present this work.
Therefore, I want to point on a seminar at TU Berlin. On Feb. 14 Ihab and Merlin will present RailSim in our Seminar. This seminar has 2 hours to present and discuss. So much more time as on MUM or any conference. The seminar will be in our department, but we also stream via Zoom.
If you want to join, you can register for the seminar invitations here: [https://www.tu.berlin/en/vsp/study-teaching/seminars](https://www.tu.berlin/en/vsp/study-teaching/seminars)

PR #3083, opened 2024-01-30
feat: let DefaultPassengerEngine respond to multiple modes

sebhoerl:
We are about to prepare a PR for the `shared_mobility` contrib in which we want to reuse the `DefaultPassengerEngine`. However, we need to respond to two types of trips (owner and users in a peer-to-peer car-sharing system). For that purpose, we would like to push this PR which makes `DefaultPassengerEngine` respond to a configurable list of passenger modes with one service mode.

@michalmac Any concerns?


michalmac:
> @michalmac Any concerns?

Not really. It is hard for me to have an opinion here. I do not know `shared_mobility` and I lack the context (mainly the motivation). Specifically, I do not know the difference between `mode` and `departureMode`.

I am open to discuss it further, but I am also fine with merging the proposed changes (assuming it solves your problem).

michalmac (approved):

sebhoerl:
Hi Michal, ok, great. To give some background: We model two modes (being a car sharing user, no own car; and being a car sharing user with an own car) and all the cars in the system are handled via DVRP. This means that we want agents to drive around with cars from the "p2p" fleet when they depart either as "p2p:user" or "p2p:owner". So there are two modes that want to use one fleet (one DVRP mode).

michalmac:
> Hi Michal, ok, great. To give some background: We model two modes (being a car sharing user, no own car; and being a car sharing user with an own car) and all the cars in the system are handled via DVRP. This means that we want agents to drive around with cars from the "p2p" fleet when they depart either as "p2p:user" or "p2p:owner". So there are two modes that want to use one fleet (one DVRP mode).

@sebhoerl Thanks for the background. Looks good to me.

Issue #3082, opened 2024-01-30
Events in the queue are not ordered chronologically due to #2995


steffenaxer:
I'm facing maybe some corner case problems, probably in the last second of the simulation, in which eDrt vehicles might generate battery related events. Those events are potentially emitted during afterSim() (with time set to Infinity). Than I have a regular stuck event that says, that an other event has set the simulation time already to infinity. This problem is related to #2995.

Error msg looks as follows:
`2024-01-29 18:39:50.171,"Caused by: java.lang.RuntimeException: Events in the queue are not ordered chronologically. This should never happen. Is the SimTimeStepParallelEventsManager registered as a MobsimAfterSimStepListener? LastEventTime = Infinity currentEvent.time = 108001.0 currentEvent.type = stuckAndAbort full event: 	<event time=""108001.0"" type=""stuckAndAbort"" person=""091740000000_091620000000_13774"" link=""1290642-1290644"" legMode=""car""  />"`

steffenaxer:
Does it make sens to handle and to produce events within afterSim()?
I would personally not create any more events. One might simply clean queues or even do nothing. @michalmac

mrieser:
There should never be events with `time="Infinity"`

> Does it make sens to handle and to produce events within afterSim()?

Yes, usually `PersonStuckEvents` are generated in afterSim to report about agents not being able to finish their plan before the simulation ended.

michalmac:
- We generate these events for consistency so that they reflect things that happened during simulation.

- Why are they generated after the simulation ends? Because of race conditions, we cannot emit them immediately when "something" occurs, that's why they are delayed to the next sim step. However, this "something" occurs in the last step, so there is no next sim step and we need to emit the corresponding events after the simulation ends.

I wrote a bit more about the motivation in this PR: https://github.com/matsim-org/matsim-libs/pull/2995. At the end, I wrote that this is something I came up with, but maybe there are other/better ways of handling this type of issues.

@mrieser Is there any convention about timestamps we should use in the aferSim()? From the error message I guess the stuck events are generated with the `endTime + 1` timestamp.

michalmac:
> @mrieser Is there any convention about timestamps we should use in the aferSim()? From the error message I guess the stuck events are generated with the endTime + 1 timestamp.

I just checked the code. We always take the current time step (and the timer is not incremented after reaching `endTime`). That means the stuck events are generated with `endTime` as the timestamp and the same should be done for EV-related events.

[...]

PR #3084, opened 2024-01-30
Fix DriveDischargingHandler

steffenaxer:
This PR is a fix for #3082 and replaces the infinite time with the qsim.getSimTimer().getTimeOfDay() in order to avoid wrong event order.


contribs/ev/src/main/java/org/matsim/contrib/ev/discharging/DriveDischargingHandler.java
```diff
@@ -77,9 +78,10 @@ private boolean isOnFirstLink() {
 
 	private final Queue<LinkLeaveEvent> linkLeaveEvents = new ConcurrentLinkedQueue<>();
 	private final Queue<VehicleLeavesTrafficEvent> trafficLeaveEvents = new ConcurrentLinkedQueue<>();
-
+	private final QSim qsim;
```
michalmac, line 81: It would be nice to have a binding for the mobsim timer, so we do not need to inject the whole qsim. But that's something beyond this PR.

michalmac (approved):
Thanks for the fix. It looks good to me (see my comment here: https://github.com/matsim-org/matsim-libs/issues/3082#issuecomment-1917971869)

PR #3095, opened 2024-02-07
drt: attempt to handle waiting time explicitly when teleporting

tschlenther:
Attempting to explicity handle waiting time when teleporting drt (e.g. when using drt-speedup) by delaying the `PassengerPickedUpEvent` accodingly. The remaining problem is that `TeleportingPassengerEngine` delegates to the `DefaultTeleportationEngine` for throwing the arrival-related events like `TeleportationArrivalEvent`. `DefaultTeleportationEngine` therefore calls  `agent.getExpectedTravelTime().seconds()`, i.e. looks up the estimated travel time in the pre-computed route (if available), which does not include the waiting time (and should not??). This then leads to the fact that the `TeleportationArrivalEvent` is thrown without accounting for the waiting time. Is there any smooth way out without overriding `DefaultTeleportationEngine.handleDeparture` @michalmac  ??


michalmac:
I think your approach with extending the route travel time is fine. (Not sure if that answers your question).

PR #3096, opened 2024-02-07
feat(drt): handle DetourTimeEstimator through dependency injection

sebhoerl:
Which `DetourTimeEstimator` to use was kind of hardcoded before. This PR now includes the interface in dependency injection, so we can replace it by overriding the binding. The use case is that we want to compare DRT 1:1 to another algorithm, and we want to do this with the bare calculations without approximating travel times through the `TravelTimeMatrix`. So this PR gives us the possibility to override the default implementations of the `DetourtTimeEstimator` with exact routing code.


michalmac (approved):
LGTM 👍

PR #3104, opened 2024-02-09
feat(drt): add/remove vehicles dynamically during qsim

### LINKED ISSUE #3086: drt: dynamic vehicle fleets
In an upcoming PR we want to implement peer-to-peer car-sharing in the `shared_mobility` contrib. The idea is that vehicle owners can either share with other individuals or contribute their (automated) vehicle temporarily to a managed fleet. Technically, this means that we temporarily integrate new vehicles into a DVRP/DRT fleet. This works without a problem with DVRP and DRT: We just override a new `Fleet` implementation that allows adding and removing vehicles on the fly. Everything works surprisingly smoothly, only the analysis code makes certain assumptions that are not compatible with a dynamic fleet.

Therefore, we would like to propose flexibilizing the analysis with the following steps:
- Introduce an `AbstractFleetEvent` in DVRP with two specific implementations `VehicleAddedEvent` and `VehicleRemovedEvent`
- Modify the `analysis` packages in DVRP and DRT such that they detect these events
- Instead of having one `VehicleState` for each vehicle to track the activity, we introduce a list of states because vehicles can be active during multiple periods of the day

What do you think about that? Is this something that sounds interesting or should be find another solution? @michalmac 

There is one design decision we are unsure about. Currently, handling the fleet is managed differently in DVRP and DRT:
- DRT takes a `FleetDefinition` when initializing the analysis components and then response to events linked to those vehicle IDs
- DVRP tracks `TaskStartEvent`s and registers vehicles for analysis once they find a task with the right DVRP mode and `task index = 0`

We could replace this logic by publishing a VehicleAddedEvent by default at the beginning of the iteration and then respond to it in both cases. But this is not fully necessary, we can just treat added vehicles as a special case.

sebhoerl:
This is a draft for introducing dynamic vehicle fleets in DVRP/DRT. The PR fixes #3086.

Major changes:
- `Fleet` is now an object with its own logic. In particular, it generates generats events when vehicles are added and removed; `VehicleAddedEvent` and `VehicleRemovedEvent`.
- DVRP and DRT are responsive to a changing fleet size during the day. The DRT distance analysis has become totally independent of the fleet specification itself.
- Instead of providing a `Fleet` from a `FleetSpecification`, users now need to implement a `FleetCreator`. However, it works 99% analogously to the previous logic of creating a fleet. In most cases, this is changing two lines of code (see code changes in the unit tests). The change is necessary as `Fleet` has now become a container object to which vehicles are added at start-up or later on during the simulation.
- The code allows to flexibly move vehicles from one DVRP mode to another one (first removing it, then adding it to the other one).
- Adding vehicles from scratch is a bit more tricky, as we need to register the vehicle with (1) `Fleet`, (2) `QSim`, (3) `DynActivityEngine` and DRT vehicles also need to get an initial task assigned. For that, two helper classes have been introduced: `FleetExtensionHelper` (is a MobsimEngine), and `DrtFleetExtensionHelper` (delegates to the former).

All of this functionality is covered by `RunDrtExampleIT.testRunWithVariableFleetSize`. Here is an example of the integration test's output. Three vehicles are added dynamically between 8h00 and 12h00, thus, giving additional capacity during that period:

output_occupancy_time_profiles_StackedArea_drt


sebhoerl:
This is a first draft. There are probably certain things to clean up still.

tschlenther:
very nice. was thinking about this functionality long time ago. This enables to investigate exciting use cases, for example for goods+people DRT. Let's merge when the conflicts are solved :)

PR #3109, opened 2024-02-13
Add method to remove DisallowedNextLinks attribute from link

marecabo:
There was no method to remove a `DisallowedNextLinks` attribute from a link. Now there is.


steffenaxer:
LGTM

PR #3123, opened 2024-02-20
add method for detecting delimiter of csv file before reading it enti…

simei94:
similar to matsim python tools / calibration /analysis.py, a functionality was implemented, which checks the necessary csv files for their delimiter by reading in the first line of the file, before reading the entire file. This especially is useful for scenarios, which have different delimiter settings to the german ";".


steffenaxer:
Nice, thanks for the contribution. Would it be possible to provide a small unit test with some different separators, e.g. tab semicolon comma. Maybe this could be located in a more prominent package? E.g. CsvOptions?

simei94:
I'll add a unit test, you are right, this makes a lot of sense. @rakow what do you think about moving the newly added method to CsvOptions? I think it is a good idea

contribs/application/src/main/java/org/matsim/application/analysis/population/TripAnalysis.java
```diff
@@ -188,6 +188,45 @@ public Integer call() throws Exception {
...
+				reader = new BufferedReader(new InputStreamReader(new GZIPInputStream(new FileInputStream(path))));
+			} else if (path.endsWith(".csv")) {
+				reader = new BufferedReader(new InputStreamReader(new FileInputStream(path)));
+			} else {
+				log.error("Usupported file format.");
+			}
+
+			String firstLine = reader.readLine();
+
+			delimiterCounts.put(',', firstLine.split(",").length);
```
rakow, line 207: I think you could use a util library, something like `StringUtils.countMatches`. The current solution creates unnecessary new strings

contribs/application/src/main/java/org/matsim/application/analysis/population/TripAnalysis.java
```diff
@@ -188,6 +188,45 @@ public Integer call() throws Exception {
 		return 0;
 	}
 
+	private Character detectDelimiter(String path) throws IOException {
+		// Create a map to count occurrences of potential delimiters
+		Map<Character, Integer> delimiterCounts = new HashMap<>();
```
rakow, line 193: I would not create an extra map to store the occurrences. You can just create 3 variables for your delimiters and compare them directly.

rakow (changes requested):
Moving it to a more prominent class would be indeed a good improvement.

contribs/application/src/main/java/org/matsim/application/analysis/population/TripAnalysis.java
```diff
@@ -188,6 +188,45 @@ public Integer call() throws Exception {
...
+
+		BufferedReader reader = null;
```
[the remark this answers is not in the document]
simei94: I do not know what you mean by this. Please have a look at my newest commit and tell me if this is solved already

PR #3135, opened 2024-03-01
account for group bookings in customer stats

nkuehnel:
Since the introduction of group bookings in drt, the customer stats were a bit inconsistent.
-> the "rides" column would count every passenger
-> the "rejections" are counted per request

A proposal would be to fix the customer stats such that:
-> there is a column rides_request which counts the number of served rides at the request level (similar to rejections)
-> and an additional column rides_pax which counts the total number of served passengers
(optionally) -> groupSize_mean which is the average group size of served requests (the division of the former two variables)

Any feedback is welcome

### DIFF
--- a/contribs/drt-extensions/src/main/java/org/matsim/contrib/drt/extension/dashboards/DrtAnalysisPostProcessing.java
+++ b/contribs/drt-extensions/src/main/java/org/matsim/contrib/drt/extension/dashboards/DrtAnalysisPostProcessing.java
@@ -119,7 +119,7 @@ public final class DrtAnalysisPostProcessing implements MATSimAppCommand {
 	}
 
 	private static void printDemandKPICSV(Table customerStats, Table tableSupplyKPI, Path output) throws IOException {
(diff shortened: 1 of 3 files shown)


steffenaxer:
I like this approach. Thanks @nkuehnel for this optimization!

nkuehnel:
> The word rides_request is than maybe misleading as it does not count any ride

not sure I fully agree - can't a ride still be for multiple passengers? I think the definition of a ride meaning exactly one passenger is a MOIA/VW thing..

In any case, thinking about it we might jsut keep "rides" as before to ensure backwards compatibility, as then it doesnt change if no group bookings are modeled (in this case rides is defined at the request level). And then we could just have an additional column for rides_passengers that actually counts the number of persons.

What do you think?

steffenaxer:
I agree. Thanks for this proposal!

michalmac (approved):
LGTM

steffenaxer (approved):

PR #3139, opened 2024-03-04
guarantees sorted modes per link

### LINKED ISSUE #3134: Differently sorted modes in network.xml
I noticed that while juggling networks (e.g. recreating with pt2matsim) a network that did not change at all still has a different xml representation because the order of transit modes per link is not fixed.

So I end up with diffs like this:
```diff
< <link id="229329" from="1095288715" to="337401315" length="42.46496260420861" freespeed="8.333333333333334" capacity="600.0" permlanes="1.0" oneway="1" modes="bus,car" >
---
> <link id="229329" from="1095288715" to="337401315" length="42.46496260420861" freespeed="8.333333333333334" capacity="600.0" permlanes="1.0" oneway="1" modes="car,bus" >
```

I would love to get rid of this behavior. We could change this in `NetworkWriterHandlerImplV2` but I believe a more efficient way is to change it at the source - in `LinkImpl`: https://github.com/matsim-org/matsim-libs/blob/b78e7c7f1e6762a14e872a4fb942691d368b4771/matsim/src/main/java/org/matsim/core/network/LinkImpl.java#L284C4-L284C73 

Instead of an unsorted `Set.copyOf` I propose to use create a sorted set there, e.g. using `com.google.common.collect.ImmutableSortedSet.copyOf` (or something equivalent)

markusstraub:
fixes #3134

### DIFF
--- a/matsim/src/main/java/org/matsim/core/network/LinkImpl.java
+++ b/matsim/src/main/java/org/matsim/core/network/LinkImpl.java
@@ -36,6 +36,8 @@ import org.matsim.core.gbl.Gbl;
 import org.matsim.utils.objectattributes.attributable.Attributes;
 import org.matsim.utils.objectattributes.attributable.AttributesImpl;
 
+import com.google.common.collect.ImmutableSortedSet;
+
 /*deliberately package*/ class LinkImpl implements Link {
 
 	private final static Logger log = LogManager.getLogger(Link.class);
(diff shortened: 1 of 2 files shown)


mrieser:
I'm fascinated by the fact that no other tests needed to be adapted. I would have expected we have some tests that just check the checksum of files, and that those are now different. Well, better this way!

mrieser:
Thanks for the fix!

markusstraub:
> I'm fascinated by the fact that no other tests needed to be adapted. 

Me too, that was a pleasant surprise :)

Thanks for the fast merge!

PR #3141, opened 2024-03-06
feat(srr): transit routing by transport mode utilities

sebhoerl:
This is a relatively simple PR that adds a new option to `RaptorParameters` called `useTransportModeUtilities`. If it is set to `true` (default is `false`), SwsissRailRaptor will calculate the utility of a leg while routing based on the `transportMode` that is defined in the schedule for the respective `TransitRoute`, rather than using the generic `pt`. 

Such a behavior could already be achieved by defining "mode mappings", but this forces the user to introduce many individual leg modes, which means that one needs to deal with them in analysis. Plus, the mode mapping approach doesn't allow distinguishing between scoring generic transit trip vs. deciding which path to take while routing (which is not the same).

Another alternative would be to use `RaptorInVehicleCostCalculator`, but the interface, currently, does not pass the `TransitRoute` or anything else into `getInVehicleCost` that would allow returning a `transportMode` dependent cost. We could discuss extending the interface instead of the present PR, but I think it would have additional implications for other places such as the `OccupancyTracker`, etc.

See the new unit test for an example.


sebhoerl:
Actually, not sure anymore if it is the best solution. For now, we rather continue passing a custom RaptorParam

jfbischoff:
Hi, sorry for the lenghtly response, I was sitting at the beach 🏖️ 
I think this is very reasonable, thanks for the idea.

jfbischoff (approved):

PR #3146, opened 2024-03-07
Feature/subpopulations in annealer

jfbischoff:
This PR allows full support for subpopulations in the ReplanningAnnealer. The previous setup could silently overwrite settings. To clarify things, the subpopulation-Config parameter has been adapted. Sorry if that breaks anyone's config 💯


manserpa:
Hi @jfbischoff ,

This is cool stuff - just a question for my understanding: does this code change allow for setting annealing rates for multiple subpopulations? 
I think many MATSim users have multiple subpopulations such as low-/medium-/high-income subpopulations, and as of my understanding it was not supported to set an annealing rate for all of them.

Cheers!

jfbischoff:
Hi @manserpa 
Yes, that is what it is meant for. You can now define the same annealing parameters for each subpopulation. Previously, one could only set different annealing strategies for each subpopulation, but not the same.

PR #3156, opened 2024-03-11
fix: default vehicle type id

### LINKED ISSUE #3155: Clash of IDs in Vehicles between two successive unit tests
I'm a bit lost with this problem right now. Here is a simple set-up of two unit tests in current MATSim master. They are exactly the same, but the second one fails. The last `addVehicleType` gives

``` 
java.lang.IllegalArgumentException: Vehicle type with id = custom already exists.
	at org.matsim.vehicles.VehiclesImpl.addVehicleType(VehiclesImpl.java:115)
```

When I look at the code in debug mode, the line `VehicleType vehicleType = VehicleUtils.getDefaultVehicleType();` indeed creates the correct object, and the `Id` in `vehicleType` is correct. Then I call `addVehicleType` and I look into its member `vehicleTypes` I see that the key in the `IdMap` is *custom*.

So there seems to be some issue with the ID management and the static default vehicle type? I tried adding `Id.resetCaches()` but it doesn't seem to help.

```java
package org.matsim;

import org.junit.jupiter.api.Test;
import org.matsim.api.core.v01.Id;
import org.matsim.api.core.v01.Scenario;
import org.matsim.core.config.Config;
import org.matsim.core.config.ConfigUtils;
import org.matsim.core.scenario.ScenarioUtils;
import org.matsim.vehicles.VehicleType;
import org.matsim.vehicles.VehicleUtils;
import org.matsim.vehicles.Vehicles;

class MyTest {
	@Test
	void testA() {
		Id.resetCaches();
		
		Config config = ConfigUtils.createConfig();
		Scenario scenario = ScenarioUtils.createScenario(config);

		Vehicles vehicles = scenario.getVehicles();

		VehicleType vehicleType = VehicleUtils.getDefaultVehicleType();
		vehicles.addVehicleType(vehicleType);

		VehicleType vehicleType2 = vehicles.getFactory().createVehicleType(Id.create("custom", VehicleType.class));
		vehicles.addVehicleType(vehicleType2);
	}

	@Test
	void testB() {
		Id.resetCaches();
		
		Config config = ConfigUtils.createConfig();
		Scenario scenario = ScenarioUtils.createScenario(config);

		Vehicles vehicles = scenario.getVehicles();

		VehicleType vehicleType = VehicleUtils.getDefaultVehicleType();
		vehicles.addVehicleType(vehicleType);

		VehicleType vehicleType2 = vehicles.getFactory().createVehicleType(Id.create("custom", VehicleType.class));
		vehicles.addVehicleType(vehicleType2);
	}
}
```

sebhoerl:
Fixes #3155. Anybody sees any issues with that? Letting CI run to see if this change causes any other issues.


matsim/src/main/java/org/matsim/vehicles/VehicleUtils.java
```diff
@@ -75,7 +71,12 @@ public static Vehicles createVehiclesContainer() {
 	}
 
 	public static VehicleType getDefaultVehicleType() {
```
marecabo, line 73: Ì would suggest to rename this method to `createDefaultVehicleType()`, because the user should not expect to receive _the_ single static `VehicleType` object anymore.

sebhoerl: I agree, changed!

PR #3157, opened 2024-03-11
Make StringConverter threadsafe

steffenaxer:
This PR makes the StringConverter threadsafe. There are currently hints that indicate a potential race condition as the same AttributeConverter references are used by all worker threads within the ParallelPopulationReaderMatsimV6

### DIFF
--- a/matsim/src/main/java/org/matsim/utils/objectattributes/attributeconverters/StringConverter.java
+++ b/matsim/src/main/java/org/matsim/utils/objectattributes/attributeconverters/StringConverter.java
@@ -21,6 +21,7 @@ package org.matsim.utils.objectattributes.attributeconverters;
 
 import java.util.HashMap;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 
 import org.matsim.utils.objectattributes.AttributeConverter;
 
@@ -29,7 +30,7 @@ import org.matsim.utils.objectattributes.AttributeConverter;
  * @author mrieser
  */
 public class StringConverter implements AttributeConverter<String> {
-	private final Map<String, String> stringCache = new HashMap<String, String>(1000);
+	private final Map<String, String> stringCache = new ConcurrentHashMap<>(1000);
 	@Override
 	public String convert(String value) {
 		String s = this.stringCache.get(value);


matsim/src/main/java/org/matsim/utils/objectattributes/attributeconverters/StringConverter.java
```diff
@@ -29,7 +30,7 @@
  * @author mrieser
  */
 public class StringConverter implements AttributeConverter<String> {
-	private final Map<String, String> stringCache = new HashMap<String, String>(1000);
+	private final Map<String, String> stringCache = new ConcurrentHashMap<>(1000);
 	@Override
 	public String convert(String value) {
 		String s = this.stringCache.get(value);
```
Janekdererste, line 36: With `ConcurrentHashMap` you also have to change the insertion logic, as you might have a race condition here as well, with two processes getting a `null` value at the same time and then updating the cache. You could use 
```
return map.computeIfAbsent(key, k -> k);
```

Janekdererste:
In general, I would expect an attribute converter to be stateless. Especially here: Why is there a cache if `String convert(String value)` returns the same value?

If the state is really necessary, I would think that the workers should have their own instance of these things, instead of sharing references. I.e., taking care of parallelism and independent data structures should probably be solved on the level where parallel execution is introduced.

mrieser:
The (Concurrent)HashMap here is used for (manual) deduplication of strings. Reading attributes from XML returns different Strings, even if the string-content is always the same. Having the map ensures we store identical strings only once in memory, typically reducing memory consumption by a large margin.

Janekdererste:
> Reading attributes from XML returns different Strings

Interesting! I assumed that this was solved via interning, but of course, the XML parser allocates `new Strings(bytes)` I guess. 

Could we use something like the following?
```
var cached = value.intern();
```
Then the JVM would take care of this for us. (Just being curious)

mrieser:
At least in older versions of Java, the data structure used for the `intern()` had a fixed size, performing poorly when a large number of different Strings were interned. It was thus usually advised to rather use a ConcurrentHashMap for better performance.

E.g. see https://stackoverflow.com/questions/10624232/performance-penalty-of-string-intern, https://www.baeldung.com/java-string-pool#performance-and-optimizations (look for `-XX:StringTableSize`)

PR #3158, opened 2024-03-11
preserve possibility to run converter without shp file

simei94:
since ShpOptions.getShapeFile() no longer is a path but a string, we have to check if it is defined before creating SumoNetworkConverter to preserve the possibility to run the converter without a shp file, otherwise, when calling Path.of(shp.getShapeFile) a NullPointerException is caused


simei94:
@michalmac I do not know why you appear here. Must be some IntelliJ bug.

PR #3159, opened 2024-03-12
drt wait stats with .gpkg Geopackage

nkuehnel:


nkuehnel:
This PR is a follow-up to #3143 and changes the drt wait stats output to geopackage.
However, geopackage does not allow _not_ having a CRS set. Would it be ok to skip the creation if the global config crs is not set? I'm not sure if most people actually set it.

Another idea would be to fall back to .shp again in this case. Or to default to WGS84 or sth...

Opinions?

luchengqi7:
It also looks good to me. Does that mean the .shp & related files. will be replaced by a single .gpkg file? That would be great! And will the commonly used visualisation softwares (e.g., VIA, QGIS, SimWrapper) also recognize the new file type?

luchengqi7 (approved):

nkuehnel:
@luchengqi7 

> Does that mean the .shp & related files. will be replaced by a single .gpkg file?

Exactly, it is just one .gpkg file.

> That would be great!

It is! :D

> And will the commonly used visualisation softwares (e.g., VIA, QGIS, SimWrapper) 

I do not know about SimWrapper. However, I know that via and QGIS already support it.

PR #3161, opened 2024-03-12
Make memory observer interval configurable

nkuehnel:


nkuehnel:
This simple PR allows to configure the interval of the memory observer log statement. 
This can become handy in large simulations running for days and low "simulated vs real" time values.
Especially when running on services like AWS where the cloudstream logs are harder to stream/skim through.

PR #3167, opened 2024-03-13
Add support for H3 spatial indexing in DRT zonal systems

nkuehnel:
This PR adds support for the H3 hexagonal geospatial indexing system developed by uber
https://h3geo.org/

For now, it is only implemented for DRT zonal systems (i.e., for rebalancing zones / wait stats analyses). In the future I would like to extend it for DVRP zones (i.e., travel time matrices) and possibly other use cases as well (there is a lot of duplication of zones in matsim in drt, dvrp, accessibiilty, analysis, noise... that might could be unified at some point).

The geospatial index introduced here has some nice features:
- every real-world geo-location is referenced by a unique, reproducible id. This means that given a certain resolution, every position will always be addressed by the same hexagon. This makes comparisons across simulations easier even if network boundaries or service areas etc. change
- in contrast to square grid systems, hexagonal systems' neighboring cells all have the same distance to each other, which might reduce biases for travel time matrices / rebalancing algorithms
- the cells can be easily (dis-)aggregated to a (higher) lower resolution. This is useful for post-process analyses. It is also useful for zone systems with varying levels of resolutions, e.g. based on network or demand density

In total, there are 15 resolutions, ranging from 1280 km to 0.0005 km average edge length.

The api is well documented.


nkuehnel:
This is an overview of resolution levels

https://h3geo.org/docs/core-library/restable/
Resolution| Total number of cells | Number of hexagons | Number of pentagons
-- | -- | -- | --
0 | 122 | 110 | 12
1 | 842 | 830 | 12
2 | 5,882 | 5,870 | 12
3 | 41,162 | 41,150 | 12
4 | 288,122 | 288,110 | 12
5 | 2,016,842 | 2,016,830 | 12
6 | 14,117,882 | 14,117,870 | 12
7 | 98,825,162 | 98,825,150 | 12
8 | 691,776,122 | 691,776,110 | 12
9 | 4,842,432,842 | 4,842,432,830 | 12
10 | 33,897,029,882 | 33,897,029,870 | 12
11 | 237,279,209,162 | 237,279,209,150 | 12
12 | 1,660,954,464,122 | 1,660,954,464,110 | 12
13 | 11,626,681,248,842 | 11,626,681,248,830 | 12
14 | 81,386,768,741,882 | 81,386,768,741,870 | 12
15 | 569,707,381,193,162 | 569,707,381,193,150 | 12


Tests are included in the PR. The picture shows an example system for the Kelheim test scenario on resolution level 9.
Screenshot 2024-03-13 at 16 37 04

nkuehnel:
@michalmac I put it under drt-extensions for now. But if there are no objections we could also add it to base drt, which would reduce the code duplication of modules etc a bit. Also, I could re-use the drt zonal systems params and add a new option, including a configurable resolution.

JWJoubert:
Thanks for introducing this, @nkuehnel. H3 has been on my radar for a while now, even if it is only for DRT right now, still good to see it finding some traction.

sebhoerl:
It's quite useful, just for mapping the disadvantage is that in a Cartesian projection the hexagons are not necessarily "hexagons" I think :)

nkuehnel:
@sebhoerl 

>  in a Cartesian projection the hexagons are not necessarily "hexagons" I think :) 

Yea, a bit of distortion might be there. However, when creating the grid, the defined crs is used to convert to WGS84 and back. It should look ok in most resolutions (see plot above for EPSG:25832 at resolution 9).

> It's probably more complicated for H3 I think?

Not sure I understood correctly but finding the containing hexagon for any given point is as easy as:
```
H3Core h3 = H3Utils.getInstance();
LatLng latLng = H3GridUtils.coordToLatLng(ct.transform(link.getToNode().getCoord()));
String s = h3.latLngToCellAddress(latLng.lat, latLng.lng, resolution);
```

Finding the parent hexagon at resolution 8 is 
`String h3Id = h3.cellToParentAddress(h3Address, 8)`

sebhoerl:
Yes, I meant wrt to the computation behind `latLngToCellAddress`, I think this is quite a bit more expensive than working with Euclidean distances in a Cartesian system. Let's not discuss it here, it is a very useful addition in any case ;)

steffenaxer (commented):
LGTM! Thanks for this contribution!

contribs/drt-extensions/src/main/java/org/matsim/contrib/drt/extension/h3/drtZone/H3ZonalSystem.java
```diff
@@ -0,0 +1,75 @@
...
+ */
+public final class H3ZonalSystem {
+
+	static final Logger log = LogManager.getLogger(H3ZonalSystem.class);
+
+
+	public static DrtZonalSystem createFromPreparedGeometries(Network network,
+															  Map<String, PreparedGeometry> geometries,
+															  String crs,
+															  int resolution) {
```
michalmac, line 35: Please improve formatting here (and check other places too).

contribs/drt/src/test/java/org/matsim/contrib/drt/analysis/zonal/DrtZonalSystemTest.java
```diff
@@ -62,23 +62,23 @@ void test_gridWithinServiceArea(){
...
+		Map<String, PreparedGeometry> grid = DrtGridUtils.filterGridWithinServiceArea(DrtGridUtils.createGridFromNetwork(createNetwork(), 100), serviceArea);
 		DrtZonalSystem zonalSystem = createFromPreparedGeometries(createNetwork(),
 				grid);
 
 		assertEquals(2, zonalSystem.getZones().size());
 
 		//link 'da' is outside of the service area
 		Id<Link> id = Id.createLinkId("da");
-		assertThat(zonalSystem.getZoneForLinkId(id) == null);
+		assertThat(zonalSystem.getZoneForLinkId(id) == null).isTrue();
```
michalmac, line 73: Or `assertThat(zonalSystem.getZoneForLinkId(id)).isNull()`?

michalmac (approved):
Nice! Looks good to me.

michalmac:
> @michalmac I put it under drt-extensions for now. But if there are no objections we could also add it to base drt, which would reduce the code duplication of modules etc a bit. Also, I could re-use the drt zonal systems params and add a new option, including a configurable resolution.

@nkuehnel I think it could even go to DVRP, where the grid zone is.

BTW. Zones (both in dvrp and drt) haven't undergone any cleanups/refactorings recently. If you see that something could be improved, feel free to change it.

nkuehnel:
@michalmac makes sense to me.
I will spend some more time on this and implement it for dvrp as well.

How about we move the whole zonal stuff to the "common" contrib? From there it could be used in other places later as well (thinking about accessibility, noise and emissions again). We could do it step by step

marecabo:
> How about we move the whole zonal stuff to the "common" contrib? From there it could be used in other places later as well (thinking about accessibility, noise and emissions again). We could do it step by step

Commons would be great, as it will benefit so many features of MATSim. Looking forward to this and really appreciate your effort you are putting into this! 🚀

marecabo (approved):

contribs/drt/src/test/java/org/matsim/contrib/drt/analysis/zonal/DrtZonalSystemTest.java
nkuehnel: Done.

nkuehnel:
I updated the PR and created a zone package inside the common contrib and refactored the drt zone code to use the zones from there. The h3 resolution can now be controlled with the drt zonal systems params.

I would use this as a first step towards generalization.

Merging with dvrp zones would be a next step in a separate PR.

michalmac (commented):
Awesome! 🚀

PR #3173, opened 2024-03-18
install the drt zonalsystem module independently of the rebalancing module

nkuehnel:
the zonal system is not only used for rebalancing but also for the zonal wait stats output (and possibly other analyses in the future)


tschlenther:
we have tests failing now in downstream projects with error message saying that the ConfigGtoup DrtZonalSystemParams does not accept unknown parameters. This PR didn't have to change any config, which means there is no input test config around the ZonalSystem ?? Not so nice

tschlenther:
and this PR is not backwards-compatible - we have to change all conifgs.
Not super problematic but avoidable if we would have had test configs

nkuehnel:
@tschlenther I don't think this is related to this PR.
It is probably because of this PR https://github.com/matsim-org/matsim-libs/pull/3224
And AFAIK there are tests around zonal systems

PR #3174, opened 2024-03-18
Move method to PTWaitingTimeEstimator interface

marecabo:
This change promotes the `double estimateWaitingTime(List<? extends PlanElement> elements)` method of the `ScheduleWaitingTimeEstimator` to its `PTWaitingTimeEstimator` interface as default implementation.

It was not part of the interface `PTWaitingTimeEstimator` before despite being public and being the most convenient way to use `ScheduleWaitingTimeEstimator`.

@sebhoerl Please let me know, if that is fine for you?


sebhoerl:
The stuff in `utils` is basically just for convenience, and I'm not sure where exactly you are using it and how. In any case, the interface should stay an interface without any logic, please avoid using `default`. But I'm fine with changing the interface to `double estimateWaitingTime(List<? extends PlanElement> elements)` or adding a second method to the interface.

marecabo:
Thanks for the feedback, I'll just keep the signature then in the interface.

sebhoerl (approved):

PR #3175, opened 2024-03-18
Improve agent score comprehensibility

rakow:
In complex scenarios, it can be quite difficult to understand why agents chose a certain plan over another. We may see the total score at the end, but may not be aware what led to that score. In my opinion, a good understanding about all separate factors in the scoring function is very useful, especially when calibrating or debugging scenarios.

This PR tries to address this by adding a `explainScore(StringBuilder out)` method to the scoring interface. By implementing this method, scoring functions may write out individual components of the scoring process. 
These score explanations are attached as attributes to a plan if `ScoringConfigGroup.writeScoreExplanations` is enabled.
The implementation tries to not cause any overhead if the feature is disabled, in this case no strings are constructed.
In this PR the functionality has been implemented for the default `CharyparNagelScoring`.
Scores can be slightly different now at the last significant digits (~10e-15), because the order of floating point operations is now different.

Feedback for the implementation or the idea in general is very welcome.


jfbischoff:
I think this would be totally appropriate to merge, anything against it @rakow ?

rakow:
No, I think it would be great to merge if others can make use of it as well. Just be aware, scores change, so very restrictive tests might fail now.

jfbischoff:
Oh, indeed - now it's failing :-D

rakow:
Locally it worked fine, I restarted the test on GitHub and now it just works. The test did not seem related to the changes, but  still very strange behavior.

PR #3182, opened 2024-03-22
add triggerPattern option to dashboard header

simei94:


simei94:
the option to add a triggerPattern to the dashboard header was missing, so I added it

rakow (approved):

PR #3224, opened 2024-04-19
Consolidate dvrp/drt/taxi zones

nkuehnel:
This PR consolidates dvrp and drt zones and makes zones available in the "commons" contrib.

It's definitely a breaking change.

Zone systems can be defined through zone system params (currently defined for square grid, gis input files and h3 grid). These zone system params can be used at different places. For example, DvrpTravelTimeMatrixParams and DrtZonalSystemParams both have a pointer to the same definition of ZoneSystemParams.

**Some regression tests had to be updated, as the creation of square grid in dvrp and drt were slightly different.** 

It turned out to be more extensive then initially estimated, happy for comments and reviews.

It's now possible to have individual zone systems and sizes for dvrp travel time matrices which might come handy in large simulations with different reuirements of spatial accuracy.


sebhoerl (approved):
Hi Nico, I quickly looked through it. Looks good from my side. It's a pity that some of the integration tests needed to be updated, but it's probably even better that this is uniform now. When I have the time, I will maybe implement a proper Euclidean hexagon grid ;-)

steffenaxer:
Test it on my mac m1. It runs on my machine. The test seems to fail in PopulationUtils.comparePopulations(expected, actual) with a different result.

michalmac (approved):
Nice big refactoring!

PR #3243, opened 2024-04-30
Do not create GeometryFactory in GeometryUtils too often

marecabo:
This change reuses a static `GeometryFactory` in `GeometryUtils` which might improve performance for repeated calls to `createGeotoolsPoint` and `createGeotoolsLineString`.


Janekdererste:
We also have `MGC` which provides conversions between MATSim geotools abstractions such as `Coord` and `Point`. Possibly, this could be done there as well?

(Just a guess, I haven't checked myself)

marecabo:
Fortunately, this is already done there: 
https://github.com/matsim-org/matsim-libs/blob/ef7e4343a7805aea87dd4b19b483500348a78587/matsim/src/main/java/org/matsim/core/utils/geometry/geotools/MGC.java#L47

PR #3252, opened 2024-05-03
factor out drt optimization constraint params from drt config group 

nkuehnel:
preparatory work for service differentiation within a single service (i.e., different service classes express/wheelchair/disabled... or spatially dependent constraints)

### DIFF
--- a/contribs/commercialTrafficApplications/src/main/java/org/matsim/contrib/commercialTrafficApplications/jointDemand/examples/RunJointDemandDRTExample.java
+++ b/contribs/commercialTrafficApplications/src/main/java/org/matsim/contrib/commercialTrafficApplications/jointDemand/examples/RunJointDemandDRTExample.java
@@ -141,9 +141,9 @@ class RunJointDemandDRTExample {
 		MultiModeDrtConfigGroup multiModeDrtConfigGroup = ConfigUtils.addOrGetModule(config, MultiModeDrtConfigGroup.class);
 
         DrtConfigGroup drtCfg = new DrtConfigGroup();
-        drtCfg.maxWaitTime = 2 * 3600;
-        drtCfg.maxTravelTimeAlpha = 5;
-        drtCfg.maxTravelTimeBeta = 15 * 60;
+        drtCfg.getDrtOptimizationConstraintsParam().maxWaitTime = 2 * 3600;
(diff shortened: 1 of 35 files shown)


nkuehnel:
Another breaking change, but highly relevant for us

michalmac (commented):
Some high-level feedback:
1. I agree that the drt config is a bit messy now and requires some restructuring, the proposed grouping goes in that direction.
2. It would be nice if the drt config had a structure that enables easy plugging in of different optimisers (so something more than this PR proposes)
3. Because breaking the backward compatibility at the xml level may be quite painful for the users, I would consider maybe drafting "drt config v2" with a nice pluggability of different optimizers, while still keeping the current drt config backward compatible.

nkuehnel:
Based on https://github.com/matsim-org/matsim-libs/pull/3027 I added a method to handle old config group files by using the handleUnknownParam method. There's also a test included which runs a test on the older version of mielec_drt_config_v1.xml


> It would be nice if the drt config had a structure that enables easy plugging in of different optimisers (so something more than this PR proposes)
> Because breaking the backward compatibility at the xml level may be quite painful for the users, I would consider maybe drafting "drt config v2" with a nice pluggability of different optimizers, while still keeping the current drt config backward compatible.

Not voting against it, but it would require some more effort.. let me know if this backwards compatibility works as well for now.

 ( @marecabo @steffenaxer fyi)

michalmac:
> Based on #3027 I added a method to handle old config group files by using the handleUnknownParam method. There's also a test included which runs a test on the older version of mielec_drt_config_v1.xml

Thanks! As long as this PR is backward compatible, I do not have any objections. BTW. I will not be able to review this PR by the end of next week, so please do not wait for my approval.

michalmac:
> > It would be nice if the drt config had a structure that enables easy plugging in of different optimisers (so something more than this PR proposes)
> > Because breaking the backward compatibility at the xml level may be quite painful for the users, I would consider maybe drafting "drt config v2" with a nice pluggability of different optimizers, while still keeping the current drt config backward compatible.
> 
> Not voting against it, but it would require some more effort.. let me know if this backwards compatibility works as well for now.

Yeah, that was meant as food for thought, but not a blocker for this PR.

nkuehnel:
Alright, thank you

PR #3254, opened 2024-05-06
Noise dashboard

frievoe97:
- Adds a default noise dashboard
- stores the noise data in avro files

Bildschirmfoto 2024-05-06 um 11 11 16


contribs/simwrapper/src/main/java/org/matsim/simwrapper/viz/MapPlot.java
```diff
@@ -14,6 +14,8 @@ public final class MapPlot extends Viz {
 	public Double zoom;
 
 	public Display display = new Display();
+	public double minValue;
```
rakow, line 17: Need to be `Double`

PR #3258, opened 2024-05-07
fix: personMoney event reference lookup missing in EventsReader

mfrawley-moia:
We (Moia) discovered that the reference attribute for personMoney events was always null when parsed by the EventsReaderv1 class. I added a hopefully-appropriate unit test for it in EventsReadersTest, and adjusted the events.xml fixture accordingly. 

Feedback is more than welcome!  cc @nkuehnel  @fzwick


nkuehnel:
Thanks Mark, looks good to me : )

nkuehnel (approved):

PR #3275, opened 2024-05-16
Allow for storing multiple DrtOptimizationConstraintsParams

marecabo:
This is a follow-up to #3252 from @nkuehnel and me which allows to store multiple `DrtOptimizationConstraintsParams` in `DrtConfigGroup` while ensuring, there is always a default `DrtOptimizationConstraintsParams` present.


michalmac (approved):
Conceptually, LGTM as this is a continuation of another PR. (I haven't gone through the code thoroughly)

PR #3283, opened 2024-05-23
railsim: Test Scenario for a construction site between Lausanne and Geneva

NicDul:
This PR adds a further integration test case for railsim

**Network**
Network based on the infrastructure between Lausanne and Geneva, only the stations needed for the long-distance trains have been modeled based on https://network.sbb.ch/en/
All links are mesoscopic, with tracks capacities following the reality. 
The portion of the line where the construction site happens is modelled in a microscopic manner between Morges (MOR) and Renens (REN).

**Schedule**
The schedule is built on the typical hour schedule available here : https://network.sbb.ch/en/ repeating itself every hour between 8am and 12pm (noon). 

**Construction site**
In order to model the capacity restriction on one track of the construction zone, a construction train is used, staying at at "construction Stop" for the duration wanted. In this case a construction train leaves Morges (MOR) at 10am and blocks one link (MOR2-REN1-1) for one hour, and then drives to Renens (REN).

This blocks the track and the trains therefore need to reroute if needed. The needed links in MOR and REN have been put to Entry and Exit to allow the rerouting. 

@rakow @ikaddoura @munterfi

### DIFF
--- a/contribs/railsim/src/test/java/ch/sbb/matsim/contrib/railsim/integration/RailsimIntegrationTest.java
+++ b/contribs/railsim/src/test/java/ch/sbb/matsim/contrib/railsim/integration/RailsimIntegrationTest.java
@@ -373,9 +373,9 @@ public class RailsimIntegrationTest {
 			type.setMaximumVelocity(30);
 			type.setLength(100);
 		}
-		
+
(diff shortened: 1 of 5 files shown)


munterfi:
Perfect, thanks! 🚀

PR #3307, opened 2024-06-04
Performance improvements for PrebookingStopActivity

marecabo:
This change improves the performance of `PrebookingStopActivity` a bit.
* Use PriorityQueue for `PrebookingStopActivity.leaveTimes`.

### DIFF
--- a/contribs/drt/src/main/java/org/matsim/contrib/drt/prebooking/PrebookingStopActivity.java
+++ b/contribs/drt/src/main/java/org/matsim/contrib/drt/prebooking/PrebookingStopActivity.java
@@ -3,6 +3,8 @@ package org.matsim.contrib.drt.prebooking;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.PriorityQueue;
+import java.util.Queue;
 import java.util.Set;
 import java.util.function.Supplier;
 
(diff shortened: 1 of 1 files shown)


contribs/drt/src/main/java/org/matsim/contrib/drt/prebooking/PrebookingStopActivity.java
```diff
@@ -84,27 +86,30 @@ protected void beforeFirstStep(double now) {
...
-			var entry = iterator.next();
-
-			if (entry.getValue() <= now) { // Request should leave now
-				passengerHandler.dropOffPassengers(driver, entry.getKey(), now);
-				prebookingManager.notifyDropoff(entry.getKey());
-				onboard -= dropoffRequests.get(entry.getKey()).getPassengerCount();
-				iterator.remove();
-			}
+		while (!leaveTimes.isEmpty() && leaveTimes.peek().time <= now) {
+			RequestIdTime rit = leaveTimes.poll();
```
sebhoerl, line 98: Actually, you don't even need `rit`, you can just save the `id` as `leaveTimes.poll().id`

contribs/drt/src/main/java/org/matsim/contrib/drt/prebooking/PrebookingStopActivity.java
```diff
@@ -84,27 +86,30 @@ protected void beforeFirstStep(double now) {
...
+			passengerHandler.dropOffPassengers(driver, rit.id, now);
+			prebookingManager.notifyDropoff(rit.id);
+			onboard -= dropoffRequests.get(rit.id).getPassengerCount();
 		}
 
-		return leaveTimes.size() == 0;
+		return leaveTimes.isEmpty();
+	}
+
+	private record RequestIdTime(Id<Request> id, double time) implements Comparable<RequestIdTime> {
```
sebhoerl, line 107: I would call it something like `QueuedRequest`

sebhoerl (changes requested):

marecabo:
@sebhoerl Thanks for the comments, just addressed them.

sebhoerl (approved):

sebhoerl:
Great, thanks!

PR #3311, opened 2024-06-11
Configurable DRT constraints

nkuehnel:
this is the final change towards more flexible drt optimization constraints - the drt route creator uses a constraint set chooser and constraints calculator to create and update routes.

There are certainly other ways to achieve this, I'm happy for any discussion / feedback.

Use case:
- spatio-temporal differences in constraints
- service classes / mixed fleets (express, wheelchair, ...)
- non-linear constraint calculations


nkuehnel:
I had to borrow the concept from the multimodedrtconfiggroup which allow defiing a custom supplier for drt config groups for materializing / reading back in configs. 
The supplier is required for DrtConstraintsSets inside DrtConstraintsParams and needs to be passed to the DrtConfigGroup which is why I added an additional constructor.

It works well but is another change to the drt config group.

nkuehnel:
Ah and there's a test added that shows how to do it:
https://github.com/matsim-org/matsim-libs/blob/57fa8aa5a414519dad374b1e9650b927d3ffb0fc/contribs/drt/src/test/java/org/matsim/contrib/drt/config/ConfigBehaviorTest.java#L92-L167

luchengqi7 (approved):

steffenaxer (approved):
LGTM. Thanks Nico for this very valuable contribution.

marecabo:
LGTM!

paulheinr:
During the MATSim class we (@kainagel) noticed that config switch. Now you need to write

```java
DrtConfigGroup drtCfg = new DrtConfigGroup();
DefaultDrtOptimizationConstraintsSet defaultConstraintsSet =
        (DefaultDrtOptimizationConstraintsSet) drtCfg.addOrGetDrtOptimizationConstraintsParams()
                .addOrGetDefaultDrtOptimizationConstraintsSet();
```
instead of 
```java
drtCfg.getDefaultDrtOptimizationConstraintsParam();
```
I think in general it would be good, if we first deprecate these methods first and delete them after a year or so. Right now, switching to another version immediately breaks code. 

Moreover, I think it's a bit odd, that `addOrGetDefaultDrtOptimizationConstraintsSet()` doesn't return the type `DefaultDrtOptimizationConstraintsSet` but needs to be casted.

paulheinr:
Answer Paul (via mail nov 7th, 2024)

> The possibility to have multiple constraint sets for a single service makes sense. After your explanation and reading the test case org.matsim.contrib.drt.config.ConfigBehaviorTest#testMaterializeAfterReadCustomDrtConstraints I think I now understand your interpretation of the addOrGetDefaultSet-Method.
>  
> I think the issue is, that “default” now has to meanings:
> In DefaultDrtOptimizationConstraintsSet it refers to the „default attribute types” a constraint set could have (like “maxTravelTimeAlpha” or “maxTravelTimeBeta”)
> For the anonymous ConstraintSetChooser in org.matsim.contrib.drt.run.DrtModeRoutingModule#install, default refers to the constraint set which has the name “default”
>  
> So e.g., one could have two constraint sets: (i) of type “DefaultConstraintSet” and (ii) of type “CustomConstraintSet”, but (i) should be used in specific cases and (ii) in general, thus “default”, cases.  
>  
> The current implementation also allows to bind DrtRouteConstraintCalculators to classes that can’t handle a specific type of DrtOptimizationConstraintsSet. But shouldn’t the type of the constraint set determine the calculator?
>  
> We actually had the same type of problem not long ago with PT fares. Our use case was the following: within a certain region, the local PT fare should be used, out of this region (NRW), the NRW fare should be used and out of NRW the Deutschlandtarif should be used.
>  
> We did it this way:  
> * An abstract class org.matsim.contrib.vsp.pt.fare.PtFareParams, which has two child classes. They are the equivalent of your DrtOptimizationConstraintsSet. But they don’t have a name, but an order.
> * In org.matsim.contrib.vsp.pt.fare.PtFareModule, for each PtFareParams, we instantiate a corresponding calculator (depending on the type) and chain them based on the order.
> * The calculation happens in org.matsim.contrib.vsp.pt.fare.ChainedPtFareCalculator. This class goes through all other calculators and as soon as it gets a result, it returns it.
>  
> Of course, this could have been done differently. In particular, each PtFareParam knows the region to which it is to be applied. I see, that this may not be what we want with the DrtConstraintSets. But one could define these things outside or make it optional inside.
>  
> If the Drt Config would follow this architecture, one wouldn’t need the definition (2) of “default” and thus could return type DefaultDrtOptimizationConstraintsSet for the “addOrGetDefaultConstraintSet” function.
>  
> I don’t know if this is the right place for the discussion. I will copy this text to github – feel free, to answer there or here, if you have thoughts on this.

nkuehnel:
Thanks for your reply, just one thought - there culd also be multiple calculators that work with the same constraint set types, right? In that regard one has to always _explicitly_ glue everything together correctly?

paulheinr:
No, right now, there is exactly one calculator implementation per parameter type: 
```java
if (p instanceof FareZoneBasedPtFareParams fareZoneBasedPtFareParams) {
  ptFareCalculator.addBinding().toInstance(new FareZoneBasedPtFareCalculator(fareZoneBasedPtFareParams, context));
} else if (p instanceof DistanceBasedPtFareParams distanceBasedPtFareParams) {
  ptFareCalculator.addBinding().toInstance(new DistanceBasedPtFareCalculator(distanceBasedPtFareParams, context));
} else {
  throw new RuntimeException("Unknown PtFareParams: " + p.getClass());
}
```

nkuehnel:
But there _could_ be additional calculators that operate on the same constraints, right? Also does it work in case you dont know either the constraints or the calculator implementation yet? Lets say we have our MOIA calculator that works on the traditional constraint sets but we cannot open source it?

kainagel:
I am sorry but I haven't followed the technical details of the discussion.  It is addressed as follows in core matsim (examples are timeMutation, changeTripMode, subtourModeChoice):
* There is a place where one can recruit some functionality (e.g., use timeMutation as a strategy)
* There is some other config module (see `TimeAllocationMutatorConfigGroup` where that function can be configured.
From a teaching perspective, there is a lot to be said for always using the same pattern, even if other solutions are possible.  Since as long as it is the same pattern, I have to teach it only once.  For every new pattern, I need a new teaching unit.

PR #3315, opened 2024-06-13
Debug noise analysis

tschlenther:
`NoiseAnalysis` matched `vehicles` from the run directory. In case of runs with dvrp, there are multiple of files matching that pattern. We need to make sure that the one, complete `output_vehicles` file is read.


kt86:
@tschlenther I just realize this now, because of the current related discussions: There is a `output_allVehicles.xml.gz`, which should contain all vehicles and their types, once this vehicle was used in the mobsim....
We had introduced this a few years ago, because we had similar issues, with freight vehicles not showing up in the regular "output_vehicles" file... With that "fix" we were able to do the emissions analysis (and some stuff more) 
Would this also help you for your noise analysis?

PR #3333, opened 2024-06-24
Update DRT estimate and teleport module

luchengqi7:
The new DRT estimate and teleport module (a continuation based on PR #3160) will generate a trip estimation (estimated waiting time, travel time and the probability of being served) for each DRT request. The DRT trips will then be teleported during the MobSim. A more detailed documentation will follow shortly. Some reformatting will be performed before this PR is merged into master branch. Ideas and comments are welcomed!!!

A brief intro to this implementation is provided here: 
There are several advantages of doing this:
1. Speed up the procedure
2. Enhancing the reliability of the results
3. Easier for transport planning
4. Being able to match real-world operational data

**Current implementation**
1. Interface for typical waiting time estimator
   a. Constant waiting time estimator
   b. Shape-file based waiting time estimator

2. Interface for typical ride duration estimator
   a. Network based ride duration estimator
   b. Euclidean distance based ride duration estimator

3. Distribution generator
   a. Log-normal distribution generator
   b. Normal distribution generator 
   c. No distribution (currently can be set via normal distribution with std = 0)

4. Probability of being accepted
  a. Uniform value
  b. shape-file-based (to-do)

**How to use:**  
To build a DRT estimator, we need to choose a waiting time estimator (1) and a ride duration estimator (2). For each of the two estimators, we can match it with an independent distribution model (3). Finally, we choose an estimation model for the probability of being accepted. 

The parameters for the estimators and the distribution models can be fitted according to real-world operational data or systematic offline DRT simulations (i.e., based on fix demands). Other simulation tools or optimizers can also be used for this purpose.


tschlenther:
Overall, great stuff! Thank you! I am looking forward to try this. 

Here are a few question that came into my while quickly flying over the code mind. 
Will need to dive deeper with more time.
Happy to discuss them in person, if you prefer.

1. Is this compatble with `SubtourModeChoice` at the time being (or only informed mode choice)? How does the former consider `Estimate`s? Does that work via `DrtSpeedUp`?

1. How is the rejection rate handled, later when it is not 1? In informed mode choice, I can imagine for it to be accounted in the expected score. With SMC, however, do we set waitingTime / ride duration to very high values for every Xth trip request??

3. Could you maybe provide a small RunExample which demonstrates how to set this up? That would probably be similiar to `MultiModalDrtLegEstimatorTest` - btw tests would also be very nice.

luchengqi7:
> Overall, great stuff! Thank you! I am looking forward to try this.
> 
> Here are a few question that came into my while quickly flying over the code mind. Will need to dive deeper with more time. Happy to discuss them in person, if you prefer.
> 
> 1. Is this compatble with `SubtourModeChoice` at the time being (or only informed mode choice)? How does the former consider `Estimate`s? Does that work via `DrtSpeedUp`?
> 2. How is the rejection rate handled, later when it is not 1? In informed mode choice, I can imagine for it to be accounted in the expected score. With SMC, however, do we set waitingTime / ride duration to very high values for every Xth trip request??
> 3. Could you maybe provide a small RunExample which demonstrates how to set this up? That would probably be similiar to `MultiModalDrtLegEstimatorTest` - btw tests would also be very nice.

Re: 1. The Estimate is the same one used for the informed mode choice module. So, it should be compatible with the new sub-tour mode choice. And this new DRT estimate module is designed to replace the DrtSpeedUp in the long run, as this is more systematic and more flexible. The replacement of DrtSpeedUp will be in a future PR. 

Re: 2. The rejection will be handled in the standard way, same as in the full simulation of DRT (i.e., either aborted or teleported based on walk speed). A custom rejection handling approach can also be coupled to the DRT estimate.  

Re: 3. Sure. This will be added shortly.

PR #3340, opened 2024-06-27
add router binding for complex request unscheduling

nkuehnel:
Hey @sebhoerl  I was playing around with the prebooking module - not sure if this was really missing or it was supposed to be bound elsewhere but when choosing the complex request unscheduler in prebooking there was a (modal) binding missing for the LeastCostPathCalculator

This also crashed when choosing this unscheduler in the matsim-code-example


sebhoerl:
Hi @nkuehnel great, good catch, I guess we were testing it in some other set-up. The unit tests pass, but I think the complex unscheduler is never tested in an actual IT. If you are motivated, maybe you could modify for instance `AbandonAndCancelTest::cancelLateTest` to have a `cancelLateWithComplexUnschedulerTest`? :)

sebhoerl (approved):

nkuehnel:
@sebhoerl please check if I got the test right.

I had to adapt the unscheduler because it was crashing wenn the cancelled pickup is the start waypoint of the vehicle entry (as it is the case in the test)

sebhoerl:
LGMT, thanks a lot :)

PR #3348, opened 2024-06-28
Handle mode restriction correctly

paulheinr:
If you restrict modes on links, there are in general two problems: 
1. The network can then have local dead ends, i.e. the network is not connected any more for the mode.
2. The initial plans may have routes that use the restricted links, thus agents still use the links. 

Both should be tackled with this PR. Therefore, several changes are needed. 

### New Config Parameter
In the Routing Config, there is the parameter `networkRouteConsistencyCheck`. If used with `abortOnInconsistency`, in `NetworkRoutingProvider` the network is checked against problem (1) and in `PersonPrepareForSim` the routes of the agents are checked against problem (2). 

### New Utils
For (1), there is now  the class `NetworkModeRestriction`, which applies mode restriction and cleans the network. There is a corresponding function in `NetworkUtils`

For (2), there is now the class `PersonRouteCheck`, which deletes routes, if they use links which do not allow the mode any more. There is a corresponding function in `PopulationUtils.`

### To discuss: 
- [x] The default value of the config is `abortOnInconsistency`. This is quite restrictive, but helpful. Unfortunately, lots of tests are broken because of that :-( -> corrected all tests in the contribs. I will continue with the core, if this approach is accepted.
- [x] Check the result of `ModeRestrictionTest` against whole events file and not against first leg. Need to comment the result very well then.


paulheinr:
@kainagel @rakow what do you think about this approach?

matsim/src/main/java/org/matsim/core/network/NetworkUtils.java
```diff
@@ -1018,8 +1019,30 @@ public static boolean addDisallowedNextLinks(Link link, String mode, List<Id<Lin
...
+		link.setAllowedModes(modes);
+	}
+
+	/**
+	 * Removes the given modes from the links and runs the network cleaner afterwards. Thus, some more links may be restricted to keep the network consistent.
+	 * That means, each link can be reached from each other link.
+	 * @param network the network
+	 * @param modesToRemoveByLinkId map of modes that should be removed from the links
+	 */
+	public static void restrictModesAndCleanNetwork(Network network, Map<Id<Link>, Set<String>> modesToRemoveByLinkId) {
```
rakow, line 1045: Using a lambda Function<Id<Link>, Set<String> here is probably a lot more flexible.

paulheinr: Yes, good idea!

PR #3379, opened 2024-07-19
Make DRT shifts compatible with prebooking, drt shift maintenance

nkuehnel:


nkuehnel:
WaitForShiftStayTask has been promoted to WaitForShiftTask which is a stop base task now (instead of stay task).
Requests may be allowed to be isnerted in upcoming shifts if they are already assigned to a vehicle and the vehicle is waiting at the hub. (For this set "considerUpcomingShifts" in ShiftParams to "true" and set the shiftScheduleLookAhead to a larger value than the prebooking slack times.)

Test with shifts + prebooking is included.

Requests may not be inserted if they would preclude a meaningful shift break for unscheduled shift breaks (e.g. if a prebooking would slice the break corridor in half and neither half would be long enough for the complete shift break).

Added some documentation for drt operations.

Fixed some minor bugs and cleaned up some code

PR #3382, opened 2024-07-24
add fare zone based pt pricing

simei94:
This is the result of a calibration task force meeting, where we agreed to this distance  + fare zone based pricing scheme.


simei94:
@rakow please have a look here

simei94:
@GregorRyb please check out this branch and test if it works for your scenario

PR #3385, opened 2024-07-29
make drt work with electric vehicles, shifts and prebooking

nkuehnel:
This setup _should_ now work with electirc vehicles, shifts and prebooking

Lesson learned after hours of debugging:
priority queues do not iterate in the order as one may hope for T__T


sebhoerl:
> priority queues do not iterate in the order as one may hope for T__T

We also figured this out recently :D

PR #3398, opened 2024-08-07
NoiseAnalysis: set the scale factor for according to sample size

tschlenther:


tschlenther:
all of this would need proper testing...

PR #3404, opened 2024-08-09
DRT: add option to reject found insertions (e.g. by customer)

nkuehnel:


nkuehnel:
We already had the DrtOfferAcceptor that mainly used to narrow down constraints for accepted requests.
With this small modification, implementations of the interface can also reject an offer y returning an Optional.empty(). In this case a rejection event is thrown with an "offer_rejected" cause.

In the future, more information such as price, waiting time or detour could directly be passed through the interface, in addition multiple offers could be offered to the customer, from which the customer can select one or none.

steffenaxer (approved):

steffenaxer:
LGTM thanks 😊 Nico for this improvement. Very helpful for fleet simulations!

sebhoerl (approved):

PR #3405, opened 2024-08-11
fix DRT: insertion scheduler bug in prebooking scenarios

nkuehnel:


nkuehnel:
TL;DR: as my drum teacher used to say: "timing is everything"


Fixes a rare bug in prebooking scenarios in the following case:

Assume a vehicle is currently relocating (drive task) and already has a prebooked request scheduled for pickup, which requires a wait/stay task before PU, i.e. assume following schedule:

1 drive task RELOCATE (current task)
2 stay wait task
3 drive task to prebooked PU
4 stop task for prebooked PU
5 drive task to prebooked DO
6 stop task for prebooked DO

A new adhoc request insertion is found with a PU before the prebooked PU (so between task 1 and 4 above) and a DO after the prebooked PU (after task 4). 

Now the default request insertion scheduler will schedule PU and DO tasks. Since the current task is a drive task, it is diverted to the new PU. After that a new PU stop is inserted, followed by a new wait and drive task for the existing prebooked PU. I guess the problem is that the scheduler so far "thinks" that a currently ongoing drive task with an existing future stop _must_ be the direct drive to that stop, n which case a diversion would be enough. This is not the case here, as it is a relocation followed by a wait for a future prebooked stop. So after the insertion, the schedule looks like this:

1 drive task RELOCATE (current task, diverted to new adhoc PU)
2 stop task (*new PU)
3 stay wait task task (*new wait before drive)
4 drive task to prebooked PU (*new, from PU 2)
5 stay wait task task (old task 2 above)
6 drive task to prebooked PU (old task 3 above)
7 stop task for prebooked PU (old task 4 above)
8 drive task to new adhoc DO (*new)
9 stop task for new adhoc DO (*new)
...

At some point in between insertions there is a call to scheduleTimingUpdater.updateTimings..(..), in which the old stay task (now task 5) gets cancelled out as its start time is before the end time of the new drive task, resulting in:

1 drive task RELOCATE (current task, diverted to new adhoc PU)
2 stop task (*new PU)
3 stay wait task task (*new wait before drive)
4 drive task to prebooked PU (*new, from PU 2)
5 drive task to prebooked PU (old task 3 above)
6 stop task for prebooked PU (old task 4 above)
7 drive task to new adhoc DO (*new)
8 stop task for new adhoc DO (*new)
...

Now there are still two consecutive drive tasks (4 and 5) which causes the verifyStructure(...) method to throw an exception.

The fix basically just removes the old wait and drive tasks in between.

sebhoerl:
Thanks :)

PR #3411, opened 2024-08-16
Add chained pt fare handler

paulheinr:
This PR changes the PT fare configuration logic. Everything what was possible before is still possible. Additionally, now you can chain certain pt fare calculators.

Right now, there are two types of calculators: 
- `DistanceBasedPtFareCalculator`
- `FareZoneBasedPtFareCalculator`

While the first one only takes the distance into account, the second one looks whether start and end are in the same fare zone based on the given shape file. Distance based pt fare can also be conditioned by a shape file.

You can chain any number of such calculators and configure them independently from each other. The calculators must be strictly ordered by the `order` parameter. As soon as a calculator is able to calculate a fare, this is taken. If no calculator can calculate a fare, an exception is thrown.

What still needs to be done:  
- [ ] adapt informed mode choice module
- [x] take fare zone of stops rather than stop coordinates (this was also not the case before, so it would change the behaviour) -> moved to #3412


paulheinr:
An example config now looks like this: 
```xml
<module name="ptFare" >
	<param name="applyUpperBound" value="true" />
	<param name="upperBoundFactor" value="1.5" />
	<parameterset type="ptFareCalculationDistanceBased" >
		<param name="order" value="10" />
	</parameterset>
	<parameterset type="ptFareCalculationFareZoneBased" >
		<param name="order" value="5" />
	</parameterset>
</module>
```

paulheinr:
@simei94 fyi

vsp-gleich:
would be nice to get the parametersets sorted by order in the output config file

PR #3417, opened 2024-08-19
tourEndTime always > tourStartTime

simei94:
tourStartTime is always set to 0, which can lead to the end time of a tour being earlier than the start time of the tour... this is fixed now


rewertvsp:
Thanks @simei94 for finding the bug with the plan generation after I changed someting one month ago.
Now I did fixed this and also did this:

- fix Bug with commercial plan generation (Tour start time, activity durations and tour end time) and adjust test
- remove static 
- move tour characteristic data to separate class
- use only EnumeratedDistribution instead of own implementation

PR #3422, opened 2024-08-20
Drt: fix prebooking insertion bug

nkuehnel:
fixes another rare bug in prebooking scenarios:

given a vehicle is assigned a prebooked request such that the following schedule emerges (with wait before drive set to true):

1 Relocate (diverted)
2 Stay
3 Drive
4 Pickup
5 Drive
6 Dropoff

In congested scenarios and in cases where the task 2 Stay above is short it may happen that the task 1 relocate ends after the initial stay end time due to delays. In this case the schedule updater will at one point remove the task 2 Stay leading to:

1 Relocate (diverted)
2 Drive
3 Pickup
4 Drive
5 Dropoff

When the next insertion is checked (can be completely unrelated to the existing one) the verifyStructure in the insertion scheduler will throw an exception (similar to https://github.com/matsim-org/matsim-libs/pull/3405).

This PR adjusts the verifyStructure method to explicitly check whether two subsequent drive tasks actually connect. In addition, the first of the two tasks has to be a diverted task. In this case the above schedule is valid.


sebhoerl (approved):

sebhoerl:
thanks

PR #3423, opened 2024-08-20
Adding geometry-free zone-system

luchengqi7:
Sometimes geometry-based zonal system cannot handle the irregularity in the network (e.g., rivers/bridges, one-way-roads, ramps). A geometry-free zonal system may be desirable. 

In this PR, an interface for the generation of a geometry-free zonal system is implemented. The zone-system will be built according to the special attribute stored in the network (node and link). 

A network preparation script based on a simple heuristic has been implemented (MaxTravelTimeBasedZoneGenerator). A run script to use the generator is also added (PrepareMaxTravelTimeBasedZonalSystem). These two scripts are in the application-contrib.


nkuehnel (approved):
LGTM,nice :)

PR #3425, opened 2024-08-22
Drt shifts update

nkuehnel:


nkuehnel:
This PR introduces a ShiftScheduler interface. This scheduler is meant to take over all the house keeping of shifts across and within iterations. Instead of installing DrtShiftsSpecification once, the shift scheduler may
1. change the specifications across iterations
2. come up with new shifts within an iteration

the scheduler is polled once initially per iteration and then in each time step by the shift dispatcher. 
Currently aimed to schedule operational hours for AV on the fly

steffenaxer (approved):
LGTM.

PR #3431, opened 2024-08-28
DRT: fix insertion scheduler bug

nkuehnel:


nkuehnel:
Fixes a rare bug in the insertion scheduler when prebooking params are present and scheduleWaitBeforeDrive is true.

Essentially it appears when, suddenly, an existing stop is reached after inserting a new request before.

This can happen if, e.g., the pickup of the new request can be merged with another existing stop (pickup time loss = 0) and the travel time to the existing stop suddenly becomes shorter (e.g., through network change events) AND/OR very coarse travel time matrices are used in which predicted travel times may behave more inconsistent.

With this PR, the existing stop will not stay fixed, but may be reached earlier if there is not earliestStart time of a pickup.
In addition, if scheduleWaitBeforeDrive is true, then there cannot be an additional stay task _after_ the drive.

The PR includes a first test for the scheduler which can be extended in the future.

PR #3444, opened 2024-09-03
Fix a (severe) bug in MSA scoring

tthunig:
Before the bugfix of this branch, MSA was not active, even if it was switched on via the config. The reason for this is that the NewScoreAssigner was recreated every iteration and therefore could not save any information across iterations. There was a test for MSA, but it did not create a complete controller, but only tested the scoring via EventsToScore individually, where the NewScoreAssigner was bound differently (correctly).

With the changes to this branch there is 1. a test that tests MSA scoring on a standard controller run (this test fails without the changes to this branch) and 2. the NewScoreAssigner is now bound as a singleton in the PlansScoringModule and is no longer recreated in PlansScoringImpl.

This change will probably affect a lot of your runs!

### DIFF
--- a/matsim/src/main/java/org/matsim/core/scoring/PlansScoringImpl.java
+++ b/matsim/src/main/java/org/matsim/core/scoring/PlansScoringImpl.java
@@ -52,11 +52,11 @@ final class PlansScoringImpl implements PlansScoring, ScoringListener, Iteration
 	@Inject private OutputDirectoryHierarchy controlerIO;
 	@Inject private ScoringFunctionsForPopulation scoringFunctionsForPopulation;
 	@Inject private ExperiencedPlansService experiencedPlansService;
+	@Inject private NewScoreAssigner newScoreAssigner;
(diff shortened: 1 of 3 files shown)


kainagel (approved):
looks good to me.  Thanks for writing a new test for this.

neuma:
We are not using MSA in our runs but activated it anyway for our larger regression test[^1]. We did not find any significant changes in scores, mode shares, and travel distances, i.e. the largest change is a 0.09% drop in the pt share and hence an 8m increase in the average leg distance.

However, the regression test might not be susceptible to MSA since it does not contain drt nor is there any capacity restrictions for public transit vehicles. Thus, there is less room for flip-flop situations and plans either work or consistently fail.

[^1]: Setup: 21k agents, 10pct real-world scenario with modes car/pt/bike/walk/ride, 280k links network, no capacity restriction for pt, ensemble runs with each scenario run 10 times with different random seeds

PR #3447, opened 2024-09-03
update prebooking stop activity

nkuehnel:
includes a performance related update by @sebhoerl proposed here:
https://github.com/matsim-org/matsim-libs/tree/feat/improve-prebooking-stop-activity

in addition, it adds a check to notifyPassengersAreReadyForDeparture in the prebooking stop activity such that only pickups are queued if they are still contained in expectedPickups. Otherwise it can happen that a request is abandoned in the same second as it is picked up, throwing an exception:

1. a request is part of a prebooking stop activity and registered in the expectedPickups map.
2. the vehicle arrives before the agent
3. the vehicle waits until the abandon voter votes for abandonment (e.g., after 10 mins) at time t, the request is removed from the expected pickups map and registers the abandonment at the prebookingmanager (no immediate action)
4. the DefaultPassenger engine handles a departure event for the expected agent also at time t
5. the prebooking stop activity gets notified about the departure and picks up the passenger, setting the RequestItem in the prebookingmanager to onBoard=true
6. at the end of the time step, the prebooking manager goes through all registered abandonments (from step 3 above) and tries to actually abandon the request. At this time the Verify(onBord!=true) fails with an exception as the pickup already happened.

With this fix, step 5 is extended to check whether the request is still contained in expected pickups, from which the agent should be removed in step 1, if abandonment happens in the same second.

### DIFF
--- a/contribs/drt/src/main/java/org/matsim/contrib/drt/prebooking/PrebookingStopActivity.java
+++ b/contribs/drt/src/main/java/org/matsim/contrib/drt/prebooking/PrebookingStopActivity.java
@@ -1,15 +1,15 @@
 package org.matsim.contrib.drt.prebooking;
 
-import java.util.HashSet;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.PriorityQueue;
 import java.util.Queue;
-import java.util.Set;
 import java.util.function.Supplier;
 
 import org.matsim.api.core.v01.Id;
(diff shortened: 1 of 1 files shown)


mfrawley-moia (approved):

sebhoerl:
I just wanted to have a look, but I think something just went wrong with the last push ;)

nkuehnel:
oopsie, @sebhoerl please check again : )

sebhoerl (approved):

PR #3459, opened 2024-09-09
Revert to old pseudo pt network creator

rakow:
@davibicudo @jfbischoff: The new pt pseudo network creator with loop-links is creating invalid schedules according to the TransitScheduleValidator. Especially, the network routes seem to be incorrect now.

I would appreciate, if you fix this issue and also add a test case for it. In this PR, I restored the old functionality and moved your new code to a separate class.


jfbischoff:
Whoops. Thanks for finding out, we'll fix it.

PR #3460, opened 2024-09-09
add turn restriction test case

paulheinr:


matsim/src/test/java/org/matsim/core/router/AbstractLeastCostPathCalculatorTestWithTurnRestrictions.java
```diff
@@ -54,46 +52,60 @@ void testCalcLeastCostPath_TurnRestrictions() throws SAXException, ParserConfigu
 		assertEquals(network.getLinks().get(Id.create("5T", Link.class)), path.links.get(5));
 	}
 
+	@Test
```
paulheinr, line 55: @mrieser I would have expected, that this test case returns a correct path - but it doesn't. I think the reason is that node 3 is in the middle of a turn restriction sequence. Thus, the router does not visit node 3 during the search but the newly inserted colored node. This looks like a bug to me. What do you think?

mrieser: I'll try to look at this soon

mrieser: Took a bit longer than hoped, but I've finally had the time to look at this. Thanks for finding these cases!

mrieser: Thinking about the issue even more, it gets interesting...

Currently, node 3 can only be reach coming from link 23 which has the turn restriction so it is not possible to use link 4T. But in a slight different network, there might be other links to arrive at node 3.

Now what if an agent starts at link 34? Would the agent be allowed to use link 4T? Probably yes, because we don't know that the agent has arrived on link 34 coming from link 23 with the restriction.  This could lead to the situation where an agent arrives at an activity on link 34 coming from 23, but then departs later on and taking the (originally forbidden) link 4T. On the other hand, one could probably assume that if the agent has an activity on such a link being part of a turn restriction, then the turn restriction might not apply also in reality because to park at the activity, the agent has to somehow leave the road area... So hopefully this edge case can be neglected.

paulheinr: Thank you very much! Christian and I were also thinking about this case and came to the same conclusion.

PR #3466, opened 2024-09-09
use CsvOptions.detectDelimiter instead of hard coded ';'

simei94:


simei94:
@frievoe97 please note that for a while now there is the method CsvOptions.detectDelimiter. Thus, there is no need to hard code delimiters for csv files when reading them. In most of our scenarios the delimiter in fact is ';', but for international scenarios this might not be true. I would like to ask you to use the detect delimiter method for the future.

PR #3478, opened 2024-09-12
feat: flexibilize DrtRouteConstraintsCalculator

sebhoerl:
Our use case is that we want to have individual wait time, travel time, etc. constraints per request. So far, we were implementing a custom `DrtRouteCreator` for that, but now there is `DrtRouteConstraintsCalculator`. However, the interface right now doesn't seem to be very useful, because almost no information is passed on. This PR flexibilizes the interface by passing the person and the trip attributes.

As a sidenote, I'm a bit sceptical about how the `DrtOptimizationConstraintsSet` is passed on to this interface. My feeling is that this is actually something that should be negotiated behind that interface. Using the `DrtOptimizationConstraintsSet` sounds like the "default implementation", while any other implementation will specifically be more complicated than using this data structure. I think the main reason is that the `DrtOptimizationConstraintsSet` combines certain things that seem to be handled by the interface (travel time, ride time) while other parameters in `DrtOptimizationConstraintsSet` concern more specifically the `DrtRouteCreator`. In my mind, some kind of internal version of `DrtOptimizationConstraintsSet` (with various properties) should be the output of `DrtRouteConstraintsCalculator` instead of having multiple methods, and then one specific implementation would obtain the information from the configuration with the `ConstraintSetChooser` and so on. What do you think @nkuehnel?

### DIFF
--- a/contribs/drt/src/main/java/org/matsim/contrib/drt/routing/DefaultDrtRouteConstraintsCalculator.java
+++ b/contribs/drt/src/main/java/org/matsim/contrib/drt/routing/DefaultDrtRouteConstraintsCalculator.java
@@ -1,7 +1,9 @@
 package org.matsim.contrib.drt.routing;
 
+import org.matsim.api.core.v01.population.Person;
 import org.matsim.contrib.drt.optimizer.constraints.DefaultDrtOptimizationConstraintsSet;
 import org.matsim.contrib.drt.optimizer.constraints.DrtOptimizationConstraintsSet;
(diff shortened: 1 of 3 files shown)


nkuehnel:
HI @sebhoerl yes makes sense to me. The rationale behind not passing person and trip attributes so far was that the constraints set chooser would already take this into account and pass the appropriate configuration.

Your proposed solution does seem more flexible, though, I agree. Thanks!

PR #3479, opened 2024-09-12
feat: flexibilize structure of drt route constraints

sebhoerl:
See #3478. This is my proposal to have things more streamlined (at least in my opinion, but maybe there are good arguments against it). What do you think @nkuehnel?


nkuehnel:
I like that, thank you!

nkuehnel (approved):

PR #3483, opened 2024-09-13
chore: default zoning params in DvrpTravelTimeMatrixParams

tkchouaki:
@nkuehnel 

In the DvrpConfigGroup class, the getTravelTimeMatrixParams() method creates a DvrpTravelTimeMatrixParams on the fly if it doesn't already exist (meaning it wasn't in the XML input or it wasn't added by the script that created the DvrpConfigGroup). 

However, when a DvrpTravelTimeMatrixParams is created in the above mentioned method, no ZoneSystemParams is created inside it. Which then causes a NullPointerException whenever the ZoneSystemParams  is requested and used. 

To address this, this PR continues the same approach of creating the params on the fly by creating a SquareGridZoneSystemParams with a cellSize of 200m, as was the default in the old implementation.


nkuehnel:
LGTM, thanks!

PR #3493, opened 2024-09-24
Add different cases for carrier when solving a VRP; fix behavior of addPlan(..) / setSelectedPlan(..)

rewertvsp:
Adds 3 cases for the situation that a carrier already has a plan and the method runJsprit is called
Now there are these option:
- overwrite all existing plans (default)
- only create solutions for the carrier without plans
- create solutions for all carrier and add new plans and set as selected if already a plan exists


kt86:
I slightly changed the names of the Options.

AND 

The methods `addPlan(...`) and `setSelectedPlan(...)` in the `CarrierImpl` are now working as it is implemented for Persons. 
This means, that 

- `setSelectedPlan(..) `will NOT add the plan any longer. It moreover expects, that the plan is add beforehand. 
- `addPlan(..)` add the plan and makes it selected only if there is no other selected plan set.

kt86:
Please note: the latter change may crash existing code....

PR #3496, opened 2024-09-26
refactor CleanPopulation

simei94:
This is my proposal to make functionalities of CleanPopulation accessable from the outside to not have to duplicate code when wanting to remove unselected plans, remove route etc.


simei94:
@rakow Any reason why we should not make those functionalities accessable?

rakow (approved):

PR #3497, opened 2024-09-30
Adjust some carrier stuff

rewertvsp:
- set the selected plan to null if when using clearPlans
- add method to check if all jobs are handled in the selected plan


kt86 (approved):
Thanks. I just added one line in javaDoc, to clearify, that only the number of jobs is checked.

Two remarks for the future:
1.) This method should compare based on the job Ids and notify, which jobs are missing.
2.) I have the idea in my mind, to have some kind of "afterRun" consistency checker, where such a method could/should be called. What do you think?

PR #3499, opened 2024-10-01
Added a converter for xyt csv files to avro

frievoe97:
The new CSVToAvroConverter class converts an xyt csv file into an avro xyt file. The paths for both the input CSV and the output AVRO file need to be provided. Additionally, the projection and the name of the dataset (e.g., "Emissions") can be passed as optional parameters.


simei94 (approved):
looks good to me

simei94:
@frievoe97 @rakow at some point we should think about a AvroUtils class, which then has writeAvro / readAvro methods. I wrote my own read / write methods on several occasion, this could be generalized.

PR #3505, opened 2024-10-09
fix: considering the case when current element is a leg when aborting a rejected prebooking request

tkchouaki:
Hi @nkuehnel,

This follows ##3442. 
I performed some tests, I think that when the current plan element is a Leg, that's when the agent is walking towards the pickup location. So I think the agent should remain stuck in the following interaction activity.

Best regards,
Tarek.


nkuehnel (approved):
sounds good to me

PR #3536, opened 2024-11-02
feat: generalize multimodal link chooser

sebhoerl:
Currently, the access/egress link of a car trip is only chosen based on the origin/destination facilities of the trip. We have use cases, where we know the location of a car (at a P&R station), so the "access link" should not be the activity location, but rather the location of the vehicle / station. The generalization of the interface as proposed in this PR allows implementing this kind of custom behavior (for instance by passing additional information through the Attributes of the RoutingRequest).


Janekdererste:
Do we need the other `decideOnLink` method then?

sebhoerl:
@Janekdererste This is just in the default implementation, because this one doesn't care about access/egress, but only about the facility location. So decideAccessLink and decideEgressLink delegate to decideOnLink, otherwise we duplicate the same code. But yes, we can make this one `private` in the default implementation.

Janekdererste:
Sorry, @sebhoerl, I didn't read carefully enough. I thought the method is still in the interface. You implemented it just like I would have expected :-)

d-roeder:
I really like the idea of ​​making this configurable.

However, when looking at the code I noticed one thing that might be problematic or lead to unintended behaviour. In https://github.com/matsim-org/matsim-libs/blob/b487b41ce55ef2c1a949d1981a5ca1fc74e40485/matsim/src/main/java/org/matsim/core/router/NetworkRoutingInclAccessEgressModule.java#L206 and following the behavior of this PR is overridden when the facility is assigned to a link. This could possibly be mitigated if accessEgressModeToLink is used. Not sure. However, I think there should be no hidden magic when whe have the explicit LinkChooser.

Another point (whith a scope somehow outside of this PR) is that in  https://github.com/matsim-org/matsim-libs/blob/b487b41ce55ef2c1a949d1981a5ca1fc74e40485/matsim/src/main/java/org/matsim/core/router/NetworkRoutingInclAccessEgressModule.java#L191 nearestPointOnLink is used. I think this leads to wormholes unless relativePositionOnLink is used, as the agent walks to any point on the link, but de facto starts at the end.

sebhoerl:
Hi @d-roeder, I just took a quick look at the code. The `toFacility` is actually the destination of the car trip, and `egressActLink` is the link chosen by the link chooser. So I think everything is fine, or am I mistaken?

d-roeder:
oh, you're right 👀 

It could just be a bit confusing if the toFacility has no link set and NetworkRoutingInclAccessEgressModule then assumes that startLink==endLink. This makes sense in the default logic, but if arbitrary links can be set, then this is no longer necessarily correct.

PR #3546, opened 2024-11-06
fix: make ChargingHandler a MobsimEngine

### LINKED ISSUE #3545: ev: issues with ChargingHandler.notifyMobsimAfterSimStep
I have a use case where I react on `ChargingStartEvent`s using a simple event handler, and it is critical that timing is right. In particular, I collect various events in time step `T` and then process them in a `MobsimEngine.doSimStep` in time step `T + 1`. This way, we are sure to handle everything that happened in `T` at once and not have events arriving in parallel to the logic implemented in `doSimStep`. This is a common pattern that we use in various contribs (for instance, sharing).

Now for `ChargingStartEvent` this mechanism breaks, because the events are generated over several stations inside `ChargingHandler.notifyMobsimAfterSimStep`:

- The `ChargingStartEvent` is generated at the end of `T` in `ChargingHandler.notifyMobsimAfterSimStep`
- However, this happens after the event handlers have been instructed to synchronize on time step `T`
- So the event (which has `time=T`) still may arrive when everything is in the state of time step `T`, then my event collection logic still adds the event to the list of events collected in `T`
- But it may happen that publication of the event takes slightly longer, so it will arrive after I have flushed my collected events from `T` and it will be added to the collected events of `T + 1`
- In consequence, everything else in my business logic happens one time step later, i.e. the charging event will only be processed in `T + 2`

My conclusion from that would be:
- We should never write events inside `notifyMobsimAfterSimStep`. Somehow I even think that I noticed during development of the `sharing` contrib that this used to be forbidden (enforced?) in some way.
- In the particular case of the `ChargingHandler`, I'd like to transform this class into a MobsimEngine rather than and generate the events in a `doSimStep`. Or has this been tried already and there have been issues? @michalmac

sebhoerl:
This PR fixes #3545. The changes are minimal and looking at the unit test results, some events are shifted by one second at max.

### DIFF
--- a/contribs/ev/src/main/java/org/matsim/contrib/ev/charging/ChargingHandler.java
+++ b/contribs/ev/src/main/java/org/matsim/contrib/ev/charging/ChargingHandler.java
@@ -17,35 +17,48 @@
  *                                                                         *
  * *********************************************************************** */
 
-package org.matsim.contrib.ev.charging;
+ package org.matsim.contrib.ev.charging;
 
-import org.apache.logging.log4j.LogManager;
-import org.apache.logging.log4j.Logger;
-import org.matsim.contrib.ev.EvConfigGroup;
-import org.matsim.contrib.ev.infrastructure.Charger;
-import org.matsim.contrib.ev.infrastructure.ChargingInfrastructure;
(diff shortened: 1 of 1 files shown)


sebhoerl:
Found another fix, see #3545, closing this for the time being.

PR #3560, opened 2024-11-12
add networkMode to defaultVehicleType

rewertvsp:
- needed for the latest update of #3532 
- for the defaultVehicleType we should set a networkMode to get no exception


matsim/src/main/java/org/matsim/vehicles/VehicleUtils.java
```diff
@@ -81,7 +81,7 @@ public static VehicleType createDefaultVehicleType() {
 				.createVehicleType(Id.create(DEFAULT_VEHICLE_TYPE_ID, VehicleType.class));
 
 		defaultVehicleType.getCapacity().setSeats(4);
-
+		defaultVehicleType.setNetworkMode("car");
```
rakow, line 84: You can use TransportMode.CAR here

rewertvsp: Done :)

Issue #2764, opened 2023-09-13
ZStandard produces surprisingly large population files


neuma:
The PopulationWriterHandler, e.g. PopulationWriterHandlerImplV6, [flushes the buffered writer after each person](https://github.com/matsim-org/matsim-libs/blob/c80717b6b675b29d85a09f7b8c4d8425f28f6c6c/matsim/src/main/java/org/matsim/core/population/io/PopulationWriterHandlerImplV6.java#L122).
This results in small buffers written to file, if the person has only a few or very short plans.
Dictionary-based compression algorithms like [zst](https://en.wikipedia.org/wiki/Zstd) seem to have a problem building efficient dictionaries for small buffers and hence produce larger compressed files than regular deflate gzip.

Proposal
Move the flush to endPlans.

Possible Pitfalls
Flushing after each person guarantees a recoverable file in case of a crash - at least for uncompressed files.

---

Some Benchmarks
Given an uncompressed plans file with empty legs that it read in via StreamingPopulationReader and immediately streamed out via StreamingPopulationWriter. Output is written once as standard gz and once as zst with the default compression level of 6.

![image](https://github.com/matsim-org/matsim-libs/assets/1352652/02435767-16f2-4cd9-bc99-90f7347cea06)

Given the current situation with the flush at the end of writePerson zst produces a 15% larger file compared to the regular gz implementation of MATSim. Moving the flush to endPlans produces the expected file size reduction compared to gz. Note that gz seems to be unaffected by the flush at the end of writePerson.

PR #3564, opened 2024-11-14
Flush FacilitiesWriter only at the end to reduce Filesize

d-roeder:
The FacilitiesWriter creates surprisingly large files when using ZST compression, because flush is called too early. This PR fixes the problem, analog to https://github.com/matsim-org/matsim-libs/issues/2764.


steffenaxer:
Good catch! Thanks for this improvement!

PR #3565, opened 2024-11-14
feat: make chargers attributable

sebhoerl:
This PR makes `ChargerSpecification` and indirectly `Charger` attributable. The functionality is provided to read and write chargers with attributes just like it works for persons, facilities, etc. 

While doing this, the PR also adds a general unit test for reading/writing chargers, plus another one where attributes are used.

The DTD for the chargers is changed in this PR. I don't think we need to introduce a v2, because the basic structure is not changed, only the attributes are added as an optional feature. But may somebody who is more familiar with how DTDs are handled can comment on this?


kt86:
Only my opinion re DTD: 
Additional attributes in the DTD does **not** require a version change. Because it is not breaking anything. Thus, we do not need it, and it makes effort without gaining anything.
I only do all the work with a new version, if I change/remove existing stuff, that will break anything. 
-> Please go ahead


Thanks for adding tests!

PR #3567, opened 2024-11-17
feat(dvrp): cached free-speed matrices

sebhoerl:
When working with large scenarios, calculating the dvrp global and mode-specific free-speed matrices takes a lot of time, especially when parametric studies are run in which this has to happen at every simulation start-up. 

This PR introduces the possibility to cache the calculated free-speed matrices using the `cachePath` parameter in the respective `DvrpTravelTimeMatrixParams`. If the file does not exist, the matrix calculation is performed as usual and saved. If the file exists already, the matrix is read from the file. This way, when setting a global path, the calculation will be performed once in the parametric simulation series and then only read for all subsequent runs. There are various verifications included in the reading code to make sure that the travel times contained in the source file correspond to the present network and zoning system. The matrices are saved in a condensed custom binary format. Unit tests are provided.


nkuehnel:
Nice! Something that was on our list for weeks. Thank you so much!

PR #3574, opened 2024-11-19
Update deprecated expression in pom of decongestion contrib

kt86:
Use §{project.parent.version}, because the expression ${parent.version} is deprecated.
(This warning came from the log of the weekly builds: 
grafik
)


kt86:
@Janekdererste :
Shouldn't we use this expression` §{project.parent.version}` in all poms instead of the hard-code version 2025-.... in all dependencies in the different poms? 
Then, the version it is only there once per pom.

I know, that in practice this does not make a huge difference, as soon, as the update is done by search/replace all... but is looks a bit cleaner for me.

PR #3579, opened 2024-11-25
DRT: Zone based rebalancing target sensitivities

nkuehnel:
adds the option to set the target sensitivities alpha and beta of the min cost flow rebalancing strategy _per zone_ via attributes.
this complements the previous option to have flexible drt constraints to allow more spatial diversity within a single service.

includes a test for the min cost flow rebalancing (had to make the method package private for testing)


steffenaxer (approved):
Thanks for this improvement @nkuehnel. LGTM

mfrawley-moia (approved):

Issue #3583, opened 2024-11-26
Weekly deployment failures due to non-up-to-date dtds


jfbischoff:
Hi
I have received an error message that the weekly deployments have not been running for 2 weeks:
https://github.com/matsim-org/matsim-libs/actions/workflows/deploy-weekly.yaml

The reason for this is a test failure in 
[org.matsim.contrib.ev.infrastructure.ChargerReaderWriterTest.testReadWriteChargersWithAttributes()](https://github.com/matsim-org/matsim-libs/blob/8969b53e4879a344e3691536ae02dbe84cd6e890/contribs/ev/src/test/java/org/matsim/contrib/ev/infrastructure/ChargerReaderWriterTest.java#L57)

This failure occurs because the chargers_v1.dtd in http://matsim.org/files/dtd/chargers_v1.dtd is not up to date, it has also been changed recently: https://github.com/matsim-org/matsim-libs/blob/8969b53e4879a344e3691536ae02dbe84cd6e890/matsim/src/main/resources/dtd/chargers_v1.dtd#L12 by @sebhoerl 

I'm not sure how the dtds make it to the website, but apparently that pipe is broken.
@mrieser - you probably have an idea?

Bests,
Joschka

kt86:
As far as I see, the workflow for pushing the dtds to the website has never run: https://github.com/matsim-org/matsim-libs/blob/master/.github/workflows/deploy-dtds.yaml 
![grafik](https://github.com/user-attachments/assets/93f48849-f573-491b-b2b8-942c20d3a9f8)

kt86:
@jfbischoff Thanks for diving deeper in and finding the cause for the missing weekly build. 
I tried to update the script, that is responsible for updating the dtds on the website -> see #3585.

kt86:
The workflow for publishing the DTDs is now fixed #3597 
I manually (re-)started the job for creating the weekly build. Let's see, what happens there

PR #3585, opened 2024-11-26
Update trigger for DTD deployment job.

kt86:
The previous script was never run (which seems to be the cause for #3583), so there seems to be an issue with the triggers:

Therefore I changed the following: 
- Run on merged PRs.
- fix syntax …for target path.

@mrieser : Can you have a look in it, and if ok, merge it? 

And then I assume, we need a PR, changing something in the dtd path, to see, if it works (and updating the ev's dtd at the same time) ... 

Update: I have prepared such a PR: #3586


mrieser:
I don't have much experience with GitHub actions. Comparing your suggested changes with https://github.com/matsim-org/matsim-libs/blob/master/.github/workflows/deploy-on-pr-merge.yaml I see that the `deploy_on_pr_merge` also has the condition that it is only triggered when merging into `master` branch. Currently, the dtd-deploy would be triggered on ANY merge request. Do we want that?

Otherwise... as I said, I don't have much experience, so it's okay for me. I guess, we'l have to try it out and see what happens.

jfbischoff:
The pipeline is now running, unfortunately something seems wrong with the paths for rsync:

rsync: [sender] change_dir "/github/workspace/matsim/src/main/resources/dtd" failed: No such file or directory (2)

Here is the full log:
https://github.com/matsim-org/matsim-libs/actions/runs/12039650110/job/33567897764

jfbischoff:
I think the path should probably be "/github/workspace/**matsim-libs/**matsim/src/main/resources/dtd

kt86:
> I think the path should probably be "/github/workspace/matsim-libs/matsim/src/main/resources/dtd 

Can you just change it, please? 
Unfortunately I have no idea, with the path information. I am just happy that it has worked to start the pipeline;)

kt86:
Now it works :) 

grafik

The final solution to ass the checkout command as well: #3597

Issue #3318, opened 2024-06-17
Refine OSM parsing for bicycles


JWJoubert:
From the [OSM Wiki](https://wiki.openstreetmap.org/wiki/Bicycle) (S6 lower down the page), stairs can be tagged to include (or not) a ramp for bicycles. We want to add this check for ramps.

Two possibilities:
1. We exclude stair links that are not explicitly tagged with `ramp:bicycle=yes`, or
2. We include stairs and only omit those stair links where it is tagged `ramp=no`.

I would prefer (1), but I would like your input, specifically referring to @dziemke.

marecabo:
My preference is also 1. By default, I would rather consider them an obstacle for bike.

dziemke:
Thanks @JWJoubert I also think that 1 is the better option, as this option is directly using the given data, not be based on any implicit assumption. I would also expect that stairs without any ramp specification on OSM will by tendency not have a ramp in reality.

JWJoubert:
A quick update. I replaced [here](https://github.com/matsim-org/matsim-libs/blob/5420b79c69ed3d03b29e0064ab899d3ce24485bf/contribs/osm/src/main/java/org/matsim/contrib/osm/networkReader/OsmBicycleReader.java#L29) and [here](https://github.com/matsim-org/matsim-libs/blob/5420b79c69ed3d03b29e0064ab899d3ce24485bf/contribs/osm/src/main/java/org/matsim/contrib/osm/networkReader/OsmBicycleReader.java#L139) the `OsmTags.STEPS` with `OsmTags.RAMP` and `OsmTags.RAMP_BICYCLE`, but now the test fails.

However, and I would like @Janekdererste to just confirm, the test file `expected-result.xml.gz` is not correct since there are many steps that are now considered navigable by bicycle where I would argue they're not. Some examples:
- [Steel stairs](https://www.google.com/maps/@42.5782585,1.6465186,3a,75y,342.43h,77.61t/data=!3m6!1e1!3m4!1sLGfMAaOtL88e9syL4hiiqQ!2e0!7i13312!8i6656?coh=205409&entry=ttu) and [here](https://www.openstreetmap.org/way/143299078#map=19/42.57830/1.64681) in OSM;
- [Here](https://www.openstreetmap.org/way/191582679) is a case where the OSM tags suggest it is a path, yet the `XML` indicates it as steps (line 21235).

PR #3589, opened 2024-11-26
Ignoring steps for the cycling network.

JWJoubert:
Only if there is an explicit OSM tag indicating an accompanying (cycling) ramp, should it be included. This closes issue #3318.


JWJoubert:
@dziemke and @marecabo, sorry this took so long. I had trouble getting my Git client to wrangle through the new GitHub authentication... I couldn't create and publish a branch. Then, in the end, when getting around today to actually _doing it_, realised the problem was simply that I had an `https` checkout of the repository and not an `ssh` copy. Problem solved.

PR #3614, opened 2024-12-01
Freight: rename CarrierControler... to CarrierController...

kt86:
..., rename package accordingly, and change usage of the (core) `Controler `to (core) `Controller`.

Also run "optimize imports" on the freight contrib.


kt86:
With this PR, we will fix the spelling error of `controler` in the freight contrib.
In contrast to the changes in core (see #3511), we do **not** provide the option for the controller with 'll'. **We directly changed it.** This avoids another step somewhere in the future and having both options flying around for the time being.
So in the end, this will be **code-braking**, at least due to the renaming of the package.
So please be aware of it @JWJoubert, @tschlenther

PR #3620, opened 2024-12-03
Cleanup  freight contrib

rewertvsp:
- delete deprecated infrastructure (e.g. CarrierVehicleTypeLoader, CarrierPlanWriter V1, V2)
- delete plan writer V1 and V2, because we only support V2_1
- use recent VehicleEngineInformation setter for fuelType and consumption
- reduce vehicleTypes.xml duplicates


rewertvsp:
The changes should not break code which was not based on deprecated methods.

kt86 (approved):

kt86:
Thanks a lot for all that cleanup! It really helps to remove that old code from time to time :)

PR #3629, opened 2024-12-05
fix(drt): names in attribute-based prebooking

sebhoerl:
A colon was missing in the attribute-based prebooking logic. And a convenience setter has been added.


nkuehnel:
Thanks, appreciated!

Issue #2817, opened 2023-10-09
Change default network routing algorithm to SpeedyALT


mrieser:
Currently, the default network routing algorithm is `AStarLandmarks`.

I propose to change the default to `SpeedyALT`, which is considerably faster and has proven to be reliable in the last 2 years.

kainagel:
+1

neuma:
Hi @mrieser, one of our regression tests showed non-deterministic behavior - very same setup, code, input data results in different routes. This seems to be the case whenever there is an alternative route with the same (best) travel time. I nailed it down to the sequence of link preprocessing of the SpeedyGraph. Not sure, if there is a better solution than [fixing the sequence of processing.](https://github.com/matsim-org/matsim-libs/commit/4e57a602a46f2261f7afdb56a810164922c25fe5)

michalmac:
IIRC after switching to SpeedyALT some test became flaky, so this may be related to the issue reported by @neuma.

neuma:
Fix had been integrated into master https://github.com/matsim-org/matsim-libs/pull/3638

PR #3638, opened 2024-12-13
Make Speedy Graph Builder more deterministic

Royal2Flush:
We have noticed inconsistent car routing when multiple routes with equal travel time/disutility are available. This fix adds the links to the `SpeedyGraphBuilder` in order of their sorted id, not in order of their index as the standard iterator does. This makes the graph independent of any changes of the order in which the links are created (which to my understanding determines the relationship index<->id)


Royal2Flush:
The current state of the pull request causes several failed tests (see https://github.com/matsim-org/matsim-libs/actions/runs/12318669861):
- `matsim`: Some tests (`SpeedyALTTest`, `SpeedyDijkstraTest`, `SpeedyGraphTest`, `ModeRestrictionTest`) using the `SpeedyGraph` fail on tests where there is a symmetrical scenario with multiple equal routes. The router now in some cases chooses the other possibility due to the different ordering. This is in my opinion not at all worrying and I would just adjust the tests to pass with the new ordering
- `application`: This is a failure in `CreateScenarioCutOutTest`, where unrouted plans get routes on a chessboard(!) network and then the network is cut based on the links used. Since the chessboard has a lot of different route options it is also not surprising that something changes here. Again, I would just replace the reference test results
- `taxi`: This one is a lot more complex as it has high level integration tests failing with a lot of moving parts (literally). As far as I can see it is mostly small changes in travel time and score for a handful of agents. This might be caused by a different route choice and/or its effects downstream, even if the network used here is not symmetrical.

Royal2Flush:
Before I just replace the expected outcome of the failing tests, could @mrieser as the original author of the SpeedyALT have a quick look at this change?

Also, fyi @michalmac , @nkuehnel and @Aleksander1234519 since it would affects tests in contribs which I think are mostly developed by you

nkuehnel:
Could you explain why that works? I mean it does change it, apparently, but isn't the index just the index of the Id object? and isn't that index determined at the time of creation of the id object? I'm struggling to understand :)

nkuehnel:
I understand how the iteration order changes when building the graph (i.e., calls to addLink) and I understand how the index is different for a given link in different runs. 

How does it affect the speedy graph when addLink is called? Doesn't the graph just add the link to the data structure based on
```
int fromNodeIdx = link.getFromNode().getId().index();
int toNodeIdx = link.getToNode().getId().index();
int linkIdx = link.getId().index();
```
? Doesn't it just take what is returned by index()? So regardless of the order of the iteration for calls to addLink, shouldn't the index be the same (even though possibly different between runs)? Really curious what I'm missing.

michalmac:
Wouldn't adding links in a different order impact the order of in/out links for a given node? I assume it does, and hence the Dijkstra-like path search would visit nodes in a different order. So you get a min cost path, but no guarantee it's exactly the same one always.

Royal2Flush:
Exactly, the `setInLink()` and `setOutLink()` methods are sensitive to the call order. They fill the `nodeData` array in a kind of "first come, first serve" way if the node gets assigned multiple inlinks or outlinks. So the router later probably tests the outlinks in a different order.

nkuehnel:
Understood, thanks for clarifying ✌️

neuma:
Thanks to all of you for getting this pushed to the master.

---

Cross referencing the original comment and the commit therein https://github.com/matsim-org/matsim-libs/issues/2817#issuecomment-1923347664

Issue #3600, opened 2024-11-27
Freight (Carrier): Simply infrastructure of CarrierService and CarrierShipment


kt86:
Get rid of some duplicated infrastructure to handle "Services" and "Shipments" on the MATSim side.

(This is an extracted issue from the original #3510 )

kt86:
see also #3509 for making things more consistend

kt86:
Updating here to make it a bit more clear and differentiate it from #3509 :

IMO we could build a general class (CarrierActivity ???), containing the location, StartTimeWindow and Duration.
This could be used for both, `CarrierService` and `CarrierShipment` (Pickup and Delivery).

This would allow us, to reduce a lot of duplicated code.

When updating this, please consider for a new Version of the XML scheme.

kt86:
I think the most of it, which is not part of #3601 is prepared in #3645. 
So this issue gets closed by #3645

Issue #3509, opened 2024-10-10
Rename things in CarrierService / CarrierShipment, to have this consistent. 


kt86:
Also adapt the fields in the xml file as well. 

This is a first step on the way to maybe introduce something lika a "CarrierJob" on MATSim-level, which can have 1 or 2 locations. If it is a 1-location job, jsprit uses "Services", if it has 2 locations, it uses "Shipments".

kt86:
This will lead in the end to a new carrierDefinition XSD...

rewertvsp:
- einheitliche Benennung zwischen Service und Shipment (z.B. TimeWindows, Demand/Size...)

Issue #3601, opened 2024-11-27
Freight (Carrier): Generalize CarrierJobs 


kt86:
Later it is maybe more generalized and can have 1 or to 2 locations with all the settings (LinkId, StartTimeWindow, handlingDuration).
Depending on the number of locations, the jsprit algorithm is chosen: If it is a 1-location job, jsprit uses "Services", if it has 2 locations, it uses "Shipments".

A therefore needed, central Interface is created in #3599. 

Maybe something for the hackathon, or the code sprint... but definitely not urgent, but code-breaking.

(This issue was extracted from #3510.)

kt86:
@JWJoubert @rewertvsp I have talked to Kai yesterday, and we agree, that this could make the code and would be a small improvement. But at the same time, we agreed, that this change is most-probably a "nice-to-have".
So I put in a comment in the code with that idea, but we will close the issue for the time beeing.

**If you disagree, and that is something, we should really do**, please reopen the issue and we can have a meeting to discuss the details :)  
And/or bring it to the code-sprint in march.

PR #3645, opened 2024-12-20
Freight (carrier): Rename internal variables and getter/setter to make them more consitent

kt86:
In this PR I changed the names of some variables and the corresponding getter and setter to make them more consistent between `CarrierService` and `CarrierShipment`.

I kept the old methods as deprecated for not breaking outside code.

Moreover, the XML-schemes were kept unchanged, until we decide, for a more general approach in defining the CarrierActivities (service, pickup, delivery).
See also #3601


kt86:
This PR will close #3509 and close #3600.

PR #3649, opened 2024-12-30
Merge disallowed links

nkuehnel:


nkuehnel:
Adds _one_ way of merging disallowed links when merging duplicated links.
The code will permissively retain only those restrictions where the last link of the sequence exists in both restrictions while only keeping the sequence of the first link.

After all links have been processed, a final sweep replaces link ids of sequences by the merged replacements.

nkuehnel:
update:
the merging of turn restrictions is now handled after all links have been merged to also catch the case where the target link of turn restrictions is replaced

steffenaxer:
LGTM

steffenaxer (approved):

PR #3650, opened 2024-12-30
Add Tarjan SCC network cleaner algorithm

nkuehnel:
Adds a new network cleaner based on Tarjan's strongly connected components algorithm to identify strongly connect clusters.
https://en.wikipedia.org/wiki/Tarjan%27s_strongly_connected_components_algorithm

The reason is that it is considerably more performant than the default network cleaner which allows to incorporate a simple mechanism to create strongly connected clusters with the recently introduced multi-step and stackable turn restrictions (aka disallowed next link sequences https://github.com/matsim-org/matsim-code-examples/wiki/turn-restrictions)

The idea is to make use of the turn restrictions context introduced in the speedy graph builder which already expands the turn restrictions into colored subgraphs. For the adjacency matrix in Tarjan's algorithm, I simply add end nodes of colored subgraphs to the adjacency list of the start node of a sequence. As the adjacency may be pruned in the course of the algorithm, the implementation re-iterates until two consecutive passes result in the same network size.

For this, the TurnRestriction context had to be refactored into its own package (is that okay? @mrieser @marecabo )

Tested on multiple large networks, the largest having ~1.2 million links and 600k nodes with roughly 1 million turn restrictions (a lot of forbidden u - turns). For this largest network, the algorithm had to pass in 15 iterations, which took about 30s.

The implementation is not multimodal so far. As it is currently implemented using a lot of deep recursion it might be required to increase the stack size of the jvm (eg., -Xss100m) (or we need to slightly adjust it to an iterative approach)

I haven't tested every "edge" (badumm tss) case, leaving this PR for discussion for now.

### DIFF
--- a/matsim/src/main/java/org/matsim/core/network/algorithms/TarjanSCCNetworkCleaner.java
+++ b/matsim/src/main/java/org/matsim/core/network/algorithms/TarjanSCCNetworkCleaner.java
@@ -1,0 +1,214 @@
+package org.matsim.core.network.algorithms;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.matsim.api.core.v01.Id;
+import org.matsim.api.core.v01.network.Link;
+import org.matsim.api.core.v01.network.Network;
+import org.matsim.api.core.v01.network.Node;
+import org.matsim.core.router.turnRestrictions.ColoredLink;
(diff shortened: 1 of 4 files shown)


nkuehnel:
Ah, also I do not have a warranty that this approach does not "overprune" a given network but it does seem to produce strongly connected networks _with_ turn restrictions

nkuehnel:
Unfortunately, it does not work as intended, closing this for now

PR #3651, opened 2025-01-03
Use URL instead of path string in matrix cache

nkuehnel:
I had some trouble with file paths that contained spaces " "


steffenaxer:
I have no idea why this `SelectVsRandomSubtourTest` is failing. This change is only related to some `Path` changes isnt it?

rakow:
Please update, the failing test should be fixed now.

nkuehnel:
The IOUtils stream also improves performance by a factor of 20x-30x since it's using a buffered input stream that works better with the double-looped getInt() calls

PR #3652, opened 2025-01-03
Drt: streamline constraints and separate soft vs hard constraint handling

nkuehnel:
- streamlined constraints
-- extended DrtRouteConstraints to include the late diversion and pickup delay constraints, which can now be set per request
-- re-use the immutable constraints object in the DrtRoute class to avoid unneccessary duplication
-- updated the handling of route description json-ification (@steffenaxer ?) to have a nested constraints representation. not sure how stable that is for future udpates
-- we could think about further re-using the constraints object in the drt request object itself
- separated the max ride duration, late pickup and diversion threshold constraints in the handling of the cost calculation. until now these three constraints will always lead to rejection even if rejection for soft constraints is turned off


steffenaxer:
As far as I can remember, the description elements are needed within pricing context. Especially the unshared route length and optionally the unshared link sequence might be required. Anyway as long as those elements are available things should work. Thanks @nkuehnel

PR #3655, opened 2025-01-06
Remove old network routing module

Aleksander1234519:


contribs/informed-mode-choice/src/test/java/org/matsim/modechoice/ScenarioTest.java
```diff
@@ -28,6 +34,22 @@ public void setUp() throws Exception {
 		prepareConfig(config);
 
 		controler = MATSimApplication.prepare(TestScenario.class, config, getArgs());
+
+
+		// We need to add a vehicle, it however does not affect the results
```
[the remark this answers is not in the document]
Aleksander1234519: For now I added a note to all fixes of this type. We can decide later if using the none-option is a better solution.

.github/workflows/verify-push.yaml
```diff
@@ -14,7 +14,7 @@ jobs:
     if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
 
     runs-on: ubuntu-latest
-    timeout-minutes: 45
+    timeout-minutes: 90
```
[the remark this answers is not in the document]
Aleksander1234519: I think I found the cause of this problem: The "TerminationTest.testCustomConverenceCriterion()" Implements its own termination-criterium with "countCar > countWalk". This worked with the old Router, but the new router has much more walk-counts than car-counts, due to the access and egress legs, thus the simulation never terminates -> endless loop.

matsim/src/main/java/org/matsim/core/population/algorithms/PersonCalcTimes.java
```diff
@@ -69,7 +70,7 @@ public void run(Person person) {
 			int max = plan.getPlanElements().size();
 			for (PlanElement pe : plan.getPlanElements()) {
 				cnt++;
-				if (pe instanceof Activity) {
+				if (pe instanceof Activity && !TripStructureUtils.isStageActivityType(((Activity) pe).getType())) {
```
paulheinr, line 73: why are we filtering out stage activities?

matsim/src/test/java/org/matsim/core/controler/ControlerIT.java
```diff
@@ -147,7 +147,7 @@ void testConstructor_EventsManagerTypeImmutable() {
 	@ParameterizedTest
 	@ValueSource(booleans = {true, false})
 	void testTravelTimeCalculation(boolean isUsingFastCapacityUpdate) {
-		Fixture f = new Fixture(ConfigUtils.createConfig());
+		Fixture f = new Fixture(this.utils.loadConfig((String) null));
```
paulheinr, line 150: I think `ConfigUtils.createConfig()` was fine.

matsim/src/test/java/org/matsim/core/router/old/PlanRouterTest.java
```diff
@@ -52,6 +55,7 @@
 
 import java.util.Arrays;
 import java.util.List;
+import java.util.Set;
 
 public class PlanRouterTest {
```
paulheinr, line 60: I think it would be good to have these test cases parameterized, so that we check the `none` case and the `accessEgress` case

matsim/src/test/java/org/matsim/core/router/RoutingIT.java
```diff
@@ -156,6 +159,23 @@ private void doTest(final RouterProvider provider) {
 		final String inPlansName = utils.getClassInputDirectory() + "/plans.xml.gz" ;
 		new PopulationReader(scenario).readFile(inPlansName);
 
+		com.google.inject.Injector injector = Injector.createInjector(scenario.getConfig(), new AbstractModule() {
```
paulheinr, line 162: I think this test should also be parameterized

matsim/src/test/java/org/matsim/core/router/TripRouterFactoryImplTest.java
```diff
@@ -127,12 +134,10 @@ public void install() {
 				new LinkFacility( l1 ),
 				new LinkFacility( l3 ),
 				0,
-				PopulationUtils.getFactory().createPerson(Id.create("toto", Person.class)), new AttributesImpl());
+				p, new AttributesImpl());
+
+		Leg l = (Leg) trip.get(2) ;
```
paulheinr, line 139: If we parameterize this test, we can check both cases as it was done before

matsim/src/test/java/org/matsim/modules/ScoreStatsModuleTest.java
```diff
@@ -90,65 +90,45 @@ public void notifyShutdown(ShutdownEvent event) {
 			//            }
 
 			// yy the following is retrofitted from an older double[][] data structure and thus messy.  Please improve it.  kai, nov'16
-			if ( config.routing().getAccessEgressType().equals(AccessEgressType.none) ) {
+			// yyyy these change with the access/egress car router, but I cannot say if the magnitude of change is plausible. kai, feb'16
```
paulheinr, line 93: I think we can leave the test as it is and use the existing parameterization. but i don't really get, why the scores didn't change...

contribs/sbb-extensions/src/test/java/ch/sbb/matsim/mobsim/qsim/SBBQSimModuleTest.java
```diff
@@ -72,6 +77,15 @@ void testIntegration() {
         new ConfigReader(config).parse(new ByteArrayInputStream(xmlConfig.getBytes(StandardCharsets.UTF_8)));
         config.controller().setOutputDirectory(this.utils.getOutputDirectory());
         Scenario scenario = ScenarioUtils.createScenario(config);
+
```
paulheinr, line 80: I am wondering whether we still need this as the router doesn't necessarily need a network any more.

paulheinr (changes requested):
Looks good, thank you! I think that last thing to do is the parameterization of the core router tests. They would be good to have and hoefully not too hard to implement.

paulheinr:
This PR sets the default value of access egress type in the routing config from `none` to `accessEgressModeToLink`. The removal of the network module without access/egress will be done later in another PR.

PR #3664, opened 2025-01-14
Freight (carrier): renamings in CarrierService/CarrierShipment

kt86:
As a follow-up and partly revert of #3645, this PR renames some internal variables and the corresponding getter/setters.
The new names introduced now, some from a discussion with @rewertvsp and @kainagel.

The getters/setters (before the changes in #3645) are still there, marked as deprecated, and point to the new methods.

- size / capacityDemand / demand --> `capacityDemand`
- <Service/Pickup/delivery>Start(s)TimeWindow -->`<Service/Pickup/delivery>Start**ing**TimeWindow `

Moreover, I added a new `getInstance` method for the `CarrierService.Builder`, allowing to set the `capacityDemand` already in the Builder. The method with the current signature is marked as deprecated.
This is done to move a step forward, to have the `capacityDemand` of the job immutable .


rewertvsp (approved):
Thanks for the changes

PR #3669, opened 2025-01-15
Update of EvConfigGroup

fastGeorge:
An update to EvConfigGroup so that it is formatted in the preferred way according to what is said in ReflectiveConfigGroup. + A small change in RunEvExampleWithLTHConsumptionModel so that a local config file can be found.

I just happened upon these classes and expect I might have use for them in further work in an ongoing master thesis, so I attempted to update them. Please ignore this suggestion if my request in any way diverges from how you prefer feedback to be put forth.


jfbischoff:
Thanks!

fastGeorge:
I see, so the usage of StringGetters and StringSetters is no longer preferred? Then I guess EvConfigGroup was mostly correct to begin with. There is a comment about enums being preferred over booleans in the config file, is this still something that is true? Not that it would be relevant to change that in EvConfigGroup at this moment.

jfbischoff:
Probably the comment in ReflectiveConfigGroup is a bit old. Back when the concept of ReflectiveConfigGroups was new (long time ago) the comments in there were right. Sorry about that misleading information :(

fastGeorge:
No worries, I guess one shouldn't try to change what isn't broken! Thanks for your helpful and quick response!

PR #3679, opened 2025-01-22
Add network cleaner that respects turn restrictions

nkuehnel:
Moved out the turn restrictions context of the speedy graph builder.

With this, add a network cleaner version that accounts for turn restrictions:
- Use the context to explicitly expand the network by replacing start links and adding the colored subgraphs.
- Run the network cleaner over the expanded network.
- Collapse the expanded network and re-apply remaining turn restrictions.

Move everything in a turn restrictions subpackage.

Needs some more testing


nkuehnel:
For re-applying the restrictions the idea is the following:

- starting from a replaced link (i.e., the beginning of a colored subgraph)
- recursively follow colored links along the subgraph, keeping track of the route from the replaced start link to the current link/node
- if we hit a colored node which has fewer outlinks than the original uncolored node, add a turn restriction sequence to the start link

Issue #3687, opened 2025-01-25
problem with regression testing running times?


kainagel:
Hello, I just had a PR regression test failed because of 
```
Error: The operation was canceled.
```
Since it took 45 min to get there, my intuition is that we have once again hit the github resource limit.  I have looked a bit around and find the same fail in other PRs.  Could others please confirm (from their experience or their knowledge) that this is indeed the issue?

I can't say if we should reduce tests ... maybe some of them are too long and should be shortened?  Any insights here?

Another thing might be to split the "matsim" tests.  This would be a good thing anyways since in many cases now all the contribs have run through before "matsim" is done.  Any insights here?

kt86:
I had a similar feeling, that we hit a limit, when one of my PRs aborted yesterday - without any changes, just after updating the branch... (was there any change in the last time, increasing the test duration?)

Splitting up the tests sounds good, but I personally have no idea how to do it.
That the contribs are faster than MATSim is totally fine for me and IO like it, because many changes are done on contrib level and now we get a quick feedback without first waiting to matsim :)

PR #3688, opened 2025-01-25
reduce to 10000 persons

kainagel:
We had a couple of regression test failures, presumably because of time limits, see #3687 .  On my local machine, `RaptorDeterminismTest#testRaptorDeterminism` took so long that I interrupted it.  I counted 20sec per 10000 agents, and it goes  to 100'000 agents (i.e. 200sec ≈ 3min 30sec).   I don't know how computing times on my laptop compare to github actions, but this seems more than we can afford.  **As a quick fix, I reduced the test to 10'000 agents.**  

However, I am sceptical in general if we can have such "shotgun tests" as part of the normal regression suite.  We used to have the *IT tests, which were run separately and only about once a week.  I do not know if that functionality still works.

### DIFF
--- a/matsim/src/test/java/ch/sbb/matsim/RaptorDeterminismTest.java
+++ b/matsim/src/test/java/ch/sbb/matsim/RaptorDeterminismTest.java
@@ -150,7 +150,7 @@ public class RaptorDeterminismTest {
 
 		logger.info("Comparing stop RaptorStopFinder.findStops alongSide ptRouter.calcRoute(pt, ...)");
 
-		for(int personIndex=0; personIndex<personLists[0].size(); personIndex++) {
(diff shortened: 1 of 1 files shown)


steffenaxer:
@mrieser and I investigated this yesterday also. We had the assumption that the mem req. of this test is quite high and thus GC kicks in, which causes much longer runtimes on github.

steffenaxer:
@kainagel you could also reduce the  scenarioSamples = 10 to e.g. 5?

kainagel:
> @mrieser and I investigated this yesterday also. We had the assumption that the mem req. of this test is quite high and thus GC kicks in, which causes much longer runtimes on github.

Thanks.  Please feel free to improve further; I just did a quick fix to get the merge queue work again.  

The "20 sec per 10'000 agents" came from my local machine after increasing memory.  Without increasing memory, it was indeed again much slower.

PR #3689, opened 2025-01-25
Try to increase mvn memory

steffenaxer:


rakow:
Was the intention here to give more memory to the tests? Tests run in a forked jvm (usually) and giving maven more memory will not affect them. It rather needs to be set in the `maven-surefire-plugin` config. 

`
<argLine>@{argLine} -Xmx6500m -Djava.awt.headless=true -Dmatsim.preferLocalDtds=true</argLine>
`

As far as I know there are 7G available, so it should be split between the maven and test process.

steffenaxer:
Thanks @rakow. Yesterday we had the problem, that certain tests were not running due to memory limitations. I thought that this setting did the trick as they could be executed now. Should i revert this commit?

rakow:
I would propose to revert it but looking at the runtimes and see if it has any effect. 

Then increasing the actual memory for the tests is probably a good idea, as they are only allowed to use 1 GB.
https://github.com/matsim-org/matsim-libs/blob/b4f1bdd980ce3fdc7fdbd58865a5082717f74939/matsim/pom.xml#L61

PR #3707, opened 2025-02-03
make sure that landmarks always have data in turn restricted networks

nkuehnel:


marecabo:
Thanks for addressing the observed runtimes!

marecabo (approved):

Issue #3698, opened 2025-01-29
look into RunFissDrtScenarioIT.  Is it truly necessary that all the DRT config groups need to be cast?  didn't we want to have a separate MOIA config group?


kainagel:

paulheinr:
@luchengqi7 @nkuehnel could you give us an update about reverting/adapting the DRT config so that a cast is not necessary?

nkuehnel:
@paulheinr @kainagel I personally did not have the time yet, sorry for that. I will try to come up with a proposal next week

kainagel:
cast supposedly no longer necessary. need to evaluate

PR #3709, opened 2025-02-04
consistently return default drt constraints

nkuehnel:
@luchengqi7 @kainagel @paulheinr 

Could you have a look at this PR? 
This would remove the unnecessary cast to default constraints sets and still offer the flexibility to add additional constraints sets params.
see #3698 and #3311


contribs/drt/src/main/java/org/matsim/contrib/drt/optimizer/constraints/DrtOptimizationConstraintsParams.java
```diff
@@ -54,12 +54,14 @@ public List<DrtOptimizationConstraintsSet> getDrtOptimizationConstraintsSets() {
                 .toList();
     }
 
-    public DrtOptimizationConstraintsSet addOrGetDefaultDrtOptimizationConstraintsSet() {
-        Optional<DrtOptimizationConstraintsSet> drtOptParams = getDrtOptimizationConstraintsSets().stream()
+    public DefaultDrtOptimizationConstraintsSet addOrGetDefaultDrtOptimizationConstraintsSet() {
+        Optional<DefaultDrtOptimizationConstraintsSet> drtOptParams = getDrtOptimizationConstraintsSets().stream()
                 .filter(params -> params.name.equals(defaultConstraintSet))
```
paulheinr, line 59: If someone implements another type of `DrtOptimizationConstraintSet` and assign "default" as name of this set, then this function would create another optimization constraint set with the name "default" and with the type `DefaultDrtOptimizationConstraintSet`. Since we at VSP will probably never use other parameter sets than the `Default...` one, this is okay I guess. But this can be confusing.

paulheinr (approved):
Thank you for this hotfix!

luchengqi7 (approved):

luchengqi7 (approved):
Thanks for the update!

jfbischoff:
@nkuehnel I think you need to update the base branch for a merge

nkuehnel:
@jfbischoff yes, plz dont merge yet, I am thinking of changing it a bit further :)

contribs/drt/src/main/java/org/matsim/contrib/drt/optimizer/constraints/DrtOptimizationConstraintsParams.java
nkuehnel: @paulheinr thanks for the hint! I removed the set name based lookup and introduced a dedicated default constraints field in the params

PR #3715, opened 2025-02-10
add analysis re act start and end times

simei94:


simei94:
@rakow fyi

contribs/application/src/main/java/org/matsim/application/analysis/CheckPopulation.java
```diff
@@ -164,35 +166,98 @@ public Integer call() throws Exception {
...
+//				increase counters
+				if (hasNoStartTime || hasNoEndTime) {
+					noStartTime += hasNoStartTime ? 1 : 0;
+					noEndTime += hasNoEndTime ? 1 : 0;
+					noStartAndEndTime += (hasNoStartTime && hasNoEndTime) ? 1 : 0;
+				}
 
-			for (String act : Sets.newHashSet(activities)) {
-				haveActivity.mergeInt(act, 1, Integer::sum);
+				double duration = 0.;
```
tschlenther, line 217: this should probably use methods from TimeInterpretation, as this is what we _should_ use later in the sim, as well.
See also #3725

PR #3718, opened 2025-02-11
fix turn restriction related bug in LeastCostPathTree

nkuehnel:
the comingFrom[nodeIndex] data structure is exposed and used for path construction in the OneToManyPathCalculator. However, some nodes may internally be reached via colored nodes which are not known outside the speedy graph structure. So make sure to always return the uncolored node id index

### DIFF
--- a/matsim/src/main/java/org/matsim/core/router/speedy/LeastCostPathTree.java
+++ b/matsim/src/main/java/org/matsim/core/router/speedy/LeastCostPathTree.java
@@ -204,7 +204,6 @@ public class LeastCostPathTree {
         this.data[index + 2] = distance;
     }
 
-
     public int getComingFrom(int nodeIndex) {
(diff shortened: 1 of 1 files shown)


marecabo (approved):

matsim/src/main/java/org/matsim/core/router/speedy/LeastCostPathTree.java
```diff
@@ -204,8 +204,14 @@ private void setData(int nodeIndex, double cost, double time, double distance) {
         this.data[index + 2] = distance;
     }
 
+
     public int getComingFrom(int nodeIndex) {
-        return this.comingFrom[nodeIndex];
+        if(graph.hasTurnRestrictions()) {
```
mfrawley-moia, line 209:
```suggestion
        if (graph.hasTurnRestrictions()) {
```

PR #3721, opened 2025-02-12
Use path iterator instead of exposing internal id indices for constructing paths outside of least cost path tree

nkuehnel:


marecabo (approved):
Thank you!

PR #3733, opened 2025-02-17
feat(drt): optionally updates routes in drt

sebhoerl:
This PR solves a code TODO from February 2020 :) 

**Problem**: Currently, the schedules of the vehicles are updated in each time step. However, drive tasks are simply shifted, and the initial route is kept. In particular, this means that the vehicles don't adapt to information in the `TravelTime` that may be provided to the operator (for instance, a short-term prediction of congestion).

**Solution**: This PR introduces the `DriveTaskUpdater` which is called by the `ScheduleTimingUpdater`. This was already foreseen in the code, but never implemented. The default implementation `DriveTaskUpdater.NOOP` does nothing, but the `DrtRoutingDriveTaskUpdater` can be activated by setting `updateRoutes` in the `drt` config group to `true`.

**Note**: Since all vehicle routes are rerouted in every time step, this may cause some overhead for larger simulations, but there are only limited cases where this actually makes sense. Also, the overhead is only linear in the number of routes and vehicles.


nkuehnel:
nice one!

PR #3742, opened 2025-02-20
Improve write speed in FreeSpeedTravelTimeMatrix

mfrawley-moia:
# Background
During simulations at MOIA we noticed that for some scenarios, when there is no pre-calculated freespeed cache, it can be costly to create on the fly. For a filtered service area depicting Hamburg it took 122s just to write the freespeed TT cache file (excluding calculation time). 

With this implementation that falls to 18 seconds, albeit still using only a single core so there is still room for improvement. 
 I *could* also provide a strawman parallelised implementation brings the time down to ~2s (on an 8 core Macbook M2), but it adds significant complexity to the code.


nkuehnel (approved):
TY!

sebhoerl:
Great :)

PR #3743, opened 2025-02-21
fix: properly initializing DvrpLoadParams and DrtEstimatorParams in DrtConfigGroup

tkchouaki:
The `DvrpLoadParams` and `DrtEstimatorParams` parameter sets in the `DrtConfigGroup` are initialized directly with the attributes declaration. This causes an inconsistency with the base `ConfigGroup` class as these two parameter sets are not recorded in the `ConfigGroup.parameterSetsPerType` attribute. Which can cause a crash if the method `removeParameterSet` is called with one of the two sets mentioned above.

This PR fixes this by initializing the parameter sets in the constructor by calling the `addParameterSet` method.

### DIFF
--- a/contribs/drt-extensions/src/main/java/org/matsim/contrib/drt/extension/preplanned/run/PreplannedDrtModeModule.java
+++ b/contribs/drt-extensions/src/main/java/org/matsim/contrib/drt/extension/preplanned/run/PreplannedDrtModeModule.java
@@ -64,13 +64,13 @@ public final class PreplannedDrtModeModule extends AbstractDvrpModeModule {
 
 		bindModal(DvrpLoadFromTrip.class).toProvider(modalProvider(getter -> {
 			DvrpLoadType loadType = getter.getModal(DvrpLoadType.class);
-			return new DefaultDvrpLoadFromTrip(loadType, drtCfg.loadParams.defaultRequestDimension);
+			return new DefaultDvrpLoadFromTrip(loadType, drtCfg.getLoadParams().defaultRequestDimension);
 		})).asEagerSingleton();
 
 		install(new FleetModule(getMode(), drtCfg.vehiclesFile == null ?
 				null :
 				ConfigGroup.getInputFileURL(getConfig().getContext(), drtCfg.vehiclesFile),
-				drtCfg.changeStartLinkToLastLinkInSchedule, drtCfg.loadParams));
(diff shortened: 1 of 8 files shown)


tkchouaki:
The first fix described above doesn't work. After looking more in details there seems to be an issue with the structure of the ``.
Currently, If we have a `ReflectiveConfigGroupWithConfigurableParameterSets` and we want to set a default value for one of the parameter sets, there's no clean way of doing it. 
- If we initialize it directly as an attribute, it's not recorded in the level of the base class `ConfigGroup` and for instance we can't call `removeParameterSet`.
- If we call `addParameterSet` in the constructor, the problems mentioned above are fixed but another one appears. When loading the config file and copying the parameter sets from the generic `ConfigGroup` object to the specialized one, the `ReflectiveConfigGroupWithConfigurableParameterSets::addParameterSet` method throws an exception because such a parameter set already exists.

I would propose modifying the `ReflectiveConfigGroupWithConfigurableParameterSets::addParameterSet` to just override the existing parameter set and not throw an Exception but I am not sure about the MATSim-wide implications. 

@steffenaxer

steffenaxer:
Hm modifying the ReflectiveConfigGroupWithConfigurableParameterSets::addParameterSet and just overwriting something silently is not so cool. In my eyes the DvrpLoadParams and DrtEstimatorParams are somehow like the DrtInsertionSearchParams. They are very fundamental one and needs to be defined. 

- [ ] Option 1: Enforce NotNull leads to lots of minor code changes to ensure DrtConfigGroup has been correctly configured.
- [ ] Option 2: Let them private Nullable with a public getter would allow a lazy init. No further code changes would be required.

I would prefer Option 2 as this is also for non experts easier.

steffenaxer:
@tkchouaki thanks. LGTM.

tkchouaki:
@steffenaxer 
This lazy initialization fixes the problem for the DvrpLoadParams which always has to be present anyway. 
But for the `DrtEsttimatorParams` which is initialled in the attribute declaration but is not required to be present, the problem still persists that one cannot call `removeParameterSet`. A lazy initialization in the getter would prevent one from running without that parameter set.

tkchouaki:
I removed the initialization of the `drtEstimatorParams` in the `DrtConfigGroup` and added a fast fail in the `DrtModeModule`. 
I also modified the relevant tests to manually add the `drtEstimatorParams` when necessary. 
@luchengqi7

PR #3744, opened 2025-02-24
fix: prebooking setting agents stuck

sebhoerl:
The implementation for making agents stuck if their prebooked request is rejected was still not ideal:
- When an agent is on a leg, the next activity is set to an infinite end time, which makes the ActivityEngine let the agent go to sleep like on a regular end of day (without a specific *stuck* event). In that case, it would be nice to have the source of ending the day being identified as being stuck.
- When an agent is on an activity, we ended the activity in the same way, but additionally created a *stuck* event (nice!). However, we call `AgentCounter.decLiving` which is also called by the `ActivityEngine` which sends the agent to sleep. In consequence, the counter is called twice and in some particular situations (no fixed end time), this makes the simulation stop too early because the counter falls to zero.

The current PR should now fix the issues:
- In both cases, a *stuck* event is generated that clearly names DRT as the source of why the agent is stuck.
- The counting of agents should now be correct.

Internally, the logic is as follows:
- There is now a `stuckUponNextActivity` set in the `PrebookingManager` that manages letting agents get stuck.
- If an agent is on an activity, the agent is added to that set 
- If an agent is on a leg, the next activity in the schedule is searched, the end time is set to "never" (MAX_VALUE, which is different from POSITIVE_INFINITY which would let the agent simply go to sleep), and the agent is added to the set
- After processing all potential stuck agents, we go through the `stuckUponNextActivity` set to check whether the agent is in an activity, and if so, we make the agent stuck by letting the ActivityEngine send them to sleep, but we also create an explicit *stuck* event

### DIFF
--- a/contribs/drt/src/main/java/org/matsim/contrib/drt/prebooking/PrebookingManager.java
+++ b/contribs/drt/src/main/java/org/matsim/contrib/drt/prebooking/PrebookingManager.java
@@ -378,6 +378,7 @@ public class PrebookingManager implements MobsimEngine, MobsimAfterSimStepListen
 
 	// Rejections
 
+	private static final String ABORT_STUCK_REASON = " rejected prebooked request";
 	private final IdSet<Person> stuckUponActivity = new IdSet<>(Person.class);
 
 	private void processRejections(double now) {
@@ -456,7 +457,7 @@ public class PrebookingManager implements MobsimEngine, MobsimAfterSimStepListen
 		((HasModifiablePlan) agent).resetCaches();
 		internalInterface.getMobsim().rescheduleActivityEnd(agent);
 		eventsManager.processEvent(new PersonStuckEvent(now, agent.getId(), agent.getCurrentLinkId(),
-				this.mode));
+				null, mode + ABORT_STUCK_REASON));
 
 		internalInterface.getMobsim().getAgentCounter().incLost();
 
--- a/contribs/drt/src/main/java/org/matsim/contrib/drt/prebooking/PrebookingManager.java
+++ b/contribs/drt/src/main/java/org/matsim/contrib/drt/prebooking/PrebookingManager.java
@@ -378,6 +378,8 @@ public class PrebookingManager implements MobsimEngine, MobsimAfterSimStepListen
 
 	// Rejections
 
+	private final IdSet<Person> stuckUponActivity = new IdSet<>(Person.class);
+
 	private void processRejections(double now) {
 		for (Id<Request> requestId : rejectedEventIds) {
 			RequestItem item = requests.remove(requestId);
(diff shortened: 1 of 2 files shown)


sebhoerl:
Not sure how to properly test that, but if @nkuehnel and @tkchouaki could do a quick sanity check, I'd be happy :)

sebhoerl:
Actually, the proposal messed up the leg histograms (and they were already messed up before) if aborting agents in the prebooking manager was activated. The reason is that a stuck event with a mode is counted as an arrival. But when we let the agent get stuck after a prebooking attempt, there hasn't been any departure yet.

So the latest commit sets the mode to null. However, to trace which stuck events refer to prebooking, I added a `reason` attribute to the stuck event which may also come in handy at other places in the future.

sebhoerl:
I tested with some of our simulations, and it looks like it works properly now. I'll merge, but @nkuehnel please let me know if you notice anything problematic in your simulations.

nkuehnel:
Thanks a lot! in fact we experienced this issue in recent weeks and were wondering what happened. We will let you know if we find any other issues :)

nkuehnel:
@sebhoerl our simulation crashes at processRejections in the prebooking manager:
```
Iterator<Id<Person>> stuckIterator = stuckUponActivity.iterator();
		while (stuckIterator.hasNext()) {
			Id<Person> personId = stuckIterator.next();
			MobsimAgent agent = internalInterface.getMobsim().getAgents().get(personId);

			PlanElement planElement = WithinDayAgentUtils.getCurrentPlanElement(agent);
			if (planElement instanceof Activity activity) {
				abortAgent(agent, activity, now);
				stuckIterator.remove();
			}
		}
```
because the MobsimAgent agent is null.

Did you experience anything like that?
CC @mfrawley-moia

sebhoerl:
No, we didn't. Maybe there is some way that the agent can get twice in that iterator because rejections or something else are processed at different times? So we try to abort an agent that has already been aborted?

I'm wondering if we tested with a simulation (and in the unit tests) where all agents are prebooked for at least a few seconds. Maybe the PassengerEngine already sets some of them stuck?

nkuehnel:
I observed two things:
- the rejectedRequestsEvents in the DefaultPassengerEngine is only cleared if there as an entry in activePassengers (in which case there is iterator.remove()). Most of rejections never go there in our sims such that the rejectedRequestsEvents continue to grow and are iterated over in every time step again in doSimStep:
```
Iterator<PassengerRequestRejectedEvent> iterator = rejectedRequestsEvents.iterator();

while (iterator.hasNext()) {
	PassengerRequestRejectedEvent event = iterator.next();
	if (event.getTime() == time) {
		// There is a potential race condition wrt processing rejection events between doSimStep() and handleEvent().
		// To ensure a deterministic behaviour, we only process events from the previous time step.
		break;
	}

	List<MobsimPassengerAgent> passengers = activePassengers.remove(event.getRequestId());

	if (passengers != null) {
		// not much else can be done for immediate requests
		// set the passenger agent to abort - the event will be thrown by the QSim
		for (MobsimPassengerAgent passenger: passengers) {
			passenger.setStateToAbort(mobsimTimer.getTimeOfDay());
			internalInterface.arrangeNextAgentState(passenger);
		}
		iterator.remove();
	}
}
```
- regarding the crash, it seems to be the case that the agent is set to abort by 					`passenger.setStateToAbort(mobsimTimer.getTimeOfDay());` in the above snippet in time step x. then later in the time step, the prebooking manager tries to set the agent to stuck as well but then it is already removed from the qsim

nkuehnel:
I found the issue, it happens when the endtime of the preceding activity equals exactly the submission time of the prebooking leg (end time == prebooking:submissionTime:drt), e.g.:

```
  <person id="req_59380_0">
    <attributes>
      <attribute name="groupID" class="java.lang.String">req_59380</attribute>
      <attribute name="seatType" class="java.lang.String">seat</attribute>
      <attribute name="timestamp" class="java.lang.Integer">515515</attribute>
    </attributes>
    <plan selected="yes">
      <activity type="origin" x="567222.4194006884" y="5937060.3589411415" end_time="143:11:55" >
        <attributes>
          <attribute name="prebooking:submissionTime:drt" class="java.lang.Double">515515.0</attribute>
          <attribute name="prebooking:plannedDepartureTime:drt" class="java.lang.Double">515895.0</attribute>
          <attribute name="serviceClass" class="java.lang.String">classic</attribute>
        </attributes>
      </activity>
      <leg mode="drt" />
      <activity type="destination" x="567766.4213050185" y="5935876.584494734" />
    </plan>
  </person>
```

In that case, I get a rejection event in timestep 515516, which is handled by the prebooking manager. In there the rejection is handled by adding it to the stuckUponActivity queue, as the current plan element is already the leg ( the agent has started walking that second). In the same time step the DefaultPassengerEngine adds the rejected request event to rejectedRequestsEvents. In the next time step, in 515517, first the agent is first set to stuck by the passenger engine. After that, the prebooking manager tries to access the agent taking it from the stuckUponActivity queue, but the agent is no longer present.

@sebhoerl not sure about the obvious fix. One idea would be to just do nothing and remove the agent from the stuckUponActivity once we realize the agent is not present in the qsim anymore. WDYT?

sebhoerl:
I think it is fine to do nothing, but we should put a comment in the code.
The agent not being in the QSim anymore is a clear sign that the agent is already stuck, imo.

We could cover this case by a unit test in the PrebookingTest. If I understand correctly, this is just the case if there is no prebooking slack (= 0 seconds) which generates a prebooked request that is kind of processed just like an immediate request?

In that sense, another approach could be to make sure that a prebooked request is always (strictly) in the future (departureTime > currentTime), otherwise it should not be consumed by the PrebookingEngine. On the end of the PrebookingManager, we could delete prebooked requests for which the simulation has already passed the departure time to clean up memory.

nkuehnel:
It's not an immediate request nor a is the prebooking slack 0 - the difference is that the agent already starts walking to the stop in that second. In MATSim, immediate requests (slack of 0) would be submitted only when the agent arrived at the stop after walking.

I will create a PR then that will do nothing if the agent is null (already removed from the qsim)

PR #3758, opened 2025-03-04
fix handling of prebooked stuck agents

nkuehnel:
see discussion in #3744

### DIFF
--- a/contribs/drt/src/main/java/org/matsim/contrib/drt/prebooking/PrebookingManager.java
+++ b/contribs/drt/src/main/java/org/matsim/contrib/drt/prebooking/PrebookingManager.java
@@ -435,11 +435,15 @@ public class PrebookingManager implements MobsimEngine, MobsimAfterSimStepListen
 			Id<Person> personId = stuckIterator.next();
 			MobsimAgent agent = internalInterface.getMobsim().getAgents().get(personId);
 
-			PlanElement planElement = WithinDayAgentUtils.getCurrentPlanElement(agent);
-			if (planElement instanceof Activity activity) {
-				abortAgent(agent, activity, now);
-				stuckIterator.remove();
+			if(agent != null) {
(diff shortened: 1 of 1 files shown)


contribs/drt/src/main/java/org/matsim/contrib/drt/prebooking/PrebookingManager.java
```diff
@@ -435,11 +435,15 @@ private void processRejections(double now) {
...
-			PlanElement planElement = WithinDayAgentUtils.getCurrentPlanElement(agent);
-			if (planElement instanceof Activity activity) {
-				abortAgent(agent, activity, now);
-				stuckIterator.remove();
+			if(agent != null) {
+				PlanElement planElement = WithinDayAgentUtils.getCurrentPlanElement(agent);
+				if (planElement instanceof Activity activity) {
+					abortAgent(agent, activity, now);
+				}
+			} else {
```
mfrawley-moia, line 443: Would it be worth maintaining a counter of "stuck" agents and printing it at a suitable time after the loop?

mfrawley-moia (approved):

contribs/drt/src/main/java/org/matsim/contrib/drt/prebooking/PrebookingManager.java
nkuehnel: i guess all stucked agents are in the regular qsim log statement anyways:
https://github.com/matsim-org/matsim-libs/blob/aad969d99479e8fa32c61ce37d7dfeb69cf7ea81/matsim/src/main/java/org/matsim/core/mobsim/qsim/QSim.java#L544-L558
(lost agents)

sebhoerl: Yes, I think that is the only thing that the agentCounter.incLost() is used for

sebhoerl (approved):

PR #3760, opened 2025-03-05
feat: Parking in QSim

paulheinr:
@GregorRyb @jakobrehmann @kainagel 

This PR introduces parking search to the QSim. For fast computational execution, we decided to parameterize the parking search. 

Currently, there are 4 different parking contribs:

1. **Parking Proxy**: Parameterizes parking search, but relies on data from previous iterations. See below for further explanations.
2. **Parking Search**: Explicitly simulates parking search via within-day replanning.
3. **Parking Costs**: Integrates parking costs based on link attribute data.
4. **Parking Choice**: Old, to our knowledge outdated package.

# Why yet another parking model?
We want a parking model that can react on current parking occupancy. Fast execution is important. 

Thus: 
- Option 4 is outdated => doesn't fit
- Option 3 introduces parking cost, but no parking search time => doesn't fit
- Option 2 is out due to computational time. A parameterization is what we want. => doesn't fit.
- Option 1 seems to be plausible, but we faced some issues, which we want to overcome with this approach.

## Why not Parking Proxy?
Further details about this: https://www.matsim.org/conferences/mum2024/presentations/MUM2024_Kohl_ParkingProxy.pdf

First, let's have a look what the parking proxy does. 

Based on the number of parked cars within a certain area (500m x 500m grid cells) it estimates the parking search time.

**This works as follows:** 
	0. Register the parked cars via event handling during the previous iteration (custom initialization for it. 0)
	1. Change the selected plan: For each car leg, add time $x$ to the egress walk duration and subtract $x$ from next activity duration. 
	2. Perform Mobsim (penalty for longer parking search time is subsequently applied due to changes made in 1.)
	3. Reverse the changes on the plan such that parking search time only occurs in experienced plans, not in the memory.

**Issues:**
- Step 1+3 are very implicit (and auto-magic) what is in general not wanted. Moreover, changing the plans and reverse the changes does only work, because this is the only module doing this (otherwise there would be a fight about earliest/latest execution)
- Parking proxy does not take into account total number of parking lots within the grid cell
- It relies on data of the previous iteration and NOT from parking occupation of the current iteration. There is custom initialization for it.0. Initialization draws cars from home locations
- There is only one regression test

## How the new parking module works
Goals: 
- Parking search time depends on current occupation of parking lots, i. e. apply Belloche function (see https://doi.org/10.1016/j.trpro.2015.03.024)
- Clear integration into mobsim

To tackle this, we introduced the following components: 
- Similar to the ``waitingQueue``, ``AbstractQLink``  now has a ``parkingQueue``
- We use the already existing ``VehicleHandler`` interface to determine parking or not
- Call ``ParkingSearchTimeCalculator`` within the ``AbstractQLink`` for calculation of parking search time
	- ``ParkingOccupancyObservingSearchTimeCalculator`` implements event handler that registers parked vehicles on the network and based on that calculates the 
	- Other implementations possible

Configuration: 
Parking spots on links are initialized by link attributes
  - "onstreet_spots"
  - "offstreet_spots"
For Berlin, we estimate these attributes based on the number of total parking lots in Berlin (https://www.berlin.de/aktuelles/9275007-958090-12-millionen-parkplaetze-neue-karte-zeig.html#:~:text=In%20ganz%20Berlin%20gibt%20es,wurden%2C%20wie%20die%20Senatsverkehrsverwaltung%20mitteilte) and OSM parking amenities (see https://github.com/matsim-scenarios/matsim-berlin/blob/parking_search_osm/src/main/R/parking_osm.Rmd) 

# Open Things 
- [x] Any thoughts on the design are very welcome
- [x] Where to put the link attribute definition?
- [x] How do we organize the new and existing parking approaches? (this is maybe something for the code sprint)
- [x] Configuration is currently possible only via Code. This is a deliberate decision since we don't know how stable the new module will be at the beginning and we want explicit bindings in code. Should we have a ``ParkingModule`` and ``ParkingQsimModule`` as reference in the core? For now, there is only the ``ParkingTests`` class where one can see how to configure the new parking module. 
- [x] Check occupancy calculation with scaled demand.
- [ ] Does commercial traffic do parking search? 
- [ ] Plan-based initialization does not filter persons only, thus considers commercial traffic


kainagel:
Many thanks!

Just a small comment re "where to put attributes": Normally we hide those into `ParkingUtils` or similar.  See, e.g. `VehicleUtils#set/getAccessTime(...)`.  If this approach is not acceptable, could you please talk to me?  Thanks ...

paulheinr:
I discussed this approach with @mrieser and we created a follow up Issue #3846 for better architecture (not just parking). I guess this won't be the last PR touching the new code as we need to calibrate the settings based on the Berlin scenario. Finding a realistic initial parking occupancy seems harder expected.

PR #3764, opened 2025-03-06
feat(profiling): add profiling contrib

stendler:
Work-in-progress draft PR for easier review.

Based on a past commit to be compatible to matsim-berlin (6.4) for manual testing.


contribs/profiling/src/test/java/org/matsim/contrib/profiling/aop/AspectJTest.java
```diff
@@ -0,0 +1,88 @@
...
+
+		@Override
+		public int getIteration() {
+			return 6;
+		}
+	}
+
+	@BeforeEach
+	public void setUp() {
+		System.setOut(new PrintStream(outputStreamCaptor));
```
paulheinr, line 43: I totally see, why you do this manipulation, bit it seems a bit hacky. I think, this is ok, as long as there is a comment explaining why you need this.

contribs/profiling/src/test/java/org/matsim/contrib/profiling/aop/AspectJTest.java
```diff
@@ -0,0 +1,88 @@
...
+import org.matsim.vehicles.Vehicle;
+import org.matsim.vehicles.VehicleUtils;
+
+import java.io.ByteArrayOutputStream;
+import java.io.OutputStream;
+import java.io.PrintStream;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class AspectJTest {
```
paulheinr, line 20: I think it would be good to have a comment that explains how the aspects tested here are registered.

jfbischoff:
@stendler are you still working on this?

PR #3777, opened 2025-03-13
ev: improve documentation, provide utility function and comments

tschlenther:
1. This PR mainly provides additional documentation on the `ev` contrib.
2. Some typos are fixed.
3. A utility method to define a vehicle type to be electric is introduced in `ElectricFleetUtils.setElectricVehicleType` (which delegates to `VehicleUtils`).
4. Some potential bugs in the sevc packages are commented. I will point out those below.
5. The ReadMe is updated. 

It does feel a little ... confusing that one has to call the` _Within_DayEVEngine` to activate strategic charging behavior (to me and to people from Lund TH). Having said that, it does make sense that the charging behavior functionality resides in the within-day package, rather than in the strategic package. I wonder whether it should be in its own package, but probably not.

### DIFF
--- a/contribs/ev/src/main/java/org/matsim/contrib/ev/strategic/StrategicChargingUtils.java
+++ b/contribs/ev/src/main/java/org/matsim/contrib/ev/strategic/StrategicChargingUtils.java
@@ -287,11 +287,11 @@ public class StrategicChargingUtils {
     }
 
     /**
-     * Sets up the controller
+     * Sets up the controler
      */
-    static public void configureController(Controler controller) {
-        controller.addOverridingModule(new WithinDayEvModule());
-        controller.addOverridingModule(new StrategicChargingModule());
+    static public void configureControler(Controler controler) {
+        controler.addOverridingModule(new WithinDayEvModule());
+        controler.addOverridingModule(new StrategicChargingModule());
     }
 
     /**
--- a/contribs/ev/src/main/java/org/matsim/contrib/ev/strategic/StrategicChargingUtils.java
+++ b/contribs/ev/src/main/java/org/matsim/contrib/ev/strategic/StrategicChargingUtils.java
@@ -274,7 +274,7 @@ public class StrategicChargingUtils {
      * <li>Agents still need to be activated using the `activate` method</li>
      * </ul>
      */
-    static public void configureStanadlone(Config config) {
+    static public void configureStandalone(Config config) {
         configure(config);
 
         config.replanning().setMaxAgentPlanMemorySize(1);
--- a/contribs/ev/src/main/java/org/matsim/contrib/ev/strategic/StrategicChargingUtils.java
+++ b/contribs/ev/src/main/java/org/matsim/contrib/ev/strategic/StrategicChargingUtils.java
@@ -290,11 +290,11 @@ public class StrategicChargingUtils {
     }
 
     /**
-     * Sets up the controler
+     * Sets up the controller
      */
-    static public void configureControler(Controler controler) {
-        controler.addOverridingModule(new WithinDayEvModule());
-        controler.addOverridingModule(new StrategicChargingModule());
+    static public void configureController (Controler controller) {
+        controller.addOverridingModule(new WithinDayEvModule());
+        controller.addOverridingModule(new StrategicChargingModule());
     }
 
     /**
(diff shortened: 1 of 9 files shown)


contribs/ev/src/main/java/org/matsim/contrib/ev/strategic/scoring/ChargingPlanScoring.java
```diff
@@ -444,6 +445,12 @@ private void initializeDetours() {
 			AtomicDouble travelDistance = new AtomicDouble(0.0);
 
 			for (Leg leg : TripStructureUtils.getLegs(person.getSelectedPlan())) {
+				// don't we need to filter for legs with the EV, i.e. of the chargingMode?
+				// while the finalizeDetour() method takes max(0,observedDetour) and observedDetour considers trips with EV only,
+				// the fact that we do not filter here means that detours with EV are first compensated by legs with other modes, before they are scored, right?
+				// Moreover, I ran into situations where leg.getTravelTime was undefined, even though that the travel time was set in the route (e.g. walk) -
+				// maybe that doesn't belong here, but should we handle undefined times here?
+				// ts, march 2025
```
tschlenther, line 453: @sebhoerl : 

I do not really know the reason, but it seems that VSP has scenarios (e.g. Berlin v6.4) where the output plans contain legs for which the travel time is undefined even though that they have a route (with a travel time). 

This is not handled here.

I am currently trying to set up a standalone `SEVC` simulation based on the Berlin output. This is why I stumbled upon that issue.

tschlenther: I only realised now that the code here relies on the route being available for the distance, anyway. I will propose to rely on the route for the travel time for a first try and on the leg attribute as a fallback, as well as provide a meaningful error message when both are undefined.

tschlenther: I still think we need to filter the leg for `chargingMode` in order to obtain correct detours (haven't checked neither changed the code yet)

contribs/ev/src/main/java/org/matsim/contrib/ev/strategic/StrategicChargingUtils.java
```diff
@@ -287,11 +290,11 @@ static public void configureStanadlone(Config config) {
     }
 
     /**
-     * Sets up the controller
+     * Sets up the controler
```
sebhoerl, line 293: Change back to "controller" to avoid wrong writing. We even encourage people to change the name in the config group now :) 
https://github.com/matsim-org/matsim-libs/blob/506e4627b8c449a7ddf920da67730ffff405dab4/matsim/src/main/java/org/matsim/core/config/ConfigAliases.java#L28

contribs/ev/src/main/java/org/matsim/contrib/ev/strategic/scoring/ChargingPlanScoring.java
```diff
@@ -257,6 +257,7 @@ private void handleEnergy(double now, Id<Vehicle> vehicleId, double endCharge) {
 		EnergyEntry entry = this.energy.get(vehicleId);
 		double initialSoc = entry.current / entry.total;
 
+		//wouldn't an NPE be thrown above (and below) if entry was null? ts, march 2025
```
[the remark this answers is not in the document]
sebhoerl: Maybe some artifact from an earlier version. I think NPE cannot happen because of `initializeEnergy`. So we could remove the `if`.

contribs/ev/src/main/java/org/matsim/contrib/ev/strategic/scoring/ChargingPlanScoring.java
sebhoerl: Yes, I think you're right, we should filter for the chargingMode when initializing the detours

sebhoerl: For me its fine to get the travel time from the route

contribs/ev/src/test/java/org/matsim/contrib/ev/strategic/utils/TestScenarioBuilder.java
```diff
@@ -486,11 +486,11 @@ private Scenario prepareScenario(Config config) {
 		return scenario;
 	}
 
-	private Controler prepareController(Scenario scenario) {
-		Controler controller = new Controler(scenario);
+	private Controler prepareControler(Scenario scenario) {
```
sebhoerl, line 489: Same as before, two L ;)

sebhoerl (commented):
Thanks! :)

PR #3778, opened 2025-03-14
DRT Shits: make shift assignment logic available via injection

nkuehnel:


nkuehnel:
Hey @kt86 would you know why the freight contrib fails?
Thanks in advance :)

kt86:
Dear @nkuehnel ,
I have no idea, because the changes of this PR should not cause any change to the results there.

I will try to investigate it locally on Monday.

nkuehnel:
@kt86 thanks!

PR #3779, opened 2025-03-14
DVRP: swap variable names in DVRP load from trip

nkuehnel:
very minor thing, but I believe the variable names were supposed to be swapped


sebhoerl (approved):
thanks, nice catch :)

PR #3783, opened 2025-03-17
Application contrib usable with URLs

rakow:
Configs for run classes with the application contrib can now be loaded directly from URL.


rakow:
@tschlenther

PR #3787, opened 2025-03-20
fix: Update PessimisticDrtEstimator.java

luchengqi7:
While looking for examples of how to get maxTravelTimeAlpha and maxTravelTimeBeta after the re-sturcture of the DRT constraint, I have noticed a minor thing that can be improved. The estimated travel distance is now depending on the ratio between the estimated travel time and the direct ride time, which is now consistent with other implementations of the DrtEstimator. 

Besides, @nkuehnel, is this the desired way to get and set the maxTravelTimeAlpha and maxTravelTimeBeta in the DRT config group? Is there any shortcut (i.e., before it was only one line, for example double alpha = drtConfigGroup.maxTravelTimeAlpha)? 

For example, I have added a small function that can get the DefaultDrtOptimizationConstraintsSet directly from config, which can reduce the needs for repeated check and casting. What do you think?


nkuehnel:
@luchengqi7 
please have a look at. there is now a getter for the default implementation

luchengqi7:
Oh, right, This was actually already solved in the previous PR. I will revert the commit 6a2f5b9. 

But I would like to keep the commit b59b19c. I will try to update this PR. Thanks!

luchengqi7:
While looking for examples of how to get maxTravelTimeAlpha and maxTravelTimeBeta after the re-sturcture of the DRT constraint, I have noticed a minor thing that can be improved. The estimated travel distance is now depending on the ratio between the estimated travel time and the direct ride time, which is now consistent with other implementations of the DrtEstimator.

Issue #3811, opened 2025-03-24
Adapt existing config groups to use the new `@Parameter` pattern


mrieser:
See discussion in #3810

kt86:
Does it make sense to also add the `@StringGetter` and `@StringSetter` to the corresponding getter/setter in this step, if they are not annotated yet?
See the definition of them in the `ReflectiveConfigGroup`-- same location as for the `@Parameter` pattern.

marecabo:
> Does it make sense to also add the `@StringGetter` and `@StringSetter` to the corresponding getter/setter in this step, if they are not annotated yet?

I would say no. The `@Parameter` annotation was introduced to reduce the boilerplate/code duplication with trivial setters/getters and trivial StringSetters/-Getters.

Issue #3810, opened 2025-03-24
Discussion: explicit getters and setters


paulheinr:
I am seeing more and more places (possibly in various configs) where fields are accessed directly instead of via getters and setters.  Are we moving away from getters and setters?  Or only in the specific context of certain config groups?

Main reason why I am asking is because I would like to know how to teach this.  I find it difficult to teach students to use getters and setters when the first layer of MATSim with which they interact does _not_ use getters and setters consistently.  What should I teach them (besides the usual ``you need to muddle through'')?

Thanks for any help ...

paulheinr:
Use private fields with `@Parameter` Annotation and generate setters and getters. See: `org.matsim.modechoice.InformedModeChoiceConfigGroup`

PR #3818, opened 2025-03-24
chore(ev): update to setters and getters

sebhoerl:
Based on the discussionsin #3810 and #3811, update `ConfigGroup` for `ev` with setters and getters.

### DIFF
--- a/contribs/ev/src/main/java/org/matsim/contrib/ev/strategic/costs/DefaultChargingCostsParameters.java
+++ b/contribs/ev/src/main/java/org/matsim/contrib/ev/strategic/costs/DefaultChargingCostsParameters.java
@@ -16,17 +16,57 @@ public class DefaultChargingCostsParameters extends ReflectiveConfigGroup implem
 	}
 
 	@Parameter
-	public double costPerUse = 0.0;
+	private double costPerUse = 0.0;
 
 	@Parameter
-	public double costPerDuration_min = 0.0;
+	private double costPerDuration_min = 0.0;
 
 	@Parameter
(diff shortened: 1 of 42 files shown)


contribs/ev/src/main/java/org/matsim/contrib/ev/strategic/costs/DefaultChargingCostsParameters.java
```diff
@@ -16,17 +16,57 @@ public DefaultChargingCostsParameters() {
...
+
+	public double getCostPerUse() {
+		return costPerUse;
+	}
+
+	public void setCostPerUse(double costPerUse) {
+		this.costPerUse = costPerUse;
+	}
+
+	public double getCostPerDuration_min() {
```
mfrawley-moia, line 41: If we are introducing a "new" API here anyway, is there a reason to continue to deviate from camelCase?

tduberne: Hi Mark, just saw this randomly when looking at the changes from the sprint, and thought I would answer: the pattern `getNameOfVariable_unit()` or `nameOfVariable_unit` is quite standard in MATSim. It is only used for adding the unit at the end of a variable or getter/setter. It is used a lot in the scoring code. 

It avoids making mistakes such as mixing times in minutes with times in seconds without conversion. Similar to the idea behind the infamous "hungarian notation", see e.g. https://www.joelonsoftware.com/2005/05/11/making-wrong-code-look-wrong/

PR #3830, opened 2025-03-25
Improve rejection handling with drt estimator

luchengqi7:
I just noticed that the rejection handling for DRT estimator is not complete. And now it is completed. 

Can you please briefly check about that? Thanks! 

Besides, it seems that it is not possible to use the public static String from another contrib. For example, I define the term 
public static final String EST_RIDE_TIME = "est_ride_time" in the drt-contrib. But I cannot use it in the dvrp-contrib. 
Is there a solution to that?


luchengqi7:
By the way, I am not officially participating in the code-sprint. I happen to come across this today. 😂

I hope this won't mess up with the code-sprint. And we can also have a brief discussion on the DRT estimator topics if you still have time and energy. 😝

rakow:
The DVRP contrib does not depend on DRT, but the otherway round. So you need to define these attributes somewhere in DVRP so you can access it in the DRT estimator.

rakow (approved):

PR #3841, opened 2025-03-25
chore: add messages to some assert statements

mfrawley-moia:
## What?
This PR attempts to add useful messages to some of the many assertions in the matsim codebase to aid in log-based debugging.

## Why?
The matsim codebase contains many assert statements which prevent unwanted behaviour, for many there is no explanatory message attached as to which invariant was violated. Often invariants are failed lookups (!= null asserts) and adding a little context about which id lead to the null may be useful.


Janekdererste:
I changed the base branch, to test whether the ci works with the new `main` branch. Don't mean to interfere.

PR #3848, opened 2025-03-26
feat(core): add option to add "scale" attribute to MATSim containers and introduce helper methods for population

nkuehnel:
The scale attribute introduced here is meant for documentation purposes for now but could be consumed for analyses or for code consuming the data.


marecabo (approved):
Thanks, looks good to me! 🚀

PR #3850, opened 2025-03-26
feat(core): Only consider mode-specific turn restrictions when cleaning the network

marecabo:
This change 
* adds a `mode` argument to the `TurnRestrictionsNetworkCleaner` (#3679) to only consider turn restrictions (#2855) of that mode.
* adds a central [`NetworkUtils.cleanNetwork(Network network, Set<String> modes)`](https://github.com/matsim-org/matsim-libs/blob/6d36770428a2eb0fe6a6ff23afc714203d2a68bc/matsim/src/main/java/org/matsim/core/network/NetworkUtils.java#L829-L855) that contains all currently recommended sanitization steps for cleaning a network.  
  This method calls the `TurnRestrictionsNetworkCleaner` so turn restrictions are considered.  
  https://github.com/matsim-org/matsim-libs/blob/6d36770428a2eb0fe6a6ff23afc714203d2a68bc/matsim/src/main/java/org/matsim/core/network/NetworkUtils.java#L829-L855
* deprecates the non-mode-aware `NetworkCleaner` and
* tries to change all usages of non-mode-aware `NetworkCleaner` to the new `NetworkUtils.cleanNetwork` method. Where no other cleaning modes where obvious, "car" was used, as historically, MATSim only had car traffic.
* informs in the comments about the mode-aware-but-not-turn-restriction-aware `MultimodalNetworkCleaner`
* cleans invalid turn restrictions after filtering a network with the [`NetworkFilterManager`](https://github.com/matsim-org/matsim-libs/pull/3850/files#diff-3861d74ca7250777ef0086239c66ed0eaa47bc8d5637b5cc455650ad0d5387d5)


nkuehnel (approved):
Nice! Lgtm

Issue #3840, opened 2025-03-25
Shall we improve the handling of FileNotFound for shape file reading?


luchengqi7:
Currently, if the path to a shape file (shp) is wrongly specified, it will return the error: 
Exception in thread "main" java.lang.IndexOutOfBoundsException: Index 0 out of bounds for length 0
	at java.base/jdk.internal.util.Preconditions.outOfBounds(Preconditions.java:100)
	at java.base/jdk.internal.util.Preconditions.outOfBoundsCheckIndex(Preconditions.java:106)
	at java.base/jdk.internal.util.Preconditions.checkIndex(Preconditions.java:302)
	at java.base/java.util.Objects.checkIndex(Objects.java:385)
	at java.base/java.util.ArrayList.get(ArrayList.java:427)
	at org.geotools.data.shapefile.ShapefileFeatureSource.buildFeatureType(ShapefileFeatureSource.java:477)
	at org.geotools.data.store.ContentFeatureSource.getAbsoluteSchema(ContentFeatureSource.java:339)
	at org.geotools.data.store.ContentFeatureSource.getSchema(ContentFeatureSource.java:308)
	at org.geotools.data.store.ContentFeatureCollection.<init>(ContentFeatureCollection.java:69)
	at org.geotools.data.store.ContentFeatureSource.getFeatures(ContentFeatureSource.java:571)
	at org.geotools.data.store.ContentFeatureSource.getFeatures(ContentFeatureSource.java:106)
	at org.matsim.core.utils.gis.GeoFileReader.getSimpleFeatures(GeoFileReader.java:145)
	at org.matsim.core.utils.gis.GeoFileReader.getSimpleFeatures(GeoFileReader.java:114)
	at org.matsim.application.options.ShpOptions.readFeatures(ShpOptions.java:190)
        ......

This may give users an expression that there may be a problem with the shape file, rather than it is because of FileNotFound (i.e., the path is wrong). Can it be improved?

kainagel:
Please go ahead.  Thanks ...

luchengqi7:
It seems that the problem is only with the path in the URL format. 

If the path (String) is in the normal format, a FileNotFound exception will already be returned if the path is wrong. 
(e,g., String shpFile =  "/Users/xxx/path/to/shp/abc.shp";)

[...]

PR #3853, opened 2025-03-26
fix: Improve shp reading

luchengqi7:
This PR addresses the issue #3840. A FileNotFoundException will be thrown if the path to the Shapefile is not correctly specified. This is clearer than the error message it is returned currently.

### DIFF
--- a/matsim/src/main/java/org/matsim/application/options/ShpOptions.java
+++ b/matsim/src/main/java/org/matsim/application/options/ShpOptions.java
@@ -200,12 +200,12 @@ public final class ShpOptions {
 	public List<SimpleFeature> readFeatures() {
 		if (shp == null)
 			throw new IllegalStateException("Shape file path not specified");
-		if (!Files.exists(Path.of(shp)))
-			throw new RuntimeException(shp + " does not exist! Please double check the path for the shp!");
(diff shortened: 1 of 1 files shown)


luchengqi7:
> This fix is not correct, because the shp can be an URL pointing to any address.

And actually, the FileNotFound Exception will be thrown already if the path is incorrect. 

But it's weird that I got some strange error message from the Lausitz scenario, when the shp does not exist. I am trying to figure out what was going on there...

luchengqi7:
So, if the String is converted from a URL, then no FileNotFound Exception will be given, even the path is wrong:
e,g,:
Assume the file /Users/luchengqi/Documents/MATSimScenarios/Lausitz/shp/xxx/lausitz.shp does not exisit. Then I try to get geometry from the shp file with Geometry geometry = new ShpOptions(x, null, null).

Then, I will have: 
String x = "file:/Users/luchengqi/Documents/MATSimScenarios/Lausitz/shp/lausitz.shp" --> shp file has no feature
String x ="/Users/luchengqi/Documents/MATSimScenarios/Lausitz/shp/lausitz.shp" --> FileNotFound Exception

luchengqi7:
So we cannot do too much with the URL? I will then close this PR...

PR #3857, opened 2025-03-27
Cleanup bicycle contrib

GregorRyb:
Restructuring of the bicycle contrib:
- Created a bicycle default params implementation: BicycleParamsDefaultImpl with default values for surface, comfort, infrastructure factor.
-BicycleParams can now be bound to explicit values with your own implementation of the BicycleParams interface. 
- Added test for the BicycleParamsDefaultImpl. 
- Moved compute surface factor method from bicycle link speed calculator to bicycle params.
- Deprecated some methods. 
- Move methods getComfortFactor(), getGradient(), getInfrastructureFactor() from BicycleUtils to bicycle params.


simei94:
@dziemke we just wanted you to be aware of our changes. We think that the BicycleContrib now is more flexible :)

PR #3868, opened 2025-03-28
docs(GitHub): Add explanation of available types for Conventional Commits

marecabo:
I'd suggest to provide the list of possible/recommended types with a short explanation in the PR template. This is taken from the error message, when the Conventional Commits check fails.


Janekdererste (approved):
Good idea. I wanted to keep the template short, but this information is probably essential

PR #3870, opened 2025-03-28
refactor(core): use links for routing in path calculator (and tree)

nkuehnel:
As decided during the code sprint (see #3804 ) the least cost path calculator interface can now operate at the link level, to allow considering turn restrictions at the start link. The node based methods are set to deprecated for now.

We deliberately chose to change to link level, even though we acknowledge the fact that most routing algorithms operate at the node level. However, the internal implementations will still use the nodes (but may choose to start on colored nodes if there is a starting turn restriction). From the outside it is now actually more "consistent" with MATSim's default interpretation that everything is somehow happening "on a link" instead of "at a node"


sebhoerl:
I didn't check all the PRs, but I have the feeling that @nkuehnel wins the Code Sprint in the category "most files changed" :1st_place_medal:

nkuehnel:
@rakow 
could you check if the railsim actually depends on routing _from node_ of the from link to the _to node_ of the to link?
in default MATSim this is usually the other way around. If railsim expects it the other way around we might need to keep and support the node to node interface

The same applies for the application contrib, where the scenario cut out routes from the _to node_ of the from link to the _to node_ of the to link. but this actually looks more like a typo. Could you check? :)

rakow:
In railsim there can be diversion points after which the trains are allowed to reroute. I think we can adapt to the new interface, but I would prefer to keep the old method for now.

In application there is a method to generate reference routes, which are supposed to have a clearly defined start and end points. This was easy to achieve using the nodes and their corresponding coordinates. The same could be done with links aswell, but it needs to be cleared defined from which coordinates the routing takes places and how the distance is then calculated. Are the start and end link included in the calculation?

nkuehnel:
@rakow by default the implementations route from the to node of the origin link to the from node of the destination link.

I see your use case may be different currently, so I will adapt the calls in application and railsim to the old method for now :)

nkuehnel:
@rakow actually the TrainRouter class routes like default matsim (to node of origin link to from node of destination link)

It's just the RailsimTestutils class that does it differently. Is that on purpose?
https://github.com/matsim-org/matsim-libs/blob/b93dd55b19fc9a54a25d6e2065fb594311303857/contribs/railsim/src/test/java/ch/sbb/matsim/contrib/railsim/qsimengine/RailsimTestUtils.java#L88


Same question here:
https://github.com/matsim-org/matsim-libs/blob/b93dd55b19fc9a54a25d6e2065fb594311303857/contribs/application/src/main/java/org/matsim/application/prepare/network/SampleNetwork.java#L236

nkuehnel:
@rakow 
for the railsim, we reverted to the previous way of using the router using the _from_ node of the _from_ link routing to the _to_ node of the _to_ link.
This actually seems a bit suspicious, as the TrainRouter uses it the opposite way, just like the rest of MATSim.
https://github.com/matsim-org/matsim-libs/blob/1c1fa658e902b31cce6706ca502d06ca9faad00b/contribs/railsim/src/main/java/ch/sbb/matsim/contrib/railsim/qsimengine/router/TrainRouter.java#L69-L74

We added a TODO comment in the RailsimTestUtils class.

@rakow 
We also updated the test reference files for the scenario cut out tests as we think that the router call there should also be "to node" to "from node".

marecabo (approved):
Thanks, @nkuehnel for pushing forward with the final missing "link" ;) on fully supporting turn restrictions in MATSim! 🚀 

As this actually fixes an issue (https://github.com/matsim-org/matsim-libs/issues/3804), I would be in favor of merging it now and address any potential inconsistencies in tests in later PRs. @rakow Is that fine with you?

rakow:
Please go ahead for now. I will take a look at the railsim routing another time in the near future.

PR #3871, opened 2025-03-31
feat(freight): logistics: use CarrierShipments for delivery and some maintenance around the LSP

kt86:
## Problem:
Currently, the delivery is planned based on `CarrierService`s. 
VRPs with Services cannot use a vehicle for more than one subtour, so it runs only once. As a consequence there are - for the last mile -  too many vehicles on the road, leading to too much fixed costs.

## Solution:
With this PR, it is now possible to plan the delivery based on `CarrierShipment`s, allowing the Vehicle to return to the depot/hub and reload goods for the next (sub-)tour.

## Other changes:
- Some formatting: mostly avoid line breaks
- Replace some hard-coded strings by use of (existing) constants
- Add a `isWithinBound(...)` method for the tests to reduce code-duplication in various tests.


sebhoerl:
Please use `feat(freight)` (contrib name)

kt86:
> Please use `feat(freight)` (contrib name) 

@sebhoerl : I can change it.
My idea was to specify to which of the two parts of the freight contrib this PR belongs to.

The reason: Carriers is the old freight contrib, used from some groups independently, while logistics is a new package.

sebhoerl:
ah ok, if there is a good reason, fine for me ;)

PR #3872, opened 2025-03-31
chore: respect log level environment var

mfrawley-moia:
## What?
Configure log4j to optionally override the log level of the root logger **if** the LOG_LEVEL environment variable is set, **if unset** it will continue to default to info.

## Why?
To make it easy to switch log levels when executing matsim remotely without recompiling code, for example to get more context on a crash. 

Relates to: #3801


nkuehnel (approved):
lgtm!

PR #3875, opened 2025-04-01
New hbefa integration test (WIP)

Aleksander1234519:
- EXTERNAL: Fixed encrypted hbefa-tables in public-svn as they were corrputed and resulted in faulty emission-calculations. This may change emissions-results!
- Added PHEMTest, which checks if the emissions-calculations are valid
    -> TODO: The file-paths need to be changed
- added tools for fixing HBEFA-tables with missing entries (EmissionUtils)
    -> TODO: This tool is very slow and makes tables exponentially larger. Maybe add a new option for allowing tables with missing entries instead?
- Added a ConsistencyChecker to check for corrupted HBEFA-Tables
- Improved some error-messages for better comprehensibility

TODO: Remove the .R file
TODO: Resolve the Stop&Go-Fraction-Problem


kt86:
> * Maybe add a new option for allowing tables with missing entries instead?
We decided at VSP, that we want to the user to provide a complete table. This avoids some auto-magic, that caused issues in the past. 
Why is that getting slow? IMO, we only need to create the missing entries once, add them to the file, we are reading in and that's it.

Aleksander1234519:
> Why is that getting slow? IMO, we only need to create the missing entries once, add them to the file, we are reading in and that's it.

The current implementation only allows tables that include every possible combination of road categories, traffic situations, vehicle categories, vehicle attributes, and pollutants. For example, when adding a new pollutant, a table entry must be created for every other scenario, road, vehicle, etc.

This becomes increasingly inefficient for larger tables. I had a table that contained all the necessary information, but it did not include all possible combinations. After adding the zero-value entries, the table size increased from 19MB to 2.4GB. Reading it now takes significantly more time, which is very frustrating when developing or debugging a test/scenario. Since it's essentially the same table but filled with 99% "blank" information (entries with a 0 EFA), the behavior of the contrib should not change if we instead check for the key and return 0 if it does not exist. 

We could make this a non-VSP-default setting to prevent accidental activation. It could even serve as a debugging tool: saving all the missing keys and printing them out at the end. This way, the developer knows exactly which entries need to be added.

kt86:
oh, wow. This is indeed a huge increase. 
And it will get even worse, once we include other trafficSituation and/or grades.... argh.

Please also bring this issue to your hackathon-talk

PR #3880, opened 2025-04-02
build: enable manual triggering of code coverage workflow

michalmac:


steffenaxer:
Nice to see @michalmac with a PR 😀

PR #3885, opened 2025-04-04
fix: small fixes in SpatialRequestFleetFilter

tduberne:
Details:
- in some simulations it happened that the vehicles kept in the spatial tree were not available anymore in the simulation, resulting in an NPE
- not all DrtTastTypes have a baseType. This case is now handled by ignoring those tasks, rather than crashing.


nkuehnel (approved):
Thanks for updating!

PR #3893, opened 2025-04-09
fix(ev): Prevent SOC from exceeding 100% during energy recuperation

mattiasingelstrom:
**Summary**

When an electric vehicle starts at 100% SOC and recuperates energy (e.g., driving downhill), the current implementation allows the battery to exceed its maximum capacity. This can lead to runtime errors.

**Motivation**

In scenarios using energy consumption maps where negative energy values (i.e., recuperation) are possible, the battery can exceed 100% SOC. This previously caused runtime exceptions. With this change, recuperation is only allowed if the battery is not full, and only up to the remaining capacity.

**What this PR does**

Adds a check in DriveDischargingHandler that limits recuperation so that the battery charge never exceeds capacity.

Skips recuperation entirely when SOC is already at 100%.

**Testing**

There is currently no unit test for this behavior, but I verified the implementation by creating an energy consumption map with only negative energy values. The simulation runs without errors, and the resulting SOC profiles correctly cap at 100%.

Please let me know if a unit test needs to be added or if you'd prefer to make this behavior configurable.

### DIFF
--- a/contribs/ev/src/main/java/org/matsim/contrib/ev/discharging/DriveDischargingHandler.java
+++ b/contribs/ev/src/main/java/org/matsim/contrib/ev/discharging/DriveDischargingHandler.java
@@ -161,10 +161,25 @@ public final class DriveDischargingHandler
 			double energy = ev.getDriveEnergyConsumption().calcEnergyConsumption(link, tt, eventTime - tt) + ev.getAuxEnergyConsumption()
 				.calcEnergyConsumption(eventTime - tt, tt, linkId);
 			//Energy consumption may be negative on links with negative slope
-			ev.getBattery()
-				.dischargeEnergy(energy,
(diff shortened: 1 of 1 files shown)


jfbischoff:
Two PRs with the same fix in 8 hours :D 
https://github.com/matsim-org/matsim-libs/pull/3894

Which one should we merge?

fastGeorge:
When comparing the two solutions we would say that both achieve the same thing, but with our implementation being slightly more concise. Then again Mattias is our supervisor so it is his call! ;)

/ Gustav & Ylva

mattiasingelstrom:
Whichever works for me! However, mine has superior documentation 😉
But yes, theirs is more concise.

jfbischoff:
I threw a coin and merged the other one :-D

PR #3901, opened 2025-04-11
feat(pt): Add support for chained departures in transit schedules

rakow:
This PR introduces the concept of "chained departures" to MATSim's transit schedule, allowing one departure to be connected to another. This feature allows to properly model real-world transit operations like:
 
 - dividing or merging trains ("Flügelung")
 - through services where vehicles continue on a different line ("Durchbindung")
 - Circular routes

## Features

- New `ChainedDeparture` interface and implementation
- Extended `Departure` interface with methods to get/set chained departures
- Updated XML reader/writer to support extensions to the existing format
- Added validation to ensure referenced departures exist
- Enhanced the `TransitScheduleFactory` with methods to create chained departures

## XML Format

The transit schedule XML format has been extended to support chained departures. Example:

```xml
<departures>
  <departure id="d1" departureTime="09:24:00" vehicleRefId="vehicle1">
    <!-- Simple case - chain to a departure in the same line/route -->
    <chainedDeparture toDeparture="d2"/>
  </departure>
  
  <departure id="d2" departureTime="10:14:00" vehicleRefId="vehicle2">
    <!-- Complex case - chain to a departure in a different line/route -->
    <chainedDeparture toTransitLine="line2" toTransitRoute="route5" toDeparture="d3"/>
  </departure>
</departures>
```

The `toTransitLine` and `toTransitRoute` attributes are optional if the chained departure is in the same line/route.

The PR adds validation to the `TransitScheduleValidator` to ensure that all referenced chained departures actually exist in the schedule, preventing broken references.

## Next Steps

This PR only adds the data structures and file format. Future changes will extend the router to consider chained departures when planning trips and update QSim and scoring.


matsim/test/input/org/matsim/pt/transitSchedule/chained_departures_schedule.xml
```diff
@@ -0,0 +1,102 @@
...
+				<link refId="pt_221075_bus-pt_90277_bus"/>
+				<link refId="pt_90277_bus"/>
+				<link refId="pt_90277_bus-pt_196643_bus"/>
+				<link refId="pt_196643_bus"/>
+			</route>
+
+			<departures>
+				<departure id="d1" departureTime="09:14:00" vehicleRefId="pt_842---32640_0_0">
+
+					<chainedDeparture toTransitLine="line" toTransitRoute="line1_r1" toDeparture="d2"/>
```
vsp-gleich, line 93: I assume it should be toTransitLine="line1", right?

mrieser:
very nice, thanks a lot! I like that the reader automatically fills in transitLine and transitRoute if they are missing in the XML, so that in the code those are always set. And that the writer automatically filters those out!

One suggestion: The TransitScheduleValidator could also test that the last StopFacility of the route is the same as the first StopFacility of the chained route. In my understanding, this should always be the case, otherwise it's not real chaining if you could to a trip between two stops without an official line/route serving it.

PR #3904, opened 2025-04-11
fix: SpatialRequestFleetFilter can now also understand tasks without BaseType

tduberne:
check class of DRTTask instead of BaseType

charging tasks have to base task. This way of operating covers those cases, and is also logically what is needed.


nkuehnel (approved):
Thx!

PR #3905, opened 2025-04-13
build(deps): Bump geotools to 33.0

steffenaxer:
This PR bumps geotools to the next major release. In order to make this happen I needed to update several conflicting dependencies.

### DIFF
--- a/contribs/shared_mobility/src/main/java/org/matsim/contrib/shared_mobility/run/SharingServiceConfigGroup.java
+++ b/contribs/shared_mobility/src/main/java/org/matsim/contrib/shared_mobility/run/SharingServiceConfigGroup.java
@@ -4,13 +4,14 @@ import java.util.Arrays;
 import java.util.Map;
 import java.util.stream.Collectors;
 
-import javax.annotation.Nonnegative;
 import jakarta.validation.constraints.NotNull;
 import jakarta.validation.constraints.Positive;
 
 import org.matsim.core.config.Config;
 import org.matsim.core.config.ReflectiveConfigGroup;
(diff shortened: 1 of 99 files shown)


contribs/shared_mobility/src/main/java/org/matsim/contrib/shared_mobility/run/SharingServiceConfigGroup.java
```diff
@@ -4,14 +4,15 @@
...
 
-import javax.annotation.Nonnegative;
 import jakarta.validation.constraints.NotNull;
 import jakarta.validation.constraints.Positive;
 
 import org.matsim.core.config.Config;
 import org.matsim.core.config.ReflectiveConfigGroup;
 import org.matsim.core.config.consistency.BeanValidationConfigConsistencyChecker;
 
+import javax.annotation.Nonnegative;
```
mrieser, line 14: shouldn't this also be `jakarta.*`?

mrieser (commented):
looks fine except that one place where still `javax.annotation.*` is used, while everywhere is it got replaced with `jakarta.annotation.*`

contribs/shared_mobility/src/main/java/org/matsim/contrib/shared_mobility/run/SharingServiceConfigGroup.java
steffenaxer: The Nonnegative annotation is only existing in the jsr305 dependecy from google. Jakarta does not support this annotation.

PR #3913, opened 2025-04-24
feat(pt): Support chained depatures in routing

rakow:
This PR is a follow-up to #3901. The transit router now supports chained departures and finds connections without transfers if departures are chained. 

`DefaultTransitPassengerRoute` was modified to support these new routes that may consist of multiple sections.
Note that they are not handled by pt passengers and driver agents yet. Simulating these will be implemented in a later PR.

### DIFF
--- a/matsim/src/main/java/ch/sbb/matsim/routing/pt/raptor/RaptorRoute.java
+++ b/matsim/src/main/java/ch/sbb/matsim/routing/pt/raptor/RaptorRoute.java
@@ -75,6 +75,12 @@ public class RaptorRoute {
         this.ptLegCount++;
     }
 
+	void addChainedPart(TransitStopFacility fromStop, TransitStopFacility toStop, TransitLine line, TransitRoute route, String mode, double depTime, double boardingTime, double vehDepTime, double arrivalTime, double distance) {
+		RoutePart part = new RoutePart(fromStop, toStop, mode, depTime, boardingTime, vehDepTime, arrivalTime, distance, line, route, null);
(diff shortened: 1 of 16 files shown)


matsim/src/main/java/ch/sbb/matsim/routing/pt/raptor/RaptorRoute.java
```diff
@@ -75,6 +75,12 @@ void addPt(TransitStopFacility fromStop, TransitStopFacility toStop, TransitLine
         this.ptLegCount++;
     }
 
+	void addChainedPart(TransitStopFacility fromStop, TransitStopFacility toStop, TransitLine line, TransitRoute route, String mode, double depTime, double boardingTime, double vehDepTime, double arrivalTime, double distance) {
+		RoutePart part = new RoutePart(fromStop, toStop, mode, depTime, boardingTime, vehDepTime, arrivalTime, distance, line, route, null);
+		this.travelTime += (arrivalTime - depTime);
+		editableParts.getLast().chainedPart = part;
```
mrieser, line 81: What if we have 3 or more departures chained, e.g. in circular routes? Shouldn't the new part be attached to the  end of already chained parts? something like:

```java
RoutePart lastPart = editableParts.getLast();
while (lastPart.chainedPart != null) {
  lastPart = lastPart.chainedPart;
}
lastPart.chainedPart = part;
```

mrieser (approved):
Except the one remark about 3 or more chained departures this looks great! Thanks!

PR #3923, opened 2025-04-30
feat(core): add optional exception message to shutdown event

nkuehnel:
for information purpose in shutdown event consumers


mfrawley-moia (approved):

mrieser (approved):
Besides the comment to use the full exception object instead of just a String, this looks good.

PR #3930, opened 2025-05-06
refactor(drt): better interpretable drt stats

nkuehnel:
in recent simulations we sometimes had a hard time understanding some of the DRT output such as wait stats, drt legs etc in prebooking scenarios.

The problem is that the analysis code previously defined wait time as "dropoff/arrival time minus departure time". Now the problem of this is that "departure time" here is actually the departure event of a person starting a drt leg (after access walk in stop based services). And here it starts getting a bit blurry and being about semantics.

In the prebooking case, the DRT submission may- for example- be made during the preceding activity of the agent. The submission is made for a departure in the future, with an earliest pickup time defined. In a stop-based service, the agent will then start walking to the stop, with a drt interaction event upon arriving at the stop. After arriving at the stop, the agent starts the drt leg with another departure event. However, this departure may not necessarily mean that the earliest pickup time of the prebooked request has already passed - the agent may just have arrived early at the stop ( in this case the current 'departure time' in the drt legs output can even be before the 'earliestDepartureTime', which is confusing, as the former is basically 'agent arrived at stop and is ready for pickup' whereas the latter is more a 'earliestPickupTime'.

So in this scenario, we have
1) submission
2) activity end 
3) access departure
4) arrive at stop
5) drt interaction
6) drt departure
7) pickup
In the door to door case, steps 3-5 do not exist.

We now propose to change the definition of wait time in the drt analysis to 'pickup time - earliestDepartureTime' instead of 'pickup time - drt leg departure time'. In the prebooking case, this means that only time after the earliest time for pickup is counted as wait time. In the regular adhoc case, agents will first walk to the stop and only then submit their request, such that earliestDepartureTime == drt leg departure time.

One could argue that from the agents point of view it may still be waiting time if the agent arrives early. In reality one would probably not base evaluations of a service based on people arriving earlier than agreed on ;) 
Also, with this update, wait times are closer to what is defined in max wait constraints in the optimizer. Because previously our drt legs file had legs with waiting times consistently above the allowed maximum wait time. The same thing happened in the constraints_drt_waiting_time plot.

In addition to the interpretation of wait times, we also changed the column name of the drt legs csv from "departure time" to "readyForPickupTime" which is more closely to what it is in the simulation - the agent being ready to be picked up at the pickup location (with or without preceding access walk).

@sebhoerl @tkchouaki  this PR mostly changes things in prebooking scenarios as the result does not change in adhoc scenarios as explained above.

ATTENTION:
this changes some output csv files of DRT scenarios used for downstream analyses


nkuehnel:
(we can also change readyForPickupTime to something else, e.g. 'wait start time' which might also be a good expressive candidate)

jfbischoff:
Aww, that was probably the last DRT KPI that has remained unchanged since 2017 :-) 
(Makes a lot of sense though, so please go ahead)

PR #3935, opened 2025-05-09
fix(core): update turn restrictions network cleaner to account for temporarily removed nodes

nkuehnel:
some modes might get removed in the cleaning but get re-inserted if only colored versions of it persist.


marecabo (approved):
Good catch!

PR #3938, opened 2025-05-09
refactor(pt): Make RouteDescription public again

marecabo:
@rakow The `RouteDescription` class was made private in #3913 without a comment, however, it is quite useful for analyzing transit vehicle usage during simulation time. Thus, I would suggest to make it public again.

### DIFF
--- a/matsim/src/main/java/org/matsim/pt/routes/DefaultTransitPassengerRoute.java
+++ b/matsim/src/main/java/org/matsim/pt/routes/DefaultTransitPassengerRoute.java
@@ -19,7 +19,6 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 
 public class DefaultTransitPassengerRoute extends AbstractRoute implements TransitPassengerRoute {
 	protected final static int NULL_ID = -1;
-	private final static ObjectMapper OBJECT_MAPPER = new ObjectMapper();
 
 	final public static String ROUTE_TYPE = "default_pt";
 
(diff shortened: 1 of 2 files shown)


rakow:
In the matsim repo the `RouteDescription` isn't used in any API, neither internally nor public. It is purely used in the transit route itself for serialization. I suppose you must have used your own json mapper to read it? Your code may break when any changes to the format are made internally. I would propose to add some public static read/write methods to make sure this use-case is supported, and the class has a reason to be public. At least, there should be some javadoc stating the use of the `RouteDescription` and why it is public.

marecabo:
Thank you for your comments, @rakow. Yes, we just read the RouteDescription objects with a plain ObjectMapper. I followed your suggestion and added serialization and deserialization methods as well as a small test.

kainagel:
I agree that you would need to discuss with Marcel if this can be made public.  I can't tell but it "feels" awkward.  If there is important material, it should rather be Attributable, but maybe that is too slow for routes.  Thanks ...

On 9. May 2025, at 18:27, rakow ***@***.***> wrote:

[https://avatars.githubusercontent.com/u/6706801?s=20&v=4]rakow left a comment (matsim-org/matsim-libs#3938)<https://github.com/matsim-org/matsim-libs/pull/3938#issuecomment-2867158904>

In the matsim repo the RouteDescription isn't used in any API, neither internally nor public. It is purely used in the transit route itself for serialization. I suppose you must have used your own json mapper to read it? Your code may break when any changes to the format are made internally. I would propose to add some public static read/write methods to make sure this use-case is supported, and the class has a reason to be public. At least, there should be some javadoc stating the use of the RouteDescription and why it is public.

steffenaxer:
Dear @mrieser do you have any concerns making this public again with @marecabo changes? It is currently a roadblock. 🙏

kainagel:
Did you give up since you did not receive an answer?

steffenaxer:
No, I found out that one could just use the defaultpassegertransitroute to get the required info. Thanks @kainagel

kainagel:
ah good thanks

PR #3941, opened 2025-05-10
feat(core): teleported mode speed by person

sebhoerl:
This PR allows users to define a `personSpeedAttribute` in the `RoutingParameter`s. If the option is set and a person has the corresponding attribute, the value will be used for teleportation routing instead of the global parameter, which was the only option previously. This allows to define speeds, for instance, based on the age of the person.

### DIFF
--- a/contribs/drt/src/test/java/org/matsim/contrib/drt/routing/DrtRoutingModuleTest.java
+++ b/contribs/drt/src/test/java/org/matsim/contrib/drt/routing/DrtRoutingModuleTest.java
@@ -82,7 +82,7 @@ public class DrtRoutingModuleTest {
 		final double networkTravelSpeed = 0.83333;
 		final double beelineFactor = 1.3;
 		TeleportationRoutingModule walkRouter = new TeleportationRoutingModule(TransportMode.walk, scenario,
-				networkTravelSpeed, beelineFactor);
+				networkTravelSpeed, beelineFactor, null);
 		DrtConfigGroup drtCfg = DrtConfigGroup.getSingleModeDrtConfig(scenario.getConfig());
 		String drtMode = "DrtX";
 		drtCfg.setMode(drtMode);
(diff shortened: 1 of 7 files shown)


steffenaxer:
Do you have a public source for active mode speeds by age?

kainagel:
Thanks a lot.

Wanted to say that in the past (e.g. with bicycle _but_ for routing on the network) we had moved this into the vehicle speed.  I.e. different persons may have different bicycle types, and they have different maximum speeds.

And that is picked up in the router.

I can see, though, that having it as a person attribute is more "natural".

Still, we should keep this in mind when revisit the bicycle contrib or similar.  @dziemke

antonindanalet:
In the Swiss [Mobility and Transport Microcensus (MTMC)](https://www.mzmv.bfs.admin.ch/), the average speed for a trip leg ("Etappe") on foot is 4.22 km/h (+/- 0.03). It is based on 45'149 persons (who did - at least - a trip leg on foot during the reference day).

In the [main report](https://www.are.admin.ch/dam/are/de/dokumente/verkehr/dokumente/mikrozensus/mzmv-hauptbericht2021.pdf.download.pdf/Hauptbericht%20MZMV%202021%20DE.pdf), you also see the speeds by bike, slow ebike and fast ebike (pp.28-29, figure G3.3.1.6).

These are the numbers for 2021. You can also get the results for previous years (2015, 2010, ...).

This value can be differentiated by age (my own calculation based on the MTMC 2021): speed_by_age

jfbischoff:
Thanks for posting, @antonindanalet! 
And thanks for the PR @sebhoerl , we'll certainly use this feature :-)

antonindanalet:
Another source is [Buchmueller, S., Weidmann, U. (2006): Parameters of Pedestrians, Pedestrian Traffic and Walking Facilities, IVT-Report Nr. 132, IVT, ETHZ](https://www.research-collection.ethz.ch/bitstream/handle/20.500.11850/47950/[sr132.pdf](https://www.research-collection.ethz.ch/bitstream/handle/20.500.11850/47950/sr132.pdf?sequence=8&isAllowed=y)?sequence=8&isAllowed=y) (p.17) (indeed from 1993)

antonindanalet:
The code for generating the figure above is here: https://github.com/antonindanalet/simba-python-choice/blob/master/src/simba/mobi/mzmv/pedestrian_speed/descr_stats.py

PR #3943, opened 2025-05-12
fix(core): Make TurnRestrictionsContext independent of link order

marecabo:
Adding some more tests


nkuehnel:
...and a first fix.. (some tests still fail)

We observed that the cleaner sometimes produced correct results and sometimes not, depending on the order of the links in  the network. The reason is the optimization in the turn restriction context builder. So far, when a new turn restriction is processed, existing colored versions of the starting link are checked and eventually re-pointed to the new colored subgraph. However, if - based on the order of the links - the existing colored versions don't exist yet and are thus never updated. Our proposal is to catch the case where an "exit" link of a colored subgraph directly enters into a new colored subgraph of an existing turning restriction and then directly pointing to the existing subsequent colored node.

nkuehnel (approved):

PR #3946, opened 2025-05-13
feat(scoring consistency check): check consistency between routing and scoring parameters

### LINKED ISSUE #3927: Add consistency check if for any mode there are no corresponding scoring parameters
If for any mode there are no scoring parameters, the consistency check should warn the user. Otherwise, it leads to errors like this: https://github.com/orgs/matsim-org/discussions/1230

paulheinr:
Closes #3927.

If there is a routing mode defined but no corresponding scoring parameter, the consistency check will complain.


paulheinr:
Hm, seems to be harder than expected. There are default routing modes which are not used in the simulation. One could add scoring parameters for those as well.

kainagel:
This is a reaction to unreadable error messages (*) during teaching after introducing new routing modes.

Suggestion to instead improve the error messages (*) rather than checking a consistency.  Possibly catch the null pointer exception before it happens.

paulheinr:
fixed in #5070 5071

PR #3949, opened 2025-05-14
fix(drt): fix stats bug introduced in previous PR

nkuehnel:
travel time calculation is now based on actual pickup time instead of beginning of the time window, so no longer need to subtract wait time


mfrawley-moia (approved):

contribs/drt/src/main/java/org/matsim/contrib/drt/analysis/DrtAnalysisControlerListener.java
```diff
@@ -262,7 +262,7 @@ public void notifyIterationEnds(IterationEndsEvent event) {
 					format.format(leg.toCoord.getY()),//
 					leg.waitTime + "",//
 					leg.arrivalTime + "",//
-					(leg.arrivalTime - leg.pickupTime - leg.waitTime) + "",//
+					(leg.arrivalTime - leg.pickupTime) + "",//
```
mfrawley-moia, line 265: since we had negative inVehicleTravelTimes, that might be an assertion we could have? invehicletravelTime >=0?

PR #3955, opened 2025-05-15
feat(drt): refined the rebalancing strategy based on relative demand

ricoraber:
This PR refines the rebalancing strategy based on relative demand. 
Before, the target number of vehicles to send to each zone was  
alpha * target(zone) + beta
where target(zone) was computed based on relative demand for that zone:
First a weight is computed 
w(zone) = demand(zone) / total_demand
and then 
target(zone) = w(zone) * number of rebalancable vehicles.

This only really made sense for beta = 0, because then all the rebalance vehicles could be distributed based on this logic.

But for beta > 0, we already send beta vehicles to each zone, so there are 
beta * number of zones  
fewer vehicles available for rebalancing based on the above logic.

This PR takes this into account and subtracts 
beta * number of zones
vehicles from the number of rebalance vehicles before applying the rebalancing logic based on relative demand.


steffenaxer:
@ricoraber thanks for your contribution. Would it be able to provide a proper short description of this PR.

ricoraber:
Added description for the PR.

nkuehnel (approved):

PR #3959, opened 2025-05-20
feat(drt-extensions): add time based capacity reservation for operation facilities

nkuehnel:
Operation facilities may now reserve vehicle spots for a specific period of time, eg to know in advance when a spot will open up again in the facility

BREAKING CHANGE


nkuehnel:
superseded by https://github.com/matsim-org/matsim-libs/pull/4250

PR #3966, opened 2025-05-22
refactor(freight(carriers)): Use try-with-ressources block instead of manually call executor.shutdown

kt86:
This PR reverts the changes in #3965 and uses another solution: Try-with-ressources.
After talking to KN, this is the better way and it also makes sure, that the threads gets closed.


kt86:
@paulheinr FYI.

PR #3975, opened 2025-06-02
feat: making MultimodalLinkChooserDefaultImpl public so it can be used as delegate

mkreuschner:


kainagel:
For this to be clean, the constructor needs to be package-private.  If there is no constructor, then this means that a package-private empty constructor needs to be added.  Please do so.  Thanks.

rakow:
I think the constructor needs to be public because this class can not be injected via guice directly if somebody wants to replace it. A `MultimodalLinkChooser` can not inject a `MultimodalLinkChooser` into itself as a delegate, because it would create an unresolvable circular loop.

rakow:
I see one can bind the `MultimodalLinkChooserDefaultImpl` and inject that instead. I guess it works, but it is a bit less obvious solution, probably.

kainagel:
Yes, exactly, one needs to bind the implementation and can then use it as a delegate. 

The necessary functionality was already added via some other PR. → closing this PR again.

PR #3976, opened 2025-06-03
feat: enable experienced plans write by default and add matsim application class for postprocessing

paulheinr:
Default for writing experienced plans is now `true`. It’s a bit of a hassle when the file isn’t in the output by default. Additionally added a MATSim application script for postprocessing in these cases.


simei94:
this is really handy, thanks!

PR #3979, opened 2025-06-03
fix(drt): fix timing bugs when using custom stop durations

### LINKED ISSUE #3757: drt: incoherence between InsertionGenerator and CostCalculator for custom stop durations
Currently, variable dropoff durations are not always handled correctly during insertion. We notice clearly that if we put a non-zero dropoff duration (only available in prebooking), that this time is not taken into account in the maximum travel time of the agent. 

The reason is as follows: The logic that checks whether a request fulfills its constraints (DefaultInsertionCostCalculator) uses the arrivalTime from the DropoffDetourInfo from InsertionDetourTimeCalculator. However, this variable, currently, does describe the vehicle's arrival time at the stop, not the "dropoff time" inside the stop. In the standard configuration, both times are equivalent. If we have a non-zero stop duration (used in the PrebookingStopTimeCalculator and PrebookingStopActivity) the "dropoff time" comes a few seconds (= stop duration) later. 

So for the InsertionCostCalculator the dropoff seems to be happening earlier than it actually will. This has the effect that some requests may be inserted that arrive with a delay (= maximum the stop duration) compared to what is allowed, and existing requests may be shifted up to that limit. In turn, an existing request with dropoff time 60s can be shifted by maximum 240s if a request with that stop duration is inserted, for instance. 

I think the same is true for pickups, but we haven't seen it clearly in simulation yet. Right now, the departureTime inside PickupDetourInfo refers to the end time of the respective stop in which the pickup happens. It does not give the actual pickup time durating the stop, on which constraints should be evaluated.. However, I imagine the problem to be kind of inverse: Here the InsertionDetourTimeCalculator would be too restrictive rather than too allowing. Because the expected stop end time is an upper bound to the actual pickup time if there are multiple requests with varying stop durations in the same stop task.

I have a couple of code snippets to fix the issue, coming soon, but putting the PR to not forget about this.

sebhoerl:
Currently, when using custom stop durations (only available when prebooking is active), pickup and dropoff times can be slightly exceeded. The cause of this is that the `InsertionGenerator` and (more severly) the `DefaultInsertionCostCalculator` make the assumption that pickups happen at the end of a stop task and dropoffs happen at the beginning of a stop task. This is not the case any more when custom stop durations are used, so the "vehicle arrival time" at a stop is not any more necessary equal to the "request dropoff time", and the "vehicle departure time" at a stop is not any more necessarily equal to the "request pickup time". 

The present PR fixes the issue by renaming the old departure/arrival times to vehicleDepartureTime and vehicleArrivalTime, and by dragging the requestPickupTiome and requestDropoffTime along the dispatching process.

A comprehensive set of unit tests (`VariableStopDurationTest`) has been added.

Fixes #3757.
Ping #3922.

**TODO**
- [ ] I left some small comments in the `shifts` package. It would be great if you can have a look, I don't really have the deep understanding to say whether, in some places, you'd rather like to test for the "vehicle arrival time" or the "request dropoff time". I think it is the former and implemented it this way, but please verify.

### DIFF
--- a/contribs/drt-extensions/src/main/java/org/matsim/contrib/drt/extension/insertion/ConfigurableCostCalculatorStrategy.java
+++ b/contribs/drt-extensions/src/main/java/org/matsim/contrib/drt/extension/insertion/ConfigurableCostCalculatorStrategy.java
@@ -24,7 +24,7 @@ public class ConfigurableCostCalculatorStrategy implements CostCalculationStrate
 	}
 
 	private boolean checkHardInsertion(DrtRequest request, DetourTimeInfo detourTimeInfo) {
-		return detourTimeInfo.pickupDetourInfo.departureTime <= request.getLatestStartTime()
+		return detourTimeInfo.pickupDetourInfo.requestPickupTime <= request.getLatestStartTime()
 				&& detourTimeInfo.dropoffDetourInfo.requestDropoffTime <= request.getLatestArrivalTime();
 	}
 
@@ -33,7 +33,7 @@ public class ConfigurableCostCalculatorStrategy implements CostCalculationStrate
 			return 0.0;
 		} else {
 			double pickupDelay = Math.max(0,
-					detourTimeInfo.pickupDetourInfo.departureTime - request.getLatestStartTime());
+					detourTimeInfo.pickupDetourInfo.requestPickupTime - request.getLatestStartTime());
 
 			double dropoffDelay = Math.max(0,
 					detourTimeInfo.dropoffDetourInfo.requestDropoffTime - request.getLatestArrivalTime());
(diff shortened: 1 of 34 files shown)


nkuehnel:
@sebhoerl this looks nice, thanks! You are right. I added the dropoff duration to the stop end calculation, which should be more correct, right? Not sure if I miss that the dropoff stop could potentially last longer for existing requests, but I assume then they should have been caught in their respective insertion as well..

nkuehnel:
What I'm checking in the shift insertion class is whether the insertion overlaps with shift breaks or shift ends, so the complete stop duration should be taken into account.

sebhoerl:
There are still some shift tests that are failing now. We just update the test values?

nkuehnel:
@sebhoerl i mistakenly summed up arrival and dropfoff time and fixed it now to only use dropfoff time

sebhoerl:
Thanks!

PR #3981, opened 2025-06-04
feat(freight(carrier)): run CarrierConsistencyCheck when using CarrierUtils.runJsprit()

### LINKED ISSUE #3982: Freight(Carriers) Include ConsitencyCheckers in workflow
Start with placing the checks into CarrierUtils.runJsprit(), since this is the most convenient workflow.

kt86:
This PR now checks the consistency of the carriers when using ``CarrierUtils.runJsprit(...)``.
The ``ConstistencyChecker`` was added in #3888. 

There are two checks:
1.) before running jsprit: Check if the problem is solvable.
2.) after running jsprit: Check if all jobs are in exactly one tour.

I decided to start with just writing an ERROR. This may be changed to throw a RTE. So all the code should be still run through and nothing should break.

On the way, I improved the logging a bit.

There are two methods in ``CarrierUtils`` also doing similar things. I decided to deprecate them, so the users can switch to the more general (and powerful) ``CarrierConstitencyChecker``.
@rewertvsp: Please have a look, if you can replace this usage.

This PR will close #3982.


rewertvsp:
I would prefer to let methods allJobsHandledBySelectedPlan and createListOfCarrierWithUnhandledJobs in CarrierUtils instead of removing them.
The advantage is that you get a return of unsolved carriers or information about a certain carrier, which can be helpful within a simulation. The checker only checks and writes unsolved carriers in the log, which cannot be used in the same run.
That would be my first intuition :)

kt86:
Good Point. I'am fine with that approach.
I suggest, that 
1. I revert the deprecation of that methods.
2. I add a note instead, that this is a huge overlap and this someone :) may merge the things (after thinking a bit on the direction of doing so, e.g. that the checker uses the code in CarrierUtils???, ....)

Is that a go-able way for you?

rewertvsp:
yes. You can go ahead

kt86:
@rewertvsp Please take a look, and feel free to start merging, if you are fine with it.

PR #3986, opened 2025-06-05
feat(drt-extensions): conditional generation of companions

sebhoerl:
This PR adds a feature to the `companions` extensions for DRT that allows to define for which agents companions are generated:

- If an agent has the attribute`drt:companions` and the value is `true`, companions are generated.
- If an agent has the attribute `drt:companions` and the value is `false`, comanions are *not* generated.
- If an agent does *not* have the attribute, use the `generateCompanionsByDefault` value from the config group.


steffenaxer:
@sebhoerl Well you could controll the default behaviour in the companion parameters and set, default to false - no companion generation if activation attribute is absent. So I just need to update the params and switch it to true.

sebhoerl:
Ok perfect, yes, just wanted to see if you're fine with that.
Edit: Ah ok now I understood, yes, I can add a default to the config.

steffenaxer:
Yes you got it right now 😄

PR #3991, opened 2025-06-10
fix(railsim): Extend an existing test

ikaddoura:
... to see if vehicles eventually arrive at the end of the line. Currently the test is failing.


ikaddoura:
Hey @rakow , could you have a look at this. It seems that our trains do not behave as expected... :-(

rakow:
I fixed an issue in the deadlock avoidance so the train should not get stuck anymore.

ikaddoura:
Thank you!!

PR #3997, opened 2025-06-16
feat(freight:carriers): Take roadpricing into account when running jsprit from CarrierUtils.runJsprit(...)

### LINKED ISSUE #1367: Add road pricing to NetworkBasedTrasnportCosts in FreightUtils.RunJsprit(..)
If road pricing is set in the config, it has to be added to the NetworkBasedCostCalculator in

https://github.com/matsim-org/matsim-libs/blob/063be1a154a20c1318a3d19a2374930162f87b23/contribs/freight/src/main/java/org/matsim/contrib/freight/utils/FreightUtils.java#L97-L99

by adding  `.setRoadPricingCalculator(roadPricing)` to the builder.

This is currently missing there, so this method cannot be used for usecases with road pricing enabled.

kt86:
With this PR the Utils class `CarrierUtils.runJsprit(...)` now looks whether a `RoadPricingScheme` exists or not.
If it exists, it will be considered in the calculation of the `NetworkBasedCosts`.
If not, nothing changes.

This allows, to use the utils method also when running scenarios with tolls. (Before that, I used a copy of it and manually added the rpScheme -.-)

Closes #1367

### DIFF
--- a/contribs/freight/src/main/java/org/matsim/freight/carriers/CarriersUtils.java
+++ b/contribs/freight/src/main/java/org/matsim/freight/carriers/CarriersUtils.java
@@ -40,7 +40,6 @@ import org.matsim.api.core.v01.Id;
 import org.matsim.api.core.v01.Scenario;
 import org.matsim.api.core.v01.TransportMode;
 import org.matsim.api.core.v01.network.Link;
-import org.matsim.api.core.v01.network.Network;
 import org.matsim.api.core.v01.population.Plan;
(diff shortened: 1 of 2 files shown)


rewertvsp (approved):
I changed spelling and only one simplification because the case in line 238 is not needed because the default is already null

kt86:
Danke! Auch für die Vereinfachung. Hatte das Handling der `null` erst später in der anderen Methode hinzugefügt ^^

PR #3999, opened 2025-06-17
fix(drt): Round up rebalancing vehicles

ricoraber:
For the rebalancing strategy EstimatedRelativeDemand, the target number of vehicles sent to each zone could be non-integer beforehand. For small total number of vehicles this can result in values between 0 and 1 which is later rounded down to zero, so no rebalancing is happening.

This PR rounds up the number of rebalanced vehicles to avoid this problem.


nkuehnel (approved):
lgtm!

PR #4001, opened 2025-06-17
feat(core): Consider subpopulation within sampling population

rewertvsp:
- samples all subpopulations separately
- if persons have no subpopulation, they are sampled as one subpopulation
- merge both existing PopulationUtilsTest classes
- add simple test for population sampling.
- delete some methods from PopulationUtils and inline the functionality.


rewertvsp:
@paulheinr : Can you have a quick check if this PR is plausible?

PR #4002, opened 2025-06-18
feat(drt-extension): preemptive rejection

sebhoerl:
This PR adds **preemptive rejections** to the DRT extensions. They are an efficient way of controlling overall rejection rates in the following situations:

- *Service segmentation:* A couple of requests in *business class* are rejected more frequently than others because the dispatching constraints are stronger (no pooling, shorter wait time, ...). In that case, the new code allows rejecting requests of other booking classes *preemptively* to decrease business class rejections.

- *Equity:* There exist certain requests that are rejected with higher probability because they are harder to dispatch than others (wheelchair users, larger groups, ...). Then they can be defined as an individual booking class and the new feature can be used to *preemptively* reject standard requests to reduce rejection rates of the special requests.

Technically, the following steps need to be followed:

- Preemptive rejections need to be activated by adding the module:

```java
controller.addOverridingModule(new PreemptiveRejectionModule());
```

- Preemptive rejections need to be activated via configuration:

```java
PreemptiveRejectionParams preemptiveParams = new PreemptiveRejectionParams();
preemptiveParams.setInputPath(path);
drtConfig.addParameterSet(preemptiveParams);
```

- A booking class must be defined per person as a person attribute or by calling `PreemptiveRejectionOptimizer.setBookingClass(person, bookingClass)`

- The input file for preemptive rejections is a JSON file with the following structure:

```json
{
    "rejections": [
        { "start_time": 21600.0, "end_time": 43200.0, "booking_class": "business", "rejection_rate": 0.25 },
        { "start_time": 46800.0, "end_time": 72000.0, "booking_class": "business", "rejection_rate": 0.5 }
    ]
}
```

Requests of the indicated *booking class* during the indicated *time slot* are then rejected *preemptively* according to the rejection rate. This means that the indicated share is rejected by default regardless of whether insertions of the incoming requests can be found or not.

The whole system works best if the preemptive rejection instructions are calibrated by an outside algorithm, but the framework could be extended to self-calibrate, for instance, by trying to reach certain rejection rate thresholds by booking class.

### DIFF
--- a/contribs/drt-extensions/src/main/java/org/matsim/contrib/drt/extension/DrtWithExtensionsConfigGroup.java
+++ b/contribs/drt-extensions/src/main/java/org/matsim/contrib/drt/extension/DrtWithExtensionsConfigGroup.java
@@ -19,17 +19,19 @@
 
 package org.matsim.contrib.drt.extension;
 
+import java.util.Optional;
+import java.util.function.Supplier;
+
 import org.matsim.contrib.drt.extension.companions.DrtCompanionParams;
 import org.matsim.contrib.drt.extension.insertion.spatialFilter.DrtSpatialRequestFleetFilterParams;
 import org.matsim.contrib.drt.extension.operations.DrtOperationsParams;
+import org.matsim.contrib.drt.extension.preemptive_rejection.PreemptiveRejectionParams;
 import org.matsim.contrib.drt.extension.services.services.params.DrtServicesParams;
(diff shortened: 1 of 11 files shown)


nkuehnel:
@sebhoerl this is a nice feature for greedy insertion, thank you! Would it make sense to adapt it such that the configuration is part of the DrtOptimizationConstraintsSet (of which there could be multiple per service)? Because we already use the different sets to distinguish different service classes, maybe it make sense to combine it then

sebhoerl:
Yes that could make sense

PR #4030, opened 2025-06-19
feat: added/updated flow map viz class to pt dashboard class in simwrapper contrib. Tested with SimWrapperTest.

Brendan-Lawton:


frievoe97:
LGTM :)

PR #4092, opened 2025-07-08
fix(ev): Charger power profile iteration reset and update EV routing with AccessEgress support

mattiasingelstrom:
## Summary

This PR includes two key improvements:

1. **Fixes iteration-wise accumulation in charger power profiles**
2. **Updates EV routing with AccessEgress handling**

---

### 🔧 Charger Power Profile Reset
- `ChargerPowerTimeProfileCalculator` now implements `IterationStartsListener` and clears its internal state at the start of each iteration.
- Prevents unintended accumulation of charger power data across iterations.
- Registered in `EvStatsModule` to ensure activation.

### EV Routing Fixes
- `EvNetworkRoutingModule` now filters out non-vehicle (e.g. walk) legs when estimating energy use.
- Routing logic retains the true destination even when intermediate charging activities are added.

---

## Motivation

- Charger power time profiles were previously accumulating across iterations, leading to inflated memory use and incorrect analysis outputs.
- Routing logic could incorrectly interpret intermediate activities as the final destination, especially in multimodal setups.

---

## Test Impact

Note: The updated handling of Access-Egress changes the plan structure by inserting walk legs.

- Existing tests that rely on plan equality (e.g. comparing full route XML structure) fails.
- This is **expected behavior**, and test plans should be updated accordingly.
- No impact on routing correctness or simulation logic


contribs/ev/src/main/java/org/matsim/contrib/ev/routing/EvNetworkRoutingModule.java
```diff
@@ -104,11 +104,20 @@ public List<? extends PlanElement> calcRoute(RoutingRequest request) {
 		final Person person = request.getPerson();
 
 		List<? extends PlanElement> basicRoute = delegate.calcRoute(request);
+		List<Leg> nonWalkLegs = basicRoute.stream()
+			.filter(Leg.class::isInstance) // only Leg elements
```
tschlenther, line 108: Just FYI: THere is also `TripStructureUtils.getLegs(final List<? extends PlanElement> planElements )`

contribs/ev/src/main/java/org/matsim/contrib/ev/routing/TruckEvNetworkRoutingModule.java
```diff
@@ -96,11 +94,14 @@ public List<? extends PlanElement> calcRoute(RoutingRequest request) {
         // Get basic route (no EV logic)
         List<? extends PlanElement> basicRoute = delegate.calcRoute(request);
         Id<Vehicle> evId = Id.create(request.getPerson().getId() + vehicleSuffix, Vehicle.class);
-		Leg basicLeg = (Leg)basicRoute.get(0);
 		// Skip if the main leg is not truck mode (e.g. access/egress walk)
-		if (TransportMode.walk.equals(basicLeg.getMode())) {
-			return basicRoute;
-		}
+		List<Leg> nonWalkLegs = basicRoute.stream()
+			.filter(Leg.class::isInstance) // only Leg elements
```
tschlenther, line 99: see above

contribs/ev/test/input/org/matsim/contrib/ev/example/RunEvExample/config.xml
```diff
@@ -29,7 +29,7 @@
     <module name="controler">
         <param name="outputDirectory" value="./output/evExample"/>
         <param name="firstIteration" value="0"/>
-        <param name="lastIteration" value="0"/>
+        <param name="lastIteration" value="10"/>
```
tschlenther, line 32: does that affect the run time of tests ?

tschlenther:
> As for the failing test:
> 

I now see its the plans file that is tested. Please check+update that file as well..

mattiasingelstrom:
Thanks for the input! I've made the changes, but I haven’t committed them yet since I found something else that might need adjusting.

First, I checked the `plans` and `events` files:

- The `car` legs and `car interaction` charging activities are **identical**.
- The only difference is the added `car interaction` activity and a `walk` leg, both with zero duration.

What I found:

I added the following to the config:

```xml
<module name="routing">
    ...
    <param name="accessEgressType" value="accessEgressModeToLink"/>
</module>
```

However, I then noticed that when running the example using the `...Example.main()` method, the config is loaded from the GitHub path even though a local config file exists. The issue seems to be with the `DEFAULT_CONFIG_FILE` string, which should include the `"contribs/ev/"` root path to work correctly when running the example.

If that is changed, the test then fails, since the path differs when running from the test folder.

This seems easy to fix, but I wanted to check if there’s a preferred approach to ensure consistent behavior between example runs and test executions.

tschlenther:
> However, I then noticed that when running the example using the `...Example.main()` method, the config is loaded from the GitHub path even though a local config file exists. The issue seems to be with the `DEFAULT_CONFIG_FILE` string, which should include the `"contribs/ev/"` root path to work correctly when running the example.
> 
> If that is changed, the test then fails, since the path differs when running from the test folder.
> 
> This seems easy to fix, but I wanted to check if there’s a preferred approach to ensure consistent behavior between example runs and test executions.

If I understand that right, please do not change the path in your commit (or change it back). The config on GitHub will be changed as part of your PR, when merged.

Please also ensure that the test does not run significantly longer than before. 

Thank you :)

tschlenther (approved):

PR #4094, opened 2025-07-08
feat: Update the congestion index calculation

luchengqi7:
The congestion index based on the idea of excess travel time is now added (i.e., comparable to the TomTom congestion index). This index is preferred by VSP, and should replace the previous congestion index. 

To make it backward compatible, the previous congestion index is still kept, but deprecated. For the dashboard (SimWrapper), only the new excess travel time index is displayed. 

To avoid confusion, we name the new congestion index as "excess travel time index", in order to distinguish with the existing (deprecated) congestion index.

@rakow could you please help to check the dashboard part? Shall we replace the old plot or keep the old plot? And is there any quick test I can perform?


luchengqi7:
I have a question and would like to ask for your opinions. 

We want to have a similar congestion index as in TomTom. The index is weighted by the volumes / counts of the roads. The higher the volume (or traffic count) a link has, the more it contributes to the congestion index. 

In our simulation, we have many links with very small or 0 volume during most time of the day. If we only take the links with traffic volumes (more similar to the trip-based approach in TomTom index), then we may end up with very high value in this index (e.g., 20 = 2000%, meaning 20 times of excess travel time). I am not sure if such values are useful or not.

On the other hand, if we don't take volumes into consideration, then the index will be very close to 0, because most of the link does not have any excess travel time (i.e., no volume --> excess travel time ratio = 0). This makes the index not very responsive or helpful. 

Maybe we want to have something in between. What I have implemented in the latest update is as follows:
(a) each link contribute to one unit of base weight
(b) links with volumes will contribute extra weight: 1 volume = 1 unit of extra weight
We then take the weighted average.

With this approach, the value lands in some reasonable range (see attached figure below from the unit test scenario). But here we are introducing some extra parameters (i.e., basic weight). On the other side, we may also include road length when weighting the contribution, with longer links contribute more. 

chart

What do you think? Do you have other better approaches? @kainagel @steffenaxer @rakow

kainagel:
Dear Chengqi, sorry that was dormant for so long.

I don't think that we need an additional specification for the averaging; the travel time index already does that.

A travel time index of 0.2 means that each trip in the average takes 20% longer than with the free speed travel time.  The average here is "over trips".

We now need to break this down into the pieces that we want.  The pieces are "hours" and "link types".  So a travel time index of 0.2 between 9am and 10am for "primary" would mean (in my understanding) that all vehicles that enter primary links between 9am and 10am are, in the average, 20% slower than with free speed.  I.e. you sum up the travel times of all vehicles on those links, and the free speed travel times of all vehicles on those links.  Only at the very end you divide. 

I.e. you do not take an average over links.  

Does this make sense?

luchengqi7:
Dear Prof. Nagel,

Thank you very much for the feedback! Yes, this sounds good. I will adapt that into the calculation, and update the PR. 

Best regards,
Chengqi Lu

luchengqi7:
Dear Prof. Nagel,

I have updated the calculation of the excess travel time index. I have also clean up the dashboard a little bit. Now the excess travel time will replace the old "congestion index". The table is also updated accordingly. 

Please feel free to take a look and then we can merge this PR. If there is any other things I need to change, please let me know!

Thank you very much! 

Best regards,
Chengqi Lu

PR #4097, opened 2025-07-09
fix(drt): Handle negative departure time corner case

steffenaxer:
Fix for a rare corner case situation, that can create index out of bounds error. DepartureTime might get negative caused by a large negative pickupDetourInfo.pickupTimeLoss.

### DIFF
--- a/contribs/drt/src/main/java/org/matsim/contrib/drt/optimizer/insertion/repeatedselective/DetourTimeEstimatorWithAdaptiveTravelTimes.java
+++ b/contribs/drt/src/main/java/org/matsim/contrib/drt/optimizer/insertion/repeatedselective/DetourTimeEstimatorWithAdaptiveTravelTimes.java
@@ -51,6 +51,7 @@ public class DetourTimeEstimatorWithAdaptiveTravelTimes implements DetourTimeEst
 
 	@Override
 	public double estimateTime(Link from, Link to, double departureTime) {
+		departureTime = Math.max(0, departureTime);
 
 		if (from == to) {
 			return 0;
--- a/contribs/dvrp/src/main/java/org/matsim/contrib/zone/skims/AdaptiveTravelTimeMatrixImpl.java
+++ b/contribs/dvrp/src/main/java/org/matsim/contrib/zone/skims/AdaptiveTravelTimeMatrixImpl.java
@@ -108,7 +108,7 @@ public class AdaptiveTravelTimeMatrixImpl implements AdaptiveTravelTimeMatrix {
 
 	@Override
 	public double getTravelTime(Node fromNode, Node toNode, double departureTime) {
-		int bin = this.getBin(departureTime);
+		int bin = this.getBin(Math.max(0,departureTime));
 		Double sparseValue = this.sparseTravelTimeCache.get(this.getSparseTravelTimeKey(fromNode, toNode, bin));
 		if (sparseValue != null) {
 			return sparseValue;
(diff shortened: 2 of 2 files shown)


nkuehnel:
@steffenaxer wouldnt it be more sensible to cap the loss itself further upstream or where the method is called? I would prefer that it actually crashes for negative departure times

steffenaxer:
@sebhoerl should answer this. I'm responsible for this code part 😅. The point is there might be multiple positions at which `getTravelTime` is called. I do not know at which places a negative time could happen. For now I only know [Link](https://github.com/matsim-org/matsim-libs/blob/90431b5beff5929e7b76871d0b9c639dfb7a46a5/contribs/drt/src/main/java/org/matsim/contrib/drt/optimizer/insertion/InsertionGenerator.java#L402C10-L402C33) ff. That might cause a negative departureTime.

sebhoerl:
To give some context: We are in the following situation in the schedule:

```
Start -> P0 -> D0 -> End
```

Now we insert a new request such that 

```
Start -> P1 -> P0 -> D1 -> D0 -> End
```

The pickup loss describes by how far everything behind P1 will be shifted because P1 has been inserted. By default, this loss is always positive, because we push things to the future. But our travel time estimation is rough (matrix-based), so the following can happen: 

- We estimate the existing part `-> P0 ` as T0.
- We estimate the new inserted part `-> P1 -> P0` as T1.
- The pickup slack is  S = T1 - T0, and usually we have T0 < T1. To insert D1, we need to estimate P0 -> D1 and we need to provide a departure time. Usually this would be the currently planned departure time from P0 plus the slack.
- However, since our travel time estimation is coarse (and dynamic in the case of the RepeatedSelective approach), it can happen that the slack becomes negative. This means that the departure from P0 is now assumed *earlier* than it is currently planned.
- If we are now also really early in the simulation, it can happen that this departure time is actually negative.

**It would be good to construct a unit test that actually shows this behaviour. For now, this is the best guess hypothesis.**

Regardless of any of this, I think the clean approach would be to cap the travel time to zero in the estimator. This would be consistent with the general travel time estimation in MATSim (see `TimeBinUtils.getTimeBinIndex` and its use in the `org.matsim.core.trafficmonitoring` package).

nkuehnel:
@sebhoerl sound reasonable!

> Regardless of any of this, I think the clean approach would be to cap the travel time to zero in the estimator. This would be consistent with the general travel time estimation in MATSim (see TimeBinUtils.getTimeBinIndex and its use in the org.matsim.core.trafficmonitoring package).

I agree that the capping should be further upstream. But it probably shouldn't be the travel time (since we never obtain one and crash during the lookup in the adaptive matrix) but rather just cap the departure time inside the estimateTime method, since this is what becomes negative (maybe you meant that):
```

	@Override
	public double estimateTime(Link from, Link to, double departureTime) {

		if (from == to) {
			return 0;
		}

		double duration = FIRST_LINK_TT;
		duration += this.getTravelTime(from.getToNode(), to.getFromNode(), departureTime + duration);
		duration += VrpPaths.getLastLinkTT(travelTime, to, departureTime + duration);
		return duration / speedFactor;
	}
```

..which means that the departureTime argument here should be capped at 0. @steffenaxer WDYT? This is in `DetourTimeEstimatorWithAdaptiveTravelTimes` class

steffenaxer:
Fine for me. I will change my PR. Thanks for review guys 😃

PR #4117, opened 2025-07-15
feat(aws): Add AWS contrib

nkuehnel:
Adds a new small contrib with AWS SDK dependencies.

So far, it introduces 
- a startup hook to register a URI resolver which allows to use AWS S3 URIs as input paths in the config (see test for how it works)
- a static utility method to obtain an S3 input stream for further parsing

By doing that, a config may have paths like:
"s3://<bucket-name>/network.xml.gz"

..which allows streaming input from AWS on-the-fly without the need of manually syncing content before each simulation.

Syncing output back is not as straightforward, as there is no simple output stream writer. However, we will add a shutdown hook later, which can optionally sync back from the local output directory, if desired.

see also 
https://docs.aws.amazon.com/AmazonS3/latest/userguide/UsingObjects.html

### DIFF
--- a/contribs/aws/pom.xml
+++ b/contribs/aws/pom.xml
@@ -5,9 +5,8 @@
     <modelVersion>4.0.0</modelVersion>
     <parent>
         <groupId>org.matsim</groupId>
-        <artifactId>matsim-all</artifactId>
+        <artifactId>contrib</artifactId>
         <version>2026.0-SNAPSHOT</version>
-        <relativePath>../../pom.xml</relativePath>
     </parent>
 
     <artifactId>aws</artifactId>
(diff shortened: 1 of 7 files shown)


nkuehnel:
We recently open sourced our AWS setup as well, hopefully someone finds it useful:
https://matsim.org/news/2025/matsim-aws-example-setup-repository-published/

https://github.com/moia-oss/matsim-aws

steffenaxer (approved):

jfbischoff:
Nice work, even if we can't use it as our compancy policy has it's own ideas about how to setup an AWS instance :-D
The PR creates problems in IOUtils on Windows, I'm just about to fix this.

nkuehnel:
@jfbischoff not sure what you mean, this code is not about setting up an instance but about communicating with S3 more directly, am I missing something? :) 
curious to know if anything and what prevents you from using that

nkuehnel:
ah and thanks for fixing the windows stuff!

jfbischoff:
Yeah, you're right, my comment was more about the things in https://github.com/moia-oss/matsim-aws - indeed we have been handling s3 quite similarily for a while, so we'll probably switch to this implementation.

PR #4119, opened 2025-07-15
feat: Add a new constructor for LanduseOptions

luchengqi7:
We provide LanduseOptions a new constructor, so that we can create LanduseOptions also in the script, with default options. 

Btw, I haven't changed the XYTData. It seems that IntelliJ changed that somehow by itself... Shall I revert that change? @rakow


rakow:
Please revert any other changes not related to landuse.

luchengqi7:
The unnecessary changes are now reverted.

rakow (approved):

PR #4147, opened 2025-07-28
fix(core):fix IOUtilsTest on Windows after PR4117

jfbischoff:
some URI-URL-related shenanigan

### DIFF
--- a/matsim/src/test/java/org/matsim/core/utils/io/IOUtilsTest.java
+++ b/matsim/src/test/java/org/matsim/core/utils/io/IOUtilsTest.java
@@ -42,7 +42,7 @@ import java.nio.file.Paths;
  */
 public class IOUtilsTest {
 
-	@RegisterExtension private MatsimTestUtils utils = new MatsimTestUtils();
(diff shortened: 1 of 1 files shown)


matsim/src/test/java/org/matsim/core/utils/io/IOUtilsTest.java
```diff
@@ -547,16 +547,16 @@ void testGetBufferedWriter_withPlusInFilename() throws IOException {
...
-		System.out.println(url.toString());
+		System.out.println(url);
 	}
 
 	@Test
 	void testResolveFileOrResource() throws URISyntaxException, IOException {
 
 		File jarFile = new File("test/input/org/matsim/core/utils/io/IOUtils/testfile.jar");
-		String jarUrlString = "file:" + jarFile.getAbsolutePath(); // URLs require absolute paths
+		String jarUrlString = jarFile.toURI().toString(); // URLs require absolute paths
```
jfbischoff, line 559: this is what's actually changed.

Issue #3895, opened 2025-04-09
Rejection handling when multiple DrtOptimizationConstraintsSet exisits


luchengqi7:
When running DRT simulations with multiple DrtOptimizationConstraintsSet, I noticed interesting behavior regarding request rejections. While I have set the rejectRequestIfMaxWaitOrTravelTimeViolated to false for all my custom constraint sets, there are still rejections in the simulations. 

After digging into the matsim-lib, I have realized that in the line 135-139 in DrtModeOptimizerQSimModule, whether rejection is allowed is based on the defaultOptimizationConstraintsSet, which is automatically created (with rejectRequestIfMaxWaitOrTravelTimeViolated = true) in the DrtModeOptimizerQSimModule. This may cause some confusions: while the wait time/ride time constraints are based on the custom constraint set, the rejection behavior is based on another constraint set (namely the default one). 

What I would suggest is that we make the parameter rejectRequestIfMaxWaitOrTravelTimeViolated one level higher, to the DrtOptimizationConstraintsParams. Since we only have one single setup for rejection, it would be good to have it in a higher level. @nkuehnel What do you think? 😀

nkuehnel:
Good point. Another idea would be to actually use per constraint set rejections by letting the CostCalculationStrategy decide that on the fly. This way, one could set rejections false in areas with low pt accessibility and to true in areas where drt is more of a comfortable "addon"

nkuehnel:
see https://github.com/matsim-org/matsim-libs/pull/4161

PR #4161, opened 2025-07-30
feat(drt): per request rejection / drt route constraint promotion

nkuehnel:
@luchengqi7 this PR finally introduces the consideration of constraint set dependent rejection settings (as discussed during MUM25). By doing that, we actually end up being able to decide whether we allow rejections on a request level (i.e., service classes, "premium" vs "basic service" requests, spatiotemporal differentiation..etc,..)

To do so, I also streamlined constraint handling in DRT a bit further and propagated the DrtConstraint object further up, so we copy the constraints even fewer times and re-use the object by attaching it to the request itself. This also allows us to have a more stable drt request interface in the future, because we will not need to touch it when we introduce new constraints, as they can be encapsulated in the immutable DrtRouteConstraint (record) object.

If the general idea is considered useful, it may be good to add a builder for the DrtRouteConstraint class so it is easier to make copies and adjust them slightly.

The only caveat is that this again updates the DrtRouteDescription stuff (and probably breaks the old formats, as the constraints work more on time contract stamps than on abstract max wait time and max travel time), which I dont exactly know what it is used for, maybe someone else can judge it (@marecabo @steffenaxer )

Leaving it for here for discussion

BREAKING CHANGE


nkuehnel:
ah, for group bookings, the implementation currently disallows rejections as soon as one person may not be rejected within the booking.

nkuehnel:
fixes #3895

mfrawley-moia (approved):

marecabo (approved):

PR #4167, opened 2025-08-04
feat(PtFareConfigGroup): add typed getter for different PtFareParams types

simei94:


simei94:
@kainagel is this what you meant by your comments in PtAndDrtFareModule in matsim-lausitz? One could also imagine just having a getter for PtFareParams instead, but I am unsure how to maintain the order of PtFareParams then. Although I do not really understand why the order is important when binding.

simei94:
@paulheinr I just found this PR laying around unmerged. Do you see any reason for not merging it?

paulheinr:
No, thank you for pushing this!

PR #4179, opened 2025-08-08
feat: Only throw an exception if validation is switched on (default).

JWJoubert:
#### Validating `xml` plans

Check if validation is not switched off before throwing an exception.

#### Case in point

The `PopulationReaderMatsimV6` requires, by default, that all activities either has a `Id<Link>`, a `Coord` or a `Id<Facility>` set. But, as mentioned in the Javadoc of the [`MatsimXmlParser`](https://github.com/matsim-org/matsim-libs/blob/d1e7b191275410c54032532a2e697d02cc5b3ba6/matsim/src/main/java/org/matsim/core/utils/io/MatsimXmlParser.java#L119), during development, there is a use case to switch off validation.

However, the `V6` parser [does not check](https://github.com/matsim-org/matsim-libs/blob/d1e7b191275410c54032532a2e697d02cc5b3ba6/matsim/src/main/java/org/matsim/core/population/io/PopulationReaderMatsimV6.java#L332) for this flag before throwing an exception, meaning that when you (deliberately) want/need to switch off validation, you still keep getting exceptions.

This pull request checks if the validation flag (of the `super`) is set.


JWJoubert:
The required functionality is still not achieved. The `PopulationReaderMatsimV6` does not honour the `.setValidation(false)` .

PR #4184, opened 2025-08-11
feat: move drt offline optimizer to vsp contrib

kainagel:


kainagel:
This was resolved by first moving the matsim version up in the separate repo, and only then moving it over to the vsp contrib.  See PR #4961 .

PR #4188, opened 2025-08-12
feat(discrete_mode_choice): detailed utilities writer

Dib-AEK:
**Summary**
Adds a new boolean config attribute `writeDetailedUtilities` (default: `false`).
When enabled, it writes a CSV with all considered mode options and their utilities for each considered tour, and which of the tours is finally selected for the agent plan.

**Purpose**

* Supports calibration of the discrete mode choice model
* Helps analyze and debug mode choice decisions

**Impact**

* No change to default behavior

**Config Example:**

```hocon
DiscreteModeChoice { 
     selector:MultinomialLogit {
          writeDetailedUtilities = true
                                 }
                    }
```

### DIFF
--- a/contribs/discrete_mode_choice/src/main/java/org/matsim/contribs/discrete_mode_choice/model/utilities/MaximumSelector.java
+++ b/contribs/discrete_mode_choice/src/main/java/org/matsim/contribs/discrete_mode_choice/model/utilities/MaximumSelector.java
@@ -1,5 +1,9 @@
 package org.matsim.contribs.discrete_mode_choice.model.utilities;
 
+import org.matsim.api.core.v01.population.Person;
+import org.matsim.contribs.discrete_mode_choice.model.DiscreteModeChoiceTrip;
+
+import java.util.List;
 import java.util.Optional;
 import java.util.Random;
 
@@ -7,7 +11,7 @@ import java.util.Random;
  * The maximum utility selector collects a set of candidates with a given
  * utility value and then selects the one with the highest utility. Internally,
  * always only the best candidate is held.
- * 
+ *
  * @author sebhoerl
  */
 public class MaximumSelector implements UtilitySelector {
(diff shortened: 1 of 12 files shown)


contribs/discrete_mode_choice/src/main/java/org/matsim/contribs/discrete_mode_choice/model/utilities/MaximumSelector.java
```diff
@@ -33,7 +37,7 @@ public Optional<UtilityCandidate> select(Random random) {
 
 	public static class Factory implements UtilitySelectorFactory {
 		@Override
-		public UtilitySelector createUtilitySelector() {
+		public UtilitySelector createUtilitySelector(Person person, List<DiscreteModeChoiceTrip> tourTrips) {
 			return new MaximumSelector();
```
balacmi, line 41: Why are the `person` and `tourTrips` not passed to the `MaximumSelector`? That one should also record utilities if desired. The same for all the other possible selectors: i.e.,`RandomSelector`

balacmi (changes requested):
Thank you, @Dib-AEK, for these changes; please take a look at my comments.

sebhoerl (commented):
General comment, taking up Milos' comments: It's not really necessary to modify all the `UtilitySelector` implementations. It would be cleaner to just write a decorator for a `UtilitySelector` (e.g. `WritingUtilitySelector`) that takes another `UtilitySelector` as an argument. Then the custom logic can happen in `addCandidate` and `select` and the call is forwarded to the delegate.

In fact, the decorator pattern can already be applied to the factory of the `UtilitySelector`. So a `WritingUtilityFactory` can delegate to any other specific implementation.

Also referring to Milos' comment, I think it is better to collect the candidates in each call to `addCandidate` and then only perform the writing in `select`.

sebhoerl (changes requested):
General comment, taking up Milos' comments: It's not really necessary to modify all the `UtilitySelector` implementations. It would be cleaner to just write a decorator for a `UtilitySelector` (e.g. `WritingUtilitySelector`) that takes another `UtilitySelector` as an argument. Then the custom logic can happen in `addCandidate` and `select` and the call is forwarded to the delegate.

In fact, the decorator pattern can already be applied to the factory of the `UtilitySelector`. So a `WritingUtilityFactory` can delegate to any other specific implementation.

Also referring to Milos' comment, I think it is better to collect the candidates in each call to `addCandidate` and then only perform the writing in `select`.

sebhoerl:
And please add a unit test.

PR #4189, opened 2025-08-12
feat(drt): add more flexible waypoint mechanics

nkuehnel:
Adds waypoint factories to be able to plug in new type of stop waypoints that don't necessarily rely on requests.
This will ease configuring operational stops and preplanned fixed stops without bookings

BREAKING CHANGE


steffenaxer:
Could you please explain what is expected to break, how are simulations effected?

nkuehnel:
@steffenaxer 
This is a breaking API change: VehicleEntry.stops now exposes StopWaypoint instead of Waypoint.Stop

nkuehnel:
Besides that, no simulation results should be affected really

PR #4201, opened 2025-08-14
feat(drt): relocation with preplanned/prebooked stops

nkuehnel:
Adds the possibility to relocate a vehicle that is idle, but still has other tasks in the (distant) future.
The relocation takes care of driving to the target relocation link but also ensures that the vehicle drives back to the next fixed stop in time. If the vehicle cannot make it to the next stop in time, the relocation is not performed.

Includes a few test that should cover most of the important scenarios, including when the relocation target equals the link of the next fixed stop, when there is not enough idle time to include the relocation + driving back.

Why is this important?
- currently no relocation of vehicles that a have prebooked stop in the future, even if it is hours away
- allows to plan shift breaks and changeovers at the beginning of a shift without losing the possibility of relocation

### DIFF
--- a/contribs/drt/src/main/java/org/matsim/contrib/drt/scheduler/EmptyVehicleRelocator.java
+++ b/contribs/drt/src/main/java/org/matsim/contrib/drt/scheduler/EmptyVehicleRelocator.java
@@ -19,67 +19,156 @@
 
 package org.matsim.contrib.drt.scheduler;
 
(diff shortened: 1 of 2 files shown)


nkuehnel:
The actual employment in simulations requires an additional small change in the drt schedule inquiry

sebhoerl (approved):

PR #4211, opened 2025-08-16
fix(core): Fix iterator condition and add junit test

steffenaxer:


mrieser:
The fix is correct per-se. 

I think this iterator could be completely deleted. It is nowhere used, it does not implement an interface. It only returns the internal order of the nodes, not even the stored values.

My guess is that it is a rest of some debugging I did while developing the class. It should just be deleted instead.

PR #4214, opened 2025-08-19
fix: add check for correct bicycleMode before calculating additional score

simei94:
We are using [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/), so that we are able to auto generate a changelog. The format is described below:

#### Title
`<type>[optional scope]: <description>` e.g.: fix(QSim): Correctly calculate travel times on links

Available types:
 - feat: A new feature
 - fix: A bug fix
 - docs: Documentation only changes
 - style: Changes that do not affect the meaning of the code (white-space, formatting, missing semi-colons, etc)
 - refactor: A code change that neither fixes a bug nor adds a feature
 - perf: A code change that improves performance
 - test: Adding missing tests or correcting existing tests
 - build: Changes that affect the build system or external dependencies (example scopes: gulp, broccoli, npm)
 - ci: Changes to our CI configuration files and scripts (example scopes: Travis, Circle, BrowserStack, SauceLabs)
 - chore: Other changes that don't modify src or test files
 - revert: Reverts a previous commit

#### Body
Optionally, describe the change in more detail.

If the PR introduces a breaking change write: BREAKING CHANGE at the bottom of the message

### DIFF
--- a/contribs/bicycle/src/main/java/org/matsim/contrib/bicycle/AdditionalBicycleLinkScore.java
+++ b/contribs/bicycle/src/main/java/org/matsim/contrib/bicycle/AdditionalBicycleLinkScore.java
@@ -1,7 +1,9 @@
 package org.matsim.contrib.bicycle;
 
+import org.matsim.api.core.v01.Id;
 import org.matsim.api.core.v01.network.Link;
+import org.matsim.vehicles.Vehicle;
 
 public interface AdditionalBicycleLinkScore{
-	double computeLinkBasedScore( Link link );
+	double computeLinkBasedScore(Link link, Id<Vehicle> vehicleId, String bicycleMode );
 }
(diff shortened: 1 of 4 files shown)


simei94:
@dziemke Could you do me a favor and do a quick sanity check on this? I am pretty sure that without my fix all network modes would receive additionalBicycleScores. I checked this with a test on another branch and the "car" agent received the cycle scores, too. With this here this won't happen in the future. Or am I not getting something on how the BicycleContrib is plugged together and a case with car receiving bicycle scores cannot happen?

simei94:
@kainagel @dziemke How are the reference plans and events for the tests in class BicycleTest created? After my fix on this branch, the test BicycleTest.testLinkVsLegMotorizedScoring() fails because the reference plan scores used include car legs which are scored with bicycle scores for surface type etc. This is wrong and after my fix this does not happen anymore, but the reference plans have to be updated. Please have a look into test/input/testLinkVsLegMotorizedScoring/output_events.xml.gz and search for person score events with car vehicles. You will find car legs being scored with bicycle scores. 

Thus, we need to re-create the referende data in the test/input dir. Please tell me how to do that, I could not find a method to create the reference simulation output.

PR #4222, opened 2025-08-21
fix(drt): Move StopWaypointFactory to central point

steffenaxer:
Some custom modules provide this binding, but the most central one `EDrtModeOptimizerQSimModule ` is not binding the  `StopWaypointFactory.`


steffenaxer:
@nkuehnel The `DrtModeOptimizerQSimModule` already provides the `StopWaypointFactory`. If I would ltike to do so with the 
`EDrtModeOptimizerQSimModule` one is getting errors that the `StopWaypointFactory` is already bound. It doesn't look streamlined if for non Edrt it goes over the default module and for Edrt one need to add this binding manuelly. Could we make this smarter?

steffenaxer:
Ok I think I got it @nkuehnel. I investigated the situation and sometimes the `StopWaypointFactory` have been bound in a `AbstractDvrpModeQSimModule` or a regular `AbstractDvrpModeModule`. So imho there was no need to bind it within a `AbstractDvrpModeQSimModule`.

So the initial binding point is now in the DrtModeModule - that is called verytime. Shifts are able to overwrite it in the DrtShiftsModule.

PR #4229, opened 2025-08-26
feat(drt): Schedule inquiry updates for inter-task idle times

nkuehnel:
follow up to https://github.com/matsim-org/matsim-libs/pull/4201

it allows to schedule hub returns and vehicle rebalancing for idle times _between_ tasks, e.g. before prebooked stops.

it also streamlines the schedule inquiry to include the hub return min timeout (or time elapsed) which previously lived within the optimizer. The timeout is also configurable for rebalancing. This means:
- there is a timeout configurable for rebalancing and for hub returns, meaning that the stay task has to pass for the provided amount of time before the vehicle is considered idle for relocation
- there is a min idle time gap configurable for both relocations to define how big an inter-task time gap has to be to allow the relocation

Tests included


contribs/drt-extensions/src/test/java/org/matsim/contrib/drt/extension/edrt/run/RunEDrtScenarioIT.java
```diff
@@ -129,7 +129,7 @@ void testMultiModeDrtDeterminism() {
 
 		controller.run();
 
-		assertEquals(1926, tracker.passengerPickupEvents);
+		assertEquals(1919, tracker.passengerPickupEvents);
```
nkuehnel, line 132: This small difference is explained by the fact that the previously existing hub return timeout was not inclusive, i.e. a timeout of "60" means that at least 61 seconds had to pass. This is now changed such that it is "at least" 60s. Otherwise, if the timeout is set to "0" you would still need to have a stay task of at least 1s, which means one can't really model a non-limiting timeout

steffenaxer (approved):
LGTM. Thanks for those helpful improvements.

PR #4250, opened 2025-09-03
feat(drt-extensions): operation facility with capacity management + condensed shift starts

nkuehnel:
more or less final follow up of #4229 #4201 #4199 #4195 #4189 as a major drt shifts / operations overhaul.

Adds time based reservations to operation facilities that can be reserved for certain time periods in advance.

Drt driver shift breaks and ends are now planned in advance right at the start of the shift. This way, future insertions (even prebookings) should correctly ensure shift (break) times. Also, the time based capacity at operation facilities ensures that capacity for ending shifts can be reserved at the start of the shift.

Next to reserving capacity at facilities, vehicles need to check in / check out to signal presence. Check in and out events can be thrown.

@steffenaxer DRT services so far ignore the new reservation system.

BREAKING CHANGE


nkuehnel:
- Added a couple of tests.
- merged most of the operations/shift drt code with their respective edrt versions to reduce complexity and duplication. in consequence, the whole extension is compatible with standard and e-drt by default - up to the point that it is in theory possible to simulate mixed fleets of ev/ice vehicles with operations/shifts
- charging activites can be dynamically hooked up unto existing operational stops, without the need of replacing/duplicating tasks. even already running tasks should be supported

contribs/drt-extensions/src/test/java/org/matsim/contrib/drt/extension/operations/shifts/schedule/ChangeoverActivityTest.java
```diff
@@ -0,0 +1,135 @@
...
+    @BeforeEach
+    public void setUp() {
+        when(changeoverTask.getEndTime()).thenReturn(endTime);
+        Map<Id<Request>, AcceptedDrtRequest> emptyRequests = new HashMap<>();
+        activity = new ChangeoverActivity(passengerHandler, driver, changeoverTask, emptyRequests, emptyRequests);
+    }
+
+    @Test
+    public void testGetActivityType() {
+        assertEquals("Shift changeover", activity.getActivityType());
```
mfrawley-moia, line 46: if you had the enum, the tests would be easier to maintain here regarding activity type

mfrawley-moia (commented):
I'd suggest the string enums for any kind of small well-defined set of types/actions. I like the use of optionals.

PR #4256, opened 2025-09-05
feat(simwrapper): allow dashboard provider to be bound via guice

paulheinr:
This PR introduces the possibility to bind dashboard providers with guice. Currently, they can only be injected with SPI. I find this relatively confusing as the main dependency injection framework in MATSim is guice. 

I am even thinking of deleting the SPI injection functionality. What do others think about that? @rakow as you have implemented it, I'd be happy to hear your thoughts.

This fix introduces a new config switch which needs to be set in order to make loading via spi work.


rakow:
I'm ok with adding additional providers via injection, but please don't change the default behavior. You will break nearly every scenario.

The injector has the disadvantage that you need a full MATSim scenario, setup the controller, install all the right modules for your dashboards. 

The current SimWrapper functionality only needs the MATSim config to work. The SPI allows defining default dashboards much more easily than Guice. For project specific dashboards, a Guice binding is probably sufficient and a bit easier to setup. Dashboards can already be bound via Guice, but adding Providers as well might be good.

paulheinr:
Okay, I see your point. I adapted the changes in a way that existing code doesn't break

PR #4264, opened 2025-09-09
feat: Scenario Comparison Dashboard

Brendan-Lawton:
The dashboard is being generated as long as one sets a base path to a base case output folder (See ScenarioComparisonDashboardTest.java). Currently, it takes the difference of trips and emissions data. For this reason, the analysis classes for the trips and emissions dashboards need to run before the scenario comparison. I set the priority of those dashboard classes to 1 and made the scenario comparison dashboard's priority -20. This seems to fix the issue, but I'm not completely certain. This should be tested more.

### DIFF
--- a/contribs/simwrapper/src/main/java/org/matsim/simwrapper/dashboard/EmissionsDashboard.java
+++ b/contribs/simwrapper/src/main/java/org/matsim/simwrapper/dashboard/EmissionsDashboard.java
@@ -93,4 +93,9 @@ public class EmissionsDashboard implements Dashboard {
 		// TODO: emissions by vehicle type
 
 	}
+
+	@Override
+	public double priority() {
+		return 1;
+	}
 }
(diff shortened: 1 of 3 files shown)


Brendan-Lawton:
I rebased my branch with main before pushing so there are 95 commits in this PR. That must've been the wrong procedure for this. The two commits that are for the scenario comparison are: 

@Brendan-Lawton
[ScenarioComparison class is functioning correctly for the most part. …](https://github.com/matsim-org/matsim-libs/pull/4264/commits/39661c62b654aaa20444999b1a3b13116dcd0bdb) 
[39661c6](https://github.com/matsim-org/matsim-libs/pull/4264/commits/39661c62b654aaa20444999b1a3b13116dcd0bdb)
@Brendan-Lawton
[added different priorities for these dashboards to force their analys…](https://github.com/matsim-org/matsim-libs/pull/4264/commits/5a5ac8792150d2f8e101fdecc0f791f6abbc6d45) 
[5a5ac87](https://github.com/matsim-org/matsim-libs/pull/4264/commits/5a5ac8792150d2f8e101fdecc0f791f6abbc6d45)

Brendan-Lawton:
I will make a new PR once I fix merge issues.

PR #4282, opened 2025-09-18
feat(NetworkCreation): make addition of turn restrictions optional (SumoNetworkConverter)

simei94:
docs(NetworkUtils): add comment re computation speed of cleaning process

### DIFF
--- a/contribs/analysis/pom.xml
+++ b/contribs/analysis/pom.xml
@@ -31,7 +31,7 @@
         <dependency>
             <groupId>org.apache.httpcomponents.client5</groupId>
             <artifactId>httpclient5</artifactId>
-            <version>5.5</version>
+            <version>5.5.1</version>
         </dependency>
     </dependencies>
 </project>


simei94:
@marecabo I saw that you implemented the NetworkUtils.cleanNetwork method. Do you have any idea why it is that slow / ideas on how to improve this? Or is it only slow for me? I am using class CreateNetworkFromSumo and especially the 2nd call of the above method is slow. I ran it on our computer cluster and the class took 10+ hours to finish.

marecabo:
Hi @simei94 
hm, so far I have not observed such long run times even for large city networks. Can you use VisualVM as a profiler to see where it spends a lot of time?

simei94:
I'll try that out and will come back to you! FYI: I am trying to convert a network for Dresden, which has a detailed network in the city, a coarse one around the city and a network of germany (Bundesstraße and Autobahn only). The network has 262026 nodes and 433445 links. Is this considerably bigger than the networks you used the clean method with?

simei94:
image

Hi @marecabo, apparently the network with included turn restrictions is much slower when cleaning with MultiModalNetworkCleaner than without the turn restrictions. Also, I saw that in my network with around 450k links about 245k links are included in TurnRestrictionsContexts.replacedLinks. So more than half of all the links. To me, this seems to be too high. 

What size do the networks you clean usually have? We @VSP usually have networks for whole cities including motorways + primary roads for Germany. Is this significantly bigger than the networks you use?

simei94:
I now made the addition of turn restrcitions optional when using SumoNetworkConverter / CreateNetworkFromSumo. However, the default is that they still are added, so nothing changes for users, but one has the possibility to switch off TurnRestrictions. I added this optionality as we (VSP) usually ignore turn restrictions anyways and the addition of turn restrictions seems to slow down network cleaning by a lot, see above. I still think that one would need to take a closer look onto why network cleaning is so slow with turn restrictions.

marecabo:
Hello @simei94, thanks for taking the time with VisualVM. Do you have pictures/stats(# of link, # of links w/ turn restrictions) of before and after cleaning your network? Eventually, it might have many unconnected "islands" which makes cleaning slow(?)

Our networks are usually in the (low-medium) 7-digit # of links, are generated with [pt2matsim ](https://github.com/matsim-org/pt2matsim) and have turn restrictions enabled. We have not observed such a significant increase in runtime as you have. Maybe the SumoConverter does things differently? I was told it produces more detailed intersections?

nkuehnel:
Hi @simei94 I wouldn't rule out that the input is flawed. If the network becomes severely disconnected because of turn restrictions, you will get thousands of disconnected graphs which the cleaner has to walk through. Did the result in the end look ok? 
If you have an example we can also have a look at debugging it.

simei94:
@nkuehnel The above problem appears for at least 2 different downloaded maps of germany (from OSM). The OSM maps are processed with the same pipeline until a ready to use MATSim network is created. I have not found the time to systematically compare networks with/without turn restrictions yet, but I may find some time in the coming weeks. I'll let you know.

PR #1453, opened 2021-04-15
Distributed mobsim

milekj:

PR #4255, opened 2025-09-04
feat(core): Distributed MATSim

rakow:
This PR integrates **DSim**, a distributed, message-passing mobility simulation designed as a JVM-native replacement for QSim in large-scale MATSim scenarios. It introduces a new parallelization strategy that partitions the network and distributes workloads across processes and compute nodes, while maintaining compatibility with MATSim’s existing interfaces.  

## Implementation  
- **Domain decomposition:** Simulation load is partitioned (e.g., METIS-based) according to expected vehicle flows.  
- **Message passing:** Vehicles and link capacities are exchanged across partition boundaries as messages, limiting communication to neighboring partitions.  
- **Execution model:**  
  - Each JVM instance manages multiple simulation processes plus a communicator, events manager, and task scheduler.  
  - Supports both single-node (multi-process) and multi-node distributed execution.  
- **Compatibility:** Reuses and extends MATSim interfaces; existing event handling remains supported. However, non-distributed event handlers and simulation processes will reduce the achievable speed-up.

## Outlook  
- Further optimization of event handling and scoring for distributed execution.

## References  
- See [README](contribs/distributed-simulation/README.md)
- See also [presentation slides](https://matsim.org/conferences/mum2025/presentations/laudan_dsim.pdf) and [MUM25 Paper](https://matsim.org/conferences/mum2025/abstracts/MUM25_paper_18.pdf)
- [Laudan et al. (2025): *High-Performance Mobility Simulation: Implementation of a Parallel Distributed Message-Passing Algorithm for MATSim*](https://doi.org/10.3390/info16020116)

PR #4286, opened 2025-09-19
build: Update to Java 25

rakow:
This PR updates MATSim to Java 25. No changes to the code have been necesarry, only the version numbers in CI and Maven has been increased.

Java 25 is an LTS Release with support for at least 5 years. 
Since Java 21 (MATSim's current version) a lot of new features have been added.  Some of these (e.g the new Foreign Function Interface) are particually relevant for DSim (#1453) which currently needs a newer version to run.

I would like to encourage everyone to install java 25 on their machines already so that we can upgrade MATSim rather soonish if there are no objections.

### DIFF
--- a/.github/workflows/code-coverage.yaml
+++ b/.github/workflows/code-coverage.yaml
@@ -21,7 +21,7 @@ jobs:
       - name: Setup Java
         uses: actions/setup-java@v4
         with:
-          java-version: 21
+          java-version: 25
           distribution: 'zulu'
           cache: 'maven'
 
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -21,7 +21,7 @@ jobs:
       - name: Setup Java
         uses: actions/setup-java@v4
         with:
-          java-version: 21
+          java-version: 25
           distribution: 'zulu'
           cache: 'maven'
 
--- a/.github/workflows/deploy-on-pr-merge.yaml
+++ b/.github/workflows/deploy-on-pr-merge.yaml
@@ -22,7 +22,7 @@ jobs:
       - name: Setup Java
         uses: actions/setup-java@v4
         with:
-          java-version: 21
+          java-version: 25
           distribution: 'zulu'
           cache: 'maven'
           server-id: 'matsim-releases'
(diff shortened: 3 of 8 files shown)


paulheinr:
Java 25 is now also installed on our math cluster -- I don't see a problem with merging this PR. @kainagel any objections?

jfbischoff:
I'd say we can merge this now.

steffenaxer:
Our bamboo agents will have Java 25 available within the next two weeks. 
Would it be possible to wait till we say go ❓

paulheinr:
Fine for me!

Janekdererste:
@steffenaxer you can still run your simulations with your current matsim version. As long as you don't use a SNAPSHOT build, you should still be fine until you update to a more recent version. Does this apply to you, or can we move forward and you update once your infrastructure is ready?

marecabo:
@Janekdererste 
We update to the latest weekly build every Monday and currently do not maintain an internal fork of MATSim. Instead, we contribute fixes and new features of us and our partners directly upstream. This avoids duplicate work and hopefully shows how much we value the openness of the MATSim framework.

When MATSim was updated from Java 17 to 21 (see PR #3238), it happened in May 2024, more than half a year after JDK 21 was released. We understand that this time the project might want to move faster, and we fully respect that.

Our CI infrastructure is still being updated to Java 25 in the next few weeks. We have tried to speed this up internally, but unfortunately have reached the limits of what we can do. We know this delay is on our side and not something the MATSim project needs to worry about.

Still, if possible, we would really appreciate a short delay of the dependency update so we can keep using the latest builds without interruption. Thank you a lot for considering this.

jfbischoff:
I would also be happy if we could wait for another two weeks - opposed to what I wrote up there. Turns out our Build Servers will only receive java25 in the next couple of days 💯

Janekdererste:
@marecabo that makes total sense, and in this case we don't need to push faster. From the comments I can see that everyone is working as fast as possible :-) .  I was just a little eager to merge #4255.

Janekdererste:
To push this issue on top of everyone's inbox:  

Are there updates regarding the upgrading of build infrastructure? We would still like to merge #4255 which depends on this PR. 

I guess a good strategy would be to speak up, if your infrastructure is not ready yet :-)

steffenaxer:
We are Java 25 ready 🚀

Janekdererste:
Great! Merging this then.

Issue #4314, opened 2025-10-02
Experienced plans output should contain person attributes


tschlenther:
This is helpful for many following analyzes, including anything that is related to exposures...

At the time being, not even the `subpopulation` attribute is carried over to the `experienced_plans`, making it impossible to distinguish `person` from `freight` agents...

tschlenther:
The downside would be, that we tend to write experiencedPlans every other iteration and will thus have a lot of duplicate information in our output directory

gac55:
related to this, I still wonder why output experienced plans are not the default output, and what users are sign posted to by default

paulheinr:
experienced plans output is default since #3976

kainagel:
I added the person attributes to the persons in the experienced plans in #4398 .

Re the question by @gac55 : If you adjust the plans to the execution, you potentially introduce "drift".  Assume for example an agent logic that always departs 5min later than what is written in the plan.  Then adjusting the plan to the actual departure would let the plan drift to ever later departures. 

Re the comment by @paulheinr :  I think that that was a bit of a misunderstanding, see my anser to @gac55 above.  However, the sentence, taken literally, is correct: We now by default write output_experienced_plans in addition to output_plans by default.  I.e. if you do not want to have this, you need to switch it off yourself.

PR #4315, opened 2025-10-02
feat: write person attributes in experienced plans

tschlenther:
This PR addresses #4314.

We should discuss whether we want to really merge this, because output directories might grow substantially bigger.
On the other hand, experienced plans should be used for a fair number of analyzes, especially those involving any kind of exposure with respect to socio-demographics.
Before this PR, not even the subpopulation attribute was copied to the experienced plans.


jfbischoff:
I'm in favour of this - currently, I often need to first read the experienced plans and then the output_plans for a postprocessing script 👍

kainagel:
Ok, not so great ... I introduced this independently in #4398 (since I urgently needed it for analysis).  

Meaning that this PR is probably no longer needed.  Sorry.

kainagel:
@tschlenther : I will close this since I did the same thing elsewhere, see above.

PR #4327, opened 2025-10-08
fix(drt): make drt constrains relative again

nkuehnel:
As per #4161 there's a bug for drt constraints in the case of prebooking. When the constraints were changed to absolute time stamps, the time stamps did not take into account the offset between submission time and earliest/planned departure, meaning that the time windows were too narrow for prebooked trips, causing more rejections.

The main reason was that the prebooking manager correctly passes the actual departure time to the request creator, which sets the earliest start time accordingly. However, the latest start time was already determined in the route creator and did not account for the offset between submission and earliest departure.

I changed it back such that the constraints object uses relative constraints again, but still kept re-using the constraints object for creating the requests.

Mea culpa, I added a test to ensure the behavior stays valid.


sebhoerl:
Well, if no tests failed, then it's probably the fault of the people who didn't write enough of tests in the first place :)

sebhoerl (approved):

PR #4328, opened 2025-10-08
feat(railsim): Railsim Extensions

rakow:
This PR adds several major new features to the railsim contrib. All of these features are tested and demonstrated in corresponding scenarios contained in `contribs/railsim/src/test/java/ch/sbb/matsim/contrib/railsim/integration`. 

### MinimumStopDuration

The transit schedule format was extended with a new attribute `minimumStopDuration`. Trains (and other pt vehicles) will stop for a minimum amount of time even when delayed. 

### Turn Restrictions

Railsim now considers the new turn restrictions functionality implemented in recent PRs using  `.setDisallowedNextLinks`. 
There is a corresponding utility class `ch.sbb.matsim.contrib.railsim.prepare.PrepareTurningAngleBasedRestrictions` to forbid turning maneuvers with too steep angle.  

### Reversible Trains

A new vehicle attribute `railsimReversible` was added.  This attribute indicates that a certain vehicle type can reverse (e.g., by conductor switching cabin). The attribute value defines how long this reverse operation will take. Only reversible trains are now allowed to use a direct opposite link in their route. Other trains will trigger an exception.

### NetworkChangeEvents / Time Variant Links

Railsim now supports the usual MATSim NetworkChangeEvent mechanism. Changes in `freespeed` and `railsimCapacity` are considered and can be changed during runtime. The change in attributes was enabled in an earlier PR #4140.

### Speed Profiles

Before this PR, trains always tried to reach the maximum permissible speed. This behavior can now be customized by implementing and binding the new `SpeedProfile` interface. The previous behavior remains the default for now, but there is one example implementation `AdaptiveSpeedProfile`, which lets trains use a slower speed while still being able to fulfill the next planned arrival time.

### Non-blocking Areas

Links can be marked as non-blocking by setting the `railsimNonBlockingArea` attribute. This can be used to model certain intersections and prevent trains from waiting on any of these links. The dispatcher will ensure that trains only enter the intersection if they have a reservation for the whole area and are able to leave it without blocking anyone else. 

### Train Circulation and Formations

Railsim now supports circulation (Umlauf) and chained departures which were added in the previous PRs (#3901, #3913, #4220). Additionally to simulating trains split and merging, Railsim has a feature to keep track of the individual trains units which is important for operation, maintenance and planning. On departure,  a `RailsimFormationEvent` is thrown to inform about the individual units a train is composed of.


ikaddoura:
thaaaank you!! :-)

PR #4370, opened 2025-10-21
fix: Improve two writers

jfbischoff:
Not exactly related to another:
- the datastorage in the GeoFileWriter is not closed. This could lead to problems and produces a nasty warning.
- In the HouseholdWriter the AttributesWriter Object is not re-used.


marecabo:
@jfbischoff Thanks for the HouseholdWriter fix! 🥳

PR #4372, opened 2025-10-21
test(freight:logistics): reduce runtime of test

### LINKED ISSUE #4335: Freight(logistics): reduce runtime of integrations tests
I have added some tests for the LogisticServiceProvider, running the complete simulation and comparing the results.
Unfortunately, the runtime is not the fasted, because there are too many shipments.
This was done, because I was investigating them for my thesis, and want to make sure, not messing up things while refactoring.

I would like to keep this type of test as integration test, BUT it should be changed towards problems with less shipments, so the runtime is significantly faster.

Thanks to @rewertvsp for reminding me.

kt86:
I have changed the logistic integration test in a way, that a carrier with less jobs is used.
This increases the runtime significantly, because the VRP-solving is significantly faster.

In general the test mentioned (`ExampleTwoLspsFoodMultipleChainsWTollTest`) now need about 234 sec instead of 735 sec in GH Actions.
For the 'ExampleTwoLspsFoodMultipleChainsTest` I did the same change (in src code), resulting in a speed up for the test from approx 180 sec to 85 sec.

Overall this is a speedup of approx. 600sec or 10 min.
The freight contrib tests are now down from nearly 20min to nearly 10min.

Hope this is good enough for now ;)

Closes #4335


kt86:
@rewertvsp Thanks for pushing me!

PR #4378, opened 2025-10-23
feat(simwrapper): make dashboards replacable in SimWrapper obejct

simei94:
Due to a previous PR we are now able to add DashboardProviders via Guice. In scenario specific DashboardProviders like e.g. matsim-scenarios/matsim-dresden we want to replace e.g. the default TripDashboard with a TripDashboard includin specific reference data which is needed for calibration. With the guice injection we have to give the scenario specific dashboard a different context because otherwise simwrapper will not add it because it has the same type as the default TripDashboard. This is rather ugly because the 2 dashboards are identical in terms of analysis.

With the implemented method we could replace the dashboard and would only have 1 TripDashboard (as before).


simei94:
@paulheinr what do you think about this? Is there another way of doing this?

simei94:
The new method needs to be tested, I haven't done that yet. I'd propose to also create some unit tests for the other convenience methods in the SimWrapper class, but maybe they already exist (writing this from my phone)

PR #4379, opened 2025-10-24
fix(utils): finish implementation of LazyAllocationAttributes

marecabo:
This fixes `LazyAllocationAttributes` such that it behaves like the regular `AttributesImpl` class - and adds a test for it.

`LazyAllocationAttributes` is used for the attributes of activities and legs and thus also for trip attributes (`TripStrucutureUtils.getTripAttributes(Trip)`). Previously, one could put attributes in it, but it was impossible to get them back because the implementation was incomplete. If this was meant to be like this, please let me know.

### DIFF
--- a/matsim/src/main/java/org/matsim/utils/objectattributes/attributable/LazyAllocationAttributes.java
+++ b/matsim/src/main/java/org/matsim/utils/objectattributes/attributable/LazyAllocationAttributes.java
@@ -45,36 +45,56 @@ public final class LazyAllocationAttributes implements Attributes {
 			attributes = new AttributesImpl();
 			this.consumer.accept(attributes);
 		}
-		attributes.putAttribute(attribute, value);
-		return null;
+		return attributes.putAttribute(attribute, value);
 	}
 
 	@Override
 	public Object getAttribute(String attribute) {
-		return null;
+		Attributes attributes = this.supplier.get();
+		if (Objects.isNull(attributes)) {
+			return null;
+		}
+		return attributes.getAttribute(attribute);
 	}
 
 	@Override
 	public Object removeAttribute(String attribute) {
-		return null;
+		Attributes attributes = this.supplier.get();
+		if (Objects.isNull(attributes)) {
+			return null;
+		}
+		return attributes.removeAttribute(attribute);
 	}
 
 	@Override
 	public void clear() {
+		Attributes attributes = this.supplier.get();
+		if (!Objects.isNull(attributes)) {
+			attributes.clear();
+		}
 	}
 
 	@Override
 	public Map<String, Object> getAsMap() {
-		return Collections.emptyMap();
+		Attributes attributes = this.supplier.get();
+		if (Objects.isNull(attributes)) {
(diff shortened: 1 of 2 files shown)


mrieser:
> If this was meant to be like this, please let me know.

Yes, it was meant to be this way, but on a second glance I would say it was unsafe.

The idea of the class `LazyAllocationAttributes` was, that `Attributable`-classes do not waste memory on an empty `Attributes`-object. If this would have been replaced by a `LazyAllocationAttributes`-object instead (as this commit suggests), not much is won.

Have a look at the usage of `LazyAllocationAttributes` in `ActivityImpl.getAttributes()`:

```java
public Attributes getAttributes() {
	if (this.attributes != null) {
		return this.attributes;
	}
	return new LazyAllocationAttributes(attributes -> this.attributes = attributes, () -> this.attributes);
}
```

The goal was, that `ActivityImpl.attributs` can be null as long as no attributes are stored.
In this case, a (short-living) LazyAllocationAttributes-object is returned which is always empty, return null for any `get()` etc.

The only real operation is the `putAttribute`. Then this attribute really must be stored. This is done by `LazyAllocationAttributes` creating a real `Attributes` instance, adding the to-be-stored attribute to it, and passing it to a consumer, in this case the `attributes-> this.attributs = attributes` in the return statement. So, once an attribute needs to be stored, a non-null `Attributes`-object is stored in `ActivityImpl.attributes` and then this is being used for all further operations.

----

We can still keep your improved implementation as it fixes the following edge-case:

If one keeps the object returned by `getAttributes()`, the old implementation could have returned false values, e.g.:

```java
Attributes attrs = activity.getAttributes();
attrs.putAttribute("foo", "bar");
Object value = attrs.getAttribute("foo");
// in the old implementation, value is wrongly null, while the new implementation correctly returns "foo"

activity.getAttributes().putAttribute("foo", "bar");
Object value = activity.getAttriubtes().getAttribute("foo");
// this would work in the old and the new implementation
```

That is, `LazyAllocationAttributes` was meant to be short-lived, only used for a single operation. As we cannot guarantee this usage, the improved implementation makes sense. Thanks!

mrieser:
Looking at the newly provided Test-class: it states Thibaut as author, but the commit is from @marecabo. Who's is really?

And more importantly, the test only uses `AttributesImpl`, but *never* `LazyAllocationAttributes`, so it is kind of useless in the context of this merge request.

marecabo:
@mrieser yes, I noticed this as well with the test, sorry for that, I'll fix it as soon as possible.

marecabo:
Thanks @mrieser, for providing this explanation! I noticed this behavior when I tried to put new attributes to a `RoutingRequest` which got its `Attributes` object from trip attributes - which are actually attributes of the origin activity - and thus showed this weird behavior that attributes could not be retrieved. https://github.com/matsim-org/matsim-libs/blob/294e57c715fd5fad7e1f24c8f42332f46ddd8af8/contribs/shared_mobility/src/main/java/org/matsim/contrib/shared_mobility/logic/SharingLogic.java#L384-L385

PR #4390, opened 2025-10-27
fix: unregister LegHistogramListener handler at iteration end to avoid cumulative counts (backport 2025.0)

langkoos:
In the 2025.0 release, LegHistogramListener#notifyIterationEnds does not unregister the histogram event handler. As a result, the handler remains registered across iterations and counts accumulate, producing incorrect plots and leghistogram.txt files. Discovered when trying to analyse my own scenario built on the matsim example project which uses 2025.0

Joschka fixed this on the main branch at commit 8e358f47b0b1c0867166f8d21eebd6f4b8ae8c13 so the patch need not be applied on main, only on the 2025.0 release

Happy to learn how to deal with the release process so I can help with this kind of thing in future; tagging Paul and Janek as I see they have run the release gauntlet in the past, and to update Marcel as I have mentioned it to him in an email thread.

### DIFF
--- a/matsim/src/main/java/org/matsim/analysis/LegHistogramListener.java
+++ b/matsim/src/main/java/org/matsim/analysis/LegHistogramListener.java
@@ -67,6 +67,7 @@ final class LegHistogramListener implements IterationEndsListener, IterationStar
 					LegHistogramChart.writeGraphic(this.histogram, controllerIO.getIterationFilename(event.getIteration(), "legHistogram_" + legMode + ".png"), legMode);
 				}
 			}
+			event.getServices().getEvents().removeHandler(this.histogram);
 		}
 	}
 


mrieser:
Is a back port and official release really needed, or could you just use a newer weekly or PR-based release for your work?

langkoos:
I think if it is too disruptive to go through all the motions of a fresh release, then we should perhaps alert people on the example project of a known issue with 2025.0 and suggest they use a weekly build or maybe provide a git patch on the README... 

It took me some time to discover the cause of the issue myself, so if we can save the user some time and reassure them that there is nothing systematically wrong with the software I think that shows due diligence.

So it's more of a perception thing in the end. I can add some lines suggesting course of action to the example project README to alert, if you agree that's best. If you think it may cause more harm than good, then mum's the word till the next release which is coming up soon

Issue #4396, opened 2025-10-29
srr: error in mode-to-mode specific transfer cost calculation


sebhoerl:
I think there is an error in the `DefaultRaptorTransferCostCalculator`. @mrieser, @jfbischoff, could you confirm or tell me where is my misunderstanding? :)

**Context**

- The `SwissRailRaptorCore` calculates transfer costs along the chain in this manner, for instance:

https://github.com/matsim-org/matsim-libs/blob/8253d357a24d8930819c1b416bc2ce26fdd575f7/matsim/src/main/java/ch/sbb/matsim/routing/pt/raptor/SwissRailRaptorCore.java#L739

- Or in a simplified way:

```
arrivalTransferCost = currentTransferCost + costCalculator(..., currentTransferCost)			
``` 

- So the updated transfer cost is the current transfer cost, plus what comes out of the cost calculator. And the current cost is also passed to the cost calculator (`RaptorTransferCostCalculator.calcTransferCost`) as `existingTransferCosts`:

https://github.com/matsim-org/matsim-libs/blob/8253d357a24d8930819c1b416bc2ce26fdd575f7/matsim/src/main/java/ch/sbb/matsim/routing/pt/raptor/RaptorTransferCostCalculator.java#L30

- Inside `DefaultRaptorTransferCostCalculator` there is `calcSingleTransferCost` which will calculate the cost of a "single transfer". And the specificity is that "one transfer" can get more expensive the longer the travel time is, i.e.

https://github.com/matsim-org/matsim-libs/blob/8253d357a24d8930819c1b416bc2ce26fdd575f7/matsim/src/main/java/ch/sbb/matsim/routing/pt/raptor/DefaultRaptorTransferCostCalculator.java#L39-L40

- This value is then multiplied with the *number of transfers so far* `transferCount` in the interface method `calcTransferCost` and returned in the following way (simplified):

```
costBase = getFromConfig()
return calcSingleTransferCost(costBase, ...) * transferCount - existingTransferCosts
```

https://github.com/matsim-org/matsim-libs/blob/8253d357a24d8930819c1b416bc2ce26fdd575f7/matsim/src/main/java/ch/sbb/matsim/routing/pt/raptor/DefaultRaptorTransferCostCalculator.java#L36

[...]

PR #4397, opened 2025-10-29
fix(srr): mode-specific transfer penalties

sebhoerl:
Potential fix for #4396. To be discussed.


jfbischoff:
Thanks for spotting this, looks good to me. What a complicated topic, overall 💯

PR #4405, opened 2025-10-30
feat(ev): use fileEnding from config for VehicleTrajectoryListener

marecabo:


sebhoerl:
Thanks, let's do this also for #4407 before merging. And I would prefer if we pass the file ending to the constructor instead of the whole Config :)

marecabo:
Good point, I opted for passing the `CompressionType`, hope that's fine as well :)

marecabo:
I pushed the commits to #4407, closing this

PR #4411, opened 2025-10-31
feat(common): simplify registration of zone systems

sebhoerl:
This PR introduces `ZoneSystemUtils.registerDefaultZoneSystems` which removes code duplication and simplifies the registration of our list of standard zone systems in config groups.


sebhoerl:
Do you see any caveat @nkuehnel?

nkuehnel:
nice, thank you

nkuehnel (approved):

PR #4420, opened 2025-10-31
feat(drt): introduce earliest departure time in stops

nkuehnel:
This PR introduces an earliest departure for drt stop tasks, which may be used to model drt stop tasks that do not yet have requests assigned (but the vehicle should stay there for a certain time in case a new request arrives).

In the course of that PR I moved the earliest/latest arrival/departure time calculation out of the waypoint class into the drtstoptask class, as this calculation was only dependend on task specifics anyway. The waypoint class now calls this to cache the result of the calculation (as before).

The earliest departure is achieved by the end time calculator taking it into account for shifting stop end times.

@sebhoerl I would appreciate your opinion on this

### DIFF
--- a/contribs/drt/src/main/java/org/matsim/contrib/drt/schedule/DefaultDrtStopTask.java
+++ b/contribs/drt/src/main/java/org/matsim/contrib/drt/schedule/DefaultDrtStopTask.java
@@ -128,7 +128,7 @@ public class DefaultDrtStopTask extends DefaultStayTask implements DrtStopTask {
 	@Override
 	public double calcEarliestDepartureTime() {
 		// no restriction on earliest departure time in default implementation
-		return 0;
+		return Double.NEGATIVE_INFINITY;
 	}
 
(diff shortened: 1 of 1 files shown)


contribs/drt/src/main/java/org/matsim/contrib/drt/schedule/DefaultDrtStopTask.java
```diff
@@ -94,4 +95,46 @@ public void removePickupRequest(Id<Request> requestId) {
...
+					.stream()
+					.mapToDouble(AcceptedDrtRequest::getEarliestStartTime)
+					.min()
+					.orElse(0);
+	}
+
+	@Override
+	public double calcEarliestDepartureTime() {
+		// no restriction on earliest departure time in default implementation
+		return 0;
```
sebhoerl: Should this be rather -Inf ?

sebhoerl:
Just one little comment, otherwise looks nice, much cleaner!

PR #4423, opened 2025-11-03
fix(pt): total pt route cost

tkchouaki:
PR #4164 introduced a `totalRouteCost` attribute in the `DefaultTransitPassengerRoute` class and its associated serialized `RouteDescription`. This broke compatibility of newly created population files with old versions of the `PopulationReader` that should still be on the same V6. 

To fix this, this PR removes the attribute altogether and moves the information to leg attributes. An attribute `totalRouteCost` is added into each leg resulting from Swiss Rail Raptor, even walk ones.


kainagel:
Thank you!

PR #4429, opened 2025-11-04
refactor(FISS): Generalize code in FISS module so that it can be run with QSIM and DSIM

Janekdererste:
The FISS module had fixed dependencies on the `DefaultNetworkModeDepartureHandler`, as it delegates those departures that are actually simulated on the network to the default handler. It also had fixed dependency on the `DefaultTeleportationEngine`, delegating legs that are not simulated on the network to this engine. In fact, it even created its own instance of a `DefaultTeleportationEngine`, which is a little odd, as the main Simulation already has such an engine. 

With this change the `FissModule` binds a named default `NetworkModeDepartureHandler` depending on the mobsim in use (`qsim`, or `dsim`) and then binds `FISS` as the actual `NetworkModeDepartureHandler` used by the `mobsim`. `FISS` now expects an injected named `DepartureHandler`, to which it can delegate network departures. 

Also, the `FISS` class expects a `TeleportationEngine` interface, which is either bound to `DefaultTeleportationEngine` or `DistributedTeleportationEngine`.


nkuehnel:
Nice, thanks!

PR #4444, opened 2025-11-09
feat: add vtts analysis postprocessing runner; maintenance related to that

kainagel:


sebhoerl:
Hi @kainagel, @GregorRyb, the fact that the package is called `vtts` and the folder `VTTS` is making VSCode complain quite a bit, could you rename the folders please? :)

PR #4456, opened 2025-11-12
feat: changes related to agent-wise comparison

kainagel:


kainagel:
will try to merge this now.  AgentWiseComparisonKN is far from ready and maybe never will be ready.  But there are many related maintenance changes in matsim-libs that I would prefer to merge this now and then continue on the agent wise comparison code and related.

PR #4465, opened 2025-11-13
feat(railsim): Time-distance diagram and projected delay

rakow:
This PR introduces time-distance diagram functionality and projected delay calculation for the railsim contrib. The feature enables tracking of train positions over time and comparison between scheduled (target) and actual (simulated) trajectories, with delay calculation based on the difference between planned and realized positions.

## Features
- An additional output `railsimTimeDistance.csv` is generated, containing time-distance data for all trains
    - See `time-distance-chart.md` for infos about the CSV format and column specifications
- A new `delay` attribute is included in the train state for analysis or use by more complex disposition algorithms

## Usage

The time-distance handler is an obligatory component of the railsim simulation engine. CSV files are written to the iteration output directory as `railsimTimeDistance.csv`. Delay information is available through `RailsimTrainStateEvent` events or as a new column in the `railsimTrainStates.csv`

## Example

The example shows a train, which is delayed due to a halt on the crossing. The train is able to compensate the delay later and still arrive on time.

https://github.com/user-attachments/assets/a2436189-c786-436f-9267-356333ceb498

### DIFF
--- a/contribs/railsim/src/main/java/ch/sbb/matsim/contrib/railsim/qsimengine/RailsimTransitDriverAgentImpl.java
+++ b/contribs/railsim/src/main/java/ch/sbb/matsim/contrib/railsim/qsimengine/RailsimTransitDriverAgentImpl.java
@@ -33,6 +33,9 @@ import org.matsim.pt.transitSchedule.api.TransitStopFacility;
 import java.util.List;
 import java.util.Map;
 
+/**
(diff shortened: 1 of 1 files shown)


munterfi (approved):
Thanks @rakow, nice work! 🚀 

I would propose to merge this so we can test it in our project cases.

rakow:
I clicked merged last week, but the build didn't go through because of some maven server errors. I hope it works now!

PR #4477, opened 2025-11-14
fix(ev): resolve a rare race condition in vehicle discharging

sebhoerl:
This problem happened when running a very large simulation with electric vehicles on a large number of cores. There is a race condition in `DriveDischargingHandler` that causes a `NullPointerException` for very specific circumstances.

The problem occurs when a vehicle leaves from an activity (`VehicleEntersTrafficEvent`) and then performs a zero-duration drive, because the destination is the same link. This creates a `VehicleLeavesTrafficEvent` in the same second.

Now `DriveDischargingHandler` works such that the `VehicleEntersTrafficEvent` directly starts tracking consumption by creating a `EvDrive` object in simulation step `k`. However, the `VehicleLeavesTrafficEvent`s are only tracked in simulation step `k` (saved in a queue) and processed in step `k + 1` inside the `onSimStep` of the `DriveDischargingHandler`. This class being a `MobsimEngine` the `onSimStep` is executed in parallel to event handling. This is the reason why the events are buffered: to be sure that we process all relevant events of step `k` we only process them in `k+1`.

In particular, when the `VehicleLeavesTrafficEvent` from `k - 1` are processed on `onSimStep`, the corresponding `EvDrive` object (that should exist since `VehicleLeavesTrafficEvent` always comes after `VehicleEntersTrafficEvent`) is accessed and then removed from the tracker. 

(And eventual intermediate `LinkLeaveEvents` are processed along the route, but they don't play a role here).

For the minimal error case, consider an agent that goes from link A to link A to A. So the two trips connect the exact same link, both a zero-duration drives. What happens in the simulation (without EVs) is the following:

```
- agent starts at 0.0 (VehicleEntersNetwork)
- agent arrives at 0.0 (VehicleLeavesNetwork)
- agent starts at 1.0 (VehicleEntersNetwork)
- agent arrives at 1.0 (VehicleLeavesNetwork)
```

This leads to a NPE in the current version of `DriveDischargingHandler` when the `ev` contrib is activated. However, this only happens *sometimes* as described below:

```
In second 0.0:

- event handling: VehicleEntersNetwork is processed at 0.0 -> this creates an EvEnergy object
- event handling: VehicleLeavesNetwork is tracked at 0.0 to be processed later

In second 1.0:

- event handling: VehicleEntersNetwork is processed at 1.0 -> this creates an EvEnergy object (but there is already one - no exception but already not ideal)
- onSimStep: VehicleLeavesNetwork from second 0.0 is processed -> first examines the EvEnergy object, then removes it!

In second 2.0:

- onSimStep: VehicleLeavesNetwork from second 1.0 is processed -> tries to examine the EvEnergy object, but it doesn't exist anymore -> NullPointerException
```

The key here is that the order of the two bullets in second 1.0 is random, because event handling and `onSimStep` run in parallel. In the case above, we have the pathological case. If the two bullets happen in inverse order, a new `EvEnergy` object is registered cleanly and the access in second 2.0 happens without a problem. So this is a real race condition, but the pathological case happens much more often (at least on my machine).

**Unit test**: I have added a unit test, where you can choose the number of retries. Usually, I have a NPE in one of the first three retries.

**Solution**: This PR proposes a solution by not only buffering the `VehicleLeavesNetwork` events, but now also the `VehicleEntersNetwork` events. This way, we can make sure that we process them in the right order. This most likely adds some memory overhead, but for now I don't see another option to have a clean implementation. 

A cleaner option would definitely be to not make the energy consumption calculation dependent on event handling, as already discussed (https://github.com/matsim-org/matsim-libs/issues/3545). However, this would require a much deeper rewrite of various elements in the ev contrib and beyond.

### DIFF
--- a/contribs/ev/src/main/java/org/matsim/contrib/ev/discharging/DriveDischargingHandler.java
+++ b/contribs/ev/src/main/java/org/matsim/contrib/ev/discharging/DriveDischargingHandler.java
@@ -52,7 +52,7 @@ import com.google.inject.Inject;
  * calculating the drive-related energy consumption. However, the time spent on the first link is used by the time-based
  * idle discharge process (see {@link IdleDischargingHandler}).
  */
-public final class DriveDischargingHandler
+public class DriveDischargingHandler
 	implements LinkLeaveEventHandler, VehicleEntersTrafficEventHandler, VehicleLeavesTrafficEventHandler, MobsimScopeEventHandler, MobsimEngine {
 
 	private static class EvDrive {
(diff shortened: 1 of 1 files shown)


sebhoerl:
Don't ask how much time this took me to figure out, given that this happened *sometimes* after running a simulation for two hours ;)

PR #4478, opened 2025-11-15
feat: new contrib ATAP

gunnarfloetteroed:
ATAP stands for “Agent-based/atomic dynamic traffic assignment problem”. The se.vti.atap package is a MATSim plugin for the approximation of Nash equilibria. This has been discussed with MarcelR and KayA. There is a more detailed description in the readme.md file. 

In brief, the basic logic for running an assignment with ATAP looks as follows:

    ATAP atap = new ATAP();
    Config config = ConfigUtils.loadConfig(configFileName, new ATAPConfigGroup(), new EmulationConfigGroup());
    atap.configure(config);
    Scenario scenario = ScenarioUtils.loadScenario(config);
    Controler controler = new Controler(scenario);
    atap.configure(controler);
    controler.run();

The ATAP class is the entry point for all ATAP functionality. Its configuration (through ATAPConfigGroup) is detailed in the contrib's readme file. The EmulationConfigGroup configures the emulation functionality that is necessary for ATAP to anticipate the scores of not yet executed plans; it resembles PSim and IER (and could be eventually merged with those).

atap.configure(config) modifies the replanning and selection strategies. It sets minimal choice sets (one current and one alternative plan), removes all plan selection strategies, and re-weights all innovation strategies for greatest computational efficiency (in conjunction with the ATAP logic).

atap.configure(controler) binds PlansReplanning to ATAPReplanning, which is the core ATAP functionality. ATAPReplanning does basically the following, starting out from a situation where each agent has one current plan:
1) Construct, for each agent, a candidate plan.
2) Anticipate, via emulation, for each current and candidate plan their anticipated scores, given the previous iteration(s) network conditions.
3) Approximate by how much the next iteration's network conditions will change compared to the most recent ones, as a function of which agents get to switch to their candidate plans.
4) Decide which agents are allowed to switch to their candidate plans; this is achieved by solving a mathematical problem that balances (i) switching agents that anticipate a large score improvement and (ii) avoiding switches that lead to large network condition changes.


sebhoerl:
Hi Gunnar, its fine from my side. Please add the contribu to the list in `.github/workflows/verify-push.yaml` so that the unir tests are run and checked automatically before merging.

sebhoerl:
There seems to be an issue with the distributed matsim module right now, let's wait a bit and then we can merge.

gunnarfloetteroed:
There seem to be quite a few failing builds right now that are apparently unrelated to this PR -- will keep trying to merge every once in a while.

PR #4524, opened 2025-11-27
feat(common): integration with ntfy.sh

sebhoerl:
This PR provides functionality to get notified by MATSim events (simulation or iteration starts / ends) through `ntfy.sh`. 

**What is ntfy.sh?**

It is a light-weight push message service. For instance, you can open https://ntfy.sh/my-matsim-1234. There, you see all messages that have been sent (from any tool via HTTP) to the `my-matsim-1234` topic. The nice thing is that you can subscribe to the channel via the [ntfy.sh](https://ntfy.sh) app on your phone or using your browser and you get an instant notification as soon as a new message arrives. 

**Why is this useful for MATSim?**

You can open and use as many topics as you like, so you can easily open your own person or group topic where notifications from your MATSim simulations are sent to. This way you can get notified when your simulation starts (if you are in a cluster queue system) and ends or when an iteration starts or ends (maybe to be able to then directly check the latest outputs).

**How to use this?**

In the simplest case, before starting your `Controller`, just call

```java
NtfyUtils.install(controller, "my-matsim-topic");
```

It will add the module to the controller and add a default `NtfyConfigGroup` to the configuration. Alternatively, you can also add a `NtfyConfigGroup` when creating your `Config`. This way, you can read in all configuration parameters from the `ntfy` config module in your XML.

Make sure to set the channel to the one that you would like to use. By default, you will be notified when the simulation starts, ends, and whenever an iteration ends. For the current configuration options, have a look at `NtfyConfigGroup`.

**Custom messaging**

You can also send custom messages that are preformatted indicating a potential name of your simulation which you can put in the config or your `run id` if you have set it up on the controller config group. For that, just inject the `Notifier` singleton to any of your classes:

```java
Notifier notifier;
// ...
notifier.notify("This is my custom message");
```

This way you can send a push message whenever something interesting happens in your simulation (debug outputs, error messages, ...) and you'll be notified directly.

### DIFF
--- a/contribs/common/src/main/java/org/matsim/contrib/common/ntfy/Notifier.java
+++ b/contribs/common/src/main/java/org/matsim/contrib/common/ntfy/Notifier.java
@@ -1,0 +1,94 @@
+package org.matsim.contrib.common.ntfy;
+
+import java.io.DataOutputStream;
+import java.io.IOException;
+import java.net.HttpURLConnection;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.net.URL;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+/**
+ * @author sebhoerl
+ */
+public class Notifier {
+    private final Logger logger = LogManager.getLogger(Notifier.class);
+
+    private final String run;
+    private final NtfyConfigGroup config;
+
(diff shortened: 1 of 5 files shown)


sebhoerl:
Please let me know if you think that this should go into `common` or if it should go into the MATSim repository at all. It could also go to the code examples or elsewhere.

Issue #3916, opened 2025-04-28
Test: Does walk mode cause problems when used as network mode?


paulheinr:
Search a test case for that and if it doesn't exist, write one

paulheinr:
Test case see https://github.com/matsim-org/matsim-code-examples/blob/walk-network-mode/src/main/java/org/matsim/codeexamples/integration/RunWithWalkAsNetworkMode.java 

If implemented without access and egress it works. 

If used with `walkConstantTimeToLink` as `AccessEgressType`, it crashes. The reason is because "walk" is automatically assumed as access and egress mode, not "non_network_walk". Commit de6655c6 removed `accessMode` and `egressMode` in NetworkRouteInclAccessEgressModule. 

If used with `accessEgressModeToLink`, ~~it doesn't crash but the results do not make sense~~. In org.matsim.core.router.NetworkRoutingProvider#get the router for "walk" is provided hard-coded as router for access and egress for other modes than walk. As walk is a network mode, the corresponding routing module will produce access and egress legs. Thus, car legs have a double chained access leg like: `non_network_walk -> walk -> non_network_walk -> car -> ...` 

I think for bugfixes in both cases the access/egress routing mode needs to be configurable again.

kainagel:
Maybe we should look at this together (at some point/during matsim-advanced)?  This may be more of a conceptual problem than an implementation problem: As far as I know, if you configure a `walk2` mode as a network mode, then it actually works. 

So yes, it may ("only") be that we need to be more careful about the "bushwhacking" mode:
* If it is access/egress to PT, it is indeed `walk` (which might decompose into non_network_work to the network, and then (network_)walk to the PT stop).
* If, however, it is walk_to_(nearest)_link, then it might have to be non_network_walk.  (Although even here I am not sure ... a properly configured walk mode here would use non_network_walk to the network and then network_walk, which would just have zero length.  But possibly this is overkill, and walkConstantTimeToLink explicitly is meant as "bushwhacking" and in consequence should use non_network_walk.)

kainagel:
At not so few places, typically for access/egress, there are "self-written" walk teleportation routers, i.e. routers which construct a leg with mode `walk` and values for travel time and distance, but using `GenericRoute`. --  Evidently, this crashes with a simulation that says that `walk' should be executed in the network.

[...]

PR #4541, opened 2025-12-04
feat: add test case for walk on network

paulheinr:


paulheinr:
Closes #3916

marecabo:
@paulheinr Looks like the full integration test fails on windows with this test: https://github.com/matsim-org/matsim-libs/actions/runs/20014674472/job/57389785421#step:5:3621

PR #4549, opened 2025-12-08
fix: hoping to fix test on windows (see #4541)

paulheinr:
see #4541


paulheinr:
@marecabo checked it with a colleague. Tests should be passing now.

PR #4567, opened 2025-12-11
fix(Carriers): Fix carrier plans cleaning and analysis

rewertvsp:
- Add the removal of the selected plan if we clear the plans of the carrier. --> new method in CarrierUtils
- adjust analysis if a carrier has no selected plan
- use new method for initializeValues in logistic contrib

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -347,13 +347,13 @@
             <dependency>
                 <groupId>org.mockito</groupId>
                 <artifactId>mockito-core</artifactId>
-                <version>5.20.0</version>
+                <version>5.21.0</version>
                 <scope>test</scope>
             </dependency>
 			<dependency>
 				<groupId>org.mockito</groupId>
 				<artifactId>mockito-junit-jupiter</artifactId>
-				<version>5.20.0</version>
+				<version>5.21.0</version>
 				<scope>test</scope>
 			</dependency>
 			<dependency>


rewertvsp:
Hi @kt86,
I suggested this method because clearing the plans didn't remove the selected plan, which could lead to some problems.
Because if you rerun jsprit, the new solution will only be set if no selectedPlan is set, which is != null if you cleared the existing plans before.
I used this new method as well for the initialize methods in the logistic part. Can you have a quick check if this is plausible for you?

kt86 (approved):
@rewertvsp Thanks a lot. Makes totally sense to remove the pointer to the no-longer existing plan.

So please go ahead!

The failing test seems to be only due to the fact, that you now have a "null" instead of "NaN". I'm not totally sure, if "null" in the analysis output is the better alternative? I read somewhere that one should try to avoid "null" in code whenever it is possible. Reason is what you now had in the `ConsistenyChecker`: Need a check for not-null.
Writing a "null" to text .. I'm not sure what the best way is.

rewertvsp:
We had "null" in the output files before. With the commit, I only bring the results back to the status before. 
The not-null for the carrierAnalysis is for the case if a carrier has no selected plan, which would lead to an exception. Now we have the warning, but the run will not fail.

PR #4571, opened 2025-12-12
fix(drt): fix adaptive prebooking logic

nkuehnel:
Skip DynAgents (ie drt driver agents) in adaptive prebooking logic, as they should not be touched and the simulation crashes with

Caused by: java.lang.RuntimeException: Sorry, agent is from type class org.matsim.contrib.dynagent.DynAgent which does not support getModifiablePlan(...). Aborting!
	at org.matsim.core.mobsim.qsim.agents.WithinDayAgentUtils.getModifiablePlan(WithinDayAgentUtils.java:162)
	at org.matsim.contrib.drt.prebooking.logic.AdaptivePrebookingLogic.handleEvent(AdaptivePrebookingLogic.java:60)
	at org.matsim.core.events.EventsManagerImpl.callHandlerFast(EventsManagerImpl.java:272)
	at org.matsim.core.events.EventsManagerImpl.processEvent(EventsManagerImpl.java:120)

@sebhoerl I'm afraid the adaptive logic does not seem to work when the first trip is a drt trip, because the first activity in a plan doesn't throw an activity start event and thus the prebooking is never applied :/ 

I'm not sure how this would be fixed best


sebhoerl:
Thanks, we could just make it a simstart mobsim handler that runs the same logic at the very beginning to catch the first leg.

nkuehnel:
Yea, right, I can look into this, thanks

PR #4586, opened 2025-12-17
fix: make DvrpLoads extendible

tschlenther:
The constructors in `IntegerLoad` and `IntegersLoad` were private, I propose to make them `protected` in order to allow for inheritance.

I am currently setting up an example very much similar to what is described in the `reconfiguration` package `README`, i.e. to have two different `LoadTypes` (passengers + goods).
I stumbled upon the fact, that still ONE `LoadType` is bound per drt mode.

While I am aware that I could simply use the `IntegersLoadType` (and tested that successfully), I think that having two different `LoadTypes` as proposed in the `README` would facilitate analysis and debugging.

So, I set up two new `LoadTypes`, both extending `IntegerLoadType`. In order for these to be able to instantiate heir corresponding loads using the `fromInt` method, I need to able to call the super constructor..

Or maybe `IntegerLoad` and `IntegersLoad` were not intenteded to be inheritable @sebhoerl @tkchouaki ? In this case, I wonder why the check method is `protected`, and not `package`-private.
And it would mean that one would have to copy a lot of code from `IntegerLoadType` and `IntegerLoad` in cases similar to what is described by the README.

Happy to read about your intuitions @sebhoerl @tkchouaki.

I also included some comments and changes to the `README` in this PR in order to improve documentation.

### DIFF
--- a/contribs/drt/src/main/java/org/matsim/contrib/drt/passenger/DrtRequestCreator.java
+++ b/contribs/drt/src/main/java/org/matsim/contrib/drt/passenger/DrtRequestCreator.java
@@ -45,13 +45,11 @@ public class DrtRequestCreator implements PassengerRequestCreator {
 	private final String mode;
 	private final EventsManager eventsManager;
 	private final DvrpLoadType dvrpLoadType;
-	private final DvrpLoad emptyLoad;
 
 	public DrtRequestCreator(String mode, EventsManager eventsManager, DvrpLoadType dvrpLoadType) {
 		this.mode = mode;
 		this.eventsManager = eventsManager;
 		this.dvrpLoadType = dvrpLoadType;
-		this.emptyLoad = dvrpLoadType.getEmptyLoad();
 	}
 
 	@Override
@@ -65,7 +63,7 @@ public class DrtRequestCreator implements PassengerRequestCreator {
 
         double lateDiversionThreshold = 0;
 		boolean allowRejection = true;
-		DvrpLoad load = emptyLoad;
+		DvrpLoad load = null;
 
 		Preconditions.checkArgument(!passengerIds.isEmpty());
 		for (Route route : routes) {
@@ -77,7 +75,9 @@ public class DrtRequestCreator implements PassengerRequestCreator {
 			maxRideDuration = Math.min(maxRideDuration, constraints.maxRideDuration());
             maxPickupDelay = Math.min(maxPickupDelay, constraints.maxPickupDelay());
             lateDiversionThreshold = Math.max(lateDiversionThreshold, constraints.lateDiversionThreshold());
-            load = load.add(drtRoute.getLoad(dvrpLoadType));
+			//the DvrpLoadType needs to be able to deserialize the load from the route description
+			//and the load needs to be addable to first load of all routes, i.e. the DvrpLoad implementations of all routes need to be compatible
+            load = (load == null) ? drtRoute.getLoad(dvrpLoadType).copy() : load.add(drtRoute.getLoad(dvrpLoadType));
 			allowRejection = allowRejection && constraints.allowRejection();
 		}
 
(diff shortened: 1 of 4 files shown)


tschlenther:
In order to enable the user to have multiple implementations of the same `DvrpLoad` type, **I added more changes**.

Specifically, I added a copy method in the `DvrpLoad` interface. This is necessary, in order to ensure that the `DrtRequestCreator` adds up `DvrpLoads` of the correct implementation. Previously, it would add values from the implementations to an instance of the type that is returned by the bound `DvrpLoadType.getEmptyLoad`. This led to requests always being of the 

**Now, a user can define several `DvrpLoad` implementations that all extend `IntegerLoad`**, for example `PassengerLoad` and `GoodsLoad`. The user can make sure that these loads can not be mixed by overriding the methods from `IntegerLoad`. Overriding  `getElement`, (maybe) `compareTo`, `hashCode` is not necessary.
**Be aware**: It is necessary to ensure that substraction and addition of an empty `IntegerLoad` (i.e. `IntegerLoadType.getEmptyLoad()`) is possible, otherwise request insertion will fail at `StopwaypointImpl.getOccupancyChange`

The same functionality was achievable before, by having one `IntegersLoadType` with multiple dimensions and making sure that vehicle capacities always have positive values for one dimension only. 
However, I feel that this way of modeling is more explicit and easy to understand (during analysis).
To have one more argument to merge this, the `README` in the `reconfiguration` package indicated that this should be possible, but it wasn't (or at least I wsn't able to find another way to do it).

I have a running mielec scenario example I could copy over, however this would require quite a number of implementation classes that would only be necessary for demonstration.

contribs/drt-extensions/src/main/java/org/matsim/contrib/drt/extension/reconfiguration/logic/DefaultCapacityReconfigurationLogic.java
```diff
@@ -129,7 +129,7 @@ public DefaultCapacityReconfigurationLogic(FleetSpecification fleetSpecification
 				maxServiceEndTime = vehicleSpecification.getServiceEndTime();
 			}
 			for (int slotIndex = timeToSlotIndex(
-					vehicleSpecification.getServiceBeginTime()); slotIndex < timeToSlotIndex(
+					vehicleSpecification.getServiceBeginTime()); slotIndex <= timeToSlotIndex(
```
tschlenther: i forgot to mention this bug fix here. Before, the last possible time slot wasn't handled properly

sebhoerl:
Hi, it would be good to see a concrete example. But I see a couple of caveats with the approach that you describe. In general, we should have one load type that corresponds to one load. Because the LoadType handles everything regarding deserialization (from the vehicles file, the events when reading an events file offline) and serialization (when writing events, etc.). I'm not sure how you would implement there the diversion between one or the other extension of IntegerLoad. In any case, you would need a specific LoadType, but maybe this is what you do?

tschlenther:
@sebhoerl: thanks for the feedback.
Exactly the deserialization and serialization steps are where I now faced some difficulties. I still think it might be possible, but after testing and thinking about this further, I understand that the software design is not really meant to have two types of loads using the `IntegerLoadType` implementation.
 I think some of the documentation needs adjustment, because it misled me to thinking this should be possible.
 So, instead of having two different `IntegerLoad` implementations, one should really use the array implementation `IntegersLoadType`.
 
 Overall, I think the documentation needs an update, there are several places where classes and fields are referenced, that do not even exist any more or were renamed. 
 I will try to make a contribution into this direction in the upcoming weeks, let's see.
 
 Not sure whether this PR should be merged, as it includes some steps as to the documentation improvement, but also making the constructor of `IntegerLoadType` protected which could trigger the above-mentioned misunderstanding.

sebhoerl:
Maybe let's close this PR for now, and have one that fixes the reconfiguration logic, and one that improves the documentation? And we can have another one where we make the IntegersLoad and Integer (/Type) `final`, this should make the intention clear, at least in the code.

tschlenther:
yes, thanks! I will prepare the PRs for the logic and for IntegerLoad. As to the documentation, a work-in-progress PR, but maybe it's better to collect more practical experiences and to update later

PR #4590, opened 2025-12-18
fix(drt): account for all time slots in DefaultCapacityReconfigurationLogic

tschlenther:
* the last time slot where vehicles operate was not handled.
* some code comments to facilitate understanding


tschlenther:
first part as discussed in #4586

PR #4597, opened 2025-12-23
fix(drt): insertion for completed schedule

sebhoerl:
In our eqasim unit tests, we noticed some failure for some DRT cases when trying to switch to the latest MATSim version. The reason is that `DefaultDrtRequestScheduler.insertPickup` calls `schedule.getCurrentTask` for a schedule that is already in state `COMPLETED`. So somehow the dispatcher now assigns a request to a vehicle that should not be available anymore.

This PR introduces a new unit test where I add a new request at `29:10:00`. All vehicle schedules end at `29:00:00`. This way, the error is reproduced in the standard Mielec example.

I fixed this by returning `null` in `VehicleDataEntryFactoryImpl` when the schedule status is `COMPLETED`. This way, the vehicle is not even considered in the `InsertionGenerator`, which makes sense, I think.

However, I'm wondering why the issue didn't appear before. @nkuehnel maybe you have some intuition given your recent changes? Do you agree with this fix, or should this happen somewhere later in the dispatching chain?

### DIFF
--- a/contribs/drt/src/test/java/org/matsim/contrib/drt/run/examples/RunDrtExampleIT.java
+++ b/contribs/drt/src/test/java/org/matsim/contrib/drt/run/examples/RunDrtExampleIT.java
@@ -310,6 +310,7 @@ public class RunDrtExampleIT {
 		Config config = ConfigUtils.loadConfig(configUrl, new MultiModeDrtConfigGroup(), dvrpConfigGroup,
 				new OTFVisConfigGroup());
 
+		// !!! IMPORTANT: use the plans with a late request
 		config.plans().setInputFile("plans_only_drt_1.0_with_late_request.xml.gz");
 
 		for (var drtCfg : MultiModeDrtConfigGroup.get(config).getModalElements()) {


nkuehnel:
That's weird indeed. @sebhoerl  Does your test fail without your fix? If so I can have a look why it actually comes to this

sebhoerl:
Yes, when you remove the two new lines in the entry factory the test will fail.

nkuehnel:
@sebhoerl I had a chance to look into this. 

Your fix is correct. The problem is that previously when the schedule wasn't in "STARTED" state, the vehicle entry factory assumed it is in "PLANNED" and then added _all_ the stops to the vehicle entry list of stops and slack times, even if these stops were already in "PERFORMED"  state, which is probably wrong.

Before my PR here https://github.com/matsim-org/matsim-libs/pull/4578 it was then caught in the insertion generator, which checked for `stop.getDepartureTime()` instead of `latestDepartureTime()`, effectively sorting out stops that already had passed based on their end time. However, for stops that don't have any pickups, the latestDeparture() will return positive infinity, causing the insertion generator to assume that this stop may still be extended.

One could add an additional check in the insertion generator to skip stop tasks in "PERFORMED" state, but ideally, these stops should not even be part of the stops in the vehicle entry

nkuehnel (approved):

PR #4615, opened 2026-01-08
feat: improve conistency checking if activity links and facility links are present in network

paulheinr:
When removing links from a network (e.g. when closing a bridge) one needs also clean the network, population and facilities. Consistency checks are improved to cover all cases that can cause problems.


paulheinr:
This is too old..

PR #4625, opened 2026-01-14
feat(drt): check for maxWaitTime being lower than maxTravelTimeBeta

marecabo:
I would suggest to add a constraint "`DrtOptimizationConstraintsSetImpl.maxTravelTimeBeta` > `DrtOptimizationConstraintsSet.maxWaitTime`" in `DrtOptimizationConstraintsSetImpl`.

* `DrtOptimizationConstraintsSetImpl` defines an upper bound for the travel time of a DRT trip with its `maxTravelTimeAlpha` and `maxTravelTimeBeta` values.
  https://github.com/matsim-org/matsim-libs/blob/f466c472d11b306ab4fcdd8d27f3d014fd770ae9/contribs/drt/src/main/java/org/matsim/contrib/drt/optimizer/constraints/DrtOptimizationConstraintsSetImpl.java#L10-L22

* These are used to compute a max travel time:
  https://github.com/matsim-org/matsim-libs/blob/f466c472d11b306ab4fcdd8d27f3d014fd770ae9/contribs/drt/src/main/java/org/matsim/contrib/drt/routing/DefaultDrtRouteConstraintsCalculator.java#L45

* The resulting `DrtRouteConstraint.maxTravelDuration` is then used in the `DrtRequest` as follows:
  https://github.com/matsim-org/matsim-libs/blob/f466c472d11b306ab4fcdd8d27f3d014fd770ae9/contribs/drt/src/main/java/org/matsim/contrib/drt/passenger/DrtRequest.java#L60-L61
  Obviously, it must hold that `latestStartTime` < `latestArrivalTime`.

* Tracking this back, I would guess, that `maxTravelTimeBeta` must be always greater or equal to `maxWaitTime`.

Do you have any thoughts on this? Maybe I am missing something?

Edit: The first commit is the actual change, all subsequent commits are the necessary adaption of tests.

### DIFF
--- a/contribs/drt-extensions/src/test/java/org/matsim/contrib/drt/extension/companions/RunDrtWithCompanionExampleIT.java
+++ b/contribs/drt-extensions/src/test/java/org/matsim/contrib/drt/extension/companions/RunDrtWithCompanionExampleIT.java
@@ -83,8 +83,8 @@ public class RunDrtWithCompanionExampleIT {
 		controler.run();
 
 		int[] actualRides = getTotalNumberOfDrtRides();
-		Assertions.assertThat(actualRides[0]).isEqualTo(378);
-		Assertions.assertThat(actualRides[1]).isEqualTo(706);
+		Assertions.assertThat(actualRides[0]).isEqualTo(402);
+		Assertions.assertThat(actualRides[1]).isEqualTo(752);
 	}
 
 	@Test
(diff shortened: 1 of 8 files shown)


nkuehnel:
I'm not sure I agree. There can be scenarios where one would have max wait > beta travel time. In these cases, the max wait time may only come into effect for long distances, compare the following:

image

In the right graph max wait is higher than beta, however, it can only come into effect for long trips (when alpha >1). In this case, a trip can be picked up late, but must be quite direct.

As such, latest start and latest arrival can be defined independently, only insertions that comply with both are valid.

nkuehnel:
Sorry the above graph assumed a minimum allowed detour, without that there is no shift to the right, but the argument is the same:
image

nkuehnel:
Simple example for the argument above:

given max wait of 10 mins and beta = 5min and alpha = 1.3 (30% detour)

Given a trip with a direct ride time of 30 min.
Then max travel time is 5 + 1.3 * 30 = 44 min.

This means latest arrival at departure time = 0:00 is 0:44.

This means, a direct ride without further detours lasting 30 mins does still have room for 14 mins of waiting time, which would then be capped by 10 mins max waiting time.

PR #4646, opened 2026-01-21
Integrate PCU-based counting in Cadyts car calibration.

langkoos:
feat(cadyts): Support PCU-weighted calibration for mixed traffic

This PR introduces Passenger Car Unit (PCU) support to the Cadyts integration, allowing for accurate calibration in mixed traffic simulations (e.g., cars vs. motorcycles). We have to do mixed traffic calibration for Meerut, India, at UNSW rCITI, which prompted the need for this adaptation.

Changes include:
- **New Analyzer**: Added `PcuVolumesAnalyzer` which aggregates flow based on `vehicle.getPcuEquivalents()` rather than raw vehicle counts.
- **Context Updates**: Updated `CadytsContext` and `CadytsContextI` to expose agent weights (PCU) via `getAgentWeight(Person)`.
- **Scoring Logic**: Modified `CadytsScoring` and `CadytsPlanChanger` to scale calibration offsets by the agent's PCU. This ensures that agents contributing less to congestion (e.g., 0.5 PCU) receive a proportionally smaller score correction.
- **Testing**: Added `CadytsPcuIT` integration test to verify that the calibrator correctly steers mixed populations (1.0 PCU Cars and 0.5 PCU Motorcycles) to match PCU-based traffic counts.

### DIFF
--- a/contribs/cadytsIntegration/src/main/java/org/matsim/contrib/cadyts/car/SimResultsContainerImpl.java
+++ b/contribs/cadytsIntegration/src/main/java/org/matsim/contrib/cadyts/car/SimResultsContainerImpl.java
@@ -1,3 +1,22 @@
+/* *********************************************************************** *
+ * project: org.matsim.*
+ * SimResultsContainerImpl.java
+ *                                                                         *
+ * *********************************************************************** *
+ *                                                                         *
+ * copyright       : (C) 2012 by the members listed in the COPYING,        *
+ *                   LICENSE and WARRANTY file.                            *
+ * email           : info at matsim dot org                                *
+ *                                                                         *
+ * *********************************************************************** *
+ *                                                                         *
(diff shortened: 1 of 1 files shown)


gunnarfloetteroed:
Hi Pieter,

just a small note. I interpret your approach as if you added an agent-specific size-weight to (25) in (https://svn.vsp.tu-berlin.de/repos/public-svn/publications/vspwp/2008/08-19/FloeBierNage2010TS_dec2010.pdf). Then, the "score correction" (26) and (28) is indeed scaled by that size-weight, so your approach appears correct. 

Could you perhaps include this back-link to the theory in the code description, to simplify later interpretation?

Gunnar

langkoos:
Thanks, Gunnar, will do. I realised yesterday I had pushed too early as my code cleaner wiped a bunch of comments so fixing that and will include your note in the relevant places before merging
best
Pieter

PR #4680, opened 2026-02-04
feat: chained setters; commenting out a non-compiling (VTTS) test; vsp-chec…

kainagel:


steffenaxer:
Dear @kainagel, the recent changes to relativeToleranceForSampleSizeFactors do not seem to be fully complete or consistent. The String getter and setter are missing, so the config is currently not written at all when you try to save your GlobalConfigGroup again.
The implementation here
<https://github.com/matsim-org/matsim-libs/blob/e8947ad36e1bc88fcf1b9efa10233ab44c954265/matsim/src/main/java/org/matsim/core/config/groups/GlobalConfigGroup.java#L147>
also does not seem fully correct, because the getter will always return a value significantly greater than 0.
In the SimWrapperConfigGroup, the getter is called, but within the GlobalConfigGroup it is not called.
https://github.com/matsim-org/matsim-libs/blob/e8947ad36e1bc88fcf1b9efa10233ab44c954265/contribs/simwrapper/src/main/java/org/matsim/simwrapper/SimWrapperConfigGroup.java#L102
It would be highly desirable if an exception were not thrown immediately, but instead perhaps only a warning. After all, nothing has actually changed in the QSim logic itself in recent times. So why the hard exception now?

kainagel:
I discuss this in #4706 .

Issue #4675, opened 2026-02-03
Change default compression to lz4


Janekdererste:
We are currently doing benchmarks for a client, where I tried multiple compression algorithms. Using `lz4` instead of `gzip` is considerably faster. This can be seen in the chart below, where `dsim-25%` is run with `gzip` and `dsim-25%-lz4` is run with `lz4` compression. This will improve the overall runtime of the simulation, but, most importantly, can improve the user experience. The usual workflow of both, developers and traffic planners, is to frequently start the simulation until all errors are found. This mostly includes waiting for the input files to be read. 

<img width="4800" height="3600" alt="Image" src="https://github.com/user-attachments/assets/ac1d5bb8-33d2-4b5e-8376-357b1d4aac48" />

MATSim currently supports `gzip`, `lz4`, and `zst` out of the box. `lz4` is the fastest option, but produces the largest files. `zst` with the settings used in MATSim is slower than `lz4`, but yields the best compression rate (small files). `gzip` is slowest and produces larger files compared to `zst` and should not be used anymore.

I would opt for the fastest option, as developer time and runtimes are a constant issue for MATSim. In my opinion, disk space less of an issue nowadays compared to the long runtimes of the simulation.

I would be happy about feedback!

d-roeder:
Maybe @neuma has some more datapoints. A while ago he also looked into it, but I can not really remember the outcome, except that it is a good idea to move away from gzip. We opted for zst in the end.

rakow:
I would also rather vote for zst, which I believe is more common nowadays and even supported in browser and databases. You should be able to configure zst for higher speed and slightly larger files.

mrieser:
I did a very simple test: reading the output_plans.xml.gz from the Open Berlin scenario-6.4-10pct and then writing it out in different formats. There is a method `IOUtils.setZstdCompressionLevel(level)`, but it is not wired up to any config-param or env-variable.

The default compression level that MATSim uses is 6.

```
gz:        118733 ms   size: 1082201350
lz4:        21143 ms   size: 1636769413
zst lvl-1:  25197 ms   size:  760365696
zst lvl-3:  22136 ms   size:  759569318
zst lvl-6:  38480 ms   size:  657580016
zst lvl-9:  51332 ms   size:  585052806
```

It's clearly just a single data point from a single machine, but maybe it helps with the decision.

While lz4 is very fast, I dislike the larger file.

Maybe we could set the default zst-level to 3?

[...]

PR #4699, opened 2026-02-09
refactor(Core): Change default compression to zst

Janekdererste:
### Implementation

Following the brief discussion in #4675, I have changed the default compression format from `gzip` to `zst`. [Zst](https://facebook.github.io/zstd/) seems to be faster compared to `gzip`, while it produces comparable file sizes. Also, files compressed with `zst` can be directly read with commands like `zstdless` or `zstdgrep`.

In addition to changing the default compression algorithm, this PR sets the default compression rate for `zst` compressed files to '3'. 

### Refactoring
Most tests use `gz` as compression algorithm in combination with files that contain expected results. I have set all tests to keep using `gz` for compression, so that we don't have to exchange all the expected results files. Most of the changes are actually the retrofitted tests

### Breaking Changes

I noticed that `AbstractMatsimWriter` features two routes for compressing files. It infers the desired compression by checking the provided file extension (for example gzip for `.gz`). Also, it features a `useCompression(boolean)` method which triggers the writer to gzip compress a file even though the filename does not indicate this. At least in the core project, this method is not used anymore. I marked it as deprecated, as I think it is easier to have one path for configuring the compression of files and providing the correct extension is more flexible.

### DIFF
--- a/contribs/simwrapper/src/test/java/org/matsim/simwrapper/TestScenario.java
+++ b/contribs/simwrapper/src/test/java/org/matsim/simwrapper/TestScenario.java
@@ -9,6 +9,7 @@ import org.matsim.application.analysis.population.TripAnalysis;
 import org.matsim.contrib.vsp.scenario.SnzActivities;
 import org.matsim.core.config.Config;
 import org.matsim.core.config.ConfigUtils;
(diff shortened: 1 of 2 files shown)


Janekdererste:
Closes #4675

Issue #4659, opened 2026-01-26
Full integration tests/Weekly deployment failing (sporadically)


marecabo:
Currently the full integration tests and the weekly deployment are all failing:

* **Full Integration Test (Windows)**  
  fails consistently since DSIM https://github.com/matsim-org/matsim-libs/pull/4255#issuecomment-3625772133  
  Latest log of failure @ Jan 22: https://github.com/matsim-org/matsim-libs/actions/runs/21233928977/job/61097733855#step:5:10854  
* **Full Integration Test (Ubuntu)**  
  fails sporadically when exceeding the runtime of 6h. If it finishes earlier, it succeeds.

* **Weekly Deployment**  
  also recently failed after the 6h timeout: https://github.com/matsim-org/matsim-libs/actions/runs/21123428837/job/60739996272  
  When downloading the log archive, there are a few `java.nio.file.NoSuchFileException` in the end for example `dsim/DrtIntegrationTest/runParallelInserter` (line 757289) and `dsim/ThreeLinkIntegrationTest/oneAgentThreeThreads` (line 758339).
In the end of the log file, some orphaned java processes are terminated.

I have not looked any deeper, but suspect there are some race conditions when writing output files in DSIM that lead to these tests getting stuck? 

@Janekdererste @rakow @mrieser Could you have a deeper look at these test logs?

Janekdererste:
I have fixed some errors regarding the last sync between partitions when the simulation is finished in #4611 

Does the error still persist?

marecabo:
@Janekdererste Thanks, but the latest integration tests still get stuck at 6h unfortunately due to DSIM: https://github.com/matsim-org/matsim-libs/actions/runs/21625835036/job/62325634698#step:5:7804

Janekdererste:
Yes, I started that test. At least it is a different failure now...

I will look into it!

PR #4700, opened 2026-02-09
fix(dsim): Re-enable ThreeLinkIntegrationTest

Janekdererste:
This test was disabled as it waiting for the simulation to finish without a timeout. This issue should be fixed now. Also, I have added a timeout to the awaiting of the futures in the test, which should cause the test to fail faster.


Janekdererste:
Hm, the test still seems to hang sometimes. Maybe this is related to #4659 . Keeping this PR open until I have figured out why the integration test fails sporadically.

Janekdererste:
The changes of this PR should be covered by #4705

PR #4705, opened 2026-02-10
fix(DSim): Fix message exchange at the end of the simulation

### LINKED ISSUE #4659: Full integration tests/Weekly deployment failing (sporadically)

Janekdererste:
The failing full integration CI build revealed a race condition at the end of the simulation. If one process is much slower in during the last simulation timestep than all others, it receives messages meant for consolidating state while it is still processing messages from the simulation. When that process itself reaches the consolidation phase it would wait for messages from other processes indefinetly, as the slow process would have silently processed the consolidation messages already.

We now set the internal sequence of the `MessageBroker` to `INFINITY` right after a process finishes the simulation. This way, all messages send after a process has finished its simulation calculation are distinct from those sent during regular simulation time steps. If another, slower, process receives such a message while it is still processing the simulation, it will store it as `ahead message` which are processed later. 

This PR (hopefully) fixes #4659 
This PR makes #4700 obsolete. The described race condition is why we disabled the particular test in the first place.


Janekdererste:
Waiting on the [full integration test](https://github.com/matsim-org/matsim-libs/actions/runs/21868772614/job/63117378849) to pass.

PR #4730, opened 2026-02-18
fix(core): Make freespeed teleportation threadsafe

samuelhoenle:
`LeastCostPathCalculator`s passed to `createPseudoTransitRouter()` are not guaranteed to be thread-safe (i.e. [`SpeedyALT`](https://github.com/matsim-org/matsim-libs/blob/main/matsim/src/main/java/org/matsim/core/router/speedy/SpeedyALT.java) or [`Dijkstra`](https://github.com/matsim-org/matsim-libs/blob/main/matsim/src/main/java/org/matsim/core/router/Dijkstra.java)), but are then called in a parallelised environment. By passing the factory + components, we can create an instance per routing call, and the operation becomes thread-safe.

I ran into a problem in a simulation with a _lot_ of freespeed teleportation, leading to `NoSuchElementExceptions` in [`DAryMinHeap@decreaseKey()`](https://github.com/matsim-org/matsim-libs/blob/ace48c04f2c6c4b8199fafee8133e775abcb3e75/matsim/src/main/java/org/matsim/core/router/speedy/DAryMinHeap.java#L60). However, looking at the rest of the default routing modules and how they are all just passed a non-thread-safe routing algorithm, I wonder why this isn't a problem there...

Additional automatic code formatting using Google Style + tabs + 150 char line length.

### DIFF
--- a/matsim/src/main/java/org/matsim/core/router/DefaultRoutingModules.java
+++ b/matsim/src/main/java/org/matsim/core/router/DefaultRoutingModules.java
@@ -24,6 +24,9 @@ import org.matsim.api.core.v01.population.PopulationFactory;
 import org.matsim.core.config.groups.RoutingConfigGroup;
 import org.matsim.core.config.groups.RoutingConfigGroup.TeleportedModeParams;
 import org.matsim.core.router.util.LeastCostPathCalculator;
+import org.matsim.core.router.util.LeastCostPathCalculatorFactory;
+import org.matsim.core.router.util.TravelDisutility;
+import org.matsim.core.router.util.TravelTime;
 import org.matsim.core.utils.timing.TimeInterpretation;
 
 import jakarta.annotation.Nullable;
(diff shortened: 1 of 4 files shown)


Janekdererste:
Thanks for your PR. I am having the same problems as described in #4731. Could you limit your changes to the actual code you wanted to change? Then I could review your changes and potentially merge them.

Janekdererste:
Just writing up, so I understand, what's going on: 

The `SpeedyALT` class has the following comment:

> This implementation is not thread-safe. In the case of multi-threading, every thread should use a separate instance. (But the used {@link SpeedyALTData} is thread-safe and can be shared by multiple instances).

So, what you are doing is the correct approach, since we don't really know on which thread a request is handled. 

Do I understand correctly, that the expensive part of the setup for `SpeedyALT` is to create the `SpeedyALTData` (the landmarks)? And, that this is done once per mode network and then cached in the `SpeedyALTFactory`?

If that's the case, we are good to go :-)

samuelhoenle:
Yes, that's how I understand [`SpeedyALTFactory@createPathCalculator`](https://github.com/matsim-org/matsim-libs/blob/e9544af140b13c5721262590f375a1b8f4a0f6a9/matsim/src/main/java/org/matsim/core/router/speedy/SpeedyALTFactory.java#L44-L53) (tagging the author @mrieser). There is no guarantee however that other `LeastCostCalculators` don't have significant costs to being newly created for each routing operation. [`DijkstraFactory`](https://github.com/matsim-org/matsim-libs/blob/main/matsim/src/main/java/org/matsim/core/router/DijkstraFactory.java) appears to be uncached by default. I'm not familiar with the provision of routing algorithms in other places, but in case this is a problem, we could **expand the interface to provide an explicitly cached calculator creation method** to force implementations to deal with this.

Looking through the other implementations of `LeastCostPathCalculator@createPathCalculator`, I just now noticed that a lot of them are `synchronized`, but the one in `SpeedyALTFactory` is not. What's the story here?

Additionally, the question remains why this appears only to be a problem for freespeed teleportation and not for network routes, since if I understand the whole pipeline correctly, it should also be using the A* algorithm by default. Looking through the git blame, maybe @kainagel has some insights on this?

Janekdererste:
At least the factories under `org.matsim.core.router` all have a cache. Since this is according to the specs in the router, I am merging this. Still insight from @mrieser would be interesting as to why this is not happening for network routing.

PR #4752, opened 2026-02-26
fix(core): ensure that travel time of leg and route is consistent in NetworkRoutingInclAccessEgressModule and NetworkRoutingModule

marecabo:
This change ensures that the travel time of a leg is the same as its route when routed with `NetworkRoutingInclAccessEgressModule`.

We noticed, that both got different values, which confused the [`TimeTracker`](https://github.com/matsim-org/matsim-libs/blob/main/matsim/src/main/java/org/matsim/core/utils/timing/TimeTracker.java).

https://github.com/matsim-org/matsim-libs/blob/a9bef5fc27f056236f9bb1ead8bc6130bbc620a0/matsim/src/main/java/org/matsim/core/router/NetworkRoutingInclAccessEgressModule.java#L435-L443

The `travelTimeEstimateOnToLink` was not added to the leg's travel time and there was already a comment asking why this was the case.

When comparing to other routers, the travel time of the leg and route is the same, so I think, we should also do this in the `NetworkRoutingInclAccessEgressModule`.

https://github.com/matsim-org/matsim-libs/blob/a9bef5fc27f056236f9bb1ead8bc6130bbc620a0/matsim/src/main/java/org/matsim/core/router/FreespeedFactorRoutingModule.java#L109-L131


marecabo:
@mrieser Do you think this change is reasonable from your MATSim routing experience?

mrieser:
Thanks for the change. Yeah, the Leg-TravelTime seems to have missed the travel time on the last link, so it makes sense to fix this

marecabo:
When adjusting the `RoutingIT` I noticed that the travel times differed between `accessEgressMode` = `none` and `accessEgressModeToLink` so I also adjusted the deprecated `NetworkRoutingModule` to use the same travel time computation as `NetworkRoutingInclAccessEgressModule`.

PR #4766, opened 2026-03-03
fix(dsim): Storage capacity updates on split links

Janekdererste:
Storage capacities, as well as in and outflow capacities were updated implicitly, when anything would query `isAccepting` or `isOffering` on a `SimLink`. This has the advantage that links only have to be active when they contain a vehicle. 

However, this lead to a subtle bug when `kinematicWaves` are used. In this case a vehicle leaving the q triggers a backwards-travelling-hole, which releases the occupied storage capacity at a later point. With the implicit updating, these holes could arrive anytime any caller queries `isAccepting`. Unfortunately, `SplitInLink` needs to know about storage capacity being released during a sim step, so that it can communicate this change to the upstream link. 

With this change the capacity bookkeeping is explicitly updated in the `doSimStep` method of `SimLink`s. `SimLink`s must stay active until they can accept new vehicles.


Janekdererste:
Closing this PR, as #4767 fixes this in a less complicated way

PR #4781, opened 2026-03-08
fix: attach vehicle to route in NetworkRoutingModule

marecabo:
This change writes the routing vehicle into the route in NetworkRoutingModule.


marecabo:
contained in #4752

PR #4824, opened 2026-03-23
chore(Github): Reduce dependabot frequency to once per day

Janekdererste:
Dependabot create a lot of PRs to review. Especially, with the AWS-SDK shipping a new patch version every other day this is really annoying. 

This PR lets dependabot group all dependency updates of one day. The aws dependcies are handled separately once a week. 

I have excluded `fory` from this, as that project sometimes ships breaking changes. If other dependencies do this as well, we can add them to the exclude list, so that we receive individual PRs for those dependencies, making it easier to handle those updates.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,36 +1,19 @@
 version: 2
 updates:
-  # All Maven dependencies are bundled into a single daily PR.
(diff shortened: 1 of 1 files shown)


Janekdererste:
As a compromise between me (lazy person) and @marecabo (all the updates), we have settled for grouped PRs once a day. Let's see how this works out.

PR #4831, opened 2026-03-25
fix(ev): afterMobsim actually computes open evDrives

Janekdererste:
The `DriveDischargingHandler` collects events and calculates EV discharging in the next time step due to race conditions between the execution of the `QSim` and the parallel events handling. 

Also, it overrides `afterMobsim` to finish the calculation for all `evDrives` which were not completed when the mobsim finished. It does so by passing the mobsim's timer to the calculation logic. That timer actually stops at the last time step of the mobsim, so that the `DriveDischargingHandler` never computes the remaining discharges from the mobsim's last time step, as it waits for a later time step.

This fix increases the time step passed to the calculation, so that it can calculate open `evDrives` from the last time step of the mobsim.

It also adds a test for this behavior and adjusts the expected values of the regression tests I found. 

Possible reviewers: @sebhoerl, @steffenaxer, @marecabo (One is sufficient :-))

## Breaking Changes

Simulation results will differ for EV trips that are not ended before the last time step of the simulation.

### DIFF
--- a/.github/workflows/code-coverage.yaml
+++ b/.github/workflows/code-coverage.yaml
@@ -30,7 +30,7 @@ jobs:
         run: mvn install -P jacoco --batch-mode -Dmaven.test.redirectTestOutputToFile -Dmatsim.preferLocalDtds=true
 
       - name: Push code coverage to CodeCov
-        uses: codecov/codecov-action@v5
+        uses: codecov/codecov-action@v6
         with:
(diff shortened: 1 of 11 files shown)


sebhoerl:
Looks fine to me. Did you change anything in that race condition test? I see that the explanation of the issue has been removed. Maybe we should keep it? As you prefer.

Janekdererste:
Thank you, @sebhoerl 

> Maybe we should keep it

Re-added the comment. I put a comment on top of the test, but the existing one is more detailed.

PR #4844, opened 2026-04-01
fix(drt): Compute sharing metrics per DRT mode

samuelhoenle:
## Problem
When multiple DRT modes are configured, `SharingMetricsTracker` – although instantiated once per mode – performed no mode filtering on the events it received, so every tracker processed trips from all modes and produced identical output regardless of mode.

## Solution
`SharingMetricsTracker` now accepts a `mode` constructor argument and guards all three `handleEvent` methods with an `event.getMode().equals(mode)` check, following the same pattern used by `DrtEventSequenceCollector` and other modal handlers; `SharingMetricsModule` passes `getMode()` to the constructor.

A unit test is also added.

Tagging the original author @nkuehnel for review


nkuehnel:
Thanks, didn't have the multi drt use case in mind back then. Looks good to me

PR #4865, opened 2026-04-16
fix(scenario cutout): do not delete unselected plans

simei94:


contribs/application/src/main/java/org/matsim/application/prepare/scenario/CreateScenarioCutOut.java
```diff
@@ -620,13 +620,14 @@ public void run(Person person) {
 			personsToDelete.add(person.getId());
 		}
 
+//		Please note that we DO want to keep the unselected plans because otherwise agents might need too many iterations to find a good solution. -sm0426
 		// Remove all unselected plans because these are not handled
```
utafrali, line 624: The original comment `// Remove all unselected plans because these are not handled` was left in place unchanged. It now directly contradicts the new comment added on line 623. This will confuse anyone reading this code. Remove this line (or replace it entirely with the new comment, keeping only one).

contribs/application/src/main/java/org/matsim/application/prepare/scenario/CreateScenarioCutOut.java
```diff
@@ -620,13 +620,14 @@ public void run(Person person) {
...
-		List<Plan> plans = new ArrayList<>(person.getPlans());
-		for(Plan p : plans){
-			if (p != person.getSelectedPlan()){
-				person.removePlan(p);
-			}
-		}
+//		List<Plan> plans = new ArrayList<>(person.getPlans());
```
utafrali, line 625: Commented-out code should be deleted, not left in the source. Git history already preserves the original implementation at the previous commit. Leaving these lines here implies the code might be re-enabled later and adds noise to every future reader.

Delete lines 625-630 entirely.

contribs/application/src/main/java/org/matsim/application/prepare/scenario/CreateScenarioCutOut.java
```diff
@@ -620,13 +620,14 @@ public void run(Person person) {
 			personsToDelete.add(person.getId());
 		}
 
+//		Please note that we DO want to keep the unselected plans because otherwise agents might need too many iterations to find a good solution. -sm0426
```
utafrali, line 623: The `-sm0426` author attribution embedded in the comment is informal and unnecessary in production code. `git blame` already records who wrote this line. Strip it and keep only the rationale:

```java
// Unselected plans are kept intentionally: removing them forces agents to rediscover good plans over many iterations.
```

utafrali, line 623: The original comment said unselected plans "are not handled" downstream. This PR silently reverses that decision without explaining whether that concern was invalid, or whether the downstream code was later updated to cope with multiple plans. A sentence in the PR description (or this comment) clarifying the resolution would prevent a future reader from reinstating the old code assuming the original warning was still valid.

Additionally, the existing reference file `withoutEvents/cut_population.xml` (and the `withEvents` variant) may capture output populations. If the test-input plans contain multiple plans per person, those reference files would now be stale and the tests would fail. Please verify (or add a dedicated test) that agents with multiple plans produce the expected output after cutout.

utafrali (commented):
The behavioral intent of keeping unselected plans for better replanning convergence is plausible, but the implementation has several quality problems: the old contradictory comment is left in place, the deleted logic is commented out rather than removed, and an informal author attribution is embedded in a comment. The PR also does not address the original concern that unselected plans 'are not handled' downstream, nor does it add or update tests for the changed behavior.

Issue #3828, opened 2025-03-25
[Discussion] Make use of Github Discussions?


sebhoerl:
Github now offers StackOverflow-style discussions. Should we use them in the example project instead of issues?

mrieser:
Discussion at Code Sprint 2025:
- in favor of using GitHub Discussions instead of GitHub Issues in matsim-code-examples as it is a more obvious format for support-requests
- [ ] adapt matsim.org/faq redirect to discussions once it is set up

Janekdererste:
We have a discussion board for the `matsim-org` organization now. The board is associated with the `matsim-code-examples` project. The URL is the following: https://github.com/orgs/matsim-org/discussions

kainagel:
https://matsim.org/faq should point to this.  If they new location is accepted, could you please change the re-direct?  Probably @mrieser needs to do this.  Thanks a lot ...

Janekdererste:
yes, we are going to do this.

Janekdererste:
I have updated the links on the website in [this](https://github.com/matsim-org/website-11ty/pull/5) PR. 

Now, the redirect from `matsim.org/faq` is missing. I don't have access to that configuration. I think @mrieser must do this.

Janekdererste:
Ok. learned that the redirects of the website are also configured in the website project directly. Redirect works too.

paulheinr:
I also disabled the creation of new issues in code-examples. You are now redirected to the discussion forum.

PR #4876, opened 2026-04-21
refactor(code-examples): integrate code-examples into this repository

paulheinr:
After creating the last release, I once again realized that there are many API changes in the core during one year. It can get really messy to fix all the broken code in the matsim-code-examples if this is done unregularly and not by the corresponding author of the API change. 

This PR is a proposal to integrate the matsim-code-examples as additional maven module into matsim libs. If the community agrees we can archive the code-examples and point to this repository. Having examples in the same repository as the code is not unusual for open source projects. 

I preserved the git commit history for most of the code. I moved the input files to `src/main/resources`, otherwise code referencing them can't find them, because their parent folder is not the working directory any more.

Todo: 
- [x] check whether github actions ran build and test in the code-examples module as well

What do others think about this? @Janekdererste @rakow @marecabo @sebhoerl @kt86 @mrieser @steffenaxer @nkuehnel @jfbischoff 

I have already talked with @kainagel about this.


Janekdererste:
I think this is a very good idea, to keep the examples up to date. 

While thinking about this, I noticed that we receive almost no questions for our Q&A format anymore. Most probably, this is because people just ask their LLM companion nowadays. With this in mind, is it still worth the effort to maintain an extensive set of examples like in the examles project, or is this something noone needs anymore?

paulheinr:
Personally, I use the code examples regularly. We also use this in teaching. However, I have no experience using LLMs instead of very MATSim specific code examples. 

Maybe this is the main function of the code examples for now: Being a training set for LLMs 😄

sebhoerl:
Looks like a good idea to keep things up to date and reduce maintenance effort. :+1:

kt86:
Good idea, to centralize it and have it up-to-date!
Also it reduces a bit the complexity to find things. 

I also use these examples for teachings or other people just starting to use MATSim (freight)... so, I am also in favor, to still have this examples available.

jfbischoff:
I absolutely love the idea!

paulheinr:
Thank you all for your feedback. I would then proceed with the plan. After merging this PR, I'll adapt the readme of code-examples repository and eventually archive the code-examples repository. 

Since the matsim-org discussions are set up in the code examples, I would move them as well into matsim-libs. Any objections?

paulheinr:
Reminder to myself: temporarily disable squashing of commits in branch protection rules (otherwise, git history won't be visible)

kt86:
> Since the matsim-org discussions are set up in the code examples, I would move them as well into matsim-libs. Any objections?

I remember, that there was a discussion during the code sprint a few years ago. After that we decided, to move the discussions/faq away from matsim-libs. Unfortunately, I do not remember the reasons for that move.
In case, that it will be moved back, the pointer from matsim.org/faq needs to be moved as well.

paulheinr:
> I remember, that there was a discussion during the code sprint a few years ago. After that we decided, to move the discussions/faq away from matsim-libs. Unfortunately, I do not remember the reasons for that move.

I remember this discussion as well. The notes regarding that point are relatively sparse at the corresponding issue #3828. But I don't really see an alternative to enabling discussions in matsim-libs. Since everybody was very much in favor for getting rid of the code-examples repository, I would just do it.

paulheinr:
@rakow @Janekdererste for the second time the sbb extensions checks are killed because runtime exceeded in the merge queue :( i have seen this somewhere else. Do you have an idea why this is the case?

PR #4877, opened 2026-04-22
feat: Elevation data parser tiff crs

anezkap:
Disclaimer: I am very new to MATSim, so it is possible I missed something, but the original implementation was not working for me, and hopefully this will make it clearer for anyone trying to use it in the future :)

#### Title
feat/fix(ElevationDataParser): remove hardcoded TIFF CRS and add it as a parameter

#### Body
I believe that the hardcoded CRS in the ElevationDataParser for the Tiff file shouldn't be there.

Of course, everyone can change it in the file directly themselves, but adding tiffCRS as a parameter to the ElevationDataParser feels a bit cleaner since that way, we only need to define it in CreateBicycleNetworkWithElevation. 

If we do not want to update the ElevationDataParser itself, that is fine as well, but maybe we could just add a little comment there explaining what is going on, since it took me way too much time to understand where the problem is and why it's not working for my scenario.

(Not sure if this should be marked as BREAKING CHANGE, sorry)


Janekdererste:
This looks good to me. Thank you for the effort :-)

PR #4889, opened 2026-04-27
ci: Fix failing deploy on PR workflow

Janekdererste:
Some Github API was not available causing our deploy to crash. This pr adds write permissions for the runner and forces it to continue even if the graph step fails.

Also, updates node version from 20 to 24, as 20 is going to be deprecated soon.


Janekdererste:
I guess, I'll have to merge the PR to see if the script works? 🤷‍♀️

PR #4894, opened 2026-04-28
feat: set verbosity of least-cost path calculator

sebhoerl:
This PR adds `LeastCostPathCalculator.setNoPathBehavior`.

One can now choose:
- Print a long warning (default, current default) when the calculator returns `null`
- Raise an exception
- Do nothing (mute)

The second option is useful for debugging. The third option is useful to avoid millions of log messages where the least-cost path calculator is rather used to test whether a route exists (see `pt2matsim`) rather than actually using the route.


sebhoerl:
Not sure if it is the best way to implement this, but I didn't want to mess with all the LCPC constructors. Comments welcome.

nkuehnel:
Thanks, the pt2matsim case is a relief :) @marecabo

sebhoerl:
Thanks for cleaning up, I can have a look at the remaining cases

jfbischoff:
Should be done now :)

PR #4904, opened 2026-04-30
fix(sbb-extensions): Use sim-step time for link events in SBBTransitEngine

marecabo:
Commit 1711de74 (#4896) introduced a regression in `SBBTransitEngine` where `LinkLeaveEvent` / `LinkEnterEvent` are fired at fractional precomputed transit times instead of the integer sim-step time. This causes a `RuntimeException` in `SimStepParallelEventsManagerImpl` because events from `SBBTransitEngine` (e.g. 62.4s) can appear *after* events from other engines (e.g. `QNetsimEngine` at 63.0s) within the same sim step, violating chronological ordering.

### Before: `SBBTransitQSimEngine` (deleted in #4896)

Link events were precomputed at fractional times but fired using the integer `time` parameter from `doSimStep()`:

https://github.com/matsim-org/matsim-libs/blob/7fa07295a7251424f76f19a8ddb06034c2e57f66/contribs/sbb-extensions/src/main/java/ch/sbb/matsim/mobsim/qsim/pt/SBBTransitQSimEngine.java#L152-L161

### After: `SBBTransitEngine` (introduced in #4896)

The new `handleLinkTransition()` fires link events directly at `e.time` (the fractional precomputed time):

https://github.com/matsim-org/matsim-libs/blob/1711de74666180cebacd0c5218c5ad9660c8d2d2/contribs/sbb-extensions/src/main/java/ch/sbb/matsim/mobsim/qsim/pt/SBBTransitEngine.java#L240-L248

### Approach

Store the current sim-step time in `doSimStep()` and use it for link events, exactly matching how the old `SBBTransitQSimEngine` passed its `time` parameter:

```java
private double currentSimStepTime = 0.0;

@Override
public void doSimStep(double now) {
    this.currentSimStepTime = now;
    // ...
}

private void handleLinkTransition(TransitEvent e) {
    var transition = e.context.precomputedLinkTransitions.element();
    var vehicle = e.context.driver.getVehicle();
    em.processEvent(new LinkLeaveEvent(this.currentSimStepTime, vehicle.getId(), transition.fromLink()));
    em.processEvent(new LinkEnterEvent(this.currentSimStepTime, vehicle.getId(), transition.toLink()));

    var nextEvent = e.context.computeEventOnLinkTransition(e.time);
    enqueueOrSend(nextEvent);
}
```

### Stack Trace 

```
at SimStepParallelEventsManagerImpl.processEvent(...)
at SBBTransitEngine.handleLinkTransition(SBBTransitEngine.java:243)
at SBBTransitEngine.handleTransitEvent(SBBTransitEngine.java:198)
at SBBTransitEngine.doSimStep(SBBTransitEngine.java:189)
at QSim.doSimStep(QSim.java:395)
```

### DIFF
--- a/contribs/sbb-extensions/src/main/java/ch/sbb/matsim/mobsim/qsim/pt/SBBTransitEngine.java
+++ b/contribs/sbb-extensions/src/main/java/ch/sbb/matsim/mobsim/qsim/pt/SBBTransitEngine.java
@@ -63,6 +63,7 @@ public class SBBTransitEngine
 
 	private InternalInterface internalInterface;
 	private boolean createLinkEvents = false;
+	private double currentSimStepTime = 0.0;
 
 	@Inject
 	public SBBTransitEngine(ReplanningContext context, SBBTransitConfigGroup config, TransitConfigGroup ptConfig,
@@ -180,6 +181,7 @@ public class SBBTransitEngine
 
 	@Override
 	public void doSimStep(double now) {
+		this.currentSimStepTime = now;
 
 		while (!eventQueue.isEmpty()) {
 			var head = eventQueue.peek();
@@ -240,9 +242,8 @@ public class SBBTransitEngine
 	private void handleLinkTransition(TransitEvent e) {
 		var transition = e.context.precomputedLinkTransitions.element();
 		var vehicle = e.context.driver.getVehicle();
-		double now = Math.floor(e.time);
-		em.processEvent(new LinkLeaveEvent(now, vehicle.getId(), transition.fromLink()));
-		em.processEvent(new LinkEnterEvent(now, vehicle.getId(), transition.toLink()));
+		em.processEvent(new LinkLeaveEvent(this.currentSimStepTime, vehicle.getId(), transition.fromLink()));
+		em.processEvent(new LinkEnterEvent(this.currentSimStepTime, vehicle.getId(), transition.toLink()));
 
 		var nextEvent = e.context.computeEventOnLinkTransition(e.time);
 		enqueueOrSend(nextEvent);
--- a/contribs/sbb-extensions/src/main/java/ch/sbb/matsim/mobsim/qsim/pt/SBBTransitEngine.java
+++ b/contribs/sbb-extensions/src/main/java/ch/sbb/matsim/mobsim/qsim/pt/SBBTransitEngine.java
@@ -63,7 +63,6 @@ public class SBBTransitEngine
 
 	private InternalInterface internalInterface;
 	private boolean createLinkEvents = false;
-	private double currentSimStepTime = 0.0;
 
 	@Inject
 	public SBBTransitEngine(ReplanningContext context, SBBTransitConfigGroup config, TransitConfigGroup ptConfig,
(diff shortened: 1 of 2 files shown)


marecabo:
@Janekdererste We encountered this after upgrading MATSim and suddenly hat events with non-integer times resulting in a crashing `SimStepParallelEventsManagerImpl`.

marecabo:
**`SBBTransitEngineTest.java`**: Fixed inconsistent engine ordering in `testEvents_withoutPassengers_withLinks_SesselbahnMalformed` — it registered `SBBTransitEngine` before `ActivityEngine`, unlike all other tests. This caused transit events to be delayed by one sim step, masking/triggering ordering issues. Now uses Activity-first ordering consistent with its sibling test.

contribs/sbb-extensions/src/main/java/ch/sbb/matsim/mobsim/qsim/pt/SBBTransitEngine.java
```diff
@@ -63,6 +63,7 @@ public class SBBTransitEngine
 
 	private InternalInterface internalInterface;
 	private boolean createLinkEvents = false;
+	private double currentSimStepTime = 0.0;
```
Janekdererste: I don't think we should store the time. `handleLinkTransition` is called from `doSimStep`, which receives `now` as a parameter. Instead of caching the time I would prefer to pass `now` through the call stack to `handleLinkTransition`. Then the event dispatch becomes:

```
em.processEvent(new LinkLeaveEvent(now, vehicle.getId(), transition.fromLink()));
```

contribs/sbb-extensions/src/main/java/ch/sbb/matsim/mobsim/qsim/pt/SBBTransitEngine.java
Janekdererste: I guess these two event disptaches have the same problem? We should use `now` here as well.

contribs/sbb-extensions/src/main/java/ch/sbb/matsim/mobsim/qsim/pt/SBBTransitEngine.java
```diff
@@ -240,8 +242,8 @@ private void handleDepartureAtStop(TransitEvent event) {
 	private void handleLinkTransition(TransitEvent e) {
 		var transition = e.context.precomputedLinkTransitions.element();
 		var vehicle = e.context.driver.getVehicle();
-		em.processEvent(new LinkLeaveEvent(e.time, vehicle.getId(), transition.fromLink()));
-		em.processEvent(new LinkEnterEvent(e.time, vehicle.getId(), transition.toLink()));
+		em.processEvent(new LinkLeaveEvent(this.currentSimStepTime, vehicle.getId(), transition.fromLink()));
+		em.processEvent(new LinkEnterEvent(this.currentSimStepTime, vehicle.getId(), transition.toLink()));
```
Janekdererste, line 245: As said above, I would prefer a method parameter `now`.

Janekdererste (dismissed):
Good catch! Thank you. I have added two comments.

marecabo:
@Janekdererste Thanks, I addressed the comments

Janekdererste (approved):

PR #4913, opened 2026-05-05
refactor(parking): move parking stuff from the core to the parking contrib

kainagel:


paulheinr:
Closes https://github.com/matsim-org/matsim-code-examples/issues/1184

PR #4915, opened 2026-05-05
fix(conventional-commits): fix title spec

paulheinr:
According to https://github.com/amannn/action-semantic-pull-request, the config needs to be adapted, otherwise this workflow fails


kainagel:
:-(. Danke

PR #4960, opened 2026-05-15
maintenance: permit duplicates in the ZoneSystems of DvrpModule

kainagel:
I am having some legacy code that I currently cannot fix without this.  Do not know the possible side effects.
This should not break anything.  But it might have spurious side effects in future usages of this class.


nkuehnel:
@kainagel thanks for your comments. 

So far they're only documented in the related config settings, e.g. for late diversion threshold and max allowed pickup delay:
https://github.com/matsim-org/matsim-libs/blob/main/contribs/drt/src/main/java/org/matsim/contrib/drt/optimizer/constraints/DrtOptimizationConstraintsSet.java
in short: late diversion threshold is meant to control at which remaining time to the destination no further detour for a boarded passenger is allowed (passengers dont like to be rerouted for another round around the block when they can already see their dropoff point). Max allowed pickup delay is the maximum amount of time the _initially commited_ pickup time can be further delayed by new insertions. 


Schedule wait before drive controls whether vehicles that have a stay task before the next pickup (eg in the case of prebooking) should first drive to the pickup and wait or vice versa.

I can take a look at documenting it in further places in the code

kainagel:
@nkuehnel : thanks.  I put a reference to that config group into the DrtRouteConstraints class.  This is on the kn/mvVulkaneifel... , which will take some time until it is merged.

PR #4996, opened 2026-06-09
fix: BicycleParamsDefaultImpl.java -- infrastructure factor for highway cycleway

anezkap:
Hey hey, 

I believe that that the tag highway=cycleway is actually not usually used together with the cycleway tag, since the cycleway tag specifies a cycle path that is part of a road, whereas highway=cycleway means that it is already a completely separate cycle path. https://wiki.openstreetmap.org/wiki/Tag:highway%3Dcycleway

Hence, highway=cycleway will fall into the hasNoCycleway if branch and will be assigned 0.95 by default, while it should have the best possible infrastructure factor of 1. 

In my network (Brussels), I only ever saw type=cycleway and cycleway not being set or type=anything else and cycleway being set.

### DIFF
--- a/contribs/bicycle/src/main/java/org/matsim/contrib/bicycle/BicycleParamsDefaultImpl.java
+++ b/contribs/bicycle/src/main/java/org/matsim/contrib/bicycle/BicycleParamsDefaultImpl.java
@@ -48,6 +48,7 @@ public class BicycleParamsDefaultImpl implements BicycleParams {
 		// janek may '23
 
 		if (type == null) return 0.85;
+		if (type.equals("cycleway")) return 1.0;
 		if (hasNoCycleway(cyclewaytype)) {
 			return switch (type) {
 				case "trunk" -> 0.05;


paulheinr:
@simei94

simei94:
Hey, thanks for making us aware. We will have a look at the issue / discuss it internally. As a workaround you could bind your own BicycleParams implementation instead of changing the default one if you want to move forward quickly.

anezkap:
Hey, yes, no worries, I already changed it locally, so it is not a blocker for me, just a small suggestion. So take your time!

Also, it is just a tiny thing that does not change the behaviour drastically, just more of a fyi :)

d-roeder:
Hey @vizsim,

could you also have a look at this?

Thank you and best regards,
Daniel

vizsim:
Thanks @anezkap, good catch!
The fix is correct: a standalone `highway=cycleway` has no `cycleway` tag, so it was wrongly hitting the 0.95 default instead of 1.0.

The broader question of how `DefaultBicycleParams` should categorize cycling infrastructure (and the naming ambiguity behind `cycleway`) is already tracked in #5002, so let's keep that discussion there and merge this small fix as-is.

vizsim (approved):

d-roeder:
thanks @vizsim

PR #4999, opened 2026-06-09
feat: parsing cycleway tags with side attributes in OsmBicycleReader

anezkap:
Hi all,

I noticed (and this was also discussed in one issue a long time ago), that the OsmBicycleReader only reads tags called exactly 'cycleway' while the recommendation is to use this tag together with the side attribute, e.g., cycleway:left (https://wiki.openstreetmap.org/wiki/Key:cycleway).

In my scenario, for Brussels, from the links that had the cycleway tag (even though there were only few of them), only around one quarter were using 'cycleway', the rest were using the side attribute as well, and therefore these were ignored by the reader.

So I just thought this might be helpful for others as well. It needs to be noted, though, that the logic assumes right-side driving traffic.

### DIFF
--- a/contribs/osm/src/main/java/org/matsim/contrib/osm/networkReader/OsmBicycleReader.java
+++ b/contribs/osm/src/main/java/org/matsim/contrib/osm/networkReader/OsmBicycleReader.java
@@ -77,9 +77,26 @@ public final class OsmBicycleReader extends SupersonicOsmNetworkReader {
 	}
 
 	private static void setCycleWay(Link link, Map<String, String> tags) {
-		if (tags.containsKey(OsmTags.CYCLEWAY))
-			link.getAttributes().putAttribute(OsmTags.CYCLEWAY, tags.get(OsmTags.CYCLEWAY));
-	}
+        // Note: this logic assumes right-hand traffic.
+        // For forward links, the right side corresponds to the link direction.
+        // For reverse links, the left side corresponds to the link direction.
+        String cycleway = null;
+        String linkId = link.getId().toString();
+
+        if (tags.containsKey(OsmTags.CYCLEWAY)) {
+            cycleway = tags.get(OsmTags.CYCLEWAY);
+        } else if (tags.containsKey(OsmTags.CYCLEWAY_BOTH)) {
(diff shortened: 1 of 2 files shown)


Janekdererste:
Maybe @vizsim can review and merge, if correct?

vizsim:
Hey Janek,

I had a closer look at #4999. A few thoughts.

The problem is real. The bare `cycleway` tag is the older, coarser way of tagging, and these days OSM mostly uses the side specific `cycleway:left` / `cycleway:right` / `cycleway:both`, so the current reader misses quite a bit of infrastructure. Definitely worth fixing.

The solution looks plausible for the case it targets and seems to go in the right direction.

One thing I wasn't sure about: the direction is currently derived by parsing the link ID suffix (`linkId.endsWith("f")` / `"r"` / `"_bike-reverse"`). But `handleLink` already gets a `Direction` parameter, so would it maybe make sense to use that directly instead? My (possibly naive) worry is that the suffix approach relies on the superclass's internal ID naming (`segmentId + "f"/"r"`), which feels a bit fragile to me. But you'll know much better than I do whether that's actually a concern.

There also aren't any tests. Honestly I'm not sure how necessary that is for a change like this, would that be important to add?

On the practical side, I've never actually merged a PR before, at least not within MATSim, so could you walk me through how it works here?

Last thing: I'm reworking this whole area (and a lot more) in a planned bigger rewrite of the bicycle contrib and the network reader, now as a pipeline in the bicycle contrib. One thing I ran into there: the link attribute this PR (and the current implementation) writes into is also called `cycleway`, same as the raw OSM tag, even though at that point it's really a processed MATSim value and has to cover more than just cycleways (bicycle roads / Fahrradstraßen, direction infos and so on). So in the rewrite I rename it to `bicycle_infra` to make clear it's a MATSim side attribute and not the OSM tag, which means it's basically a fresh implementation anyway. The old `cycleway` attribute can just keep living next to this change for now though, so I don't think this PR needs to wait for the rewrite. WIP for reference:
https://github.com/matsim-org/matsim-libs/tree/bicycle-osm-infra-reader/contribs/bicycle/src/main/java/org/matsim/contrib/bicycle/network

Cheers,
Simon

PR #5017, opened 2026-06-16
perf(io): unsynchronized FastBufferedWriter for MATSim file writing

marecabo:
Replaces the JDK's synchronized `BufferedWriter` in MATSim's writers with an unsynchronized buffered writer, making scenario serialization ~3× faster with byte-identical output.

### Problem

JFR profiling showed ~72% of self-time during scenario serialization is spent in `java.io.BufferedWriter.write`'s per-call synchronization, because the XML writers emit each element as dozens of tiny `write(...)` calls. This single-threaded synchronization cost dominates writing for large scenarios (persons, households, vehicles, network, facilities).

### Solution

Add `FastBufferedWriter`, an unsynchronized drop-in `BufferedWriter` with a 64 KB buffer, and wire it into `AbstractMatsimWriter` so all writers benefit transparently. A property-gated A/B microbenchmark (enable with `-Dmatsim.benchmark=true`) confirms **3.13× faster writing (68% less time)**, and a 17-case differential test verifies byte-identical output against `java.io.BufferedWriter`.

The relevant change in `AbstractMatsimWriter` is here:
https://github.com/matsim-org/matsim-libs/pull/5017/changes#diff-fcfc872dfbfdd6424d93f45ac798a618bd905158aa344b1a71729025852cdc5dR71-R93

### DIFF
--- a/contribs/analysis/src/main/java/org/matsim/contrib/analysis/vsp/qgis/QGisFileWriter.java
+++ b/contribs/analysis/src/main/java/org/matsim/contrib/analysis/vsp/qgis/QGisFileWriter.java
@@ -1,7 +1,7 @@
 package org.matsim.contrib.analysis.vsp.qgis;
 
-import java.io.BufferedWriter;
 import java.io.IOException;
+import java.io.Writer;
 import java.nio.file.Paths;
 
 /**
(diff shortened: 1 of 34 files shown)


Janekdererste:
what is a property-gated A/B microbenchmark?

Janekdererste:
Also, super interesting, that the Documenation doesn't even mention that `BufferedWriter` is thread-safe/synchronized.

marecabo:
> what is a property-gated A/B microbenchmark?

@Janekdererste The benchmark is a test that is only run with `-Dmatsim.benchmark=true`: https://github.com/matsim-org/matsim-libs/pull/5017/changes#diff-13e6d5229e8deb0ee8ff5bd03f2780d84c1a2740641fb354be17f4276ab318d7R70-R71

PR #5023, opened 2026-06-18
fix(EV): avoid generating events "in the future" in aftermobsim

tduberne:
#### fix(EV): avoid generating events "in the future" in aftermobsim

this lead to crashes when other modules called afterwards generated events "in the present".

The fix works by separating the cutoff time to process events from the time of the events themselves.


sebhoerl:
Looks good to me, but @Janekdererste was the last one to touch this nasty logic, so maybe he has a different view :)

sebhoerl (approved):

Janekdererste:
Sorry for the late response. I resorted to implement a mirroring `DistributedDriveDischargingHandler` for `DSim`, so this is not touched by my changes. Unortunately, both implementations are necessary at the moment as `QSim` and `Dsim` handle global state differently and also provide different garuantees about when events are passed to event handlers.

PR #5063, opened 2026-07-15
feat: add default params for intermodal walk access/egress to pt for urban and rural regions

simei94:


contribs/vsp/src/main/java/org/matsim/contrib/vsp/pt/intermodal/VspDefaultIntermodalWalkAccessEgressParams.java
```diff
@@ -0,0 +1,70 @@
...
+import org.apache.logging.log4j.Logger;
+import org.matsim.api.core.v01.TransportMode;
+
+import java.util.HashSet;
+import java.util.Set;
+
+public class VspDefaultIntermodalWalkAccessEgressParams {
+	private static final Logger log = LogManager.getLogger(VspDefaultIntermodalWalkAccessEgressParams.class);
+
+	public static void setAndAddExplicitIntermodalityParamsForRuralWalkToPt(SwissRailRaptorConfigGroup srrConfig) {
```
jfbischoff, line 30: curiously looking at this. Should this radius perhaps not be a scenario wide setting but something more flexible? We end up having these discussions too :)

vsp-gleich: Probably yes, especially in mixed scenarios (urban area with rural surroundings). But that would need additional resources to implement... (even if we would use existing code, such as SBB's intermodal).

simei94: I agree with you. But thinking from a pragmatic point of view: Wouldn't a mixed scenario just mean that we want to ensure that the stops for the rural agents are findable? Hence, we would take the upper bound := rural radius settings. For the urban agents we would find enough stops in the first try anyways.

jfbischoff: Perhaps, we should think the problem the other way around, i.e., from the stop's perspective rather than from the person's perspective. You are more likely to walk 1.5km to a central rail station than to a bus stop in the city. In that case, only an upper bound needs to be set in the config (let's say, 10km or whatever the highest catchment area of any stop is) and every stop that has a lower catchment area is sorted out.

vsp-gleich: @simei94 rural settings -> bigger initial search radius -> much more computation time. Can we afford that in say Dresden + half of Saxony?
@jfbischoff Here the task was to get along with the current code. But yes, there are ways to improve the stop search in code.

PR #5085, opened 2026-07-28
fix: adapt parking search time at zero parking availability and updated co…

GregorRyb:
…rresponding test.


paulheinr (commented):
Have you measured what impact this change has? 

> To model parking scarcity, provide an explicit positive parking capacity. 

I don't find this obvious. Imagine you would like to implement a "parking spot free zone". The natural approach would be, to just remove the whole parking capacity. With this change, how would I model it? 

If you really want 0 parking search time in case of no parking supply, I would add a config switch for that.

GregorRyb:
Hi, I updated the comment based on our discussion today. Hope this is now sufficient.

paulheinr (approved):

PR #5086, opened 2026-07-29
feat: policies

GregorRyb:
During a project, the need arose to model simple policies in a standardized way within MATSim.


contribs/vsp/src/main/java/org/matsim/contrib/vsp/scenario/CornerCases.java
```diff
@@ -0,0 +1,259 @@
...
+
+			bike.setMaximumVelocity(
+				bike.getMaximumVelocity() * factor
+			);
+
+			log.info(
+				"Modified network bike speed by factor {}",
+				factor
+			);
+		}
```
d-roeder, line 98: if neither option1 nor option 2 is valid this is silently doing nothing. Maybe add a return to the second if-statement and throw an exception if you reach this point here?!

contribs/vsp/src/main/java/org/matsim/contrib/vsp/scenario/CornerCases.java
```diff
@@ -0,0 +1,259 @@
...
+			ShpGeometryUtils.loadPreparedGeometries(
+				IOUtils.resolveFileOrResource(shp.toString())
+			);
+
+		scenario.getNetwork().getLinks().values().stream()
+			.filter(link -> link.getAllowedModes().contains(TransportMode.car))
+			.filter(link -> isInShape(link, geometries))
+			// TODO: The "type" attribute is optional. Guard against null and non-string
+			// values here; otherwise an untyped car link aborts the operation.
+			.filter(link -> !((String) link.getAttributes().getAttribute("type")).contains("motorway"))
```
d-roeder: maybe pass typesToExclude as argument to make this configurable

contribs/vsp/src/main/java/org/matsim/contrib/vsp/scenario/CornerCases.java
```diff
@@ -0,0 +1,259 @@
...
+			.forEach(link ->
+				link.setFreespeed(link.getFreespeed() * factor)
```
d-roeder: maybe add a marker to the attributes. Otherwise it is complicated to see which links were modified later

contribs/vsp/src/main/java/org/matsim/contrib/vsp/scenario/CornerCases.java
```diff
@@ -0,0 +1,259 @@
...
+		);
+	}
+
+	/**
+	 * Reduce the capacities of car links in the specified area,
+	 * excluding motorways and trunks.
+	 */
+	public static void reduceCarCapacities(
+		Scenario scenario,
+		List<PreparedGeometry> areaFilter,
```
d-roeder: The other methods expect a path to a shapefile and load the PreparedGeometry on their own. Why not here?

contribs/vsp/src/main/java/org/matsim/contrib/vsp/scenario/CornerCases.java
```diff
@@ -0,0 +1,259 @@
...
+
+			.filter(link -> link.getAllowedModes().contains(TransportMode.car))
+			.filter(link -> isInShape(link, areaFilter))
+
+			.filter(link -> {
+				String typeObj = (String) link.getAttributes().getAttribute("type");
+				// TODO: Missing link types are currently excluded, although this method
+				// is documented to exclude only motorways and trunks.
+				return typeObj != null
+					&& !typeObj.contains("motorway")
```
d-roeder: maybe make this configurable as well?

contribs/vsp/src/main/java/org/matsim/contrib/vsp/scenario/CornerCases.java
```diff
@@ -0,0 +1,259 @@
...
+
+				// TODO: Validate that reductionFactor is finite and in the intended range.
+				// Very small or non-positive values can produce invalid network properties.
+				if (link.getCapacity() > 0) {
+					link.setCapacity(link.getCapacity() * reductionFactor);
+				}
+
+				if (link.getNumberOfLanes() > 2.0) {
+					// TODO: Consider clamping the result to at least one lane.
+					link.setNumberOfLanes(link.getNumberOfLanes() * reductionFactor);
```
d-roeder: Do we really want to work with a non-integer number of lanes?

contribs/vsp/src/main/java/org/matsim/contrib/vsp/scenario/CornerCases.java
```diff
@@ -0,0 +1,259 @@
...
+			"Reduced car capacities by factor {} in selected area",
+			reductionFactor
+		);
+	}
+
+	// Returns true if either endpoint of the link is in any of the given geometries.
+	// TODO: This misses links that cross a geometry while both endpoints are outside.
+	// Boundary points may also be excluded by the underlying "contains" test.
+	private static boolean isInShape(Link link, List<PreparedGeometry> geometries) {
+	    return ShpGeometryUtils.isCoordInPreparedGeometries(
```
d-roeder: Depending on the number of geometries this might be very slow.

contribs/vsp/src/main/java/org/matsim/contrib/vsp/scenario/CornerCases.java
```diff
@@ -0,0 +1,259 @@
...
+			});
+
+		log.info(
+			"Reduced car capacities by factor {} in selected area",
+			reductionFactor
+		);
+	}
+
+	// Returns true if either endpoint of the link is in any of the given geometries.
+	// TODO: This misses links that cross a geometry while both endpoints are outside.
```
d-roeder: Since we do not work with high-resolution geometries the approacch with from/to-node seems to be very plausible to me. 

Assume we have a very long link like an "Umgehungsstraße". Since we do not use detailed geometries this link might be caught by an area where it actually does not belong to and this might produce more problems.

d-roeder (approved):

PR #5089, opened 2026-07-29
feat(core): per-activity typical durations in CharyparNagelActivityScoring

michaz:
Allows scoring each activity against its own typical duration instead of its activity type's, replacing the duration-binned-activity-types workaround (`home_30`, `home_60`, ...).

- New `TypicalDurationCalculator` strategy interface, analogous to `OpeningIntervalCalculator`; `ActivityAttributeTypicalDurationCalculator` reads the `typicalDuration` activity attribute (config fallback when absent).
- The zero-utility duration is recomputed consistently with the type's configured computation mode and priority, via the new `ActivityUtilityParameters.computeZeroUtilityDuration_h`.
- Activities handed to the scoring at runtime are reconstructed from events and carry no attributes, so `CharyparNagelActivityScoring` can optionally be given the `Person` and then aligns them with the selected plan's main activities (resolved lazily after replanning; warns and falls back to config parameters on divergence).
- No config switch: enabled by constructing/binding the scoring accordingly. Default constructors are behaviorally unchanged.
- vsp VTTS analysis: `MarginalSumScoringFunction` and `VTTSHandler` accept an activity-scoring factory, create fresh scoring instances per marginal computation, and give their event-reconstructed activities the attributes of the corresponding selected-plan activities.

[AI detection: 100% generated](https://claude.com/claude-code)

### DIFF
--- a/contribs/vsp/src/main/java/org/matsim/application/analysis/population/MarginalSumScoringFunction.java
+++ b/contribs/vsp/src/main/java/org/matsim/application/analysis/population/MarginalSumScoringFunction.java
@@ -30,6 +30,8 @@ import org.matsim.core.scoring.functions.ActivityUtilityParameters;
 import org.matsim.core.scoring.functions.CharyparNagelActivityScoring;
 import org.matsim.core.scoring.functions.ScoringParameters;
 
+import java.util.function.Supplier;
+
 /**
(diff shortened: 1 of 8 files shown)


jfbischoff:
Nice and overdue addition :)

Issue #3150, opened 2024-03-11
SBB PT skims issues


panostsolerid:
Hi,

This is an issue somewhat related to a previous open one [(here)](https://github.com/matsim-org/matsim-libs/issues/2115). I am trying to make sense of SBB skims for PT specifically (no issues with car skims). I have summarised my issues/findings below.

1) Do the travel times in `pt_travel_times.csv.gz` refer to the total travel time for an OD pair or just the in-vehicle time? The previous issue I have attached above seems to imply that it is the former. But at the same time the documentation in [here](https://github.com/SchweizerischeBundesbahnen/matsim-sbb-extensions?tab=readme-ov-file#skim-matrices) says that it is the in-vehicle time from the first to the last stop.

2) Are the travel times in `pt_travel_times.csv.gz` (and also in the rest e.g. `pt_egress_times.csv.gz` and `pt_access_times.csv.gz`) normalised in any way e.g. by hour or by frequency of services for a specific time period? I have tried several different time periods, namely i) 07:00-08:00, ii) 08:00-09:00, iii) 07:00-09:00 and iv) 07:00-10:00.
For i) and ii) I am getting x mins for an OD, for iii) i am getting almost 2x mins and for  iv) i am getting almost 3x mins. 

Initially, I thought that i would have to normalise by the number of hours for each specified period, but then we ran some comparisons with google maps with a colleague.

From those experiments, it appears that when we divide the travel time for an OD in `pt_travel_times.csv.gz` with the frequency of services in `pt_frequencies.csv.gz`, we are getting the same travel time that Google suggests for a station to station trip. 

So does this mean that the travel times in `pt_travel_times.csv.gz` are indeed in-vehicle times and that we need to normalise them by the frequency (number of services) for that time period? Are the travel times in `pt_travel_times.csv.gz` the cumulative time from all services running in that OD during that period?

Any help would be greatly appreciated !

Thanks and best regards!
Panos

steffenaxer:
@mrieser or @jfbischoff you both might be the best to answer this issue.

Issue #2115, opened 2022-07-27
Issues with SBB skim matrices


markusstraub:
I believe (maybe I understood something wrong, @mrieser ?) I found the following issues with the SBB skim matrices, which, by the way, are a very helpful feature :)

## pt access/egress calculation time is not normalized to 1 hour

When calculating access/egress times [RooftopUtils.calcConnectionShares](https://github.com/matsim-org/matsim-libs/blob/43f4951f2e48c1a222457a32879e07d98e129166/contribs/sbb-extensions/src/main/java/ch/sbb/matsim/analysis/skims/RooftopUtils.java#L142) does not normalize the given duration. This means the sum of all shares is 1 for a 1 hour time window (as expected), but 2 for a 2 hour time window. This then results in the access/egress times (in the matrices) being too large.

## pt travel time

The [javadoc](https://github.com/matsim-org/matsim-libs/blob/43f4951f2e48c1a222457a32879e07d98e129166/contribs/sbb-extensions/src/main/java/ch/sbb/matsim/analysis/skims/PTSkimMatrices.java#L32) explains, that the travel time matrices such as `pt_travel_times.csv.gz` contain the pure travel time. This does not seem to be the case, it includes access and egress times - as expected when looking at [PTSkimMatrices](https://github.com/matsim-org/matsim-libs/blob/43f4951f2e48c1a222457a32879e07d98e129166/contribs/sbb-extensions/src/main/java/ch/sbb/matsim/analysis/skims/PTSkimMatrices.java#L247)

That's no big deal but the documentation and code are out of sync.

PR #5123, opened 2026-08-13
fix(sbb-extensions): pt skim normalization and documentation

Hussein-Mahfouz:
## Summary

Normalizes SBB PT skim connection shares by the selected departure-time window instead of assuming a one-hour window. 

Relevant issues: #2115, #3150

This fixes cases where PT skim values such as travel time, access time, and egress time scaled roughly with the number of hours in the requested skim period.

  ## Details

`RooftopUtils.calcConnectionShares(...)` previously divided connection coverage by `3600`. That worked for one-hour windows, but produced shares summing to about `2.0` for two-hour windows and `3.0` for three-hour windows.

Those shares are used as weights when calculating PT skim values. When they summed to about `2.0` for a two-hour window, values such as `pt_travel_times.csv.gz`, `pt_access_times.csv.gz`, and `pt_egress_times.csv.gz` could be about twice as large as expected.
  
The fix uses the actual skim window duration:
  ```java
  maxDepartureTime - minDepartureTime
```

The PR also updates SBB skim documentation/Javadoc to clarify that pt_travel_times.csv.gz currently represents origin-to-destination PT travel time, including access, egress, in-vehicle time, and transfer time. Initial waiting/adaption time is reported separately in pt_adaptiontimes.csv.gz.

cc: @panostsolerid who explained the issue to me

### DIFF
--- a/contribs/sbb-extensions/README.md
+++ b/contribs/sbb-extensions/README.md
@@ -404,7 +404,7 @@ This repository contains code to calculate a variety of skim matrices based on M
   - travel time, routed on network (loaded or unloaded)
   - travel distances, routed on network (loaded or unloaded)
 - for public transport:
-  - travel time (from departure at first stop to arrival at last stop of a trip)
+  - travel time (from origin to destination, including access and egress time)
   - access time (from origin to first stop)
   - egress time (from last stop to destination)
   - number of transfers required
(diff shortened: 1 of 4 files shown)


jfbischoff:
Great find, thanks :)

PR #5143, opened 2026-08-26
feat(drt): time-dependent service regimes (service times, stop networks, service areas)

nkuehnel:
## Why

A real DRT service has service hours, and often a time-dependent stop set or service area (only the main
corridors at night, a reduced area in the afternoon). MATSim knows one static configuration per
`DrtConfigGroup`: `operationalScheme`, `transitStopFile` / `drtServiceAreaShapeFile` and `maxWalkDistance` apply
to the whole day.

This PR adds an optional `serviceRegimes` parameter set that makes the service time and the served stop set
time-dependent. Without it nothing changes. Builds on #5142, which introduced the `RoutingRequest` in
`AccessEgressFacilityFinder` that this PR needs for the departure time.

```xml
<parameterset type="serviceRegimes">
    <parameterset type="serviceRegime">
        <param name="serviceRegimeName" value="peak"/>
        <param name="startTime" value="06:00:00"/>    <!-- inclusive, undefined = open start -->
        <param name="endTime" value="10:00:00"/>      <!-- exclusive, undefined = open end -->
        <param name="stopNetwork" value="peakStops"/> <!-- null = all stops -->
    </parameterset>
</parameterset>
```

Windows are half-open and must be disjoint; times beyond 24 h are allowed, so there is no wrap-around at
midnight. `stopNetwork` names a value of the existing `stopNetworks` facility attribute. Instead of
hand-attributing a network or stop file, the container also takes a `serviceAreaFile` whose polygons name the
regimes they belong to, and `DrtServiceAreas` stamps those names onto the stops. The served area is therefore
derived rather than configured twice.

## Two enforcement layers

`TimeDependentAccessEgressFacilityFinder` returns `Optional.empty()` outside all windows, so the trip becomes a
`drt_fallback` walk leg. That is what mode choice learns from, but it only covers requests that pass through
this run's routing. `DrtServiceRegimeRequestValidator` turns the same rule into a hard constraint and also
catches prebookings, within-day replanning, input plans with ready-made `DrtRoute`s and `estimateAndTeleport`.
Both read one `DrtServiceRegimes`, so the rule lives in one place.

Only the earliest departure time is checked, so the end of service is soft. Rejections use the regular DRT
channel and show up in `drt_customer_stats_*.csv`.

Behaviour change for anyone enabling this: rejecting an immediate request strands the agent, like any other DRT
rejection today. Input plans with DRT legs outside the service time therefore produce stuck agents in iteration
0, until routing replaces those legs.

## Also in here

- The analysis and rebalancing zone systems and the dashboard assumed a non-null `drtServiceAreaShapeFile` for
  `serviceAreaBased`. They now take the union of the service areas through one shared helper, and are unchanged
  without service areas.
- `DrtStopFacilityImpl.createFromLink` dropped the link attributes, so a hand-attributed `serviceAreaBased`
  network could not carry stop networks at all.
- `AttributeBasedStopFinder.parseStopNetworks` is now public, and `CompositePassengerRequestValidator` (dvrp)
  allows a union of validators. It is only bound when `serviceRegimes` is present.

## Tests

Unit tests for the finder, the validator, every config consistency rule and the service areas, plus
`DrtServiceRegimesIT` on Mielec: a single all-day window as non-regression, service time taking effect, stop
networks per window, door2door, validator rejections where routing does not reach, prebooking, and
`serviceAreaBased` built from polygons.

## Not in scope

Rebalancing per regime, constraints or `operationalScheme` per regime, regime identity in the route,
`door2door` service areas, and fleet/shifts per regime (already expressible via shifts).


contribs/drt/src/main/java/org/matsim/contrib/drt/run/DrtServiceConfigurationParams.java
```diff
@@ -0,0 +1,131 @@
...
+import org.matsim.core.utils.misc.OptionalTime;
+import org.matsim.core.utils.misc.Time;
+
+/**
+ * One time-dependent configuration of a DRT service, i.e. one service time window and the set of stops served within
+ * that window. Bookings with a desired departure time outside all configured windows are not served.
+ *
+ * @author nkuehnel / MOIA
+ */
+public final class DrtServiceConfigurationParams extends ReflectiveConfigGroup {
```
mfrawley-moia: is this not also a regime?

mfrawley-moia (approved):

contribs/drt/src/main/java/org/matsim/contrib/drt/run/DrtServiceConfigurationParams.java
mfrawley-moia: @nkuehnel I'd consider summarizing the changeset somewhat, claude tends to be quite verbose in explaining the what, but the why is what we want to read.

SarraDaknou (approved):
Overall looks good, some minor things to improve: 

Code nits:                                                   
  - DrtServiceConfigurations is not final, no obvious reason to leave it open given all the sibling classes are.
  - In DrtModeRoutingModule, the door2doorFinder nullable + ternary inside the stream lambda is a bit hard to follow. Separating the door2door and stopbased/serviceAreaBased cases before building timeWindows would make the intent clearer.                         

  Missing test cases:
  - stopbased + serviceAreaFile + a window with stopNetwork=null (the "serve all stops" fallback), the behaviour is correct per the javadoc but has no dedicated test.                    
  - DrtServiceAreas.intersects() path used by the analysis/rebalancing zone systems with a multi-area config.

PR #5153, opened 2026-09-02
fix(network): keep every change event when several share a start time

langkoos:
## Defect

`TimeVariantLinkImpl` stores change events in a `TreeMap` keyed by start time, so a second event on the same link at the same second **evicts** the first. The evicted event's attribute counter has already been incremented by `incChangeEvents()`, so `recalc()` then finds fewer events than it expects and throws:

```
Expected number of change events (2) differs from the number of events found (1)!
```

Worse than the throw: both `VariableIntervalTimeVariantAttribute` and `FixedIntervalTimeVariantAttribute` publish their arrays **before** validating. Once published, `isRecalcRequired()` compares equal, no further recalc is attempted, and the half-filled arrays are read as if valid. The unwritten slot returns **0.0** — so a link reads zero lanes at every time, including times before the change, permanently. A zero-lane link is a network cut.

This is reachable from within the mobsim via `NetworkChangeEventsEngineI.addNetworkChangeEvent`, not only from a change-events file. The two colliding events do not need to change the same attribute: a lanes change and a freespeed change at the same second collide, because the map is keyed by time alone.

## Fix

- Change events are grouped by start time (`TreeMap<Double, List<NetworkChangeEvent>>`). All events at a time are applied in registration order and collapse into a single value per attribute, keeping `aTimes` strictly increasing for the binary search in `getValue()`.
- An attribute's counter is incremented only when an event is the first at that time to change that attribute.
- Both implementations now validate before publishing, so a future mismatch fails loudly instead of leaving a corrupt link.

Behaviour for events at distinct times is unchanged. Existing `TimeVariantLinkImplTest`, `QSimIntegrationTest` (timevariantnetworks) and `NetworkChangeEventsEngineTest` pass unmodified.

## Tests

Adds `NetworkChangeEventLanesTest`, 6 tests:

- a lanes change event alters QSim storage capacity (previously the only untested change type — freespeed and flow capacity were covered, lanes was not)
- a lanes change leaves flow capacity alone
- the time-variant link reports the changed lane count
- an event injected during the sim takes effect immediately
- two events at the same second on one link both apply
- two events at the same second on one attribute apply in registration order

Regression: `org.matsim.core.network.**`, `org.matsim.core.mobsim.qsim.**`, `org.matsim.integration.timevariantnetworks.**` — 546 tests, 0 failures.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/matsim/src/main/java/org/matsim/core/network/FixedIntervalTimeVariantAttribute.java
+++ b/matsim/src/main/java/org/matsim/core/network/FixedIntervalTimeVariantAttribute.java
@@ -20,7 +20,9 @@
 package org.matsim.core.network;
 
 import java.util.Arrays;
-import java.util.TreeMap;
+import java.util.List;
+import java.util.Map;
+import java.util.NavigableMap;
 
 import org.matsim.core.network.NetworkChangeEvent.ChangeValue;
 import org.matsim.core.trafficmonitoring.TimeBinUtils;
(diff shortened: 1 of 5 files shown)


langkoos:
@kainagel @rakow just raising this in case it somehow may interfere with anything in Distributed MATSim, as it seems Christian recently touched the change events engine for that

PR #5167, opened 2026-09-09
refactor(dsim): decouple serialization and message type registry

Janekdererste:
We want `DSim` to be usable as a drop in replacement for the `mobsim`. This means users should be able to use it without extra dependencies. 

This PR moves the `fory` dependency into the `distributed simulation` contrib toghether with all serialization related code. The serialization is only needed if users want to execute a simulation on multiple compute nodes for which the contrib is required anyway. 

The former `SerializationProvider` is split into a `MessageTypeRegistry` which remains in the core library, while the serialization part stays in `SerializationProvider` and is moved to the contrib.

With this refactor, users can use `DSim` with only matsim-core on the classpath for local simulation setups.


Janekdererste:
Code review was done on private channel.

